PR #815, opened 2019-08-30
feat(http sink): Add support for unverified HTTPS

### LINKED ISSUE #796: Add an option to allow unverified SSL certificates in `http` sink
Some internal HTTPS services might be using self-signed certificates or there might be problems with installation of CA certificates. For such cases it might be handy to have an option in `http` sink to not check the SSL certificate.

Ref #795

bruceg:
Closes #796 

Note that this contains no additional tests, as Vector currently has no tests for HTTPS behavior, and I could not see an simple way of turning existing HTTP tests into HTTPS.

### DIFF
--- a/src/sinks/http.rs
+++ b/src/sinks/http.rs
@@ -42,6 +42,8 @@ pub struct HttpSinkConfig {
     pub request_rate_limit_num: Option<u64>,
     pub request_retry_attempts: Option<usize>,
     pub request_retry_backoff_secs: Option<u64>,
+
+    pub verify_certificate: Option<bool>,
 }
 
 #[derive(Deserialize, Serialize, Debug, Eq, PartialEq, Clone, Derivative)]
@@ -108,6 +110,7 @@ fn http(config: HttpSinkConfig, acker: Acker) -> Result<super::RouterSink, Strin
     let headers = config.headers.clone();
     let basic_auth = config.basic_auth.clone();
     let method = config.method.clone().unwrap_or(HttpMethod::Post);
+    let verify = config.verify_certificate.unwrap_or(true);
 
     let policy = FixedRetryPolicy::new(
         retry_attempts,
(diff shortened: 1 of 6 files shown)


src/sinks/http.rs
```diff
@@ -108,48 +110,52 @@ fn http(config: HttpSinkConfig, acker: Acker) -> Result<super::RouterSink, Strin
     let headers = config.headers.clone();
     let basic_auth = config.basic_auth.clone();
     let method = config.method.clone().unwrap_or(HttpMethod::Post);
+    let verify = config.verify_certificate.unwrap_or(true);
```
LucioFranco: can we add an if branch below this to spit out a warning that this setting was enabled?

src/sinks/util/http.rs
```diff
@@ -21,11 +21,64 @@ pub struct HttpService {
...
     where
         F: Fn(Vec<u8>) -> hyper::Request<Vec<u8>> + Sync + Send + 'static,
     {
-        let https = HttpsConnector::new(4).expect("TLS initialization failed");
+        Self::builder().build(request_builder)
+    }
+}
+
+/// A builder for `HttpService`s
+pub struct HttpServiceBuilder {
```
LucioFranco: Generally, I name these just `Builder` and they can only be constructed from some `builder` fn on the main struct.

src/sinks/util/http.rs
```diff
@@ -21,11 +21,64 @@ pub struct HttpService {
...
+    }
+
+    /// Verify the remote server's certificate
+    pub fn verify_certificate(&mut self, verify: bool) -> &mut Self {
+        self.verify_certificate = verify;
+        self
+    }
+}
+
+impl<F> From<(&HttpServiceBuilder, F)> for HttpService
```
LucioFranco: I don't think this from impl makes much sense can we move this to the build fn in Builder?

src/sinks/util/http.rs
bruceg: The `new` method is private, so it can only be built by code in this module.

LucioFranco: ah good catch, I missed that 😄

src/sinks/util/http.rs
bruceg: Done

src/sinks/http.rs
bruceg: Sure. Is what I just pushed out what you are looking for?

src/sinks/http.rs
```diff
@@ -118,6 +118,12 @@ fn http(config: HttpSinkConfig, acker: Acker) -> Result<super::RouterSink, Strin
         HttpRetryLogic,
     );
 
+    if !verify {
+        warn!(
+            message = "Turning off verify_certificate in http sink can introduce security vulnerabilities"
```
LucioFranco: I'd actually say something like `'verify_certificate' is DISABLED this may lead to security vulnerabilities.`

bruceg: Done

PR #817, opened 2019-08-31
fix(udp source): Update to `tokio-udp` v0.1.5

LucioFranco:
This change introduces a fix that allows us to use
BytesDelimitedCodec without needing to wrap it.


lukesteensen:
Nice! We may also want to revisit the other places we're parsing out of UDP packets and see how this affects them (e.g. syslog, statsd).

PR #840, opened 2019-09-09
chore: Add vscode folder to gitignore

mre:
The `.vscode` folder contains the settings for the VSCode IDE. Similar to the `.idea` folder for JetBrains IDEs, we should ignore any changes in there to keep the repository clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,6 @@ sample.log
 miniodat
 .DS_Store
 .idea/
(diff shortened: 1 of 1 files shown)


LucioFranco:
This also means we should probably remove `idea` too 😄

mre:
Ah I see. Did not know about that. Thanks!

PR #850, opened 2019-09-11
feat(new transform): add split transform

### LINKED ISSUE #799: Extend tokenizer transform to support arbitrary split separator
## Status Quo

As of now, the tokenizer transform splits the input on ~a single whitespace character~ whitespace.
To support more use-cases, the tokenizer could allow for arbitrary separators however.
This would be quite helpful when migrating from Logstash, which [supports that feature](https://www.elastic.co/guide/en/logstash/current/plugins-filters-mutate.html#plugins-filters-mutate-split).

## Proposal

To make this backward-compatible, we could add an optional `separator` config option to the tokenizer that would default to whitespace.

## Alternative approaches

* Using the existing `regex_parser` and `grok_parser` transforms could work, but writing a regular expression for this might be repetitive and could be avoided for this simple use-case.
* Writing a `lua` script is another option, but requires programming experience and might cause some overhead due to crossing the language barrier.
* Create a new split transform and leave the tokenizer transform untouched. In this case we have to clearly communicate the different use-cases of both transforms and there will probably be some code duplication.

mre:
This adds a split transform, which allows to split a given field by the given separator. 
If no separator is given, then it [splits on whitespace](https://doc.rust-lang.org/std/primitive.str.html#method.split_whitespace).

Fixes #799.


xcaptain:
When can this get merged, I need this feature, currently I'm using lua to do split.

mre:
Thanks for the review @lukesteensen. Applied the changes you suggested. Can you have another look?

Issue #851, opened 2019-09-12
Add a field mapping transformer


xcaptain:
A raw log may contain many fields but at last we may only care some of them and want them to have a different name, For example, a transformer produces log like

```json
{
    "a": 1,
    "c": "some non-sence data"
}
```

we may want to rename field `a` to `b` and remove field `c` and get a result like

```json
{
    "b": 1
}
```

currently the `add_fields` and `remove_fields` transformer can not achieve this goal.

lukesteensen:
Thanks for opening this @xcaptain! You're right that there is a gap in those transforms around renaming fields.

This is a good example of a situation where I'd recommend using the lua transform. Here's the equivalent of your example config:

```toml
[transforms.map_fields]
  type = "lua"
  inputs = ["transform_timestamp"]
  source = """
event["http_status"] = event["status"]
"""

[sinks.my_sink1]
  type = "console"
  inputs = ["map_fields"]
  target = "stdout"
  encoding = "json"

[sinks.my_sink2]
  type = "console"
  inputs = ["transform_timestamp"]
  target = "stdout"
  encoding = "json"
```

If you want to remove the old `status` field, you can add `event["status"] = nil`.

We are also in the process of planning a more general transform that will consolidate adding, removing, renaming, coercing, etc into a single more powerful transform (see #750).

lukesteensen:
Closing in favor of #750

xcaptain:
The lua transform is modify `event` inplace, if I want to remove other fields I have to use `remove_fields`, not very convenient but it's ok to use. looking forward to your new transform,and wish this project get better.

PR #853, opened 2019-09-12
feat(new transform): add a simple field mapping transformer

xcaptain:
It's just a simple tryout to implement a mapping transformer I suggested at #851 feel free to close these 2 issues if they are inappropriate

### DIFF
--- a/src/transforms/map_fields.rs
+++ b/src/transforms/map_fields.rs
@@ -1,0 +1,67 @@
+use string_cache::DefaultAtom as Atom;
+use indexmap::IndexMap;
+use super::Transform;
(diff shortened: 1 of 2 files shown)


lukesteensen:
Thank you for this @xcaptain!

As I [mentioned](https://github.com/timberio/vector/issues/851#issuecomment-531304065), we are in the process of planning out a more general transform that will handle this case in addition to the add/rename/coerce actions.

There is also a way to accomplish the same thing with lua right now, so I'm going to go ahead and close this. We appreciate you bringing this up! We'll definitely take it into account as we move forward with the new transform.

PR #871, opened 2019-09-16
docs: Fix some typos in file-source crate

mre:


lukesteensen:
If you wouldn't mind signing off the commit (`git commit --amend --signoff`) and re-pushing, I'll get this merged.

mre:
Sure thing. According to the `CONTRIBUTING.md`, spelling fixes don't need to be signed off however ([reference](https://github.com/timberio/vector/blob/master/CONTRIBUTING.md#what-about-trivial-changes)). Do we need to update that? 😉

lukesteensen:
Oh, good point, I just saw the red DCO check. Maybe we should just ignoring that in these cases. /cc @binarylogic

PR #893, opened 2019-09-18
feat(new source): Inital `kubernetes` source implementation 

ktff:
Implements 1. Decentralized configuration part of #260.

Implementation differs from showed example in:

- Instead of passing vector configuration as argument, vector configuration is mounted as vector.toml file, and it's path passed as argument.  

-  ~~`include_namespace` option was added as it proved to be necessary in tests.~~ (Edit: removed because it's implementation relied on unspecified behavior)

Tests are currently behind `kubernetes-integration-tests` feature flag as they require:

- Access to Kubernetes cluster.

- Tested vector image present on Kubernetes hosts. Currently they are pulling vector image from my repository. 

Before merger, find a way to:

- [x] spin up local Kubernetes cluster
- [x] build/find tested vector image and copy it to every Kubernetes host
- [x] enable tests
- [x] test and accommodate, where possible, other non Docker container runtimes. Only `distributed.yaml` can be possibly changed by this. This is required mostly because `*.log` files ,which are being read, are symlinks to other files. Which means that they also have to be mounted to be accessible.


LucioFranco:
@ktff Looking great! I will give this a proper review on Monday. For testing, I feel like setting up kube in our current test-stable job might be too much. So maybe we want to see if we can setup a specific test stage to spin up minikube and just run the kube tests.

ktff:
> For testing, I feel like setting up kube in our current test-stable job might be too much. So maybe we want to see if we can setup a specific test stage to spin up minikube and just run the kube tests.

@LucioFranco I agree.

The stage should have installed kubectl and minikube. If possible, minikube that is running directly on host so that we can avoid step 2. That would also eliminate the need for virtualization. There are guides for that in Kubernetes documentation:
- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) 
- [minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/)

Then during testing stage:

1. build/find tested vector image
2. copy it to every Kubernetes host. (this could be avoided)
3. `minikube start`
4. run tests
5. `minikube destroy`

config/kubernetes/distributed.yaml
```diff
@@ -0,0 +1,90 @@
...
+        volumeMounts:
+        - name: var-log
+          mountPath: /var/log/
+          readOnly: true        
+        - name: var-lib
+          mountPath: /var/lib
+        - name: config-dir
+          mountPath: /etc/vector
+          readOnly: true 
+      # /var/lib/vector/ needs to be created
```
spencergilbert: Is there a reason you don't just mount this dir as an emptyDir?

ktff: Yes, more specifically the case of applying changed `distributed.yaml`. With host storage, checkpoints stored in that folder will be preserved between changes which will enable us to continue collecting logs where we stopped. With `emptyDir`, some logs will be lost in that case.  

Although there are some downsides. Major one is polluting the host.

But, now I see that there is an alternative. Vector is capable of updating it's topology with new configuration, although it will need to be triggered on file change. The question is, is it possible to make Kubernetes update already existing files created from `ConfigMap` when it changes. This would be a better solution, so I'll investigate.

ktff: Hahaha, it does. I love Kubernetes. So it appears that yes, the file will change if `ConfigMap` is changed. 

I'll push a commit which uses emptyDir for storage, and open/push issue to listen for config file changes.

spencergilbert: Yes, unless the configmap is mounted with a subPath key

https://kubernetes.io/docs/concepts/storage/volumes/#configmap

src/sources/kubernetes.rs
```diff
@@ -0,0 +1,318 @@
...
+    // oldest_first false, having all pods equaly serviced is of greater importance
+    //                     than having time order guarantee.
+
+    // CRI standard ensures unique naming.
+    config.fingerprinting = FingerprintingConfig::DevInode;
+
+    // Have a subdirectory for this source to avoid collision of naming its file source.
+    config.data_dir = Some(globals.resolve_and_make_data_subdir(None, kube_name)?);
+
+    let (file_send, file_recv) = mpsc::channel(1000);
```
LucioFranco: why not use the channel provided in build?

src/sources/kubernetes.rs
```diff
@@ -0,0 +1,318 @@
...
+    let file_source = config
+        .build("file_source", globals, file_send)
+        .map_err(|e| format!("Failed in creating file source with error: {:?}", e))?;
```
LucioFranco: we should start using the errors that #811 brought.

src/sources/kubernetes.rs
```diff
@@ -0,0 +1,318 @@
...
+    })
+}
+
+fn transform_json_message() -> impl FnMut(Event) -> Option<Event> + Send {
+    let mut config = JsonParserConfig::default();
+
+    // Drop so that it's possible to detect if message is in json format
+    config.drop_invalid = true;
+
+    // Drop field
```
LucioFranco: do we need this comment?

src/sources/kubernetes.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        // Rename fields
+        let log = event.as_mut_log();
+
+        // time -> timestamp
+        if let Some(ValueKind::Bytes(timestamp_bytes)) = log.remove(&atom_time) {
+            match DateTime::parse_from_rfc3339(
+                String::from_utf8_lossy(timestamp_bytes.as_ref()).as_ref(),
+            ) {
+                Ok(timestamp) => log.insert_explicit(
```
LucioFranco: I believe this should be implicit because we are adding it?

src/sources/kubernetes.rs
```diff
@@ -0,0 +1,318 @@
...
+                    event::TIMESTAMP.clone(),
+                    timestamp.with_timezone(&Utc).into(),
+                ),
+                Err(error) => warn!(message="Non rfc3339 timestamp",error=%error),
+            }
+        } else {
+            warn!(message = "Missing field", field = %atom_time);
```
LucioFranco:
```suggestion
            warn!(message = "Missing field.", field = %atom_time);
```

src/sources/kubernetes.rs
```diff
@@ -0,0 +1,318 @@
...
+        // time -> timestamp
+        if let Some(ValueKind::Bytes(timestamp_bytes)) = log.remove(&atom_time) {
+            match DateTime::parse_from_rfc3339(
+                String::from_utf8_lossy(timestamp_bytes.as_ref()).as_ref(),
+            ) {
+                Ok(timestamp) => log.insert_explicit(
+                    event::TIMESTAMP.clone(),
+                    timestamp.with_timezone(&Utc).into(),
+                ),
+                Err(error) => warn!(message="Non rfc3339 timestamp",error=%error),
```
LucioFranco:
```suggestion
                Err(error) => warn!(message = "Non rfc3339 timestamp.", %error),
```

src/sources/kubernetes.rs
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        Some(event)
+    }
+}
+
+fn transform_cri_message() -> Result<Box<dyn Transform>, String> {
+    let mut rp_config = RegexParserConfig::default();
+    // message field
+    rp_config.regex =
```
LucioFranco: can we make sure we have tests for these?

tests/kubernetes.rs
```diff
@@ -0,0 +1,427 @@
+// NOTE: Tests assume that Kubernetes is accessable and localy available image of vector
```
LucioFranco: I think it may be better to move this to `src/sinks/kube/test.rs` then bring it in as a `mod test` under `#[cfg(test)`. The reason for this is every `tests/*.rs` we have will have to link against a whole copy of vector and this makes tests much slower. So we should be doing as much of the tests within the main binary under the `cfg(test)` attribute.

config/kubernetes/distributed.yaml
```diff
@@ -0,0 +1,84 @@
+# Everything related to vector should be in this namespace
+apiVersion: v1
+kind: Namespace
+metadata:
+   name: logging
```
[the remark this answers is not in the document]
ktff: Indeed

config/kubernetes/distributed.yaml
```diff
@@ -0,0 +1,84 @@
...
+      # Mount vector configuration from config map as a file vector.toml
+      - name: config-dir
+        configMap:
+         name: vector-config
+         items:
+           - key: vector-agent-config
+             path: vector.toml
+      containers:
+      - name: vector
+        image: timberio/vector:latest
```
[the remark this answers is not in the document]
ktff: We should

src/sources/kubernetes.rs
```diff
@@ -0,0 +1,318 @@
...
+        // Transforms
+        let mut transform_file = transform_file()?;
+        let mut parse_message = parse_message()?;
+
+        // Kubernetes source
+        let source = file_recv
+            .filter_map(move |event| transform_file.transform(event))
+            .filter_map(move |event| parse_message(event))
+            .filter_map(move |event| now.filter(event))
+            .forward(out.sink_map_err(|_| ()))
```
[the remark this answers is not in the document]
ktff: This is a nice trick

src/sources/kubernetes.rs
ktff: I am not sure I understand.

This channel is used to connect file source with transformations defined in build which are then connected to the out channel provided in build.

src/sources/kubernetes.rs
ktff: Yes, just so it's easier to deduce what fields are in the event at which point. But there is a better way to express it so I'll add that.

src/sources/kubernetes.rs
ktff: We are adding it, but the timestamp is taken by the container runtime, so this is a question of is that explicit or implicit? I had an opinion on this, but I forgot it. In the docker source it's explicit, so we should also change that if we decide the opposite.

I have taken a look at the other sources that add timestamps gotten from other side, and `journald source` is the only one of them, and it is adding timestamp as explicit.

src/sources/kubernetes.rs
ktff: I'll add a test like the `fn file_path_transform()`. Other than that, it is in the execution path of the full tests when using non Docker runtime, like containerd in microk8. Which would fail if this fails.

src/sources/kubernetes.rs
ktff: Done.

Also, this is a good use case for an error with a source error, but I don't see it used in Vector.

tests/kubernetes.rs
ktff: Done, and I have also moved source to be under the same mod.rs .

binarylogic:
@ktff just double-checking progress on this. It seems like we just need to finish testing and setting it up on CI, correct? Let us know if anything is blocking you or need help.

ktff:
>  It seems like we just need to finish testing and setting it up on CI, correct?

@binarylogic Tests are here, and they have been tested with `minikube`, `microK8s`, and `AWS EKS`. So setting up CI to run the tests is what's remaining. But, as it was the case with `docker source`, I am unable to change CI. That's why I have been communicating with @LucioFranco about it.

binarylogic:
Ah, I wasn't aware you couldn't change CI. I just invited you to our org, which should give you the access needed. Feel free to make changes as necessary.

LucioFranco:
@ktff I believe you should be able to apply changes to the circleci config?

docs/assets/kubernetes-source.svg
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
LucioFranco: @binarylogic We need to generate one for kubernetes this currently is using the `docker` source svg.

config/kubernetes/vector-daemonset.yaml
```diff
@@ -0,0 +1,85 @@
+# Everything related to vector will live under the `telemetry` namespace.
```
binarylogic: Any reason this is here? Wouldn't it make more sense to move this into the `distribution` folder?

docs/usage/configuration/sources/kubernetes.md
```diff
@@ -0,0 +1,131 @@
...
+### Environment Variables
+
+Environment variables are supported through all of Vector's configuration.
+Simply add `${MY_ENV_VAR}` in your Vector configuration file and the variable
+will be replaced before being evaluated.
+
+You can learn more in the [Environment Variables][docs.configuration#environment-variables]
+section.
+
+### High Level
```
binarylogic: This is an unhelpful title. Can we update it to reflect whatever the purpose of this section is?

binarylogic:
Noting, before we merge this I would like to approve the docs.

config/kubernetes/vector-daemonset.yaml
LucioFranco: Its more of a config than a distribution, this may not be the right place but we probably want it somewhere. It also contains the embedded example config.

docs/usage/configuration/sources/kubernetes.md
LucioFranco: Yeah, happy to take ideas, originally I named this `Concepts` but I'm not sure what a better title might be.

LucioFranco: I removed it.

LucioFranco:
@binarylogic I think the docs are ready to go. I would like to write more thorough guides for kubernetes but I don't think the docs need much more since right now this implementation is simple.

All that is left is your approval and the new svg.

binarylogic:
@LucioFranco we need to improve the docs quite a bit more. For example, there is no Kubernetes platform installation page. This is an important enough source that it warrants being thoughtful about the documentation and setup experience. I would like to review and work on the docs before we merge.

.circleci/config.yml
```diff
@@ -67,6 +67,48 @@ jobs:
...
+      MINIKUBE_WANTUPDATENOTIFICATION: false
+      MINIKUBE_WANTREPORTERRORPROMPT: false
+      MINIKUBE_HOME: /home/circleci
+      CHANGE_MINIKUBE_NONE_USER: true
+    steps:
+      - checkout
+      - run:
+          name: setup kubectl
+          command: |
+            curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/${K8S_VERSION}/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
```
binarylogic: Can you enlighten me how a developer would run these tests locally? Are there docs for this? Even better, is there a way to embed this into the current workflow?

config/kubernetes/vector-daemonset.yaml
binarylogic: The `/distribution` folder is full of configs that integrate with other platforms (docker, deb, rpm, systemd, etc). I would prefer that this live in `/distribution` since I believe it makes more sense there.

.circleci/config.yml
LucioFranco: All of the kubernetes tests are behind a `kubernetes-integration-tests` feature flag and not enabled by default.

```
minikube start
cargo test --features kubernetes-integration-tests
```

It will use your kube config that exists in `~/.kube/config`. Minikube will place one for you or you can run this command to fetch the config from an existing cluster. It just needs _some_ kube cluster.
```
curl http://<some kube cluster>/config > ~/.kube/config
```

The idea here is that if you have access to a kube cluster via `kubectl` then you can run the tests.

binarylogic: Where is this documented?

LucioFranco:
@binarylogic I'm not sure what setup there is? Most of the setup stuff I can think about would involve being a guide since they would need to talk about deployment strategies. So I was not sure what direction we wanted to take these docs versus deployment strategies. 

>  For example, there is no Kubernetes platform installation page

I'm not sure what you mean by this, do we have any examples of a platform installation page? I tried to follow how the other sources documented things.

Another thing to note is that this source only works if you run it as a `DaemonSet` and not as a sidecar which we want to support as another deployment strategy. So, in theory, this may not be the only source that users use in a kubernetes deployment. So because of that, I'm not sure this should contain our high-level kubernetes guides/setup but be a reference point for using this source with a per-node deployment style.

.circleci/config.yml
LucioFranco: Happy to add this to the contributing guide, will do that tomorrow. We can also provide a separate test command to test with minikube.

docs/usage/configuration/sources/kubernetes.md
```diff
@@ -0,0 +1,123 @@
...
+{% code-tabs %}
+{% code-tabs-item title="vector.toml" %}
+```coffeescript
+[sources.my_source_id]
+  type = "kubernetes" # must be: "kubernetes"
+```
+{% endcode-tabs-item %}
+{% endcode-tabs %}
+
+## Options
```
binarylogic: Remove this for now.

docs/usage/configuration/sources/kubernetes.md
```diff
@@ -0,0 +1,123 @@
...
+
+
+## How It Works
+
+### Delivery Guarantee
+
+Due to the nature of this component, it offers a
+[**best effort** delivery guarantee][docs.guarantees#best-effort-delivery].
+
+### Deployment
```
binarylogic: I don't think this belongs here, this is also hard to follow. Shouldn't we provide commands that user can copy/paste to quickly get setup?

docs/usage/configuration/sources/kubernetes.md
```diff
@@ -0,0 +1,123 @@
...
+Environment variables are supported through all of Vector's configuration.
+Simply add `${MY_ENV_VAR}` in your Vector configuration file and the variable
+will be replaced before being evaluated.
+
+You can learn more in the [Environment Variables][docs.configuration#environment-variables]
+section.
+
+### Metadata
+
+Each event will contain a `message` field that contains the direct output from the containers stdout/stderr. There
```
binarylogic: This isn't accurate, or the "Input/Output" section is not accurate because it does not contain the fields mentioned here. Which one is correct?

src/sources/kubernetes/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+    // // TODO: Find out if there are some guarantee from Kubernetes that current build of
+    // // TODO  pod_uid as namespace_pod-name_some-number is a somewhat lasting decision.
+    // NOTE: pod_uid is unspecified and it has been found that on EKS it has different scheme.
+    // NOTE: as such excluding/including using path is hacky, instead, more proper source of
+    // NOTE: information should be used.
+    // NOTE: At best, excluding/including using path can be an optimization
+    // TODO: Exclude whole kube-system namespace properly
+    config
+        .exclude
+        .push((LOG_DIRECTORY.to_owned() + r"kube-system_*/**").into());
```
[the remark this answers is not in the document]
ktff: I am for this, but I am also for filtering it ourselves later when we start enriching events with `namespace`.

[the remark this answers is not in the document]
LucioFranco: Right, we should also try to provide the namespace coming out of this source as well.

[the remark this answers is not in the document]
ktff: You mean trying to parse out namespace from pod_uid, if it succeeds good, if not just ignore it?
Or something else?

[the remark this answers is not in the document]
LucioFranco: Yeah, but probably makes sense to do this in a follow up PR.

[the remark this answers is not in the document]
ktff: I am for that.

binarylogic:
Just providing an update. I have merged in all of the upstream changes from `master`. This includes the new website/documentation. I am working on the installation docs as well as the use case page for Kubernetes so that we can announce it properly.

binarylogic:
Nice work on this! I'm going to merge so that we can unblock work. Documentation is being worked on.

PR #894, opened 2019-09-18
chore(config): Make encoding non-optional

### LINKED ISSUE #806: Update sink encoding to be non optional
~~Most sinks allow the user to specify an encoding type. If this option is not provided the sinks will attempt to check each event and encode it based on if it was explicitly structured by the user. Due to precidence the current implementation of the match statement will attempt to encode the event as a structured event even if the user explicitly included the `encoding = "text"` option. This should be fixed by forcing the left tuple match of the boolean value to be `None`.~~

Update:

Based on the comments below, the sinks will be modified to contain explicit options to encode their events. This means that encoding is no longer optional and a user must select their desired encoding or the sink can not be built.

Implementation wise this should be a simple fix to remove the second tuple item in this match statement https://github.com/timberio/vector/blob/master/src/sinks/aws_cloudwatch_logs/mod.rs#L268 and to change the encoding config item to not be wrapped in an option. In a separate issue, we can consolidate encoding options.

Sinks that need this fix:
- [x] aws_cloudwatch_logs
- [x] aws_s3
- [x] aws_kinesis_streams
- [x] kafka
- [x] splunk
- [x] tcp
- [x] All sinks should have tests for this behavior

LucioFranco:
This fixes the confusion around what encoding is being produced by the sink. Before, we were dynamically setting the encoding of the event per event if you did not supply the encoding option. This change changes that so that you must always provide an encoding option to ensure that it is clear what encoding is being produced by the sink.

Closes #806


LucioFranco:
@lukesteensen yeah, having trouble pulling down containers and getting ruby to work...but will make sure they are both good before we merge 👍

PR #930, opened 2019-09-25
feat(metric data model): metrics buffer

loony-bean:
Rel #920

### DIFF
--- a/src/sinks/aws_cloudwatch_metrics.rs
+++ b/src/sinks/aws_cloudwatch_metrics.rs
@@ -1,10 +1,10 @@
 use crate::{
     buffers::Acker,
-    event::{metric::Direction, Event, Metric},
+    event::{Event, Metric},
     region::RegionOrEndpoint,
     sinks::util::{
         retries::{FixedRetryPolicy, RetryLogic},
-        BatchServiceSink, SinkExt,
+        BatchServiceSink, MetricBuffer, SinkExt,
     },
     topology::config::{DataType, SinkConfig},
 };
@@ -19,16 +19,10 @@ use std::collections::HashMap;
 use std::{convert::TryInto, time::Duration};
 use tower::{Service, ServiceBuilder};
 
-#[derive(Clone, Default)]
-struct State {
-    gauges: HashMap<String, f64>,
-}
-
 #[derive(Clone)]
 pub struct CloudWatchMetricsSvc {
     client: CloudWatchClient,
     config: CloudWatchMetricsSinkConfig,
-    state: State,
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone, Default)]
@@ -85,13 +79,7 @@ impl CloudWatchMetricsSvc {
             CloudWatchMetricsRetryLogic,
         );
 
-        let state = State::default();
-
-        let cloudwatch_metrics = CloudWatchMetricsSvc {
-            client,
-            config,
-            state,
-        };
+        let cloudwatch_metrics = CloudWatchMetricsSvc { client, config };
 
         let svc = ServiceBuilder::new()
             .concurrency_limit(in_flight_limit)
@@ -100,8 +88,8 @@ impl CloudWatchMetricsSvc {
             .timeout(Duration::from_secs(timeout))
             .service(cloudwatch_metrics);
 
-        let sink = BatchServiceSink::new(svc, acker).batched_with_min(
-            Vec::new(),
+        let sink = BatchServiceSink::new(svc, acker).batched_with_max(
+            MetricBuffer::new(),
             batch_size,
             Duration::from_secs(batch_timeout),
         );
@@ -167,37 +155,16 @@ impl CloudWatchMetricsSvc {
                 Metric::Gauge {
                     name,
                     val,
-                    direction,
+                    direction: None,
                     timestamp,
                     tags,
-                } => {
-                    let delta = match direction {
-                        None => 0.0,
-                        Some(Direction::Plus) => val,
-                        Some(Direction::Minus) => -val,
-                    };
-
-                    let val = self
-                        .state
-                        .gauges
-                        .entry(name.clone())
-                        .and_modify(|v| {
-                            if direction.is_none() {
-                                *v = val
-                            } else {
-                                *v += delta
-                            }
-                        })
-                        .or_insert(val);
-
-                    Some(MetricDatum {
-                        metric_name: name.to_string(),
-                        value: Some(*val),
-                        timestamp: timestamp.map(timestamp_to_string),
-                        dimensions: tags.map(tags_to_dimensions),
-                        ..Default::default()
-                    })
-                }
+                } => Some(MetricDatum {
+                    metric_name: name.to_string(),
+                    value: Some(val),
+                    timestamp: timestamp.map(timestamp_to_string),
+                    dimensions: tags.map(tags_to_dimensions),
+                    ..Default::default()
+                }),
                 Metric::Histogram {
                     name,
                     val,
@@ -212,7 +179,7 @@ impl CloudWatchMetricsSvc {
                     dimensions: tags.map(tags_to_dimensions),
                     ..Default::default()
                 }),
-                _ => None,
+                _ => unimplemented!(),
             })
             .collect();
 
@@ -305,13 +272,8 @@ mod tests {
         let config = config();
         let region = config.region.clone().try_into().unwrap();
         let client = CloudWatchMetricsSvc::create_client(region).unwrap();
-        let state = State::default();
 
-        CloudWatchMetricsSvc {
-            client,
-            config,
-            state,
-        }
+        CloudWatchMetricsSvc { client, config }
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


src/sinks/util/metrics_buffer.rs
```diff
@@ -0,0 +1,488 @@
+use super::batch::Batch;
```
LucioFranco: we should probably create a `sinks/util/buffer/mod.rs` and add metrics and partiton into here. I think this would make sense? @lukesteensen

loony-bean:
@lukesteensen Thanks for the sketch! It was really useful. You can have a look at the updated version (note that it doesn't have support for relative Gauges yet).

> The main difference is that we have no AggregatedMetric, just normal ones.

Yeah, that was bothering me as well.

> if let Some(pos) = self.metrics.iter().position(|m|

I'm not really comfortable using Vec here, because users are talking about sending us 100's of distinct metrics at 100's of RPS, and buffer sizes could be several minutes long. So I've kept using HashMaps. Please let me know what you think.

src/sinks/util/batch.rs
```diff
@@ -65,6 +65,16 @@ where
         Self::build(inner, batch, min_size, min_size, max_linger)
     }
 
+    pub fn new_min_max(
```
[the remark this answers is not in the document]
loony-bean: Oh, thanks. That's not intended :)

loony-bean:
> I think the biggest gap right now is that it only buffers counters and gauges. That's a tricky part about using the hash maps instead of the vector, you don't really have a good place to stash things that can't be merged.

I'm starting to believe that we are close to having all of them aggregated. Could you have a look into the latest version?

`Set`s are simply deduplicated, which doesn't affect their promise to represent # of unique objects.

And for `Histogram`s I've used the idea that if two histogram observations share the same value, then we can can merge them into a single observation with their `sample_rate`s added up.

src/sinks/util/buffer/metrics.rs
```diff
@@ -0,0 +1,523 @@
...
+
+trait Aggregate {
+    fn partition(&self) -> String;
+}
+
+impl Aggregate for Metric {
+    fn partition(&self) -> String {
+        match self {
+            Metric::Counter { name, tags, .. } => format!("{}{:?}", name, tags),
+            Metric::Gauge { name, tags, .. } => format!("{}{:?}", name, tags),
```
lukesteensen: We probably need to take the type into account here. I'm also not sure that generating strings like this is the best approach.

src/sinks/aws_cloudwatch_metrics.rs
```diff
@@ -212,7 +179,7 @@ impl CloudWatchMetricsSvc {
                     dimensions: tags.map(tags_to_dimensions),
                     ..Default::default()
                 }),
-                _ => None,
+                _ => unimplemented!(),
```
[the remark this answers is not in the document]
loony-bean: No reason actually.

src/event/metric.rs
```diff
@@ -57,4 +58,85 @@ impl Metric {
...
+                },
+                Metric::Histogram {
+                    name: new_name,
+                    val: new_val,
+                    sample_rate: new_sample_rate,
+                    timestamp: new_timestamp,
+                    tags: new_tags,
+                },
+            ) => {
+                if val == new_val {
```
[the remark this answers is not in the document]
loony-bean: I agree that we might expect a lot of variation in _timings_ data, especially if its granular (ms).

But there is also another big domain of _size_ data. Like `Bytes Out`, `Content Size` kind of things, which are going to have a lot of repetition actually, especially in heavy-read workloads.

Also, if the system being logged has distinction between _processed_ and _billed_ (like [BigQuery](https://cloud.google.com/bigquery/docs/reference/auditlogs/rest/Shared.Types/BigQueryAuditMetadata) for example), then with caching enabled it will produce a lot of same values (zeros) in fields like `totalBilledBytes`.

This _processed_ VS _billed_ dichotomy can also affect timings data. Let's take [AWS Lambda](https://aws.amazon.com/blogs/mt/understanding-aws-lambda-behavior-using-amazon-cloudwatch-logs-insights/) logs for example.

```
REPORT RequestId: 8b4a32a4-7dc0-4342-a998-48b051d683e0 Duration: 182.29 ms Billed Duration: 200 ms Memory Size: 128 MB Max Memory Used: 35 MB
```

The `Billed Duration` thing here will be highly repetitive.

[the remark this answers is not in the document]
loony-bean: > If we want to support histograms, we'll need something more general than this.

The nice feature of this `sample_rate` hack is that it is a lossless transformation. I agree that having client-size (lossy) histograms aggregation is important, I just think we need a separate issue for it.

src/sinks/util/buffer/metrics.rs
loony-bean: > I'm also not sure that generating strings like this is the best approach.

I _feel_ like I understand what you mean, but maybe you could elaborate a bit? I've tried few times, but all my attempts produced much more verbose code then the 'string-typed' one.

src/event/metric.rs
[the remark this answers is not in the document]
lukesteensen: My original idea was to expand the `Histogram` variant to be able to hold more than one data point, so merging would be as simple as merging the sets of points. I'm curious how dramatic of a change that would be across the rest of the code base.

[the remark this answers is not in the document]
lukesteensen: But looking at the cloudwatch docs, it doesn't seem like that's actually a model that they support in the API.

src/sinks/util/buffer/metrics.rs
lukesteensen: Yes, that was not the most helpful advice, sorry :smile: 

I've been thinking about this a quite a bit and I _think_ I have a promising solution:

```rust
struct MetricEntry(Metric);

impl Hash for MetricEntry {
    fn hash<H: Hasher>(&self, state: &mut H) {
        mem::discriminant(&self.0).hash(state);

        match &self.0 {
            Metric::Counter { name, tags, .. } => {
                name.hash(state);
                tags.hash(state);
            }
            Metric::Gauge { name, tags, .. } => {
                name.hash(state);
                tags.hash(state);
            }
            ...
        }
    }
}

struct MetricBuffer {
    metrics: IndexSet<MetricEntry>,
}
```

So we basically use a newtype to control how metrics are hashed, which lets us use all the existing hashmap-style functionality to our advantage.

Then in the `push` implementation we can do roughly this:

```rust
if let (existing) = self.metrics.swap_take(new_metric) {
    let merged = existing.merge(new_metric);
    self.metrics.insert(merged)
} else {
    self.metrics.insert(new_metric);
}
```

What do you think?

binarylogic:
@loony-bean anything blocking you from merging? Just wanted a quick update so this doesn't linger.

src/sinks/util/buffer/metrics.rs
loony-bean: @lukesteensen I think this will work, thanks! Had to implement some traits to make it work for floats and hashmaps. Could you take a look at the latest commit?

Also this `swap_take` operation shuffles metrics order a bit, not sure if it's important for us though.

src/sinks/util/buffer/metrics.rs
```diff
@@ -0,0 +1,548 @@
...
+            Metric::Histogram { name, val, .. } => {
+                name.hash(state);
+                val.to_bits().hash(state);
+            }
+        }
+
+        self.0
+            .tags()
+            .as_ref()
+            .map(|ts| ts.iter().for_each(|t| t.hash(state)));
```
lukesteensen: Can we not just do `self.0.tags().hash(state)` here?

src/sinks/util/buffer/metrics.rs
```diff
@@ -0,0 +1,548 @@
...
+    fn eq(&self, other: &Self) -> bool {
+        let mut state = DefaultHasher::new();
+        self.hash(&mut state);
+        let hash1 = state.finish();
+
+        let mut state = DefaultHasher::new();
+        other.hash(&mut state);
+        let hash2 = state.finish();
+
+        hash1 == hash2
```
lukesteensen: This is a clever way to do it :smile:

src/sinks/util/buffer/metrics.rs
```diff
@@ -0,0 +1,548 @@
...
+        }
+
+        Self {
+            state,
+            metrics: IndexSet::new(),
+        }
+    }
+
+    fn finish(self) -> Self::Output {
+        self.metrics.into_iter().map(|e| e.0).collect()
```
lukesteensen: If we're worried about the `swap_take` and ordering, we could do a sort here. Not sure that it's worth it though.

src/sinks/util/buffer/metrics.rs
loony-bean: This is because `HashMap` does not implement `Hash`. The reason is

> To have HashMaps implement Hash, we need to iterate through their contained hashes in a consistent order.

Oh wait.. this means my naive iterative hashing is also not perfectly stable.

BTW, I was thinking some time ago about switching to IndexMap for storing tags (or maybe even something like `Vec<(String, String)>`)

src/sinks/util/buffer/metrics.rs
loony-bean: Yeah, I suppose we could use standard HashSet then?

src/sinks/util/buffer/metrics.rs
lukesteensen: Ah, I'd assumed it was a `Vec` already. I don't think we want order to matter when hashing tags, so maybe `BTreeMap`?

src/sinks/util/buffer/metrics.rs
lukesteensen: Yes, I think that's true. For some reason, I had thought that `swap_take` would maintain the order we wanted, but I don't think that's correct looking at it now.

src/sinks/util/buffer/metrics.rs
loony-bean: I believe the only place where we do something Map-ish with the tags is the `remove_tags` transform. Normally what's happening to tags is they are being constructed and carried into sinks, where they are transformed into Vec<_> or String.

I'll be fine with anything having stable order, but I'd vote for Vec because of ergonomic reasons, to remove some boilerplate converting to and from Maps.

Performance wise, because metrics do not contain many tags usually, I expect working with small Vec's should be effective even for operations like `remove_tags`.

lukesteensen: Makes sense! My only concern with a `Vec` is the case where two different metrics have the same tags but in a different order. I think we'd still want to merge them, but maybe we could adjust the hashing to account for that.

loony-bean:
@binarylogic Almost there!

lukesteensen:
:tada: Thanks for pushing this through @loony-bean!

PR #934, opened 2019-09-25
feat(kafka sink): Use PKCS#12 keys instead of JKS

bruceg:
It turns out Java 9 and later can use PKCS#12 key stores instead of the less common Java KeyStore format. Since the JKS format is much harder to deal with (only Java `keytool` can manipulate them), the PKCS#12 format is preferred.


lukesteensen:
Nice!

PR #949, opened 2019-09-30
chore(operations): Update tokio versions

LucioFranco:
This updates `tokio-reactor` and `tokio-threadpool` to the latest
versions which provide fixes for MSRV and removes extra duplicate
crates.


LucioFranco:
@lukesteensen ah damn, though this should bring us inline with some more deps. We've been cleaning up tokio 0.1 as there were some breaking changes that required us to bump.

PR #953, opened 2019-10-01
feat(elasticsearch sink): Add support for TLS options

bruceg:
This adds support for a standard set of TLS options to the Elasticsearch sink, as well as fixing some issues with other aspects of HTTPS support in that sink (the health check would not have worked as it was with some services).

Closes issue #887

### DIFF
--- a/src/sinks/util/tls.rs
+++ b/src/sinks/util/tls.rs
@@ -33,6 +33,8 @@ enum TlsError {
         filename: PathBuf,
         source: native_tls::Error,
     },
+    #[snafu(display("Must specify both TLS key_file and crt_file"))]
+    MissingCrtKeyFile,
     #[snafu(display("Could not parse X509 certificate in {:?}: {}", filename, source))]
     X509ParseError {
         filename: PathBuf,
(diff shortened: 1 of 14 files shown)


src/sinks/util/tls.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        let authority = match options.ca_path {
+            None => None,
+            Some(ref path) => Some(load_certificate(path)?),
+        };
+
+        let identity = match options.crt_path {
+            None => None,
+            Some(ref crt_path) => {
+                let identity = load_build_pkcs12(
```
LucioFranco: since we are making this generic we prob want to start thinking about other formats like pkcs8 and rsa?

src/sinks/util/tls.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+pub trait TlsConnectorExt {
+    fn use_tls_settings(&mut self, settings: TlsSettings) -> &mut Self;
+}
+
+impl TlsConnectorExt for TlsConnectorBuilder {
+    fn use_tls_settings(&mut self, settings: TlsSettings) -> &mut Self {
+        self.danger_accept_invalid_certs(!settings.verify_certificate);
```
LucioFranco: might be nice to add a warning here that this is being enabled.

src/sinks/util/tls.rs
bruceg: Absolutely. Natively supporting PKCS#12 identity files instead of creating them from a pair of PEM files would be a first step. Probably should go in a separate PR though.

LucioFranco: Yup, just thinking out loud a bit. We should make a follow-up issue for this.

src/sinks/util/tls.rs
bruceg: It is optionally enabled by setting `verify_certificate` to `false`, and there is a warning in the documentation for that setting (in the `http` sink at least, where it is visible).

... but now that I look at it, it should be an `Option<bool>` which defaults to `true`. I'll fix that part.

bruceg: Oh, and I see the `http` sink does output a warning. I'll look into consolidating the two.

bruceg:
> Maybe we can add the tls option to HttpService and then it should be really easy to expose tls to all our sinks that use HttpService.

I was thinking along those lines when I wrote `TlsOptions` but it cannot be done without breaking config file compatibility. I will open an issue to discuss this.

src/sinks/util/tls.rs
bruceg: Done

PR #960, opened 2019-10-02
fix(stdin source): Resolve inability to shutdown Vector when std…

### LINKED ISSUE #932: Update `stdin` source to use a thread instead of using the tokio version
Related to #546 and a bug in tokio where blocking annotations can not be shutdown via a drop in the task. This causes `stdin` sources to hang when a user attempts to shut it down.

# Specification

Update the `stdin` source to use a similar strategy to what the `file` source uses by spawning a thread and using [`stdin`](https://doc.rust-lang.org/beta/std/io/fn.stdin.html) directly. Then using the `out` channel passed into the `SourceConfig::build` fn to send events back into the topology. This will stop the `stdin` source from blocking shutdown.

To send events down the out channel you should use https://docs.rs/futures/0.1.29/futures/sync/mpsc/struct.Sender.html#method.try_send to send the event. 

To fetch lines off of `stdin`, https://doc.rust-lang.org/std/io/struct.Stdin.html#method.read_line should be used. Once, there is a `String` this should be converted into an `Event` as seen [here](https://github.com/timberio/vector/blob/master/src/sources/stdin.rs#L74).

AlyHKafoury:
Closes https://github.com/timberio/vector/issues/932

Signed-off-by: AlyHKafoury <aly.kafoury@gmail.com>

### DIFF
--- a/src/sources/stdin.rs
+++ b/src/sources/stdin.rs
@@ -3,13 +3,9 @@ use crate::{
     topology::config::{DataType, GlobalOptions, SourceConfig},
 };
 use bytes::Bytes;
-use codec::BytesDelimitedCodec;
 use futures::{future, sync::mpsc, Future, Sink, Stream};
 use serde::{Deserialize, Serialize};
-use tokio::{
-    codec::FramedRead,
-    io::{stdin, AsyncRead},
-};
+use std::{io, thread};
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
 #[serde(deny_unknown_fields, default)]
@@ -40,7 +36,11 @@ impl SourceConfig for StdinConfig {
         _globals: &GlobalOptions,
         out: mpsc::Sender<Event>,
     ) -> crate::Result<super::Source> {
-        Ok(stdin_source(stdin(), self.clone(), out))
+        Ok(stdin_source(
+            io::BufReader::new(io::stdin()),
+            self.clone(),
+            out,
+        ))
     }
 
     fn output_type(&self) -> DataType {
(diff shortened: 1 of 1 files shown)


LucioFranco:
@AlyHKafoury thank you for this! I will give it a proper review tomorrow/later this week!

AlyHKafoury:
> @AlyHKafoury thank you for this! I will give it a proper review tomorrow/later this week!

Thanks alot I really like the project and hope to be able to contribute alot to it.

src/sources/stdin.rs
```diff
@@ -40,32 +36,58 @@ impl SourceConfig for StdinConfig {
...
-            BytesDelimitedCodec::new_with_max_length(b'\n', config.max_length),
-        )
-        .map(move |line| create_event(line, &host_key, &hostname))
-        .map_err(|e| error!("error reading line: {:?}", e))
-        .forward(out.sink_map_err(|e| error!("Error sending in sink {}", e)))
-        .map(|_| info!("finished sending"));
+        let (mut tx, rx) = futures::sync::mpsc::channel(1024);
+
+        thread::spawn(move || {
+            println!("Loaded thread");
```
LucioFranco: I don't think this println is no longer needed?

src/sources/stdin.rs
```diff
@@ -40,32 +36,58 @@ impl SourceConfig for StdinConfig {
...
+            loop {
+                let mut buf = String::new();
+                match stdin.read_line(&mut buf) {
+                    Err(e) => {
+                        error!("Error reading from source: stdin: {}", e);
```
LucioFranco: Since we are using the `tracing` crate for logging this is the preferred way to setup printing errors.
```suggestion
                        error!(message = "Unable to read from source.", error = %e);
```

src/sources/stdin.rs
```diff
@@ -40,32 +36,58 @@ impl SourceConfig for StdinConfig {
...
+                    Ok(0) => {
+                        break;
+                    }
+                    _ => (),
+                }
+                while let Err(e) = tx.try_send(buf.clone()) {
+                    if e.is_full() {
+                        continue;
+                    }
+                    error!("Error sending : {}", e);
```
LucioFranco:
```suggestion
					error!(message = "Unable to send event.", error = %e);
```

src/sources/stdin.rs
```diff
@@ -40,32 +36,58 @@ impl SourceConfig for StdinConfig {
...
+                    break;
+                }
+            }
+        });
+
+        let source = rx
+            .map(move |line| create_event(Bytes::from(line.trim()), &host_key, &hostname))
+            .map_err(|e| error!("error reading line: {:?}", e))
+            .forward(out.sink_map_err(|e| error!("Error sending in sink {}", e)))
```
LucioFranco:
```suggestion
            .forward(out.sink_map_err(|e| error!(message = "Unable to send event to out.", error = %e)))
```

src/sources/stdin.rs
```diff
@@ -40,32 +36,58 @@ impl SourceConfig for StdinConfig {
...
-        .map(move |line| create_event(line, &host_key, &hostname))
-        .map_err(|e| error!("error reading line: {:?}", e))
-        .forward(out.sink_map_err(|e| error!("Error sending in sink {}", e)))
-        .map(|_| info!("finished sending"));
+        let (mut tx, rx) = futures::sync::mpsc::channel(1024);
+
+        thread::spawn(move || {
+            println!("Loaded thread");
+            loop {
+                let mut buf = String::new();
```
LucioFranco: I wonder if we can reuse this buf instead of allocating a new one each iteration. Another idea also is to use `bytes::BytesMut` and `BytesMut::freeze` once we've collected the whole event and want to send it.

LucioFranco (changes requested):
This looks like a really awesome start! Let me know if you have any questions, I left some comments inline.

AlyHKafoury:
@LucioFranco  Great, I'll be working on these changes in a couple of days and get back to you

AlyHKafoury:
@LucioFranco won't reusing the buffer cause us to need to clear it in each iteration, I don't know the performance of that compared to re allocating, I working on that though and please enlighten me about this part

src/sources/stdin.rs
```diff
@@ -40,32 +36,60 @@ impl SourceConfig for StdinConfig {
...
-            BytesDelimitedCodec::new_with_max_length(b'\n', config.max_length),
-        )
-        .map(move |line| create_event(line, &host_key, &hostname))
-        .map_err(|e| error!("error reading line: {:?}", e))
-        .forward(out.sink_map_err(|e| error!("Error sending in sink {}", e)))
-        .map(|_| info!("finished sending"));
+        let (mut tx, rx) = futures::sync::mpsc::channel(1024);
+
+        thread::spawn(move || {
+            let mut buf = Box::new(BytesMut::new());
```
LucioFranco: Not sure we need to box here?

LucioFranco: So we actually don't need a buffer because internally `lines()` does that. What we will want to do is use the `Bytes::from(<string>)` function to go from string -> bytes that we can send down the channel. Hopefully that makes sense :)

AlyHKafoury: Yes sure I am currently working on it

src/sources/stdin.rs
```diff
@@ -60,26 +60,23 @@ where
...
                         error!(message = "Unable to read from source.", error = %e);
                         break;
                     }
                     Ok(string_data) => {
-                        &mut buf.extend_from_slice(string_data.as_bytes());
-                        ()
+                        let msg = Bytes::from(string_data);
+                        while let Err(e) = tx.try_send(msg.clone()) {
+                            if e.is_full() {
+                                continue;
```
LucioFranco: a sleep here might be good

LucioFranco (approved):
Thanks for this!

PR #1055, opened 2019-10-18
chore(operations): Update leveldb-sys up to 2.0.5

### LINKED ISSUE #1032: Support macOS Catalina
At the moment builds with `leveldb` feature on macOS Catalina are broken because `leveldb-sys` crate cannot be compiled. Builds on macOS Mojave still work.

a-rodin:
Fixes #1032


LucioFranco:
@a-rodin looks like `2.0.5` is out can we bump to that?

a-rodin:
> @a-rodin looks like 2.0.5 is out can we bump to that?
Yeah, I wanted to bump it to 2.0.5, but 2.0.4 was cached locally and `cargo check` used it.

PR #1063, opened 2019-10-21
fix(operations): Use rdkafka crate from the upstream Git repository

a-rodin:
Fixes `rdkafka` part of #1032.

Although it solves problem with compiling `rdkafka` crate, there are still linking issues with it unsolved yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2466,7 +2466,7 @@ dependencies = [
 [[package]]
 name = "rdkafka"
 version = "0.21.0"
-source = "git+https://github.com/timberio/rust-rdkafka#9cb0245aee2769ce2113f0c35bde5dd6566578b7"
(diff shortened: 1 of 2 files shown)


a-rodin:
The problem with CI is fede1024/rust-rdkafka#126. Newer versions of `rust-rdkafka` crate require `libclang` and `libllvm`. While the requirement might seem as too much, I generally like the reasoning behind switching to generation of bindings during build time implemented in fede1024/rust-rdkafka#119 because it simplifies compilation for new platforms. For example, without it #859 would be harder to do.

benesch:
I just merged a change to rust-rdkafka that removes the build-time dependency on clang/libclang, if you want to give that a try! Details here: da77534c4cb72d8f5ce0b51fd20fc2a532f829a8

LucioFranco:
@benesch would it be possible to get a release of this?

benesch:
Sorry for the long delay here—but 0.23 is out and incorporates all the build system fixes.

PR #1092, opened 2019-10-25
chore(operations): Use leveldb dependency from a fork with portable char type

a-rodin:
Ref #1056 #1054

This PR uses the most recent versions of LevelDB and Snappy from our forks of [leveldb](https://github.com/timberio/leveldb/tree/v0.9.0) and [leveldb-sys](https://github.com/timberio/leveldb-sys/tree/v3.0.0) crates.

Because newer versions of LevelDB completely switched from Autotools to CMake, I've also added CMake dependency to docs and our builders.

### DIFF
--- a/scripts/ci-docker-images/builder-x86_64-unknown-linux-gnu/Dockerfile
+++ b/scripts/ci-docker-images/builder-x86_64-unknown-linux-gnu/Dockerfile
@@ -17,7 +17,8 @@ RUN apt-get update && \
   curl \
   git \
   openssh-server \
-  vim
+  vim \
+  cmake
 
 # Note: We do not compile or install leveldb or rdkafka libraries because
 #       those Rust creates automatically build and link the libraries for
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,6 +137,10 @@ jobs:
       - checkout
       - *install-rust
       - run:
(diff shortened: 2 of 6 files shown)


scripts/ci-docker-images/builder-x86_64-unknown-linux-gnu/Dockerfile
```diff
@@ -17,7 +17,8 @@ RUN apt-get update && \
   curl \
   git \
   openssh-server \
-  vim
+  vim \
```
LucioFranco: we can probably remove this?

a-rodin: We can, although not sure if we should if we are going to do #860.

a-rodin: Actually, the `x86_64-unknown-linux-gnu` seems to already have CMake installed, but the version in it is too old.

a-rodin:
In order to not have failing tests on `master`, I propose to wait for #860 and #946 before merging this.

a-rodin:
Closing because of https://github.com/timberio/vector/issues/1056#issuecomment-553952554.

PR #1095, opened 2019-10-25
chore(operations): Update `ctor` dependency

### LINKED ISSUE #1035: Support LLVM lld as the linker

a-rodin:
This PR uses the latest version of [ctor](https://github.com/mmastrac/rust-ctor) crate from the Git repository.

Closes #1035 
Ref #1054 #946

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -528,11 +528,11 @@ dependencies = [
 
 [[package]]
 name = "ctor"
-version = "0.1.9"
+version = "0.1.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "quote 0.6.12 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


a-rodin:
As the crate's repository doesn't have version tags, it's not clear does the release published on crates.io include all changes from the repository or not. I'll check it, and if it does, we might not need to use the Git repository.

PR #1104, opened 2019-10-28
fix(elasticsearch sink): Wrap provider call with a tokio runtime

### LINKED ISSUE #1093: ElasticSearch sink fails to get EC2 instance credentials
## Software Version

Vector: 0.5.0, RPM

## Issue

Vector doesn't seem to be able to obtain the EC2 instance profile role credentials

## Expected

Vector would get the credentials and be able to push logs to ElasticSearch

## Observations

The EC2 instance definitely has enough permissions to perform actions in the AWS ElasticSearch Service, this was confirmed by using [`aws-es-curl`](https://github.com/joona/aws-es-curl) on the instance to perform some basic operations such as looking at the cluster health and creating a new index.

When running Vector in dry-run mode with an ES sink configuration of 
```toml
[sinks.elasticsearch]
 host = "https://logging.endpoint.example.org/"
 index = "ec2-%Y-%m-%d"
 inputs = ["var_log"]
 type = "elasticsearch"
 provider = "aws"

# The region has to be provided like this, the documentation appears to be incorrect
 [sinks.elasticsearch.region]
 region = "eu-west-1"
```

Vector will eventually error out with
```
 ERROR vector::topology: Configuration error: Sink "elasticsearch": Could not
 generate AWS credentials: CredentialsError { message: "Couldn\'t find AWS
 credentials in environment, credentials file, or IAM role." }
```

However, we already verified that the instance profile role was able to perform actions within the ES cluster.

If we perform a `tcpdump` from the instance with `tcpdump -X -s 0 'tcp port 80'`, it appears that Vector (probably the Rusoto crate), never sends the `GET` request to the `169.254.169.154` meta-data endpoint. It appears to connect and disconnect, never sending the `GET`.

I am currently attempting to upgrade to Rusoto 1.38.0 and Credentials 0.17.0 to see if this has any effect, however, due to the `RusotoError` changes this is not a quick process.

LucioFranco:
This change fixes the usage of `rusoto`'s credentials
the issue arises when a user wants to fetch credentials
via the network with the `InstanceMetadataProvider` type.
This type sets a timer that depends on the `tokio-timer`
being set. Previously, this sink used the `wait` combinator
that _does_ not wrap the future with a tokio timer. Thus,
when a user tried to run this it would fail out because
the timer was not set. This changes the code to allow
the `credentials` fetch future to be wrapped in a tokio
timer.

Closes #1093

### DIFF
--- a/src/sinks/elasticsearch.rs
+++ b/src/sinks/elasticsearch.rs
@@ -130,13 +130,17 @@ impl ElasticSearchCommon {
                 if region.is_none() {
                     return Err(ParseError::AWSRequiresRegion.into());
                 }
-                Some(
-                    DefaultCredentialsProvider::new()
-                        .context(AWSCredentialsProviderFailed)?
-                        .credentials()
-                        .wait()
-                        .context(AWSCredentialsGenerateFailed)?,
-                )
+
+                let provider =
+                    DefaultCredentialsProvider::new().context(AWSCredentialsProviderFailed)?;
+
+                let mut rt = tokio::runtime::current_thread::Runtime::new()?;
+
+                let credentials = rt
+                    .block_on(provider.credentials())
+                    .context(AWSCredentialsGenerateFailed)?;
+
+                Some(credentials)
             }
         };
 


lukesteensen:
Can you explain why this will close the linked issue? It seems like this will just allow a timeout to happen, not a previously failing call to succeed.

LucioFranco:
@lukesteensen This closes the issue because when `InstanceMetadataProvider` attempts to fetch the credentials it sets up a timer to time out the http request. This timer is not wrapped in any runtime and thus bails out because there is no tokio timer set in the TLS. I was able to reproduce the error on an ec2 instance with an IAM role and once wrapping the future and using `Runtime::block_on` I was able to fetch the credentials. What I noticed was that when calling the specific provider manually I was able to get the tokio error out of it but it seems that rusoto hides this and just spits out a generic error when using the `ChainProvider`.

This type of issue with the missing timer is a very common one so its just the fact that we were using `Future::wait` which is an issue if you're trying to use any of the tokio items.

src/sinks/elasticsearch.rs
```diff
@@ -130,13 +130,17 @@ impl ElasticSearchCommon {
                 if region.is_none() {
                     return Err(ParseError::AWSRequiresRegion.into());
                 }
-                Some(
-                    DefaultCredentialsProvider::new()
-                        .context(AWSCredentialsProviderFailed)?
-                        .credentials()
-                        .wait()
```
lukesteensen: I'm pretty sure we should just swap out `wait()` for `sync()` here.

LucioFranco: Not available with the https://docs.rs/rusoto_credential/0.41.1/rusoto_credential/trait.ProvideAwsCredentials.html trait.

lukesteensen: Oh. Weird that they're not using `RusotoFuture`. Carry on.

PR #1114, opened 2019-10-29
chore(operations): Update Rusoto to 0.41.x

### LINKED ISSUE #1103: Update `rusoto` version
We are currently a few versions behind the current rusoto release. We should update to the latest version. This seems simple but they reworked their error's in `0.38` and thus we can't upgrade without changing a bunch of code. So this task is a bit more mechanical.

phyber:
Closes #1103 

This PR updates Rusoto to 0.41.x. The x is because the credentials crate is at 0.41.1, while the rest are at 0.41.0.

There was a single easily dealt with type change during this update.

### DIFF
--- a/src/sinks/aws_kinesis_streams.rs
+++ b/src/sinks/aws_kinesis_streams.rs
@@ -1,4 +1,3 @@
-use bytes::Bytes;
 use crate::{
     buffers::Acker,
     event::{self, Event},
@@ -9,6 +8,7 @@ use crate::{
     },
     topology::config::{DataType, SinkConfig},
 };
+use bytes::Bytes;
 use futures::{stream::iter_ok, Future, Poll, Sink};
 use rand::random;
 use rusoto_core::{RusotoError, RusotoFuture};
--- a/src/sinks/aws_kinesis_streams.rs
+++ b/src/sinks/aws_kinesis_streams.rs
@@ -1,3 +1,4 @@
+use bytes::Bytes;
 use crate::{
     buffers::Acker,
     event::{self, Event},
@@ -243,6 +244,8 @@ fn encode_event(
             .unwrap_or_default(),
     };
 
+    let data = Bytes::from(data);
+
     Some(PutRecordsRequestEntry {
         data,
         partition_key,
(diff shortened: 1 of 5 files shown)


src/sinks/aws_kinesis_streams.rs
```diff
@@ -243,6 +244,8 @@ fn encode_event(
             .unwrap_or_default(),
     };
 
+    let data = Bytes::from(data);
```
LucioFranco: oh cool! they updated to use bytes! https://docs.rs/rusoto_kinesis/0.41.0/rusoto_kinesis/struct.PutRecordsRequestEntry.html#structfield.data woohoo!

src/sinks/aws_s3.rs
```diff
@@ -642,7 +642,7 @@ mod integration_tests {
             Ok(_) | Err(RusotoError::Service(CreateBucketError::BucketAlreadyOwnedByYou(_))) => {}
             Err(e) => match e {
                 RusotoError::Unknown(b) => {
-                    let body = String::from_utf8(b.body.clone()).unwrap();
+                    let body = String::from_utf8(b.body.clone().to_vec()).unwrap();
```
LucioFranco: pretty minor but we probably want to use `String::from_utf8_lossy(&b.body[..])` here.

phyber:
Should be good to go!

src/sinks/elasticsearch.rs
```diff
@@ -262,7 +262,7 @@ fn es(
                     // to play games here
                     let body = request.payload.take().unwrap();
                     match body {
-                        SignedRequestPayload::Buffer(body) => builder.body(body).unwrap(),
+                        SignedRequestPayload::Buffer(body) => builder.body(body.to_vec()).unwrap(),
```
LucioFranco: we should apply that same trick above here.

phyber: Spoke too soon, on it :)

phyber: It doesn't appear that the same trick works here.

lukesteensen:
Thanks so much for this @phyber!

PR #1115, opened 2019-10-29
feat(config): Use inventory for plugins

Jeffail:


src/topology/config/component.rs
```diff
@@ -0,0 +1,180 @@
...
+            None => Err(Error::custom(format!(
+                "unrecognized type '{}'",
+                swap_out.type_str,
+            ))),
+        }
+    }
+}
+
+/// Constructs a plugin instance from a toml::Value of its configuration.
+type ComponentFromValue<T> = fn(Value) -> Result<T, String>;
```
lukesteensen: For an initial version, we could probably just use the normal derived `Deserialize` implemention to parse the remaining fields (excluding `type`). Or are there other things you had in mind here?

src/topology/config/component.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// Returns a configuration example of a plugin as a toml::Value.
+type ComponentExampleValue = fn() -> Result<Value, String>;
```
lukesteensen: It'd be nice if we could derive the examples from the config metadata we already have for building documentation instead of writing code to generate one.

src/topology/config/component.rs
```diff
@@ -0,0 +1,180 @@
...
+impl ConfigSwapOut {
+    pub fn new() -> ConfigSwapOut {
+        ConfigSwapOut {
+            type_str: "".to_owned(),
+            nested: Value::Table(BTreeMap::new()),
+        }
+    }
+}
+
+impl<'de> Visitor<'de> for ConfigSwapOut {
```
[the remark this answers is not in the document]
Jeffail: Probably a remnant of my smashing-my-head-against-the-compiler phase. Deleted.

src/conditions/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+use inventory;
+
+pub mod static_value;
+
+pub type ConditionDefinition = ComponentBuilder<Box<dyn Condition>>;
+
+pub type ConditionConfig = ComponentConfig<Box<dyn Condition>>;
+
+pub trait Condition {
+    fn check(&self, e: &Event) -> Result<bool, String>;
```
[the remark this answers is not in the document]
Jeffail: Yeah, just a new component I added as a sandbox. The idea would be for transforms, sources and sinks to work the same if we decided to go down this route.

src/topology/config/component.rs
Jeffail: Yeah I'm going to remove the example config part, I added it as a sort of showcase for stuff we can add but we don't need it right now. If we introduce this stuff later we can make it opt-in for plugins as they register themselves.

src/topology/config/component.rs
Jeffail: I was looking to support a way of separating the configuration struct from the actual implementation of the plugin. Registering a closure constructor allows you to choose whether/how they come together at the plugin level, but since it adds boilerplate I think it's fair to restrict this. I might try that out on a separate branch at first.

Jeffail:
I've removed some of the extras that weren't necessary and attempted to simplify the registering part (removed the need to define the conversion from `toml::Value` into `T`) but I'm still left with this implementation of `Into`: https://github.com/timberio/vector/pull/1115/files#diff-0fd923dfda75cfa872fa276b635c9ffcR14

Would be interested to hear if there's a way around that. The reason I'm boxing here is because inventory needs a sizable type, if there's a way around that then I'm down to let it get weird.

src/topology/config/component.rs
```diff
@@ -0,0 +1,180 @@
...
+            Ok(ConfigSwapOut {
+                type_str: type_str,
+                nested: Value::Table(nested),
+            })
+        }
+    }
+}
+
+/// Stores both a constructed plugin instance and the ConfigSwapOut used to
+/// create it, this allows us to echo back the config of a component at runtime.
```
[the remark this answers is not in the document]
Jeffail: I was thinking this would sit under `topology::config::Config` to replace the existing config structs (`TransformOuter` etc) and then the underlying `component` gets pulled out after successful parsing.

src/topology/config/component.rs
```diff
@@ -0,0 +1,113 @@
...
+        let mut types = Vec::new();
+        for definition in inventory::iter::<ComponentBuilder<T>> {
+            types.push(definition.name);
+        }
+        types.sort();
+        types
+    }
+}
+
+impl<T> Serialize for ComponentConfig<T>
```
lukesteensen: Why do we need `Serialize` and `Deserialize` for `ComponentConfig`? It feels like it'd be cleaner to drop these impls and just serialize/deserialize the config itself when needed.

Jeffail: We could do, my goal was to hide away the conversion to/from `toml::Value` so we don't have to write it out for each component type, but if it's not saving us much effort then I'm happy to do away with it.

I think it makes sense for me to make a complementary branch that demonstrates converting one of the existing component types to this method. That'll give us a chance to see it in action and decide if we need `ComponentConfig` at all.

Jeffail:
@lukesteensen I had a stab at converting all transforms to use this PR: https://github.com/timberio/vector/commit/cd97fa477dd41712eb3ed72afc1aeb4b0211c298

There's a bit of type-fu in `topology/config/mod.rs` needed but most of it is a pretty straight forward translation. Want me to try it without `ComponentConfig`?

Jeffail:
I had a stab at working this out without `ComponentConfig`: https://github.com/timberio/vector/commit/201c2285e99f784d63ec0536b7b867d788626bd7

I'm not particularly attached to either approach. I'm probably leaning closer towards having `ComponentConfig` because it pushes more checks into within `Deserialize`, which means if we're playing with segments of config later on (parsing and executing an isolated `TransformOuter`, for example) then we don't have to replicate the logic within `Config::load`.

I'm a fan of:

```rust
let foo = toml::from_str::<TransformOuter>(r#"
type = "not valid"
value = false
 "#);
```

Returning an error without requiring manual checks.

Jeffail:
I can definitely live with the idea of using both. The features we gain from `ComponentDefinition` are purely nice-to-haves so there's low impact on users if/when plugin implementors forget to add it. I'm not sure how we'd ever catch it at compile/test time, but if we end up using this functionality to partially generate documentation then it'll be obvious if it happens.

I'll update this PR to only include `ComponentDefinition`, we can always revisit this again at any time in the future.

Jeffail:
Ok I've removed everything except for `ComponentDescription<T>` which allows us to register plugins and it'll make a best attempt at deriving an example config.

For plugin configs that implement `Default` it will use that, otherwise it'll try and parse an empty string into the type and serialize the result. Nothing yet depends on the example generation but it's going to be one of the next tasks I pick up.

src/topology/config/component.rs
```diff
@@ -0,0 +1,75 @@
+use inventory;
+use serde::{Deserialize, Serialize};
+use std::marker::PhantomData;
+use toml::Value;
+
+/// Describes a component plugin storing its type name, an example config, and
+/// other useful information about the plugin.
+pub struct ComponentDescription<T: Sized> {
+    pub type_str: &'static str,
+    example_value: fn() -> Result<Value, String>,
```
lukesteensen: Why a `fn` here instead of just a `toml::Value`? I'm also curious why it would be fallible.

src/topology/config/mod.rs
```diff
@@ -160,6 +162,25 @@ pub trait TransformConfig: core::fmt::Debug {
     fn transform_type(&self) -> &'static str;
 }
 
+pub struct BoxTransformConfig {
+    pub inner: Box<dyn TransformConfig>,
+}
```
lukesteensen: I'm not totally clear why we need this type.

src/topology/config/component.rs
Jeffail: I figured it would be preferred over storing a serialized example for all components in memory, but I'm not attached so I'm happy to pre-bake it if you prefer.

Since the fallback way of generating an example is to parse an empty string into the type there's a potential for erroring out due to missing fields. I'm pretty sure serde doesn't have a way of enforcing the presence of a field during deserialize so I don't think it's currently possible to fail at that point, but I figured I'd propagate it just in case so we can expose it.

lukesteensen: Do you think that in the long term these will alway be autogenerated? Or would we eventually provide examples when adding new components as a part of documenting them?

If fallible autogeneration is the intent, I can see the argument for keeping a `Result` to show any problems. Otherwise, I think it'd make sense to reduce this to just an `Option<toml::Value>`. That way it would be present for components that have a good default or provided their own example, and simply not present for those that don't.

src/topology/config/mod.rs
Jeffail: This is a little bit of a headache but for inventory we need a sizeable type, so we need `Box<dyn Foo>`. We also want to be able to register the plugin config type itself rather than a boxed trait as that allows us to access `Serialize`, etc, in our `new` implementation.

However, rust can't infer the implementation of `Into` for a `T` into a boxed trait as it requires two jumps, the first being `T` into `dyn Foo`, the second being `dyn Foo` into `Box<dyn Foo>`, that means we can't just simply register `FooDescription::new::<BarConfig>("bar")`. But we could probably settle for something like `FooDescription::new::<Box<dyn Foo>, BarConfig>("bar")`.

I was aiming to shorten the per-plugin boilerplate, which we can do by defining  and registering `BoxTransformConfig` and implementing a generic `BoxTransformConfig::From<T>` which covers three steps, that allows us to use just `FooDescription::new::<BarConfig>("bar")`.

Although thinking about it, if it's possible to alias a function we can probably do something like:

`type RegisterFoo<T> = FooDescription::new::<Box<dyn Foo>, T>;`

Followed by lots of `RegisterFoo::<BarConfig>("bar")`. Can't check right now but if it's possible then I prefer it.

src/topology/config/component.rs
```diff
@@ -0,0 +1,75 @@
...
+            component_type: PhantomData,
+        }
+    }
+
+    /// Returns an example config for a plugin identified by its type.
+    pub fn example(type_str: &str) -> Result<Value, String>
+    where
+        T: 'static + Sized,
+        inventory::iter<ComponentDescription<T>>:
+            std::iter::IntoIterator<Item = &'static ComponentDescription<T>>,
```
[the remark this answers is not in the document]
Jeffail: It can be moved up, will do.

src/topology/config/mod.rs
Jeffail: Actually scratch that, we aren't using `T` into `Box<dyn Foo>` anymore because we aren't deserializing into it. I can get rid of the `Into<T>` trait bounds and remove `BoxTransformConfig`. I'll update the PR tomorrow morning.

src/topology/config/component.rs
Jeffail: My intention is for all configs to at least implement `Default`, and in some cases where the config structure is more complex provide a curated example. I doubt I'll manage to get all configs on `Default` right now so we need an interim but I can live with `Option`. I'm not opposed to simply returning an empty `Value::table` either.

I might also rename `new` and `new_with_default` to `new_without_default` and `new` respectively. Seems more progressive to shame the non-defaulters. We can then add `new_with_example` for configs with a curated example.

Jeffail: I've updated to return an `Option<Value>`, there's not really much point even attempting the parse of empty strings since I don't think there are any configs that would allow it but I've left it in for now.

Eventually I'll start adding curated config examples, we can support that by adding builder pattern methods to `ComponentDescription<T>` with things like `.with_example` and maybe `.with_summary`.

lukesteensen: Sounds great!

PR #1123, opened 2019-10-30
fix(observability): Improve topology tracing spans

LucioFranco:
This change improves the output of tracing spans. Span's now include the source/transform/sink type.

image

This internally refactors our topology `Task` type to include metadata.

### DIFF
--- a/src/topology/builder.rs
+++ b/src/topology/builder.rs
@@ -1,4 +1,7 @@
-use super::fanout::{self, Fanout};
+use super::{
+    fanout::{self, Fanout},
+    task::Task,
+};
 use crate::buffers;
 use futures::{
     future::{lazy, Either},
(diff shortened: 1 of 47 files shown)


lukesteensen:
[Review topology tracing spans](https://trello.com/c/yAfjqMMg/2-review-topology-tracing-spans)

src/topology/builder.rs
```diff
@@ -111,6 +111,8 @@ pub fn build_pieces(config: &super::Config) -> Result<(Pieces, Vec<String>), Vec
     for (name, source) in &config.sources {
         let (tx, rx) = mpsc::channel(1000);
 
+        let typetag = source.source_type();
```
lukesteensen: What made you decide not to get these from `typetag`?

LucioFranco: Well we have SinkOutter etc that trait objectized really early and not allowing us to call the inhernt method. So I figured this was easier.

PR #1124, opened 2019-10-30
chore(operations): Update grok to version 1.0.1

daschl:
I've released 1.1.0 of `grok`, I think you want to use it!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,7 +117,7 @@ lazy_static = "1.3.0"
 rlua = { git = "https://github.com/timberio/rlua" }
 num_cpus = "1.10.0"
 bytesize = "1.0.0"
-grok = "1.0.1"
+grok = "~1.0.1"
 nom = "5.0.0"
 uuid = { version = "0.7", features = ["serde", "v4"] }
 exitcode = "1.1.2"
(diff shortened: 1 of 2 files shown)


a-rodin:
Hi @daschl, thank you!

However, I'm somewhat reluctant about adding `onig 5.x.x` as a dependency to Vector yet because it requires `libclang` and `libllvm` to be installed system-wide until https://github.com/rust-onig/rust-onig/pull/126 is merged or https://github.com/rust-lang/rust-bindgen/issues/918 is resolved in some way. So I would prefer to depend on `onig` 4.3 for now, but just not use the unstable method `Regex::capture_names` from `onig` 4.x which doesn't work on 32-bit architectures.

From my tests, except `Regex::capture_names` method, `onig` 4.x can be built and works fine on a variety of targets, including `armv7-unknown-linux-musleabihf` and `x86_64-pc-windows-gnu`.

daschl:
@a-rodin released a 1.0.1, hope this should help

Cargo.toml
```diff
@@ -117,7 +117,7 @@ lazy_static = "1.3.0"
 rlua = { git = "https://github.com/timberio/rlua" }
 num_cpus = "1.10.0"
 bytesize = "1.0.0"
-grok = "1.0.0"
+grok = "1.0.1"
```
a-rodin: I think it should be `grok = "~1.0.1"` because by default Cargo pulls version 1.1.0.

a-rodin: I've added these changes to the branch, as soon as the tests pass, I'll merge.

PR #1135, opened 2019-11-02
fix(journald source): Flush and reset any current filter before applying new filter

dbhowell:
As per systemd man: `sd_journal_flush_matches()` may be used to flush all matches, disjunction and conjunction terms again. After this call all filtering is removed and all entries in the journal will be iterated again.

So reset the filter and then apply the `_BOOT_ID` filter.

Discussion in #1081

### DIFF
--- a/lib/journald/src/lib.rs
+++ b/lib/journald/src/lib.rs
@@ -159,13 +159,13 @@ impl Journal {
 
         // Journald does not guarantee that the following records are
         // only from the current boot, so a filter is also needed.
+        self.lib.sd_journal_flush_matches(self.journal);
         let filter = format!("_BOOT_ID={}", boot_str);
         sd_result(self.lib.sd_journal_add_match(
             self.journal,
             filter.as_ptr() as *const c_void,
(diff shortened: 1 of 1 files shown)


bruceg:
Obviously I have misunderstood the man page (and misread the systemd source, which I was looking at as I wrote this). However, given the correct reading, I don't think calling `sd_journal_flush_matches` is actually necessary. Vector does not set any journald filters other than this one, so there would never be anything to flush. Do I understand right, or am I still missing something?

dbhowell:
Correct, as long as there is no way to re-use an open journal then you wouldn't need the flush.

lib/journald/src/lib.rs
```diff
@@ -159,13 +159,13 @@ impl Journal {
 
         // Journald does not guarantee that the following records are
         // only from the current boot, so a filter is also needed.
+        self.lib.sd_journal_flush_matches(self.journal);
```
bruceg: As per our previous conversation, this is unneeded.

PR #1144, opened 2019-11-04
chore(topology): Pass `TaskExecutor` to transform

### LINKED ISSUE #1069: Add joinable transforms
# Motivation

As we further improve vector we would like to be able to "enrich" our data events with external metadata that may not be available at collection time in the source. Examples of this are enriching events based on the ec2 instance they come from, the kube pod they derive from or which ECS task they are apart of.

# Proposal

To accomplish this, we should provide the ability for our transforms to apply what we call a "stream table join" where we are able to join some external dataset onto our events. At a basic level Vector's transform trait already supports this as can be seen with the [`add_fields`] transform. 

Currently, our transform trait looks like this:

```rust
pub trait Transform: Send {
    fn transform(&mut self, event: Event) -> Option<Event>;

	fn transform_into(&mut self, output: &mut Vec<Event>, event: Event) {...}
}
```

This trait is then consumed via a `futures::stream::Map` stream combinator in the [topology transform builder]. This works well for transforms that already contain all their data at build time but this does not support dynamically fetching metadata over time. Examples of this are watching for pod metadata updates via the kube api or fetching new labels from ec2 metadata. These are items that may change over the course of running vector without the transform getting rebuilt.

To solve this we must provide the ability to run a background task that will be able to poll/collect new metadata and update some state (think hashmap or vec). This state can be updated via a simple `RwLock` that can be passed into the transform. The transform then can just acquire a short read lock to check the state and apply the metadata. The missing piece that we have here is the lack of an executor passed into `TransformConfig::build` which means we must lazily spawn the background task. While this works this is a poor pattern and it would be nicer to support this use case directly via the `TransformConfig` trait.

The change I suggest we make is to pass the tokio executor into `TransformConfig::build` to allow us to spawn a background task that can update some `RwLock` over time based on some external IO.

```rust
    fn build(&self, exec: tokio::executor::Executor) -> crate::Result<Box<dyn transforms::Transform>>;

}
```

This would then allow us to implement a transform like so:

```rust
pub struct MyTransformConfig {}

pub struct MyTransform {
    state: Arc<RwLock<HashMap<Atom, Bytes>>>,
}

impl TransformConfig for MyTransformConfig {
    fn build(&self) -> crate::Result<Box<dyn Transform>> {
		let state = Arc::new(RwLock::new(HashMap::new()));
		
		// The `Background` struct is some `Future` that can
		// loop and poll some external resource, either via a timer
		// ticker or via some "watch" type api. It will then
		// acquire a write lock on the state and update it.
		//
		// Note: we must ensure that the write lock is only acquire for 
		// a very short period to avoid other tasks locking up the runtime
		// or blocking the transform from operating.
		let bg = Background::new(state.clone());

		exec.spawn(bg);

        Ok(Box::new(MyTransform { state }))
    }

	// == snip ==
}

impl Transform for MyTransform {
    fn transform(&mut self, mut event: Event) -> Option<Event> {
        let state = self.state.read().expect("lock poisoned");
		
		update_event(&mut event, state);

		Some(event)
    }
}
```

This implementation gives us extreme freedom in how we might want to implement dynamic transforms in the future. This design was derived from how the `tokio_sync::watch` is implemented and how `fluentd` implements its stream table join transforms.

Even though this specification is long and the solution is very simple this also provides a correct framework for implementing this type of transform with a tokio/rust based futures executor.

cc @lukesteensen 

[`add_fields`]: https://github.com/timberio/vector/blob/master/src/transforms/add_fields.rs
[topology transform builder]: https://github.com/timberio/vector/blob/master/src/topology/builder.rs#L79

LucioFranco:
This passes the `runtime::TaskExecutor` to each `TransformConfig::build` to allow transforms to implement async joins.

```rust
pub struct MyTransformConfig {}

pub struct MyTransform {
    state: Arc<RwLock<HashMap<Atom, Bytes>>>,
}

impl TransformConfig for MyTransformConfig {
    fn build(&self, exec: TaskExecutor) -> crate::Result<Box<dyn Transform>> {
		let state = Arc::new(RwLock::new(HashMap::new()));
		
		// The `Background` struct is some `Future` that can
		// loop and poll some external resource, either via a timer
		// ticker or via some "watch" type api. It will then
		// acquire a write lock on the state and update it.
		//
		// Note: we must ensure that the write lock is only acquire for 
		// a very short period to avoid other tasks locking up the runtime
		// or blocking the transform from operating.
		let bg = Background::new(state.clone());

		exec.spawn(bg);

        Ok(Box::new(MyTransform { state }))
    }

	// == snip ==
}

impl Transform for MyTransform {
    fn transform(&mut self, mut event: Event) -> Option<Event> {
        let state = self.state.read().expect("lock poisoned");
		
		update_event(&mut event, state);

		Some(event)
    }
}
```

Closes #1069 


lukesteensen:
I like the idea of collapsing this and #1118 into a single struct we pass in. I don't see any reason we couldn't extend `GlobalOptions` to be something more generic like `System` or `Context` (other than coming up with a good name).

LucioFranco:
@lukesteensen I'd like to get this merged so that we can implement joinable transforms. For the context type, I'd like to expirement with one just for sink configs so that these two PRs are independent.

Jeffail:
I like the idea of a context type as it'll make it less of a headache (and backwards compatible) to make additions later on. There's likely going to be an addition coming with the conditions work on the horizon as well. We don't necessarily have to include other fields within it for this PR, but just having the wrapper added now will save us some effort later on.

LucioFranco:
@Jeffail right, I think context is a good idea, but I want to avoid a giant grab bag context type. I believe there should be different context types for sources, sinks and transforms. The first step is done with `SinkContext` in #1306. I don't think it makes sense to add a context type when we only have one item that is passed in now.

PR #1159, opened 2019-11-08
chore(operations): upgrade to rust 1.39.0

lukesteensen:


a-rodin:
A question that I wanted to ask some time ago: why is this pinning needed in the first place instead of just specifying minimal required Rust version in the docs? I mean, as I understand, Rust should be backward-compatible, so we don't have to require users of next stable releases or nightly to use exactly this version instead of a newer one with which potentially can generate code better. What

LucioFranco:
@a-rodin mainly to ensure that we are all using the same toolchain, there have been issues in the past. It also allows us to explicitly update the toolchain instead of treating it as an implicit thing.

Issue #1160, opened 2019-11-08
Unintentionally requiring `region` for ES config


lukesteensen:
It appears #1116 caused a surprising regression in how we deserialize the elasticsearch config.

Adding `flatten` seems to have caused the default value of the `region` field to go from `None` to `Some(RegionOrEndpoint::default())`. This then leads to a `MissingRegionAndEndpoint` error when neither of those fields is set, even though a region is not required for ES.

lukesteensen:
Fixed in #1161

PR #1161, opened 2019-11-08
fix(config): stop accidentally requiring region for ES

lukesteensen:
This is an at-least-temporary fix to #1160

### DIFF
--- a/src/sinks/elasticsearch.rs
+++ b/src/sinks/elasticsearch.rs
@@ -131,7 +131,7 @@ impl ElasticSearchCommon {
             format!("Basic {}", base64::encode(token.as_bytes()))
         });
 
-        let region: Option<Region> = match config.region.clone().try_into() {
+        let region: Option<Region> = match (&config.region).try_into() {
             Ok(region) => Some(region),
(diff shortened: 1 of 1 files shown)


src/sinks/elasticsearch.rs
```diff
@@ -128,9 +131,10 @@ impl ElasticSearchCommon {
             format!("Basic {}", base64::encode(token.as_bytes()))
         });
 
-        let region: Option<Region> = match config.region {
-            Some(ref region) => Some(region.try_into()?),
-            None => None,
+        let region: Option<Region> = match config.region.clone().try_into() {
```
bruceg: It's a little ugly, but this can be `(&config.region).try_into()` to avoid the clone.

bruceg:
Note there is another bug with the handling of `host` as well -- when `provider = "aws"`, the hostname is still required, even though it shouldn't be (ref all the other AWS sinks).

lukesteensen:
Good point. Would you mind opening an issue for that? I'd like to get this merged ASAP but we can circle back around and maybe come up with a better solution that covers both of those cases.

bruceg:
Definitely, go ahead.

PR #1172, opened 2019-11-12
fix(grok_parser transform): don't drop parsed field

lukesteensen:
Similar to #1166, the grok parser has a problem handling cases where a parsed field name is the same as the source field name.

### DIFF
--- a/src/transforms/grok_parser.rs
+++ b/src/transforms/grok_parser.rs
@@ -79,14 +79,11 @@ impl Transform for GrokParser {
         let mut event = event.into_log();
         let value = event.get(&self.field).map(|s| s.to_string_lossy());
 
-        let mut drop_field = self.drop_field;
         if let Some(value) = value {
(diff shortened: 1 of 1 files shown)


src/transforms/grok_parser.rs
```diff
@@ -79,10 +79,14 @@ impl Transform for GrokParser {
         let mut event = event.into_log();
         let value = event.get(&self.field).map(|s| s.to_string_lossy());
 
+        let mut drop_field = self.drop_field;
         if let Some(value) = value {
             if let Some(matches) = self.pattern.match_against(&value) {
                 for (name, value) in matches.iter() {
```
bruceg: Could `drop_field` be calculated here as:
```
let drop_field = self.drop_field and !matches.contains_key(&self.field);
```

This would elimiate the `let mut`

lukesteensen: Good call!

bruceg (approved):

PR #1189, opened 2019-11-14
chore: Add map to ServiceBuilder and s3

### LINKED ISSUE #1127: Retries in S3 sink can get new key names
Object key names created by the S3 sink can incorporate a timestamp that means they are not generated deterministically from the batch content. This is fine in isolation but causes a problem when combined with the fact that keys are regenerated when a request retries.

If a request is made to S3 to write a given batch, it will very occasionally return an error when the write actually succeeded. This will lead to a retry, which can cause a new key to be generated. If that retry also succeeds, we will have two objects in S3 with different keys but identical content.

The proper behavior here would be to ensure that all retries are sending identical requests to S3, including the key. This would make our retries idempotent and remove this possibility for duplicate data.

LucioFranco:
This one was pretty hard to get right and there were some error handling weirdness. But in the end I think this does what we want. Unfortunately `PutObjectRequest` does not implement clone so we need to wrap it. This also adds a layer implementation for `TowerRequestSettings`. This also moves all tower utils into a `service.rs` file.

Closes #1127


LucioFranco:
Since the combination of tower request settings got merged this Pr will need to modify that to allow us to put a service at the top of the stack.

src/sinks/aws_s3.rs
```diff
@@ -20,19 +20,15 @@ use rusoto_s3::{
...
 
 #[derive(Clone)]
 pub struct S3Sink {
     client: S3Client,
-    bucket: String,
-    gzip: bool,
-    filename_time_format: String,
-    filename_append_uuid: bool,
-    filename_extension: Option<String>,
+    // bucket: String,
```
lukesteensen: Probably don't need this.

src/sinks/aws_s3.rs
```diff
@@ -132,15 +128,26 @@ impl S3Sink {
...
+                build_request(
+                    req,
+                    filename_time_format.clone(),
+                    filename_extension.clone(),
+                    filename_append_uuid,
+                    compression,
+                    bucket.clone(),
+                )
+            })
+            .settings(request, S3RetryLogic)
```
lukesteensen: It's not very clear what `settings` means here (`request` is also a bad name, but I know that was there before).

LucioFranco: So settings here relate to `TowerRequestSettings` struct, which I also am not a big fan of that name but I am trying to tie it to that.

PR #1230, opened 2019-11-22
chore(operations): Build ARM archives using new `docker-run.sh` script

a-rodin:
The script was introduced in #1226. Using it would allow new changes in the builders to be applied automatically.


a-rodin:
Superseded by https://github.com/timberio/vector/pull/1247.

PR #1231, opened 2019-11-22
chore(operations): Run `test-x86_64-pc-windows-msvc` for releases

a-rodin:
This would run the tests before producing release artifacts.


a-rodin:
It seems like sometimes these tests time out. Because of that I'm not sure that we need to block releases when they fail.

a-rodin:
We already run the tests as part of normal CI process, so doing it specially for releases seems to be not necessary.

PR #1272, opened 2019-11-28
chore(operations): Increase CI output timeout

a-rodin:
It looks like CircleCI executors sometime work slower than usually, which results in broken build jobs for ARM ([example](https://app.circleci.com/jobs/github/timberio/vector/45191)). It is probably caused by CPU being used by other build jobs on the same physical machine, which explains why such failures are non-deterministic.

This PR increases max timeout that CircleCI would wait for output before failing the build from 10 minutes to 30 minutes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,6 +129,7 @@ jobs:
       - run:
           name: Build and test
           shell: bash
+          no_output_timeout: 30m
           command: |
             RUSTFLAGS=-Ctarget-feature=+crt-static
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -188,6 +189,7 @@ jobs:
           docker_layer_caching: true
       - run:
           name: Test
+          no_output_timeout: 30m
           command: cargo test --all --features docker --jobs 4 -- --test-threads 4
```
LucioFranco: while we're at it we can probably remove the thread limit here.

a-rodin: Sounds good to me, as `large` executors have 4 cores too, so this might be redundant.

PR #1313, opened 2019-12-04
fix(cli): Test & validate subcommands without args target default path

### LINKED ISSUE #1310: Enhancement: validate sub command defaults to standard configuration path
I just ran `vector validate` and expected to see some output, but everything ran successfully, no output. All good!

Unfortunately I realised a few minutes later that the config was indeed invalid, and that `vector validate` by default will do nothing.

I think a more useful default, and one that may be safer with newer users, would be to default to the default configuration path if none is provided. I don't believe there's a use-case for running validate against no configuration files, so this seems like it would be beneficial.

Jeffail:
This commit makes the test and validate subcommands target the default config path in the absense of path arguments.

Closes #1310

Signed-off-by: Ashley Jeffs <ash@jeffail.uk>

### DIFF
--- a/src/unit_test.rs
+++ b/src/unit_test.rs
@@ -1,8 +1,8 @@
 use crate::topology::{config::Config, unit_test::UnitTest};
 use colored::*;
+use lazy_static::lazy_static;
 use std::{fs::File, path::PathBuf};
 use structopt::StructOpt;
-use lazy_static::lazy_static;
 
 #[derive(StructOpt, Debug)]
 #[structopt(rename_all = "kebab-case")]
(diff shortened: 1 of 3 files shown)


src/unit_test.rs
```diff
@@ -39,10 +41,20 @@ fn build_tests(path: &PathBuf) -> Result<Vec<UnitTest>, Vec<String>> {
     crate::topology::unit_test::build_unit_tests(&config)
 }
 
+lazy_static! {
+    static ref DEFAULT_PATHS: Vec<PathBuf> = vec!["/etc/vector/vector.toml".into()];
```
lukesteensen: I think we have this in three places now, so it might be good to condense that down to one. With Rust 1.39.0 I don't think we even need `lazy_static` anymore, since `Vec::new` is a const fn.

PR #1321, opened 2019-12-06
fix(operations): Reduce test concurrency to avoid CI failures

a-rodin:
It looks like multiple test threads cause some sporadic `test-stable` failures. This PR sets number of test threads to 1.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -184,7 +184,7 @@ jobs:
       - run:
           name: Test
           no_output_timeout: 30m
-          command: cargo test --all --features docker --jobs 4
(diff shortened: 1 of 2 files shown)


LucioFranco:
So I'd like to see if we can avoid limiting the threads by increasing the sleeps on the tests or refactoring them to be deterministic.

LucioFranco:
looks like sinks one is failing though we should be able to apply what #1322 does to that test as well.

a-rodin:
Superseded by #1322 and https://github.com/timberio/vector/pull/1346.

PR #1344, opened 2019-12-10
feat(file source): Detect and read gzipped files

### LINKED ISSUE #848: Add support for gzipped files to file source

lukesteensen:
This works mostly as expected with no configuration, but there is one caveat. We cannot easily/efficiently seek into compressed files, so we expect to read them entirely before Vector is shut down. If at startup we detect a compressed file for which we've already stored a non-zero file position (i.e. we have already read data from it), we will not attempt to read any more data from the file.

Closes #848
Closes #890


lib/file-source/src/file_watcher.rs
```diff
@@ -19,7 +20,7 @@ use crate::metadata_ext::PortableMetadataExt;
 pub struct FileWatcher {
     pub path: PathBuf,
     findable: bool,
-    reader: io::BufReader<fs::File>,
+    reader: Box<dyn BufRead>,
```
LucioFranco: is there a reason you choose a trait object over a regular generic?

lib/file-source/src/file_watcher.rs
```diff
@@ -49,16 +50,34 @@ impl FileWatcher {
...
-        let file_position = if too_old {
-            rdr.seek(io::SeekFrom::End(0)).unwrap()
+        let (reader, file_position): (Box<dyn BufRead>, FilePosition) = if is_gzipped(&mut reader)?
+        {
+            if file_position != 0 || too_old {
+                // We can't accurately seek into gzipped files without manually scanning through
+                // the entire thing, so for now we simply refuse to read gzipped files for which we
+                // already have a stored file position from a previous run.
+                debug!(
+                    message = "Not re-reading gzipped file with existing stored offset",
```
LucioFranco: do we want to add the position here? Might be useful :man_shrugging:

lib/file-source/src/file_watcher.rs
LucioFranco: Ah I see here you could prob just do `impl Trait` and trait object below or you could be even more fancy and use an `Either` type of trait but that is probably overkill :)

lukesteensen: Yeah, there are actually two different types so an existential wouldn't work. One step fancier would an `Either`-like enum, which I'll probably do if we decide to implement `Seek` somewhere down the line but didn't seem worthwhile right now.

PR #1345, opened 2019-12-10
chore(operations): Build `msi` package for Vector

a-rodin:
Closes https://github.com/timberio/vector/issues/1221.

I'm unsure about one thing in the docs: should we call the package manager "MSI" or "Windows Installer"? It is officially called "Windows Installer", but I had troubles with a package manager name containing spaces, so I've set it to "MSI" for now, which is not the official name, but on the other hand is somewhat consistent with package managers like "RPM".


a-rodin:
Thank you. I've tried to run `msiexec /help` to get the official name, it shows this:

```
Windows ® Installer. V 5.0.14393.2430 

msiexec /Option <Required Parameter> [Optional Parameter]
```

So I think "Windows Installer" is a more correct name, although it looks somewhat lenghty:

image image

PR #1346, opened 2019-12-10
fix(testing): Remove sleeps from topology tests

lukesteensen:
We've been seeing some flakiness in these tests on master, so I took a stab at fixing them. These have passed a few thousand times in a row for me locally, so hopefully CI feels the same way.

Instead of relying on sleeps to ensure that our mock source has a chance to process messages, we make its input an unbuffered channel. This means that the `send` future in our test will not resolve until the source actually has possession of the input, as opposed to it sitting in the channel buffer. This should remove all need to worry about timing for these interactions.

### DIFF
--- a/tests/support/mod.rs
+++ b/tests/support/mod.rs
@@ -29,7 +29,7 @@ pub fn sink_failing_healthcheck() -> (Receiver<Event>, MockSinkConfig) {
 }
 
 pub fn source() -> (Sender<Event>, MockSourceConfig) {
-    let (tx, rx) = futures::sync::mpsc::channel(10);
+    let (tx, rx) = futures::sync::mpsc::channel(0);
     let source = MockSourceConfig::new(rx);
     (tx, source)
 }
--- a/tests/topology.rs
+++ b/tests/topology.rs
@@ -33,13 +33,6 @@ fn into_message(event: Event) -> String {
     event.as_log().get(&MESSAGE).unwrap().to_string_lossy()
 }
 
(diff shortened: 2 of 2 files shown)


tests/support/mod.rs
```diff
@@ -29,7 +29,7 @@ pub fn sink_failing_healthcheck() -> (Receiver<Event>, MockSinkConfig) {
 }
 
 pub fn source() -> (Sender<Event>, MockSourceConfig) {
-    let (tx, rx) = futures::sync::mpsc::channel(10);
+    let (tx, rx) = futures::sync::mpsc::channel(0);
```
LucioFranco: I believe this is still technically `1` slot since each sender gets a slot :( This was something that `tokio_sync::mpsc` fixed.

LucioFranco:
Curious how this differs against my approach of waiting on the channel between each send? https://github.com/timberio/vector/blob/master/tests/topology.rs#L87

lukesteensen:
> Curious how this differs against my approach of waiting on the channel between each send? [/tests/topology.rs@`master`#L87](https://github.com/timberio/vector/blob/master/tests/topology.rs#L87) [](/timberio/vector/blob/HEAD@{2019-12-11T16:40:50Z}/tests/topology.rs#L87)

I'd say it's a simpler change and still seems to work decently well :smile: 

I'm going to merge as this seems like a strict improvement over what we had. If we're still seeing test failures on CI we can do the extra work to wait on sink outputs for each of these tests.

PR #1351, opened 2019-12-11
chore(operations): Add notices for OpenSSL to the license for binary distributions

### LINKED ISSUE #1348: Add OpenSSL notice to the license distributed with Vector's binaries
Our Vector binaries contain statically linked code from the OpenSSL project. Its [license](https://www.openssl.org/source/license-openssl-ssleay.txt) requires to provide the following notices when distributing it:

> This product includes software developed by the OpenSSL Project for use in the OpenSSL Toolkit (http://www.openssl.org/)

and 

> This product includes cryptographic software written by Eric Young (eay@cryptsoft.com)

For Windows binaries it is additionally required to provide this notice:

> This product includes software written by Tim Hudson (tjh@cryptsoft.com)

So, I propose to implement such inclusion in the following way:

1. Create a text file `NOTICES` in the Vector's repository containing all required notices.
2. When building archives with Vector binaries, use concatenation of the current `LICENSE` file and the `NOTICES` file as a new `LICENSE` file. This would ensure that users opening the `LICENSE` file would see all the notices and, in the same time, keep `LICENSE` in the repository without notices, as the Vector's source code repository doesn't contain code from the OpenSSL project.

a-rodin:
Closes #1348.

The name `NOTICE` is recommended by the Apache software foundation: http://www.apache.org/dev/licensing-howto.html.

I'm also thinking about including copyright information

```
Copyright (c) 2019 Vector Authors <vector@timber.io>
```

to the `NOTICE` file, as the `LICENSE` file is a standard Apache license file and doesn't contain any copyright information.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ assets = [
   ["config/examples/*", "/etc/vector/examples/", "644"],
   ["distribution/systemd/vector.service", "/etc/systemd/system/vector.service", "644"]
 ]
-license-file = "target/debian-license.txt"
+license-file = ["target/debian-license.txt"]
 extended-description-file = "target/debian-extended-description.txt"
 
 [workspace]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ assets = [
   ["config/examples/*", "/etc/vector/examples/", "644"],
   ["distribution/systemd/vector.service", "/etc/systemd/system/vector.service", "644"]
 ]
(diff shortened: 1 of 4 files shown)


a-rodin:
I don't fully understand the advertising clause in the OpenSSL license in application to Vector. It requires to add OpenSSL notice when advertising features using SSL. For example, `http` sink uses OpenSSL, but I'm not sure is it advertising, so that we need to include the notice to docs as well, or not.

PR #1387, opened 2019-12-18
chore(operations): Add `kubernetes source` as a scope

ktff:
Needed for #1249 to pass Semantic Pull Request check.

@binarylogic I am assuming this is all that is required to be changed.

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -56,6 +56,7 @@ scopes:
   - tcp source
   - udp source
   - vector source
+  - kubernetes source
 
   # transforms
   - add_fields transform


ktff:
It can't be done like this. So I am closing it.

a-rodin:
The `.github/semantic.yml` is auto-generated from `.github/semantic.yml.erb`, which in turn uses sources from the `.meta` directory to derive valid scopes.

So it looks like the problem is that corresponding file in `.meta/sources` is called `kubernetes.toml.bak` and not `kubernetes.toml`, so that the source could not be detected by `make generate`.

ktff:
@a-rodin That's right, but by my understanding, Ben is still working on it, and by removing the `.bak` it would probably become public which isn't desired.

PR #1388, opened 2019-12-19
feat(new sink): Initial `aws_kinesis_firehose` sink

### LINKED ISSUE #559: New `aws_kinesis_firehose` sink

lukesteensen:
Closes #559 
Closes #1062 

This is another stab at an `aws_kinesis_firehose` sink. It's mostly a copied and modified version of the `aws_kinesis_streams` sinks, with some additional copying from the `elasticsearch` sink test setup (since we configure it to write to ES for the test).

There's probably a good bit of deduplication we could do between the two sinks (and ES tests), but this PR defers all of that in favor of getting something working.

I initially tried to test against an S3 delivery stream, which we should do at some point, but was foiled by some combination of localstack's S3 implementation and rusoto (https://github.com/localstack/localstack/issues/1647).

One thing that's unclear is whether or not we should add an option to encode newlines for delivery streams aimed at S3. I have not looked at how the objects look without it.

External refs: https://github.com/fluent/fluent-bit/issues/2485


binarylogic (approved):
Docs look good

src/sinks/aws_kinesis_firehose.rs
```diff
@@ -0,0 +1,374 @@
...
+};
+use serde::{Deserialize, Serialize};
+use snafu::Snafu;
+use std::{convert::TryInto, fmt, sync::Arc};
+use tower::Service;
+use tracing_futures::{Instrument, Instrumented};
+
+#[derive(Clone)]
+pub struct KinesisFirehoseService {
+    client: Arc<KinesisFirehoseClient>,
```
LucioFranco: Don't think we need to arc this https://docs.rs/rusoto_firehose/0.42.0/rusoto_firehose/struct.KinesisFirehoseClient.html#impl-Clone

src/sinks/aws_kinesis_firehose.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        let config = ElasticSearchConfig {
+            provider: Some(Provider::Aws),
+            region: RegionOrEndpoint::with_endpoint("http://localhost:4571".into()),
+            index: Some(stream.clone()),
+            ..Default::default()
+        };
+        let common = ElasticSearchCommon::parse_config(&config).expect("Config error");
+
+        let client = reqwest::Client::builder()
```
LucioFranco: I kinda wanna get rid of reqwest and just use the elastic libraries serde feature. But we can do that in the future.

LucioFranco (approved):
LGTM :shipit: :fire:

a-rodin:
This looks great! I wrote an suggestion about further improving it in https://github.com/timberio/vector/issues/1407.

PR #1404, opened 2019-12-19
feat(new sink)!: Transform existing 'tcp' sink into 'socket' sink. (#1212)

stbrody:


src/sinks/socket.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct SocketSinkConfig {
+    #[serde(flatten)]
+    pub mode: Mode,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+#[serde(tag = "mode", rename_all = "snake_case")]
+pub enum Mode {
+    Tcp(TcpSinkConfig),
+}
```
bruceg: I think it is worth writing an `impl From<TcpSinkConfig> for Mode`, which would eliminate the need for `Mode::Tcp(TcpSinkConfig)` elsewhere, and allow instead `TcpSinkConfig.into()`

src/sinks/socket.rs
```diff
@@ -0,0 +1,58 @@
...
+        Self {
+            mode: Mode::Tcp(TcpSinkConfig::new(address)),
+        }
+    }
+}
+
+#[typetag::serde(name = "socket")]
+impl SinkConfig for SocketSinkConfig {
+    fn build(&self, cx: SinkContext) -> crate::Result<(super::RouterSink, super::Healthcheck)> {
+        match self.mode.clone() {
```
bruceg: Using `self.mode.as_ref()` should work here to eliminate the need for `clone()`.

src/sinks/util/tcp.rs
```diff
@@ -25,7 +26,7 @@ use tokio_tls::{Connect as TlsConnect, TlsConnector, TlsStream};
 use tracing::field;
 
 #[derive(Debug, Snafu)]
-enum BuildError {
+enum TcpBuildError {
```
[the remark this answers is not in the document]
stbrody: This was conflicting with sinks::BuildError, which is also used in this file, causing a compile failure.  Instead I probably could have imported sinks::BuilderError as SinksBuildError or something, or just referred to it as sinks::BuildError throughout, but this seemed an equally reasonable and straightforward solution.  If you have a preference for one of the three please let me know!

src/sinks/socket.rs
stbrody: Done!

src/sinks/socket.rs
stbrody: as_ref() didn't work, but just adding an & at the beginning to make it a reference seemed to

src/sinks/util/tcp.rs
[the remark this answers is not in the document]
bruceg: That's fair enough. There is nothing magic about the name, just informing about the common naming. I had missed the `sinks::BuildError` import.

PR #1421, opened 2019-12-23
chore: Add .editorconfig

MOZGIII:
Set some common grounds for the file styling rules.
See https://editorconfig.org.

So far only rust (`.rs`) rules are added, we can add more (i.e. `.sh`, `.md`) as we see fit.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.rs]
+indent_style = space
+indent_size = 4


.editorconfig
```diff
@@ -0,0 +1,14 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
a-rodin: I think this is not true for most of our scripts, as they don't contain final newline.

a-rodin:
Thanks for submitting this!

I like the idea, but have a few thoughts:

* For Rust code the style is enforced by the `check-fmt` CI test. In order to pass this test, one typically has to run `cargo fmt` (either manually or automatically via the editor integration).
* The rules used by `cargo fmt` include not only indentation style and tab size, but also some more complex rules related to line breaking, so they are hard to be met manually without `cargo fmt` in any case.
* Also, for Rust there are standard indentation rules (used by `cargo fmt` by default), which are typically set for Rust code by most of the editors automatically by default.

So it seems like for `.rs` source files the lack of style consistency is already not an issue.

For other filetypes, such as `.sh` or `.rb` it is more interesting, but I have feeling that it might be more consistent with the rest of the project to employ the same approach as with the Rust code, e.g. to start from introducing a CI check for validating the style of these files, rather than from the editor support. This is because `.editorconfig` support is opt-in (not all contributors might be using an editor supporting it), while CI check would be mandatory for all new contributions.

MOZGIII:
You're absolutely right about the `.editorconfig` being opt-in. It's possible to introduce the editor config configuration gradually - i.e. we could merge it in first, and then gradually add the customization together with the changes.

This might be a problematic approach though:

- If we introduce a CI check, we'll immediately get it failing, because currently, the files do not comply with the rules (for example scripts currently use 4 spaces, but the config would require 2 spaces).

- When editing a non-complicit file, if the config is picked up by the editor it may cause annoyance if the file in the repo is not formatted according to the config.

To sum up, it's not a good idea to incrementally add the settings as I proposed initially, so I'm now for adding the configuration fully and CI check before the merge.

I can take care of adding the CI check, and look into all the files in the repo and roll out the initial complete config proposal, however, it's a bigger chunk of work I initially anticipated.

I'll have to get approval before I can continue with this.

LucioFranco:
Maybe I'm missing something but what benefit do we get out of this for rust files that `cargo fmt` can't handle?

.editorconfig
MOZGIII: This can be a source of all sorts of weirdnesses. According to the POSIX line definition, every line must have a newline at the end - final newline is no exception.
See 
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline
It's kind of especially important for the scripts since they are the most affected by portability issues. Realistically, even in 2019 we can still stumble upon a problematic interaction with improperly terminated files.
Good thing we can easily fix and enforce this with editor config.

MOZGIII:
`cargo fmt` solves the formatting of the rust code entirely, but we need this to enforce basic formatting rules for everything else.

```editorconfig
[*]
indent_style = space
indent_size = 2
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
```

These are pretty much the default settings one would use for a file. Ruby and JavaScript standard styling rules match these settings for example.

I added special case for Rust since it uses a different styling settings by default, but it is still true that `rustfmt` "own" the styling rules. I just made editor config ruleset compatible.

MOZGIII:
The main need for this configuration is for non-rust files. It's a pretty decent and simple way to take care of the non-essential styling rules (i.e. styling for the non-main-language file of the project). I realized this project doesn't have the editor config when I was adding the script (`signoff-git-hook.sh`). I did a quick overview of the non-rust files and found a lot of issues that are solved by just using editor config - EOFs, trailing whitespaces and so on, inconsistent spacing.

I'm not too sensitive about those things, but I think it'd be valuable to cover what's not covered by `cargo fmt`.

a-rodin:
> I can take care of adding the CI check, and look into all the files in the repo and roll out the initial complete config proposal, however, it's a bigger chunk of work I initially anticipated.

I agree that this a big chunk of work, and I didn't mean that it have to be a part of this PR. If we want to proceed with this idea, first of all we probably need to extract this into a separate issue to discuss it there and agree about the tools and desired styles for use in such CI check. I've opened issue https://github.com/timberio/vector/issues/1423 for this.

> I did a quick overview of the non-rust files and found a lot of issues that are solved by just using editor config - EOFs, trailing whitespaces and so on, inconsistent spacing.

I agree that it would be better if it was consistent, but the opt-in nature of `.editorconfig` still leaves room for such inconsistencies, especially as they are already there.

Also, officially adopting `.editorconfig` would mean that we would need to put a requirement to use it for new contributors in our contribution guidelines, which requires installation of additional third-party plugins for most of the popular editors. I think such requirement might be too much of a burden which might detract some of potential contributions, that's why I'm not very enthusiastic about it. On the other hand, if we don't put requirement to install and use plugins for `.editorconfig` in our contribution guidelines, than what would be the point for using it at all?

MOZGIII:
Well, for users that have editor config plugin installed it will make a difference, and those who don't have it installed are probably not losing anything. Adding `.editorconfig` without enforcing it via CI might still be a good idea after all. It's better than nothing.
Since it's very wide-spread at this point, I'd say it's reasonable to add it even if we don't enforce it. It's used in that fashion in some of the rust core projects, [`miri`](https://github.com/rust-lang/miri/blob/master/.editorconfig) and [`clippy`](https://github.com/rust-lang/rust-clippy) for instance:.

.editorconfig
a-rodin: > It's kind of especially important for the scripts since they are the most affected by portability issues.

Note that most of our shell scripts depend on `bash` shell (as specified in the first line of each script), so they are already not POSIX-portable. `bash` handles scripts without newline in the end without issues, at least the scripts that we use.

I'm not saying that I'm against adding the newline in the end of each shell script, but I think it might make sense to discuss this separately beforehand, as part of https://github.com/timberio/vector/issues/1423 or as a separate issue. And if a decision is made on this, it should be enforced by a CI check.

a-rodin:
It looks like GitHub web editor supports `.editorconfig` file, so using it could make it easier to contribute to the repository from the browser.

MOZGIII:
I suggest we use this for proper linting: https://github.com/jedmao/eclint (used this one before, it works great).
Alternatively, this: https://github.com/editorconfig-checker/editorconfig-checker

a-rodin:
@MOZGIII This would further increase number of development dependencies, thus making it harder for new contributors to be able to run the checks locally.

MOZGIII:
True. Maybe we can just use it in CI, as a non- failing check?

a-rodin:
@MOZGIII A non-failing check would require manual inspection of its results, so the automation would be not complete and it would be easy to forget about it.

PR #1425, opened 2019-12-23
chore(operations): Check for EOLs and trailing spaces in CI

a-rodin:
This PR adds a CI check which ensures that:

1. All text files have line feed as the last character.
2. Only Unix line endings (`\n`) are used instead of the Windows ones (`\r\n`).
3. There are no trailing spaces in the text files.

The check is added as a part of `make check-fmt`. The files which don't pass the check can be automatically fixed using `make fmt` or by invoking `./scripts/check-style.sh --fix` directly.

Ref https://github.com/timberio/vector/pull/1421#discussion_r360954943 and https://github.com/timberio/vector/issues/1423.


LucioFranco:
@a-rodin do you know if `rustfmt` adds these?

a-rodin:
@LucioFranco `rustfmt` both adds `LF` in the end of all files and remove trailing spaces from all files.

It doesn't convert Windows line endings to the Unix ones automatically, but all of the files in the reposutory already have Unix file endings.

PR #1432, opened 2019-12-25
chore(operations): Verify building of the Nix package

a-rodin:
Closes https://github.com/timberio/vector/issues/1430.

This PR is a draft, as the following items need to be done before merging:

- [x] Automatically generate `cargoSha256`.
- [x] Automatically substitute version from `Cargo.toml`.
- [x] Not only build Vector, but also run the tests.

Also, the file `distribution/nix/default.nix` is an ERB template, but it doesn't have `.erb` extension because it can't be generated with `make generate`. I'm not sure is plain `nix` extension for a template confusing or not.


a-rodin:
This is ready for merging.

PR #1457, opened 2019-12-29
enhancement(splunk_hec source)!: Rename `line` field to `messag…

### LINKED ISSUE #1409: `splunk_hec` source rename `line` field to `message`
### Description
When `splunk_hec` source receives event as a `string` it will emit it under `message` field, and it is the same case when using `raw` mode. But when it receives event as an `json` object it will just add it to Event as it is. This causes for the message to be added under different named field. In for us most important case of Docker/ECS , it will be emitted under `line` field.

Renaming `line` to `message` filed would make `splunk_hec` source behave more consistent.

### Behavior
When `splunk_hec` source receives event as an `json` object, rename field `line` to `message`.

ktff:
Closes #1409.

### DIFF
--- a/src/event/flatten.rs
+++ b/src/event/flatten.rs
@@ -1,5 +1,6 @@
-use crate::event::{LogEvent, ValueKind};
+use crate::event::{Atom, LogEvent, ValueKind};
 use serde_json::{Map, Value};
+use std::convert::AsRef;
 
 /// Recursevly inserts json values to event
 pub fn flatten(event: &mut LogEvent, map: Map<String, Value>) {
@@ -9,7 +10,7 @@ pub fn flatten(event: &mut LogEvent, map: Map<String, Value>) {
 }
 
 /// Recursevly inserts json values to event under given name
-pub fn insert(event: &mut LogEvent, name: String, value: Value) {
+pub fn insert<K: Into<Atom> + AsRef<str>>(event: &mut LogEvent, name: K, value: Value) {
     match value {
         Value::String(string) => {
             event.insert_explicit(name, string);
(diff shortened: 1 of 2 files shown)


src/event/flatten.rs
```diff
@@ -1,5 +1,6 @@
-use crate::event::{LogEvent, ValueKind};
+use crate::event::{Atom, LogEvent, ValueKind};
 use serde_json::{Map, Value};
+use std::convert::AsRef;
```
LucioFranco: I believe this is always included in the `std::prelude` so we don't need it here.

src/event/flatten.rs
```diff
@@ -9,7 +10,7 @@ pub fn flatten(event: &mut LogEvent, map: Map<String, Value>) {
 }
 
 /// Recursevly inserts json values to event under given name
-pub fn insert(event: &mut LogEvent, name: String, value: Value) {
+pub fn insert<K: Into<Atom> + AsRef<str>>(event: &mut LogEvent, name: K, value: Value) {
```
LucioFranco: Can you explain why we need to `AsRef` here, `Into<..>` should work just fine? https://docs.rs/string_cache/0.8.0/string_cache/struct.Atom.html#impl-From%3C%26%27a%20str%3E

ktff: In short, for optimization. 

It allows passing in `String` and `str` as name, which is useful for cases of `Value::Array` and `Value::Object` where Atom is not needed, and so having a way to bypass conversion to Atom is more performant. While at the same time, having an option to pass `Atom` as name because you already have it, avoids chain of conversions `Atom->String->Atom` in other `Value` cases as it was before.

src/event/flatten.rs
ktff: True

src/event/flatten.rs
LucioFranco: Not really a big fan of having two traits here but its fine for now.

PR #1458, opened 2019-12-29
enhancement(splunk_hec source): Provide timestamp

### LINKED ISSUE #1415: `splunk_hec` source provide now timestamp if not present
### Description
As stated in https://github.com/timberio/vector/issues/1409#issuecomment-568265213, splunk hec api doesn't require for a timestamp to be present, so we should add it if isn't provided. 

### Behavior
Add a timestamp of now if the timestamp isn't provided.

ktff:
Closes #1415.

### DIFF
--- a/src/sources/splunk_hec.rs
+++ b/src/sources/splunk_hec.rs
@@ -284,8 +284,8 @@ struct EventStream<R: Read> {
     events: usize,
     /// Optinal channel from headers
     channel: Option<ValueKind>,
-    /// Default time, Ok(extracted), Err(now)
-    time: Result<DateTime<Utc>, DateTime<Utc>>,
+    /// Default time
+    time: Time,
     /// Remaining extracted default values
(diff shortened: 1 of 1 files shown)


src/sources/splunk_hec.rs
```diff
@@ -284,8 +284,8 @@ struct EventStream<R: Read> {
     events: usize,
     /// Optinal channel from headers
     channel: Option<ValueKind>,
-    /// Extracted default time
-    time: Option<ValueKind>,
+    /// Default time, Ok(extracted), Err(now)
+    time: Result<DateTime<Utc>, DateTime<Utc>>,
```
LucioFranco: Generally, I don't really like using an error enum in cases where its not really an error. A custom `Time` enum could work here and be a bit more explicit.

ktff: I agree.

PR #1474, opened 2020-01-02
fix(buffers): Re-enable buffering reclaim test

### LINKED ISSUE #1253: Re-enable buffering tests on all platforms
Currently this test is enabled only on Unix platforms (see https://github.com/timberio/vector/pull/1205#issuecomment-557081933):

https://github.com/timberio/vector/blob/f9c486ce4abcd77cf61ddc7fe2fadb4aeae3b806/tests/buffering.rs#L13-L88 

The problem with it is that the first 10 messages which are stored to the disk buffer are consumed, but the last 10 messages are lost, so this assertion fails: https://github.com/timberio/vector/blob/f9c486ce4abcd77cf61ddc7fe2fadb4aeae3b806/tests/buffering.rs#L87

@LucioFranco suggested in https://github.com/timberio/vector/pull/1219#issuecomment-558161354 that resolving some futures on Windows might require reactors or timers. It is possible that problems with this test have the same nature, however I was not able to isolate them because `rt::block_on` call is there in this case.

LucioFranco:
Related to #1417

Closes #1253

### DIFF
--- a/tests/buffering.rs
+++ b/tests/buffering.rs
@@ -175,7 +175,6 @@ fn test_max_size() {
 fn test_max_size_resume() {
     let data_dir = tempdir().unwrap();
     let data_dir = data_dir.path().to_path_buf();
-    let rt = runtime();
 
     let num_lines: usize = 1000;
     let line_size = 1000;
--- a/tests/buffering.rs
+++ b/tests/buffering.rs
@@ -10,7 +10,6 @@ use vector::test_util::{
 use vector::topology::{self, config};
 use vector::{buffers::BufferConfig, runtime, sinks, sources};
 
-#[cfg(unix)]
 #[test]
 fn test_buffering() {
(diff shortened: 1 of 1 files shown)


tests/buffering.rs
```diff
@@ -317,6 +316,7 @@ fn test_reclaim_disk_space() {
         .map(|m| m.len())
         .sum();
 
-    println!("after {}, before {}", after_disk_size, before_disk_size);
-    assert!(after_disk_size < before_disk_size);
+    // Ensure that the disk after is _atleast_ more than half as small
+    // as it was before.
```
lukesteensen: "at least more than half as small" is breaking my brain

LucioFranco: haha, yeah I thought it made sense coming out of my brain but I guess hearing it out it doesn't. Let me fix that.

PR #1497, opened 2020-01-09
docs: Document `endpoint` parameter for `aws_s3` sink

a-rodin:
The `aws_s3` sink supports [either `region` or `endpoint`](https://github.com/timberio/vector/blob/0a11a36f41143f0c7f39096dc7ad19698d70231d/src/sinks/aws_s3.rs#L42) to specify the location of the target bucket. However, Vector's docs mention only the `region` option.

This PR adds the `endpoint` parameter to the documentation.


bruceg:
Does this not in fact apply for all of the AWS sinks?

PR #1505, opened 2020-01-10
chore(operations): Run `test-stable` tests using `x86_64-unknown-linux-musl` target

### LINKED ISSUE #1463: Run `test-stable` job using the `x86_64-unknown-linux-musl` target
Currently the `test-stable` job in CI is executed on `x86_64-unknown-linux-gnu` target, but our release binaries are built for `x86_64-unknown-linux-musl`. It causes issues like #1433 to be undetected by the tests.

So we need to run `test-stable` on `x86_64-unknown-linux-musl` instead.

a-rodin:
Closes #1463.


a-rodin:
The `test-stable` works job, but `test_reclaim_disk_space` fails with the same error as in #1503.

PR #1525, opened 2020-01-14
fix(docker source): Always check containers for self

ktff:
Fixes bug reported in [comment](https://github.com/timberio/vector/issues/984#issuecomment-573683760).

As mentioned in the following [comment](https://github.com/timberio/vector/issues/984#issuecomment-574225167), initial list of containers gotten from docker sometimes doesn't contain container of the vector requesting the list.  

This PR has a fix which checks all encountered containers if they are self/vector, and not only the ones from the initial list.

### DIFF
--- a/src/sources/docker.rs
+++ b/src/sources/docker.rs
@@ -191,6 +191,10 @@ struct DockerSource {
     ignore_container_id: Vec<ContainerId>,
     ///receives ContainerLogInfo comming from event stream futures
     main_recv: UnboundedReceiver<ContainerLogInfo>,
+    /// It may contain shortened container id.
+    hostname: Option<String>,
+    /// True if self needs to be excluded
+    exclude_self: bool,
 }
 
 impl DockerSource {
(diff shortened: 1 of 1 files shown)


src/sources/docker.rs
```diff
@@ -348,12 +318,31 @@ impl DockerSource {
...
 
-                // Sort for efficient lookup
-                self.ignore_container_id.sort();
                 self
             })
             .map_err(|error| error!(message="Listing currently running containers, failed",%error))
     }
+
+    /// True if passed check
+    fn self_check<'a>(&self, id: &str, image: impl Into<Option<&'a str>>) -> bool {
```
LucioFranco: I don't think you need this life time here, also the name of this and the doc don't make that much sense could you clarify what this does by updating the name/docs.

ktff: Compiler is complaining that it's `missing lifetime specifier`.

Maybe this will be clearer.

LucioFranco:
@ktff I think this needs to be rebased against the master branch I think that will fix the windows issue.

PR #1537, opened 2020-01-17
enhancement(splunk_hec sink)!: Add a new option for specifying indexed fields to `splunk_hec` sink

a-rodin:
Closes https://github.com/timberio/vector/issues/1534.

The option is added only to `json` encoding, as `text` encoding was not using `fields` previously.

I decided to not implement dropping the indexed fields from `events` mentioned in #1534 for now as I'm worried about unexpected results for existing users. It can be easily changed.

### DIFF
--- a/src/event/unflatten.rs
+++ b/src/event/unflatten.rs
@@ -46,6 +46,7 @@ impl From<HashMap<Atom, Value>> for Unflatten {
 /// This produces one path down the tree for each key that has
 /// previously been flattened. The goal here is that the return value
 /// of this function will be merged into the overall tree.
+/// h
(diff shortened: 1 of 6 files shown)


src/event/unflatten.rs
```diff
@@ -46,6 +46,7 @@ impl From<HashMap<Atom, Value>> for Unflatten {
 /// This produces one path down the tree for each key that has
 /// previously been flattened. The goal here is that the return value
 /// of this function will be merged into the overall tree.
+/// h
```
lukesteensen: Is this an oops? :smile:

a-rodin: Yes 😄

PR #1540, opened 2020-01-17
feat(new source): Initial `logplex` source implementation

lukesteensen:
Nothing too fancy, just accepts text bodies over HTTP where each line is roughly syslog-formatted. We do some basic parsing but fall back to forwarding the raw line if anything seems off.


binarylogic (approved):
Nice! Docs look good.

src/sources/logplex.rs
```diff
@@ -0,0 +1,190 @@
...
+            .and(warp::header::<String>("Logplex-Drain-Token"))
+            .and(warp::body::concat())
+            .and_then(move |msg_count, frame_id, drain_token, body| {
+                info!(message = "Handling logplex request", %msg_count, %frame_id, %drain_token);
+
+                let events = body_to_events(body);
+
+                if events.len() != msg_count {
+                    error!(message = "Parsed event count does not match message count header", event_count = events.len(), %msg_count);
+                    debug_assert!(false, "Parsed event count does not match message count header");
```
LucioFranco: Can you explain why we need this?

src/sources/logplex.rs
```diff
@@ -0,0 +1,190 @@
...
+                    .map_err(move |_: mpsc::SendError<Event>| {
+                        error!("Failed to forward events, downstream is closed");
+                        // shut down the http server if someone hasn't already
+                        trigger.try_lock().ok().take().map(drop);
+                        warp::reject::custom("shutting down")
+                })
+                .map(|_| warp::reply())
+            });
+
+        let ping = warp::get2().and(warp::path("ping")).map(|| "pong");
```
LucioFranco: is this required or was this just for testing?

src/sources/logplex.rs
```diff
@@ -0,0 +1,190 @@
...
+            .header("Logplex-Drain-Token", "drain-bar")
+            .body(body.to_owned())
+            .send()
+            .unwrap()
+            .status()
+            .as_u16()
+    }
+
+    #[test]
+    fn router_log() {
```
LucioFranco: Would it make sense to test `line_to_event` a bit more? I wonder if we could fuzz or quickcheck at the most extreme too (not now but in the future).

LucioFranco (approved):
This looks great! I left a few questions, nothing really blocking :)

src/sources/logplex.rs
lukesteensen: If we're in this branch, that means we've failed to parse the body correctly. In development, I want that to blow up, fail tests, etc. In production, I want to spit out an error but keep going so that we remain available for users.

src/sources/logplex.rs
lukesteensen: Not required, but it's convenient when setting up infrastructure to make sure the server is reachable. Think of it as a minimum viable healthcheck, since we have no other routes without side effects.

src/sources/logplex.rs
LucioFranco: Would 

```rust
if cfg_is!(test) {
	panic!(...)
}
```
be more clear here than debug_assert? Mostly, so that people know this should panic in tests.

src/sources/logplex.rs
lukesteensen: Yeah, it would. I've deliberately kept it pretty simple for now and provided fallbacks for when assumptions aren't met, but it would definitely be a good thing to fuzz.

In the meantime, I'll push a couple of other simple unit tests for it.

PR #1554, opened 2020-01-20
enhancement(aws_cloudwatch_logs sink): add assume_role

lukesteensen:
This is a minimal implementation of allowing the Cloudwatch Logs sink to assume a given IAM role at startup. I've tested it manually but couldn't come up with a good-enough-to-be-worth-the-effort automated test.

This would be useful functionality for all our AWS-based components, but we can address that in a follow-up change.


LucioFranco (approved):

binarylogic (approved):
Looks good. Let's make sure to open issues for the follow up work of supporting this in all AWS sinks.

PR #1633, opened 2020-01-28
chore(buffers): Introduce `JsonArrayBuffer` buffer

LucioFranco:
This introduces a new buffer type which is designed to be used with
sitiation where you may want to batch events and limit batch sizes by an
amount of bytes but also embed these events in some deeply nested json
object. This leverages `serde_json::value::RawValue` and makes it much
easier to nest events in some json body.

```rust
fn build_request(events: Vec<BoxedRawValue>) -> http::Request<Vec<u8>> {
    let json = json!({
	    "lines": events,
    });

    let body = serder_json::to_vec(&json).unwrap();

    // snip
}
```
This is related to #1434 

### DIFF
--- a/src/sinks/util/buffer/json.rs
+++ b/src/sinks/util/buffer/json.rs
@@ -1,0 +1,103 @@
+use crate::sinks::util::Batch;
+use serde_json::value::RawValue;
+
+type BoxedRawValue = Box<RawValue>;
+
+/// A `batch` implementation for storing an array of serialized
+/// json values.
+///
+/// To ensure that `push` does not panic pair it with the use of
+/// `JsonArray::encode` to produce the correct `Vec<u8>`.
+#[derive(Default, Debug)]
+pub struct JsonArrayBuffer {
+    buffer: Vec<BoxedRawValue>,
+    size: usize,
+}
+
+impl JsonArrayBuffer {
+    /// Encoding via this will ensure that pushing into this batch will
+    // not panic.
+    pub fn encode(value: impl serde::Serialize) -> Result<Vec<u8>, crate::Error> {
+        serde_json::to_vec(&value).map_err(Into::into)
+    }
+}
+
+impl Batch for JsonArrayBuffer {
+    type Input = Vec<u8>;
+    type Output = Vec<BoxedRawValue>;
+
+    fn len(&self) -> usize {
+        self.size
+    }
+
+    fn push(&mut self, item: Self::Input) {
+        self.size += item.len();
+        let item = RawValue::from_string(String::from_utf8(item).unwrap()).unwrap();
+        self.buffer.push(item);
+    }
+
+    fn is_empty(&self) -> bool {
+        self.buffer.is_empty()
+    }
+
+    fn fresh(&self) -> Self {
+        JsonArrayBuffer::default()
+    }
+
+    fn finish(self) -> Self::Output {
+        self.buffer
+    }
+
+    fn num_items(&self) -> usize {
+        self.buffer.len()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::json;
+
+    #[test]
+    fn multi_object_array() {
+        let mut buffer = JsonArrayBuffer::default();
+
+        buffer.push(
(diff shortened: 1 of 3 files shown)


src/sinks/util/buffer/json.rs
```diff
@@ -0,0 +1,103 @@
+use crate::sinks::util::Batch;
+use serde_json::value::RawValue;
+
+type BoxedRawValue = Box<RawValue>;
```
lukesteensen:
```
$ irb
irb(main):001:0> "BoxedRawValue".length
=> 13
irb(main):002:0> "Box<RawValue>".length
=> 13
```
:smile:

src/sinks/util/buffer/json.rs
```diff
@@ -0,0 +1,103 @@
...
+
+/// A `batch` implementation for storing an array of serialized
+/// json values.
+///
+/// To ensure that `push` does not panic pair it with the use of
+/// `JsonArray::encode` to produce the correct `Vec<u8>`.
+#[derive(Default, Debug)]
+pub struct JsonArrayBuffer {
+    buffer: Vec<BoxedRawValue>,
+    size: usize,
```
lukesteensen: Could be a little clearer to name this `total_bytes` or something like that.

src/sinks/util/buffer/json.rs
```diff
@@ -0,0 +1,103 @@
...
+impl JsonArrayBuffer {
+    /// Encoding via this will ensure that pushing into this batch will
+    // not panic.
+    pub fn encode(value: impl serde::Serialize) -> Result<Vec<u8>, crate::Error> {
+        serde_json::to_vec(&value).map_err(Into::into)
+    }
+}
+
+impl Batch for JsonArrayBuffer {
+    type Input = Vec<u8>;
```
lukesteensen: This seems wrong. Why accept raw bytes as opposed to an event or String or `Value` or something like that?

LucioFranco: So we can't do this because for our use case we would need to support  generic `Serialize` types which is not possible with the current trait since `push` is not generic. This would also mean that we have to force all buffers to be generic over `Serialize` which I don't think is a good idea.

The other option is to make JsonArrayBuffer generic over some Serialize type which would then 1) only allow you to pass one concrete type to serialize 2) would force a box if you want more than one type 3) would make this more complex since you would need to somehow name the generic Serialize type in places where you are not directly dealing with it. Being type erased is nice here.

So I decided that the simpler route makes much more sense _if_ we provide a specific encode fn that you should pair this with. Since, we are the only consumers of this I though that this is a fine assumption that is _documented_ and is easy to use.

I do think we can do some more unchecked stuff but didnt think it was directly worth the cost.

src/sinks/util/buffer/json.rs
LucioFranco: I am considering just making the type alias public since that is easier to reach for than `serde_json::value::RawValue` then boxing it.

LucioFranco:
@lukesteensen Im happy to make it a string utf unchecked but I wanted to avoid unsafe fn for now since I don't think we really need it? The cost should be cheap enough and we do spread the load on it since we don't do the entire body but just an event.

src/sinks/util/buffer/json.rs
lukesteensen: Are you planning to use it in other places? Do you just not want to import it in those places? I remain confused.

LucioFranco: Yes, I have a plan to use it in `build_request(events: Vec<BoxedRawValue>)`.

lukesteensen: Could we not use `finish` to make the output such that users don't need to worry about `RawValue`?

LucioFranco: No it needs to be boxed since the type is DST. For encoding we need a boxed version while decoding can use the dynamically sized type behind a reference. https://docs.rs/serde_json/1.0.45/serde_json/value/struct.RawValue.html#method.from_string returns a box as well which we use.

We could newtype it but I don't think that is worth it over this.

PR #1641, opened 2020-01-30
chore(http sink): Refactor internal http sink to use new `HttpS…

LucioFranco:
This refactors the internals to our `HttpService` to provide an easier to use `BatchedHttpSink` which takes some `HttpSink` implementation and some batch implementation and will implement sink that can be returned from a sink config. The `HttpSink` trait provides two methods one to encode events and one to build the overall http request. This is a common pattern among our http sinks and will make it easier to build them with less generics and less trait problems.

Sinks to convert:

- [x] `http`
- [ ] `splunk_hec`
- [ ] `elasticsearch`
- [ ] `datadog_metrics`
- [ ] `clickhouse`

Reference #1434.

## Motivation

The motivation for this change was mainly due to the unwieldiness of generics when it comes to multiple stacked sinks together. While it is nice to use `SinkExt` and a stack of 4-5 sink/service implementations and bundling them together with a builder pattern, it can start to get a bit crazy when adding some more generic things. I started to notice when making `HttpService` generic over the batch output type that it was getting difficult to align the types. This is something I've noticed with tower's layers that the more generics the worse the compiler gets at helping you. Since one of my goals here is to reduce complexity in buildings http based sinks I thought i'd be a good idea to let the user just implement a very easy trait and then tie all the composed sinks together in one inherent method on a very easy to use `BatchedHttpSink`. This should make it much easier to build these sinks and would reduce a lot of code around how we handle optional encoding etc.

cc @lukesteensen

### DIFF
--- a/src/sinks/http.rs
+++ b/src/sinks/http.rs
@@ -4,7 +4,7 @@ use crate::{
     sinks::util::{
         http::{https_client, Auth, BatchedHttpSink, HttpSink},
         tls::{TlsOptions, TlsSettings},
-        Batch, BatchBytesConfig, Buffer, Compression, TowerRequestConfig,
+        Batch, BatchBytesConfig, Buffer, Compression, TowerRequestConfig, UriSerde,
     },
     topology::config::{DataType, SinkConfig, SinkContext, SinkDescription},
 };
@@ -36,9 +36,9 @@ enum BuildError {
 #[derive(Deserialize, Serialize, Clone, Debug, Default)]
 #[serde(deny_unknown_fields)]
 pub struct HttpSinkConfig {
-    pub uri: String,
+    pub uri: UriSerde,
     pub method: Option<HttpMethod>,
-    pub healthcheck_uri: Option<String>,
+    pub healthcheck_uri: Option<UriSerde>,
     pub auth: Option<Auth>,
     pub headers: Option<IndexMap<String, String>>,
     pub compression: Option<Compression>,
@@ -48,13 +48,6 @@ pub struct HttpSinkConfig {
     #[serde(default)]
     pub request: TowerRequestConfig,
     pub tls: Option<TlsOptions>,
-
-    // A parsed uri used once the original uri field has been
-    // parsed to allow access to the uri in `HttpSink::build_request`
-    // without having to have that function parse the uri each time. Users
-    // should not be able to construct this type and it has a default impl.
-    #[serde(skip)]
-    pub uri__: Option<Uri>,
 }
 
 lazy_static! {
@@ -97,7 +90,7 @@ impl SinkConfig for HttpSinkConfig {
 
         let mut config = self.clone();
 
-        config.uri__ = Some(build_uri(&config.uri.clone())?);
+        config.uri = build_uri(config.uri.clone()).into();
         let gzip = match config.compression.unwrap_or(Compression::None) {
             Compression::None => false,
             Compression::Gzip => true,
@@ -187,11 +180,8 @@ impl HttpSink for HttpSinkConfig {
 
         builder.method(method);
 
-        builder.uri(
-            self.uri__
-                .clone()
-                .expect("Inner uri was not set, this is a bug!"),
-        );
+        let uri: Uri = self.uri.clone().into();
+        builder.uri(uri);
 
         match self.encoding {
             Encoding::Text => builder.header("Content-Type", "text/plain"),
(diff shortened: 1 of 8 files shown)


src/sinks/http.rs
```diff
@@ -48,6 +48,13 @@ pub struct HttpSinkConfig {
     #[serde(default)]
     pub request: TowerRequestConfig,
     pub tls: Option<TlsOptions>,
+
+    // A parsed uri used once the original uri field has been
+    // parsed to allow access to the uri in `HttpSink::build_request`
+    // without having to have that function parse the uri each time. Users
+    // should not be able to construct this type and it has a default impl.
+    #[serde(skip)]
+    pub uri__: Option<Uri>,
```
lukesteensen: I'm not a big fan of this pattern. I get that it's more convenient to have a single struct, but the usage becomes less clear.

src/sinks/util/batch.rs
```diff
@@ -101,6 +101,10 @@ where
         Self::build(inner, batch, min_size, min_size, max_linger)
     }
 
+    pub fn from_settings(inner: S, batch: B, settings: BatchSettings) -> Self {
+        Self::new_min(inner, batch, settings.size, settings.timeout.into())
```
lukesteensen: Do we want a minimum size to be the default? This is a little confusing, unless `BatchSettings` somehow implies minimums.

src/sinks/util/http.rs
```diff
@@ -17,20 +21,92 @@ use tower_hyper::client::Client;
...
+    fn encode_event(&self, event: Event) -> Option<Self::Input>;
+    fn build_request(&self, events: Self::Output) -> http::Request<Vec<u8>>;
+}
+
+pub struct BatchedHttpSink<T, B: Batch>
+where
+    B::Output: Clone,
+{
+    sink: Arc<T>,
+    inner: BatchSink<B, TowerBatchedSink<B::Output, HttpRetryLogic, B, HttpService<B::Output>>>,
```
lukesteensen: Are we wrapping two sinks?

src/sinks/util/http.rs
```diff
@@ -17,20 +21,92 @@ use tower_hyper::client::Client;
...
+        batch_settings: BatchSettings,
+        tls_settings: Option<TlsSettings>,
+        cx: &SinkContext,
+    ) -> Self {
+        let sink = Arc::new(sink);
+        let sink1 = sink.clone();
+        let svc = HttpService::new2(cx.resolver(), tls_settings, move |b| sink1.build_request(b));
+
+        let service_sink = request_settings.batch_sink(HttpRetryLogic, svc, cx.acker());
+        let inner = BatchSink::from_settings(service_sink, batch, batch_settings);
```
lukesteensen: This is rather difficult to follow (I know not all these methods are new, just pointing it out). A quick read seems to imply we're creating a batch sink on the first line and then wrapping it with another batch sink on the second line?

src/sinks/util/http.rs
lukesteensen: I understand now that one is actually not a sink, but I will leave the evidence of my initial confusion.

src/sinks/http.rs
LucioFranco: Yeah, the other option is to create a sub struct which I can do, just thought I'd try this out.

src/sinks/util/batch.rs
LucioFranco: We use `batch_with_min` everywhere so I figured this was the norm.

src/sinks/util/http.rs
LucioFranco: Yeah, we need to nest `BatchSink<BatchServiceSink<Service, Batch>>>` but since our service stack is pretty large I moved that to a type alias.

LucioFranco:
I think we could probably combine some our internal batching sinks a bit. The goal of the `BatchHttpSink` was to remove the need to know about all those sub types of sinks. I think maybe as we move forward we can start to prune types?

As for the name for `BatchedHtpSink` I kind of like it being a struct since it allows you to name it easily. I don't think having a method on the `HttpSink` trait would actually be a bit more confusing. I think it would also mean that we need to define more generics on it which I was trying to avoid.

src/sinks/http.rs
LucioFranco: Removed and I added `UriSerde`.

src/sinks/util/uri.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+struct UriVisitor;
+
+impl<'a> Visitor<'a> for UriVisitor {
+    type Value = UriSerde;
+
+    fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "a string containing a HTTP")
```
lukesteensen: Typo?

LucioFranco: HAHAH wow I guess my editor didn't save the rest of my sentence.

PR #1691, opened 2020-02-04
chore(operations): Use `rust-toolchain` file when building CI Docker images

a-rodin:
Closes https://github.com/timberio/vector/issues/1656.

This PR moves `rust-toolchain` file inside `scripts` directory and creates a symlink from the project root to it. It makes it possible to use the `rust-toolchain` file at the stage of building CI Docker images, thus ensuring that these images have the correct version of the toolchain inside them and containers which use them don't need to pull a different toolchain version later.


scripts/rust-toolchain
```diff
@@ -0,0 +1 @@
+1.41.0
```
LucioFranco: Can we just use the one that is in the root?

a-rodin: We would need to put entire project root as in the build context for Docker because Docker can't access parent directories.

LucioFranco: Can we do a sym link or something? or have a script copy it over?

a-rodin: Currently there is a symlink from `rust-toolchain` in the project root to `scripts/rust-toolchain`, so there is no duplication.

LucioFranco: :+1: ok just want to make sure we have one place for it.

LucioFranco: Actually, one last thing is there a way to sym link it the other direction? Right now I see it possibly sym linked (maybe? its not clear to me in this diff) from scripts -> root while I think it would be better to have the main one in the root of the project and symlinked over to scripts.

a-rodin: Docker cannot resolve symlinks which point to files outside of the context because of security reasons. The only way to use the `rust-toolchain` file in the project root is to change the Docker context to the project root and put everything except `scripts` to `.dockerignore`.

LucioFranco: Ok, I am fine with trying the current sym link solution then, we can change it if we run into issues.

PR #1697, opened 2020-02-04
feat(kafka sink): Add support for SASL

Hoverbear:
Direct copy of #1276 to see if this resolves the CI weirdness.


Hoverbear:
Okay now #1276 is fixed so I have no idea.

PR #1706, opened 2020-02-05
fix(networking): Warn log for non retryable req

LucioFranco:

### DIFF
--- a/src/sinks/util/retries.rs
+++ b/src/sinks/util/retries.rs
@@ -94,6 +94,7 @@ where
                     warn!(message = "retrying after response.", %reason);
                     Some(self.build_retry())
                 } else {
+                    warn!("request is not retryable; dropping the request.");
                     None
                 }
             }


LucioFranco:
@lukesteensen so adding a `Debug` would work but I think that makes the `warn` very noisy. I think if a user wants to know the real root problem here they should enable debug mode. Though I am happy to go either way.

The other option is to have `should_retry_request` return a string value for the non retry case.

lukesteensen:
What's the user experience right now? In the case that brought this up, for example, is this output all they'll see? If so, I don't think that's particularly useful, and I don't think saying "turn it up to DEBUG to actually see what's happening" is very empathetic to users.

Using `Debug` would be noisy, I agree. Maybe we either make it convenient to log something from `should_retry_request` or optionally return a  string to be logged? Not sure if there's a big difference between the two.

LucioFranco:
@lukesteensen I opened https://github.com/timberio/vector/pull/1712 to address that. This should provide a much better UX.

PR #1712, opened 2020-02-05
feat(networking): Add proper non retry messages

LucioFranco:
Follow up to #1706 this adds a `RetryAction` enum to allow sinks to specify why they are _not_ retrying a request.

### DIFF
--- a/src/sinks/clickhouse.rs
+++ b/src/sinks/clickhouse.rs
@@ -3,7 +3,7 @@ use crate::{
     event::Event,
     sinks::util::{
         http::{https_client, Auth, HttpRetryLogic, HttpService, Response},
-        retries::RetryLogic,
+        retries::{RetryAction, RetryLogic},
         tls::{TlsOptions, TlsSettings},
         BatchBytesConfig, Buffer, Compression, SinkExt, TowerRequestConfig,
     },
@@ -16,7 +16,6 @@ use hyper::{Body, Request};
 use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
 use snafu::ResultExt;
-use std::borrow::Cow;
 
 #[derive(Deserialize, Serialize, Debug, Clone, Default)]
 #[serde(deny_unknown_fields)]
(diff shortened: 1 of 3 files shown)


src/sinks/clickhouse.rs
```diff
@@ -190,9 +189,9 @@ impl RetryLogic for ClickhouseRetryLogic {
                 //
                 // Reference: https://github.com/timberio/vector/pull/693#issuecomment-517332654
                 if body.starts_with(b"Code: 117") || body.starts_with(b"Code: 53") {
-                    None
+                    RetryAction::DontRetry("incorrect data".into())
```
lukesteensen: Is this the clickhouse terminology? Ideally, a user should be able to google this and pretty quickly figure out what's going on.

src/sinks/util/retries.rs
```diff
@@ -141,6 +148,24 @@ impl<L: RetryLogic> Future for RetryPolicyFuture<L> {
...
+        }
+    }
+
+    pub fn is_not_retryable(&self) -> bool {
+        if let RetryAction::DontRetry(_) = &self {
+            true
+        } else {
+            false
+        }
+    }
```
lukesteensen: Nitpick, but these might be simpler as 

```rust
    pub fn is_retryable(&self) -> bool {
        match self {
            RetryAction::Retry(_) => true,
            _ => false,
        }
    }

    pub fn is_not_retryable(&self) -> bool {
        !self.is_retryable()
    }
```

src/sinks/util/retries.rs
```diff
@@ -1,21 +1,25 @@
...
 pub trait RetryLogic: Clone {
     type Error: std::error::Error + Send + Sync + 'static;
     type Response;
 
     fn is_retriable_error(&self, error: &Self::Error) -> bool;
 
-    fn should_retry_response(&self, _response: &Self::Response) -> Option<Cow<str>> {
-        None
+    fn should_retry_response(&self, _response: &Self::Response) -> RetryAction {
+        RetryAction::DontRetry("unknown".into())
```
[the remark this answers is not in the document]
LucioFranco: I guess we could just state that its `reason = "not implemented"` should be fine?

src/sinks/clickhouse.rs
LucioFranco: Yeah, we could attempt to print the code that we match on, that would be helpful. :+1:

src/sinks/util/retries.rs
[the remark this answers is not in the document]
lukesteensen: That works!

src/sinks/util/retries.rs
lukesteensen: We also already have `derive_is_enum_variant` that could replace all this.

PR #1720, opened 2020-02-05
chore(sematext sink): make host calculation more explicit

lukesteensen:
This is technically a breaking change if we reject configs that we were previously accepting, but I'm not sure we should really care in this case.

### DIFF
--- a/src/sinks/sematext.rs
+++ b/src/sinks/sematext.rs
@@ -35,28 +35,20 @@ pub enum Region {
 #[typetag::serde(name = "sematext")]
 impl SinkConfig for SematextConfig {
     fn build(&self, cx: SinkContext) -> crate::Result<(super::RouterSink, super::Healthcheck)> {
-        let mut host = None;
-
-        if let Some(region) = &self.region {
-            host = match &region {
-                Region::Na => "https://logsene-receiver.sematext.com".to_string().into(),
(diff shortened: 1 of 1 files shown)


LucioFranco:
@bruceg In this case, this is totally fine I think I like this much better since its explicit rather than implicitly doing something.

lukesteensen:
@bruceg 

> My only objection is dependant on the semantics of `host`. If it is an override, as suggested in the original code, then this is incorrect (no reason to check for both set).

Yeah, my intent here was to eliminate the case where the user gives us a configuration value that we're silently ignoring. I don't think it's a huge deal either way, but it could potentially save a headache or two.

---

I guess it's also worth noting a little bit about the general thought process. I don't necessarily find this code super aesthetically pleasing or anything like that. For me, the big benefit is that I know `match` is exhaustive and therefore all the cases must be laid out explicitly. I don't have to use any of my own mental energy thinking about cases not explicitly covered by a chained `if`, cases that fall through, etc, etc. The code ends up a little dense, but with lower cognitive overhead since I can rely on the compiler to check exhaustiveness.

Again, this is bikeshedding and very subjective in a lot of ways. I just figured it was a decent time to show how I think about it in case people find that helpful. A nice side benefit is when it flushes out cases we weren't explicitly handling before and improves the user experience ever so slightly.

PR #1722, opened 2020-02-05
feat(aws provider): Add support for `assume_role` to all AWS sinks

### LINKED ISSUE #1556: Add `assume_role` support to all AWS components
Support was added in #1554 for the Cloudwatch Logs sink to assume different IAM roles. We should generalize that functionality and apply it to all AWS-based components.

This will likely involve factoring out the relevant code from cloudwatch into our rusoto helper module.

bruceg:
Closes #1556


bruceg:
Note: I'm not entirely happy with the `aws_s3` `create_client` function. Should that test access key logic be hoisted up into the shared `rusoto::create_client` function?

binarylogic (approved):
Docs look great

src/sinks/util/rusoto.rs
```diff
@@ -17,3 +26,60 @@ pub fn client(resolver: Resolver) -> crate::Result<Client> {
...
 }
+
+pub trait RusotoNewClient {
+    fn new_client<P>(client: Client, credentials_provider: P, region: Region) -> Self
+    where
+        P: ProvideAwsCredentials + Send + Sync + 'static,
+        P::Future: Send;
+}
+
+pub fn create_client<T: RusotoNewClient>(
```
LucioFranco: I find this trait a bit overkill for this imo. I think it'd be better to let each sink construct their concrete type via `T::new_with` within the sink, then have one fn to return the client and one fn to return a credentials provider. Then from there we can do something like

```rust
pub enum AwsCredentialsProvider {
	Default(DefaultCredentialsProvider),
	AssumeRole(...),
	Static(...),
}

fn client(...) -> crate::Result<Client> { ... }
fn credentials(...) -> crate::Result<AwsCredentialsProvider> { ... }
```

Imo this would be a bit easier to grok in a per sink sense and would avoid a set of macros and a trait.

src/sinks/util/rusoto.rs
```diff
@@ -4,11 +4,20 @@ use hyper_openssl::{
...
+use rusoto_credential::{DefaultCredentialsProvider, ProvideAwsCredentials};
+use rusoto_sts::{StsAssumeRoleSessionCredentialsProvider, StsClient};
+use snafu::{ResultExt, Snafu};
 
 pub type Client = HttpClient<HttpsConnector<HttpConnector<Resolver>>>;
 
-pub fn client(resolver: Resolver) -> crate::Result<Client> {
+#[derive(Debug, Snafu)]
+enum RusotoError {
+    #[snafu(display("{}", source))]
```
LucioFranco:
```suggestion
    #[snafu(display("Invalid AWS credentials: {}", source))]
```

Do we want to add some sort of context here?

LucioFranco (commented):
Overall looks good, I left you a proposal to possible make this a bit easier to read in the future. Happy to go either way :)

src/sinks/util/rusoto.rs
bruceg: That was cloned from the `aws_cloudwatch_logs` sink. I agree it should have more context. I'll add it.

src/sinks/util/rusoto.rs
bruceg: It took a little more work this way, but the approach makes sense and is more composable. I'll see about adding the static provider to avoid that block in the S3 sink.

src/sinks/util/rusoto.rs
```diff
@@ -49,6 +54,13 @@ impl AwsCredentialsProvider {
             Ok(Self::Default(creds))
         }
     }
+
+    pub fn new_minimal<A: Into<String>, S: Into<String>>(access_key: A, secret_key: S) -> Self {
```
LucioFranco: minor nit: I would probably use `impl Trait` here, I find it much nicer to the eyes.

src/sinks/util/rusoto.rs
```diff
@@ -1,13 +1,97 @@
...
+
+            let creds = AutoRefreshingProvider::new(provider).context(InvalidAWSCredentials)?;
+            Ok(Self::Role(creds))
+        } else {
+            let creds = DefaultCredentialsProvider::new().context(InvalidAWSCredentials)?;
+            Ok(Self::Default(creds))
+        }
+    }
+
+    pub fn new_minimal<A: Into<String>, S: Into<String>>(access_key: A, secret_key: S) -> Self {
```
LucioFranco: Also curious if this should be called `new_static`?

LucioFranco (approved):
LGTM! Left some non blocking clients. :ship: it

src/sinks/util/rusoto.rs
bruceg: I thought about that, but the static provider has both `new_minimal` and a plain `new`, and I thought I'd reflect which one is being called, giving room for a `new_static` to provide the full parameters if needed.

LucioFranco: Ah ok that is fine!

PR #1726, opened 2020-02-06
feat(config): Clarify topology warnings

Jeffail:
Prints topology warnings as... warnings. Also clarifies the message about zero outputs.

Closes https://github.com/timberio/vector/issues/414


Jeffail:
Yeah I like "has no consumers"

PR #1728, opened 2020-02-06
feat(cli): Improve cyclic deps error message

### LINKED ISSUE #321: Show the detected cycle in the cyclic topology error
Instead of doing this as a separate step before building the paths to type check, we should be able to detect cycles as we build the paths (i.e. check if the current node already exists in the current path). Then we'd also have the path containing the cycle on hand to include in the error message.

Jeffail:
This allows us to report to users the exact chain of components that
causes the cycle.

The message now looks like:

`ERROR vector: Topology error: Cyclic dependency detected in the chain [ baz -> foo -> bar -> baz ]`

Closes #321

### DIFF
--- a/src/topology/config/validation.rs
+++ b/src/topology/config/validation.rs
@@ -1,12 +1,8 @@
 use crate::topology::{config::DataType, Config};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 
 pub fn typecheck(config: &Config) -> Result<(), Vec<String>> {
     Graph::from(config).typecheck()
-}
-
-pub fn contains_cycle(config: &Config) -> bool {
-    Graph::from(config).contains_cycle()
 }
 
 #[derive(Debug, Clone)]
@@ -69,12 +65,11 @@ impl Graph {
                 Node::Sink { .. } => Some(name),
                 _ => None,
             })
-            .flat_map(|node| match paths_rec(&self.nodes, node, Vec::new()) {
-                Ok(paths) => paths,
-                Err(err) => {
+            .flat_map(|node| {
+                paths_rec(&self.nodes, node, Vec::new()).unwrap_or_else(|err| {
                     errors.push(err);
                     Vec::new()
-                }
+                })
             })
             .collect();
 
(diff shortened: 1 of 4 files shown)


src/topology/config/validation.rs
```diff
@@ -59,15 +59,32 @@ impl Graph {
...
                 _ => None,
             })
-            .flat_map(|node| paths_rec(&self.nodes, node, Vec::new()))
-            .collect()
+            .flat_map(|node| match paths_rec(&self.nodes, node, Vec::new()) {
+                Ok(paths) => paths,
+                Err(err) => {
+                    errors.push(err);
+                    Vec::new()
+                }
```
lukesteensen: Not a big deal, but this is basically `Result::unwrap_or_else`

src/topology/config/validation.rs
```diff
@@ -77,7 +94,14 @@ impl Graph {
     fn typecheck(&self) -> Result<(), Vec<String>> {
         let mut errors = Vec::new();
 
-        for path in self.paths() {
+        let paths = match self.paths() {
+            Ok(paths) => paths,
+            Err(errs) => {
+                return Err(errs);
+            }
+        };
```
lukesteensen: Can this just be `let paths = self.paths()?;`?

src/topology/config/validation.rs
```diff
@@ -185,17 +209,40 @@ impl From<&Config> for Graph {
...
+        Some(Node::Transform { inputs, .. }) | Some(Node::Sink { inputs, .. }) => {
+            let mut paths = Vec::new();
+            for input in inputs {
+                match paths_rec(nodes, input, path.clone()) {
+                    Ok(mut p) => paths.append(&mut p),
+                    Err(err) => {
+                        return Err(err);
+                    }
+                }
+            }
```
lukesteensen: It _feels_ like there should be a way to do this with iterators still, but I haven't quite figured out what it is.

Jeffail:
It's definitely a very satisfying chunk of red getting rid of it. Sorry for the rebase, a date based `make generate` change was going to cause collisions with master.

PR #1769, opened 2020-02-10
feat(config): Global Default Log Schemas

Hoverbear:
This addresses #1446.

It adds a `log_schema` field to the `config.global`. It goes into `config.global` since this already gets passed to the `SourceConfig::build` event. It also exists in the `$SOURCE_source` function args as this is used in some tests.

Example:

```bash
[hoverbear@obsidian:~/git/vector]$ cat test.toml 
data_dir = '/var/lib/vector/'
dns_servers = []

[log_schema]
host_key = "instance" # default "host"
message_key = "info" # default "message"
timestamp_key = "datetime" # default "timestamp"

[sources.source0]
max_length = 102400
type = 'stdin'
[sinks.sink0]
healthcheck = true
inputs = ['source0']
type = 'console'
encoding = "json"

[sinks.sink0.buffer]
type = 'memory'
max_events = 500
when_full = 'block'


[hoverbear@obsidian:~/git/vector]$ cargo run -- --config test.toml -qqq
# ....
Feb 13 04:24:30.902  INFO vector::topology: Starting sink "sink0"
Feb 13 04:24:30.902  INFO source{name=source0 type=stdin}: vector::sources::stdin: Capturing STDIN
asd
{"info":"asd","datetime":"2020-02-13T12:24:32.632275500Z","instance":"OBSIDIAN"}

```


binarylogic:
The config syntax looks good! But shouldn't the resulting payload in your ouput have both the `datetime` and `info` keys since you didn't explicitly supply those as part of your input?

Hoverbear:
Ok! That clarifies it for me! I wasn't really sure if only the fields the source had configs for would be affected, or all would.

src/topology/config/mod.rs
```diff
@@ -118,6 +121,32 @@ pub enum DataType {
...
 
+#[derive(Debug, Deserialize, Serialize, PartialEq, Clone)]
+pub struct LogSchemaOptions {
+    #[serde(default = "LogSchemaOptions::default_host_key")]
+    pub host_key: Atom,
+    #[serde(default = "LogSchemaOptions::default_message_key")]
+    pub message_key: Atom,
+    #[serde(default = "LogSchemaOptions::default_timestamp_key")]
+    timestamp_key: Atom,
+}
```
bruceg: We've discussed reducing our `pub` exposure in the tech debt issue. Consider if you can reduce any of these to `pub(crate)` or eliminate the `pub` entirely to improve what the compiler is able to check for us.

src/topology/config/mod.rs
```diff
@@ -118,6 +121,32 @@ pub enum DataType {
...
+
+impl Default for LogSchemaOptions {
+    fn default() -> Self {
+        LogSchemaOptions {
+            host_key: event::HOST.clone(),
+            message_key: event::MESSAGE.clone(),
+            timestamp_key: event::TIMESTAMP.clone(),
+        }
+    }
+}
```
bruceg: We have used the [`Derivative` custom derive macro](https://mcarton.github.io/rust-derivative/Default.html) in other places to reduce this boilerplate. It might be useful here as well.

src/topology/config/mod.rs
```diff
@@ -331,6 +361,28 @@ impl Config {
...
+            } else {
+                self.global.log_schema.message_key = with.global.log_schema.message_key;
+            }
+            if self.global.log_schema.timestamp_key != default_log_schema.timestamp_key &&
+                self.global.log_schema.timestamp_key != with.global.log_schema.timestamp_key {
+                errors.push("conflicting values for 'log_schema.timestamp_key' found".to_owned());
+            } else {
+                self.global.log_schema.timestamp_key = with.global.log_schema.timestamp_key;
+            }
+        }
```
bruceg: I'm unclear as to what the conflicting values could come from here. Maybe it would be worth adding a comment to help my ignorance?

src/sources/stdin.rs
```diff
@@ -1,11 +1,12 @@
 use crate::{
-    event::{self, Event},
+    event::Event,
     topology::config::{DataType, GlobalOptions, SourceConfig, SourceDescription},
 };
 use bytes::Bytes;
 use futures::{future, sync::mpsc, Future, Sink, Stream};
 use serde::{Deserialize, Serialize};
 use std::{io, thread, time::Duration};
+use crate::topology::config::LogSchemaOptions;
```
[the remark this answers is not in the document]
Hoverbear: Yes! To clarify:

> This is a draft, it's not ready for review.

Hoverbear:
Thanks for this early review @bruceg  :)

Hoverbear:
@Jeffail I did it the easy way. :)

src/event/mod.rs
```diff
@@ -159,6 +158,37 @@ impl<K: Into<Atom>, V: Into<Value>> FromIterator<(K, V)> for LogEvent {
     }
 }
 
+pub fn log_schema<'a>() -> &'a LogSchema {
+    // TODO: Help Rust project support before_each
+    // Support uninitialized schemas in tests to help our contributors.
+    // Don't do it in release because that is scary.
+    #[cfg(debug_assertions)]
```
Hoverbear: I am not 100% sure on this.

src/event/mod.rs
```diff
@@ -23,9 +24,7 @@ pub mod proto {
 }
 
 lazy_static! {
-    pub static ref MESSAGE: Atom = Atom::from("message");
-    pub static ref HOST: Atom = Atom::from("host");
-    pub static ref TIMESTAMP: Atom = Atom::from("timestamp");
+    pub static ref LOG_SCHEMA: OnceCell<LogSchema> = OnceCell::new();
```
Hoverbear: Should this be in config?

LucioFranco: This doesn't need to be a lazy_static https://docs.rs/once_cell/1.3.1/once_cell/sync/struct.OnceCell.html#method.new since new is a const fn :)

src/event/mod.rs
```diff
@@ -159,6 +158,37 @@ impl<K: Into<Atom>, V: Into<Value>> FromIterator<(K, V)> for LogEvent {
     }
 }
 
+pub fn log_schema<'a>() -> &'a LogSchema {
```
LucioFranco: this life time should be elidable, I think.

src/topology/config/mod.rs
```diff
@@ -332,6 +335,35 @@ impl Config {
         self.global.dns_servers.sort();
         self.global.dns_servers.dedup();
 
+        // If the user has multiple config files, we must *merge* log schemas until we meet a
+        // conflict, then we are allowed to error.
+        let default_schema = event::LogSchema::default();
```
LucioFranco: nit: prob could just do `if with.global.log_schema = LogSchema::default()`

src/event/mod.rs
```diff
@@ -159,6 +158,37 @@ impl<K: Into<Atom>, V: Into<Value>> FromIterator<(K, V)> for LogEvent {
...
+            error!("You are not initializing a schema in this test -- This could fail in release");
+            LOG_SCHEMA.set(LogSchema::default()).ok(); // If this fails it means some other test set it while we were trying to.
+        }
+    }
+    LOG_SCHEMA.get().expect("Schema was not initialized")
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct LogSchema {
+    pub message_key: Atom,
```
LucioFranco: I would rather prefer we made these getter fns rather than pub struct items, what do you think?

Hoverbear: 🤔 I had these private, but a test wanted them public, let me revisit it.

src/topology/config/mod.rs
Hoverbear: We use it later though :o

Cargo.toml
```diff
@@ -143,6 +143,7 @@ evmap = { version = "7", features = ["bytes"] }
 logfmt = "0.0.2"
 notify = "4.0.14"
 once_cell = "1.3"
+getset = "0.1.0"
```
Hoverbear: When you get to import your own library!

Hoverbear:
@LucioFranco Take a look again?

src/sources/syslog.rs
```diff
@@ -434,6 +437,7 @@ mod test {
     }
 
     #[test]
+    #[ignore]
```
LucioFranco: why are these ignored?

Hoverbear: Indeed why!

binarylogic:
Nice work on this! I just wanted to couple doc changes with it. I'll follow up with a doc change PR.

PR #1789, opened 2020-02-14
fix(windows platform): Correct lints and tests

Hoverbear:
As reported in #1412 , Windows builds/tests have lints and errors.

This PR fixes all lints and tweaks a few tests to make it viable for Windows to pass.

* Mark any `unix` only imports in `#[cfg(unix)]`.
* Windows errors if you try to rename a file into a path that already exists. So we now use a simple file rolling procedure in those tests.
* Improves `unwrap` errors into `expect` errors to give some context to failing tests.

### DIFF
--- a/lib/file-source/src/lib.rs
+++ b/lib/file-source/src/lib.rs
@@ -19,7 +19,10 @@ mod test {
     use quickcheck::{Arbitrary, Gen, QuickCheck, TestResult};
     use std::fs;
     use std::io::Write;
+    #[cfg(unix)]
     use std::os::unix::fs::MetadataExt;
+    #[cfg(windows)]
+    use std::os::windows::fs::MetadataExt;
     use std::str;
     // Welcome.
     //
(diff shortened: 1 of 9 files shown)


lib/file-source/src/lib.rs
```diff
@@ -235,13 +245,9 @@ mod test {
...
                 FWAction::RotateFile => {
                     let mut new_path = path.clone();
-                    new_path.set_extension("log.1");
-                    match fs::rename(&path, &new_path) {
-                        Ok(_) => {}
-                        Err(_) => {
-                            assert!(false);
-                        }
-                    }
+                    new_path.set_extension(format!("log.{}", rotation_count));
```
lukesteensen: I'm sure it doesn't matter, just noting that this is a bit different from how `logrotate` works now. Bummer that windows doesn't work that way.

Hoverbear: Yeah, there were some other options but this seemed the safest.

PR #1837, opened 2020-02-18
fix(splunk_hec source): Make `token` option optional

### LINKED ISSUE #1784: Make `splunk_hec` source `token` field optional
### Description
Currently there is a discrepancy between `splunk_hec`documentation and implementation around filed `token`. 

In documentation, `token` field is mentioned as optional while in implementation it's not. 

The confusion probably started from [this comment](https://github.com/timberio/vector/issues/1088#issuecomment-557009132).

### Bug Fix
Retain documentation as it is, and implement `token` field as optional.

The reason it wasn't implemented like this was because it's less secure. But since other sources also have security as an opt in feature, this source can do the same. 

The benefit would be a simpler default configuration.

ktff:
Closes #1784

### DIFF
--- a/src/sources/splunk_hec.rs
+++ b/src/sources/splunk_hec.rs
@@ -30,12 +30,22 @@ lazy_static! {
 
 /// Accepts HTTP requests.
 #[derive(Deserialize, Serialize, Debug, Clone)]
+#[serde(deny_unknown_fields, default)]
 pub struct SplunkConfig {
     /// Local address on which to listen
     #[serde(default = "default_socket_address")]
     address: SocketAddr,
     /// Splunk HEC token
-    token: String,
+    token: Option<String>,
+}
+
+impl Default for SplunkConfig {
+    fn default() -> Self {
+        SplunkConfig {
+            address: default_socket_address(),
+            token: None,
+        }
+    }
 }
 
 fn default_socket_address() -> SocketAddr {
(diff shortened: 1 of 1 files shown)


src/sources/splunk_hec.rs
```diff
@@ -30,12 +30,22 @@ lazy_static! {
...
 pub struct SplunkConfig {
     /// Local address on which to listen
     #[serde(default = "default_socket_address")]
     address: SocketAddr,
     /// Splunk HEC token
-    token: String,
+    token: Option<String>,
+}
+
+impl Default for SplunkConfig {
```
LucioFranco: I don't think we need this + `serde(default)` since we already set a default for address and token has an Option which defaults to none.

ktff: Compiler says differently

```
error[E0277]: the trait bound `sources::splunk_hec::SplunkConfig: std::default::Default` is not satisfied
  --> src/sources/splunk_hec.rs:32:10
   |
32 | #[derive(Deserialize, Serialize, Debug, Clone)]
   |          ^^^^^^^^^^^ the trait `std::default::Default` is not implemented for `sources::splunk_hec::SplunkConfig`
   |
   = help: see issue #48214
```

I suppose `Deserialize` derive is somehow hitting [this issue](https://github.com/rust-lang/rust/issues/48214).

For note:
deriving `Default` doesn't work also

```
error[E0277]: the trait bound `std::net::SocketAddr: std::default::Default` is not satisfied
  --> src/sources/splunk_hec.rs:37:5
   |
37 |     address: SocketAddr,
   |     ^^^^^^^^^^^^^^^^^^^ the trait `std::default::Default` is not implemented for `std::net::SocketAddr`
   |
   = note: required by `std::default::Default::default`
```

PR #1852, opened 2020-02-19
enhancement(file source): Better multi-line support

### LINKED ISSUE #1431: `message_stop_indicator` for file source
In addition to `message_start_indicator`, opposite parameter can be very useful.

For example, docker json log driver splits long lines by 16k. And only very last chunk contains `\r\n`.
```
{"log":"First line\r\n","stream":"stdout","time":"2019-12-16T17:25:16.44039104Z"}
{"log":"Second...","stream":"stdout","time":"2019-12-16T17:28:28.448678893Z"}
{"log":"... line\r\n","stream":"stdout","time":"2019-12-16T17:28:28.448700857Z"}
{"log":"Third line\r\n","stream":"stdout","time":"2019-12-16T17:28:28.448700857Z"}
```

By defining `message_start_indicator = '\\n"'` I'm getting "Second..." merged to first line rather than to its continuation.

So only difference is how to combine unmatched lines — to previous or to next matched line.

MOZGIII:
Closes #1431.

TODO:
- [x] docs
- [ ] ~behavior tests?~ no behavior tests for sources
- [ ] test harness tests? probably would track this one separately


binarylogic:
@bruceg mind reviewing this?

src/sources/file/line_agg.rs
```diff
@@ -0,0 +1,519 @@
...
+    /// This is useful where a log line contains a marker indicating that it
+    /// begins a new message.
+    HaltBefore,
+
+    /// All consecutive lines, up to and including the first line matching this
+    /// pattern, are included in the group.
+    /// This is useful where a log line ends with a termination marker, such as
+    /// a semicolon.
+    HaltWith,
+}
```
bruceg: Thanks! This kind of documentation is excellent! I would only recommend putting a few words before these new types in case we run `cargo doc` on Vector some day.

src/sources/file/line_agg.rs
```diff
@@ -0,0 +1,519 @@
...
+            start_pattern,
+            condition_pattern,
+            mode,
+            timeout,
+        }
+    }
+}
+
+/// Private type alias to be more expressive in the internal implementation.
+type Filename = String;
```
bruceg: I have two reservations about this. First, the most appropriate type for filenames would be `PathBuf` instead of `String`, which is already more obviously a filename. Second, I am wondering if the newtype pattern would be more appropriate here. Since they are used as keys to a `HashMap`, that implies extra code, though. What do you think @LucioFranco?

src/sources/file/line_agg.rs
```diff
@@ -0,0 +1,519 @@
...
+    /// `Bytes` - the message data; `Filename` - file name.
+    type Item = (Bytes, Filename);
+    type Error = ();
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        loop {
+            // If we're in draining mode, short circut here.
+            if let Some(to_drain) = &mut self.draining {
+                if let Some((line, src)) = to_drain.pop() {
+                    return Ok(Async::Ready(Some((line, src))));
```
bruceg: Since this function is just one whole `loop`, I wonder if it wouldn't be better to `break VALUE` instead of `return VALUE` for this and all return statements within. That's a bit of a style thing though, so I don't have a strong opinion on this.

src/sources/file/line_agg.rs
```diff
@@ -0,0 +1,519 @@
...
+                        Some(self.buffers.drain().map(|(k, v)| (v.into(), k)).collect());
+                }
+                Ok(Async::NotReady) => {
+                    // We didn't get any lines from `inner`, so we just give
+                    // a line from the expired lines queue.
+                    if let Some(key) = self.expired.pop_front() {
+                        if let Some(buffered) = self.buffers.remove(&key) {
+                            return Ok(Async::Ready(Some((buffered.freeze(), key))));
+                        }
+                    }
```
bruceg: I don't understand why this condition wouldn't be checked unconditionally before polling the inner. If I'm understanding right, this would result in newer lines being emitted before expired lines, which will have been originally read in earlier. I see you copied this logic from the original implementation, so I'm not really asking about your change, just trying to understand the reasoning here.

src/sources/file/line_agg.rs
```diff
@@ -0,0 +1,519 @@
...
+            let condition_matched = self.config.condition_pattern.is_match(line.as_ref());
+            match self.config.mode {
+                // All consecutive lines matching this pattern are included in
+                // the group.
+                Mode::ContinueThrough => {
+                    if condition_matched {
+                        let buffered = self
+                            .buffers
+                            .get_mut(&src)
+                            .expect("already asserted key is present");
```
bruceg: Can this pattern be replaced by using the `HashMap` `Entry` API above to avoid the double `get` plus `expect`?

src/sources/file/line_agg.rs
```diff
@@ -0,0 +1,519 @@
...
+                        buffered.extend_from_slice(b"\n");
+                        buffered.extend_from_slice(&line);
```
bruceg: Given this is repeated on each mode, would it make sense to add a tiny new buffer wrapper that can manage this?

src/sources/file/line_agg.rs
```diff
@@ -0,0 +1,519 @@
...
+    /// Handle line, if we have something to output - return it.
+    fn handle_line(&mut self, line: Bytes, src: Filename) -> Option<(Bytes, Filename)> {
+        // Check if we already have the buffered data for the source.
+        if self.buffers.contains_key(&src) {
+            let condition_matched = self.config.condition_pattern.is_match(line.as_ref());
+            match self.config.mode {
+                // All consecutive lines matching this pattern are included in
+                // the group.
+                Mode::ContinueThrough => {
+                    if condition_matched {
```
bruceg: Did you consider doing this the other way around, with the `if condition_matched` outside a pair of `match self.config.mode`? This might make the double `get` usage I reference below easier to factor out.

src/sources/file/line_agg.rs
```diff
@@ -0,0 +1,519 @@
...
+
+        assert_eq!(actual, expected_mapped);
+    }
+
+    fn run_and_assert(lines: &[&'static str], config: Config, expected: &[&'static str]) {
+        let stream = stream_from_lines(lines);
+        let line_agg = LineAgg::new(stream, config);
+        let results = collect_results(line_agg);
+        assert_results(results, expected);
+    }
```
bruceg: This looks like a good job of removing a pile of common boilerplate from the individual tests.

src/sources/file/line_agg.rs
MOZGIII: Good point!
I really want `#[deny(missing_docs)]`. Maybe I’ll enable it for just this mod, as an experiment.

src/sources/file/line_agg.rs
MOZGIII: Honestly, I didn’t look into this part much, cause it’s a part of the older implementation. I decided not to change anything unrelated to the extended multi-line support.
I personally also found this odd while I was getting to know the code, and I agree that this should be in reverse order. Otherwise we don’t provide proper backpressure and the expired lines may never even flush at all if the load is super-high.

src/sources/file/line_agg.rs
MOZGIII: If we change the underlying type, I like using the `PathBuf`. It’s better to update it at the whole file source though, and maybe make `LineAgg` generic over it. It would help making it more reusable, since we already have `event::Discriminant` that could also be used there.

src/sources/file/line_agg.rs
MOZGIII: I didn’t. I can experiment with it, but overall we’re doing not just two, but four operations with HashMap - get/insert/remove. What could really help here is the `Entry` API, but one thing we beed from there, in particular, is not yet stable. I don’t know what our policy regarding unstable rust features is, so I just left it as is. I’d love to switch to `Entry` API though.

src/sources/file/line_agg.rs
MOZGIII: See https://github.com/timberio/vector/pull/1852#discussion_r381934319

src/sources/file/line_agg.rs
MOZGIII: Yes, it’s one of the potential improvements. Adding dedicated type wrappers for buffer would make sense, but even before that we could just add helper functions. Another similar improvement would be to add a wrapper around `buffers` HashMap and `timeouts` DelayQueue - to provide a more sound interface.
I decided to skip those because I didn’t want to change the approach on this part of the code beyond the scope of adding features to the multi-line support. It can be viewed as technical debt, and while I’d love to spend more time on it - I don’t think that would be a good time, since we’re trying to push 1.0.

src/sources/file/line_agg.rs
MOZGIII: Is there a reason why you prefer `break` over `return` in such cases? I understand, in this particular case it doesn’t make any practical difference at all, but there may be reasons - like `break`s being better in typical ways of how such loops evolve.

src/sources/file.rs
```diff
@@ -3,17 +3,19 @@ use crate::{
...
 use snafu::{ResultExt, Snafu};
-use std::collections::{HashMap, VecDeque};
+use std::convert::{TryFrom, TryInto};
 use std::path::PathBuf;
 use std::thread;
 use std::time::{Duration, SystemTime};
-use tokio::timer::DelayQueue;
+
+mod line_agg;
+use line_agg::LineAgg;
```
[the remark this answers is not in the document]
MOZGIII: I was thinking about moving the `line_agg` somewhere else, out of the `file` source altogether. I just couldn't find a place for it at the time.
Maybe a good place would be just a level upper - at the very `src/` dir? What do you think?

src/sources/file.rs
```diff
@@ -60,10 +70,47 @@ pub struct FileConfig {
...
+
+        Ok(Self {
+            start_pattern,
+            condition_pattern,
+            mode,
+            timeout,
+        })
+    }
+}
+
```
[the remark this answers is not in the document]
MOZGIII: I could move it in, but it'll bring the `InvalidMultilineStartPattern` and `InvalidMultilineConditionPattern` with it. That isn't bad, but I'd just leave it there for now - until we actually need to share the config.
If we move it into `line_agg`, I'd rename it `SerializableConfig`.

src/sources/file.rs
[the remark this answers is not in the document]
bruceg: If it is used in other sources, then `src/sources/util` is most appropriate; if it is used in other transforms as well then at the top level `src` is appropriate. Until there is a plan to actually use it in other places, keeping it under `src/sources/file` is still the right place.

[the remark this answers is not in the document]
MOZGIII: As I understand, we're not planning to start work on using it elsewhere now, but there were plans on using this with `socket` source. I'll leave it at `src/sources/file/` for now, and move the `file.rs` to `file/mod.rs`.

src/sources/file/line_agg.rs
MOZGIII: I did some experiments, and I think the code is more readable if we group the logic for an operation mode together. It's not worth optimizing structure cause it'll hurt maintainability.

bruceg: Looks good, and it's always nice when something like this makes our code shorter to boot.

bruceg: Oops, commented on the wrong thread there. Agreed, readability is important. Branch prediction will likely deal with any potential performance consequences.

src/sources/file/line_agg.rs
MOZGIII: I made `LineAgg` generic over the `String` - it's an easy thing to go, and it reduces the footprint of the changes if we'd want to alter the keys at the `file` source.

src/sources/file/line_agg.rs
MOZGIII: Oh yes, and I'd expect compiler to rearrange code significantly anyway, so performance aspects are hardly something we should care about.

src/sources/file/line_agg.rs
MOZGIII: I'd avoid changing that in this iteration. We better create a separate issue. There should be many more tests for the timeout and expiration logic, and once those are done - we can safely alter details like this.

bruceg: Agreed

MOZGIII: Created #1870.

src/sources/file/line_agg.rs
bruceg: I have tended to prefer the expression or functional notation over procedural. I agree this makes no practical difference here.

src/sources/file/mod.rs
```diff
@@ -195,11 +258,18 @@ pub fn file_source(
         let (tx, rx) = futures::sync::mpsc::channel(100);
 
         let messages: Box<dyn Stream<Item = (Bytes, String), Error = ()> + Send> =
-            if let Some(msi) = message_start_indicator {
+            if let Some(ref multiline_config) = multiline_config {
+                Box::new(LineAgg::new(
+                    rx,
+                    multiline_config.try_into().unwrap(), // validated in build
```
bruceg: One last thought. Is there any way for this created `line_agg::Config` to be stored where it is parsed in `build` and used here instead of re-parsing?

MOZGIII: I tried going that route - but ended up wanting to redo the whole `file` source configuration approach. I think we need to address this systematically - validate and update all the units. In short, the idea I had is to have two layers - a serializable/deserializeable config and a "processed" config, that's been validated and has values converted into the proper types. 
So, in short - I see how ugly the current approach is - and we definitely should fix this, but even if we'll do it just with the `file` source - this is one of the part where I deliberately decided to leave the current approach unchanged.

MOZGIII: I created https://github.com/timberio/vector/issues/1895

.meta/sources/file.toml
```diff
@@ -178,6 +178,88 @@ Instead of balancing read capacity fairly across all watched files, prioritize \
...
 """
 
+[sources.file.options.multiline]
+type = "table"
+category = "Multiline"
+common = true
+required = false
+description = """\
+Multiline parsing configuration. \
+If not speicified, multiline parsing is disabled.\
```
rcollette, line 188: Spelling issue.  Should be "specified."

jamtur01: @rcollette That spelling error has been addressed in later code. Thanks for letting us know though!

PR #1874, opened 2020-02-20
chore(testing): Add feature `disable-resolv-conf` to allow testing on NixOS

a-rodin:
This might resolve https://github.com/timberio/vector/issues/1803.

It is a result of a discussion with @LucioFranco we had today about ways to prevent `trust-dns-resolver` to access `/etc/resolv.conf`.

This is WIP because I haven't tested it yet on an actual NixOS machine. Note that tests on NixOS running in Docker would not be enough because Docker mounts `/etc/resolv.conf` from the host machine, so it is always present there.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -236,6 +236,8 @@ influxdb-integration-tests = []
 kubernetes = ["kubernetes-integration-tests"]
 kubernetes-integration-tests = []
 
+disable-resolv-conf = []
+
 [[bench]]
 name = "bench"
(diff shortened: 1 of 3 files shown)


a-rodin:
The changes were tested on a NixOS machine using `./scripts/verifiy-nixos.sh` and turned out to be working.

a-rodin:
The CI checks for this are not enabled because the NixOS Docker image seems to not exactly match a real NixOS system.

PR #1883, opened 2020-02-20
fix: Fix failing benchmarks

### LINKED ISSUE #1872: Benchmarks are failing with `Schema was not initialized`
Running any of the benchmarks results in the following error. This is related to #1769, as commits before that work.

```
$ cargo bench --bench bench
    Finished bench [optimized + debuginfo] target(s) in 0.20s
     Running target/release/deps/bench-063c2df790645666
Benchmarking pipe: Warming up for 3.0000 sthread 'tokio-runtime-worker-5' panicked at 'Schema was not initialized', src/libcore/option.rs:1188:5
thread 'tokio-runtime-worker-6' panicked at 'Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }', src/test_util.rs:60:78
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Canceled', src/libcore/result.rs:1188:5
error: bench failed
```

Hoverbear:
Fixes #1872

After talking to @LucioFranco we decided this is the easiest fix. Moving forward, we should consider threading through global options safely.

### DIFF
--- a/src/event/mod.rs
+++ b/src/event/mod.rs
@@ -28,6 +28,11 @@ pub static LOG_SCHEMA: OnceCell<LogSchema> = OnceCell::new();
 
 lazy_static! {
     pub static ref PARTIAL: Atom = Atom::from("_partial");
+    static ref LOG_SCHEMA_DEFAULT: LogSchema = LogSchema {
+        message_key: Atom::from("message"),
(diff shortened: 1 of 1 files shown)


Hoverbear:
@bruceg I think #1860 is better but we decided in that that further investigation is required.

PR #1886, opened 2020-02-21
enhancement(lua transform)!: respect Lua types when converting back to events

### LINKED ISSUE #857: Lua transformer type loss
Seems like int/floats get converted back into strings if manipulated via lua. Not sure if this is avoidable. My current solution is to run a coercer to get it back into an an int.

Reproduced with:
```
data_dir = "/tmp/"

[sources.input]
  type = "stdin"

[transforms.tokenizer]
  type = "tokenizer"
  inputs = ["input"]
  field_names = ["time"]
  field = "message"
  drop_field = true
[transforms.tokenizer.types]
    time = "int"

[transforms.lua]
  type = "lua"
  inputs = ["tokenizer"]
  source = """
    event['time'] = event['time'] + 1
  """

[sinks.output]
  inputs = ["lua"]
  type = "console"
  target = "stdout"
```

Run: `echo 1 | vector --config config.toml`

Expect: `{"time":2, ....`
Actual: `{"time":"2.0",...`

sciyoshi:
When using the Lua transform with a function like this:

```lua
event["number"] = 42
```

the resulting event has the value converted back to a string:

```
{ ..., "number": "42", ... }
```

This PR changes the transform to preserve types, and also adds support for setting fields to boolean values, saving the need for further processing to restore the types. It also slightly changes the behavior to drop fields when set to invalid types, rather than erroring out the function. I'm not sure what impact this will have when it comes to https://github.com/timberio/vector/issues/704, however.

Fixes #857.

### DIFF
--- a/src/transforms/lua.rs
+++ b/src/transforms/lua.rs
@@ -1,6 +1,6 @@
 use super::Transform;
 use crate::{
-    event::Event,
+    event::{Event, Value},
     topology::config::{DataType, TransformConfig, TransformContext, TransformDescription},
 };
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


cetanu:
I wonder if this can be updated to accept lua tables/arrays now that #704 is complete?

PR #1892, opened 2020-02-21
enhancement(sources): Add TLS support to socket, syslog, and vector sources

bruceg:
This is the first down payment on #1553, adding support for TLS to the three sources that accept simple TCP connections: `socket`, `syslog`, and `vector`.

Note that the `vector` *sink* does not yet support TLS connections, presumably because the source didn't support them at the time. Adding that support should be trivial, but I will put it in a follow up PR.

Similarly, this does not add support for TLS to the sources that make outgoing connections. That too will be done in a separate PR.

Note also there is a hiccup in the added `send_lines_tls` function to do with shutdown of the socket. If somebody has an idea how to resolve that, I'd like to clean it up, but it works as it is now.


src/sources/util/tcp.rs
```diff
@@ -15,6 +15,7 @@ use tokio::{
     reactor::Handle,
     timer,
 };
+use tokio_tls::TlsAcceptor;
```
LucioFranco: do we want to just start using `tokio_openssl` instead?

src/sources/util/tcp.rs
```diff
@@ -115,17 +117,47 @@ pub trait TcpSource: Clone + Send + 'static {
...
+                                                .filter_map(move |frame| {
+                                                    let host = host.clone();
+                                                    source.build_event(frame, host)
+                                                })
+                                                .map_err(
+                                                    |error| warn!(message = "connection error.", %error),
+                                                )
+                                                .forward(out)
+                                                .map(|_| debug!("TLS connection closed."));
+                                            tokio::spawn(handler);
```
LucioFranco: this has no span attached to it, we need to ensure it gets the correct span.

src/sources/util/tcp.rs
```diff
@@ -115,17 +117,47 @@ pub trait TcpSource: Clone + Send + 'static {
...
-                            })
-                            .map_err(|error| warn!(message = "connection error.", %error));
-
-                        let handler = events_in.forward(out).map(|_| debug!("connection closed."));
-
-                        tokio::spawn(handler.instrument(span.clone()));
+                        if let Some(tls) = tls.clone() {
+                            match tls.acceptor() {
+                                Err(error) => error!(message = "Failed to create a TLS connection acceptor", %error),
+                                Ok(acceptor)=> {
```
LucioFranco: can we break out this nested iflet matches a bit so that its a bit easier to follow?

src/sources/util/tcp.rs
```diff
@@ -115,17 +117,47 @@ pub trait TcpSource: Clone + Send + 'static {
...
+                                    let handler = TlsAcceptor::from(acceptor)
```
LucioFranco: there is some duplicated code here, I would suggest we do the same thing we did with sinks and enum out a MaybeTlsStream and then have one `FramedRead::new()` setup. I think that would make this a bit more clear.

src/test_util.rs
```diff
@@ -69,6 +70,48 @@ pub fn send_lines(
...
+                .map_err(|e| panic!("{:}", e))
+                .and_then(|stream| {
+                    let out = FramedWrite::new(stream, LinesCodec::new())
+                        .sink_map_err(|e| panic!("{:?}", e));
+
+                    lines
+                        .forward(out)
+                        .and_then(|(_source, sink)| {
+                            let mut stream = sink.into_inner().into_inner();
+                            // We should catch TLS shutdown errors here,
```
LucioFranco: I am not directly following why we need to handle the shutdown here? the forward future should in theory flush all the way?

src/tls.rs
```diff
@@ -146,6 +168,22 @@ impl TlsSettings {
...
+        // This data was test-built previously, so we can just use it
+        // here and expect the results will not fail. This can all be
+        // reworked when `native_tls::Identity` gains the Clone impl.
+        self.identity.as_ref().map(|identity| {
+            Identity::from_pkcs12(&identity.0, &identity.1).expect("Could not build identity")
+        })
+    }
+
+    pub(crate) fn acceptor(&self) -> crate::Result<TlsAcceptor> {
+        match self.identity() {
```
LucioFranco: minor nit: this would be better if it were just an if let imo

LucioFranco (commented):
This looks great, I left a few comments and suggestions.

src/test_util.rs
bruceg: The parallel `send_lines` function has a `shutdown`, so I tried to get it to work in the TLS version and was unable. I am uncomfortable removing one without the other.

src/sources/util/tcp.rs
bruceg: That may make sense. However, given that nothing else in Vector uses `tokio_openssl` yet, but other modules use `tokio_tls` as well (ex the parallel `src/sinks/utils/tcp.rs'), I think it would make more sense as a separate PR.

src/tls.rs
bruceg: I think this is the `match` vs `if` debate again... (ref #1720)

src/sources/util/tcp.rs
LucioFranco: Well that isn't true, I did add this :P https://github.com/timberio/vector/blob/ce326530da9235233355a86dcc5c27b626f40a10/src/sinks/util/rusoto.rs#L7

We really need to just pick one I think if we are already here we should just do it now instead of punting imo.

src/test_util.rs
LucioFranco: Yeah, so I think dropping the tcpstream should be fine enough since forward should ensure its all flushed. So I think its safe to ignore it.

src/tls.rs
LucioFranco: Though I think the reason I prefer if let here is because the option enumerates to exactly two options here therefore we don't need the extra syntax for matching None. I don't mind either way.

src/sources/util/tcp.rs
bruceg: Uh, ok, you said `tokio_openssl`, so I grepped the sources for `tokio_openssl` direct uses, of which there are still none.

I can rework the sources here, but are you also advising me to rework the sinks as well to use `tokio_openssl` in this PR?

src/test_util.rs
bruceg: The tests that use this new `send_lines_tls` have passed every time I tried, so if there is a race or other issue with this it isn't obvious. However, I left the comment there because of the discrepancy.

src/sources/util/tcp.rs
LucioFranco: No just the stuff you touch in this PR. And sorry I totally could have been more clear. `hyper-openssl` and `tokio-openssl` are basically just wrappers around each other that implement the correct traits for the `openssl` crate. So I think it makes sense to write any new code using that then going back later and updating the others.

src/sources/util/tcp.rs
bruceg: I pushed up a030d995b31a9f5a82f2ca82c67ad9324bd8473c, is that what you were suggesting?

src/sources/util/tcp.rs
bruceg: Done

LucioFranco (approved):

src/sources/util/tcp.rs
bruceg: Ok, so I looked into this, and it requires a rework of the TLS config handling that will end up being easier if I just do both acceptors and connectors at once, due to the different types involved (`native_tls` wrappers vs direct use of `openssl`).

LucioFranco: That works for me :+1:

LucioFranco (approved):
docs look good :+1:

binarylogic (approved):
Nice work. :shipit:

PR #1898, opened 2020-02-23
enhancement!: Support null values in log events

### LINKED ISSUE #1889: `json_parser` transform replaces `null` values by empty strings
Feeding this object

```json
{"x": [3, null, 5]}
```

to `json_parser` transform and then encoding the result as JSON produces the following log event:

```json
{"timestamp":"2020-02-21T18:55:15.593226889Z","x":[3,"",5],"host":"dev"}
```

So the `null` values are silently converted to empty strings.

I think it is not the correct behavior and the null values ought to stay null values after being passed through Vector.

In order to solve the issue, I propose to add a new type to the [`Value`](https://github.com/timberio/vector/blob/650ce582e2eab05f585505e5b66197eb672c556c/src/event/mod.rs#L206-L215) enum called `Null`, then use it in JSON serialization and deserialization.

a-rodin:
Closes #1889.

This is related to all components interacting with log events.


website/docs/about/data-model/log.md
```diff
@@ -258,6 +258,10 @@ Booleans represent binary true/false values.
 Timestamps are represented as [`DateTime` Rust structs][urls.rust_date_time]
 stored as UTC.
 
+#### Null Values
+
+For compatibility with JSON log events also support `null` values.
```
Hoverbear:
```suggestion
For compatibility with JSON log events, Vector also supports `null` values.
```

tests/behavior/transforms/json_parser.toml
```diff
@@ -30,3 +30,19 @@
...
+  [tests.input]
+    insert_at = "json_parser_null_value"
+    type = "raw"
+    value = "{\"a\": 3, \"b\": null}"
+  [[tests.outputs]]
+    extract_from = "json_parser_null_value"
+    [[tests.outputs.conditions]]
+      type = "check_fields"
+      "a.equals" = 3
+      "b.equals" = "<null>"
```
Hoverbear: So does this imply that `"<null>"` will be read as a `null` by Vector?

a-rodin: No, `"<null>"` will be read as a string (more precisely, `Bytes`) by Vector. So it is just a string representation produced by `to_string_lossy()` and `to_bytes()`. It is not revertible for other data types too, for example Vector timestamps are rendered as strings in ISO 8601 format, but strings containing timestamps in ISO 8601 format are read as strings, not as timestamps.

PR #1906, opened 2020-02-24
fix(clickhouse sink): Nest request fields

Jeffail:
While prodding around I noticed the request config fields aren't actually nested for the clickhouse sink.


Jeffail:
Note, this would totes mcgoats have been caught by https://github.com/timberio/vector/issues/1735

PR #1908, opened 2020-02-24
 feat(cli): Add `log-format` cli option 

LucioFranco:
Re-opening #1873 and #1907 

This is to continue the discussion around flattened fields.

As shown in #1709 the json fields that appear on `tracing::Event` should be flattened. This should be possible but will require some upstream changes to allow this type of flattening. If we are happy to just flatten the event fields but keep spans as sub-objects then we should be able to get away with this without any performance impacts.

@binarylogic if this is what we want I am happy to implement this within tracing.


LucioFranco:
I have a draft Pr up for tracing here https://github.com/tokio-rs/tracing/pull/599

LucioFranco:
This now outputs logs like so:

```json
{"timestamp":"Feb 27 19:18:32.422","level":"INFO","target":"vector","message":"Log level \"info\" is enabled."}
{"timestamp":"Feb 27 19:18:32.422","level":"INFO","target":"vector","message":"Loading configs.","path":"[\"/etc/vector/vector.toml\"]"}
{"timestamp":"Feb 27 19:18:32.434","level":"INFO","target":"vector","message":"Vector is starting.","version":"0.9.0","git_version":"v0.8.0-21-g7457edf","released":"Fri, 28 Feb 2020 00:16:51 +0000","arch":"x86_64"}
{"timestamp":"Feb 27 19:18:32.434","level":"INFO","target":"vector::topology","message":"Running healthchecks."}
{"timestamp":"Feb 27 19:18:32.434","level":"INFO","target":"vector::topology","message":"Starting source \"stdin\""}
{"timestamp":"Feb 27 19:18:32.435","level":"INFO","target":"vector::topology","message":"Starting sink \"console\""}
{"timestamp":"Feb 27 19:18:32.435","level":"INFO","target":"vector::topology::builder","message":"Healthcheck: Passed."}
{"timestamp":"Feb 27 19:18:32.435","level":"INFO","span":{"name":"source","type":"stdin"},"target":"vector::sources::stdin","message":"Capturing STDIN"}
```

PR #1917, opened 2020-02-24
enhancement(topology): POC of topology shutdown cleanup

stbrody:

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -1,0 +1,96 @@
+use futures::{Async, Future, IntoFuture, Poll, Stream};
+
+/// A stream combinator which takes elements from a stream until a future resolves.
+///
+/// This structure is produced by the [`StreamExt::take_until`] method.
+#[derive(Clone, Debug)]
+pub struct TakeUntil<S, F> {
+    stream: S,
+    until: F,
+    free: bool,
+}
+
+/// This `Stream` extension trait provides a `take_until` method that terminates the stream once
+/// the given future resolves.
+pub trait StreamExt: Stream {
+    /// Take elements from this stream until the given future resolves.
+    ///
+    /// This function will take elements from this stream until the given future resolves. Once it
+    /// resolves, the stream will yield `None`, and produce no further elements.
+    ///
+    /// If the future produces an error, the stream will be allowed to continue indefinitely.
+    ///
+    /// ```
+    /// # extern crate stream_cancel;
+    /// extern crate tokio;
+    /// extern crate futures;
+    ///
+    /// use stream_cancel::StreamExt;
+    /// use tokio::prelude::*;
+    ///
+    /// let listener = tokio::net::TcpListener::bind(&"0.0.0.0:0".parse().unwrap()).unwrap();
+    /// let (tx, rx) = futures::sync::oneshot::channel();
+    ///
+    /// let mut rt = tokio::runtime::Runtime::new().unwrap();
+    /// rt.spawn(
+    ///     listener
+    ///         .incoming()
+    ///         .take_until(rx.map_err(|_| ()))
+    ///         .map_err(|e| eprintln!("accept failed = {:?}", e))
+    ///         .for_each(|sock| {
+    ///             let (reader, writer) = sock.split();
+    ///             tokio::spawn(
+    ///                 tokio::io::copy(reader, writer)
+    ///                     .map(|amt| println!("wrote {:?} bytes", amt))
+    ///                     .map_err(|err| eprintln!("IO error {:?}", err)),
+    ///             )
+    ///         }),
+    /// );
+    ///
+    /// // tell the listener to stop accepting new connections
+    /// tx.send(()).unwrap();
+    /// rt.shutdown_on_idle().wait().unwrap();
+    /// ```
+    fn take_until<U>(self, until: U) -> TakeUntil<Self, U::Future>
+    where
+        U: IntoFuture<Item = (), Error = ()>,
+        Self: Sized,
+    {
+        TakeUntil {
+            stream: self,
+            until: until.into_future(),
+            free: false,
+        }
+    }
+}
+
+impl<S> StreamExt for S where S: Stream {}
+
+impl<S, F> Stream for TakeUntil<S, F>
+where
+    S: Stream,
+    F: Future<Item = (), Error = ()>,
+{
+    type Item = S::Item;
+    type Error = S::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if !self.free {
+            match self.until.poll() {
+                Ok(Async::Ready(_)) => {
+                    // future resolved -- terminate stream
+                    return Ok(Async::Ready(None));
+                }
+                Err(_) => {
+                    // future failed -- unclear whether we should stop or continue?
+                    // to provide a mechanism for the creator to let the stream run forever,
+                    // we interpret this as "run forever".
+                    self.free = true;
+                }
+                Ok(Async::NotReady) => {}
+            }
+        }
+
+        self.stream.poll()
+    }
+}
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -11,6 +11,14 @@ pub struct TakeUntil<S, F, O> {
     free: bool,
 }
 
+/// A stream combinator which takes a handle to hold onto for the lifetime of the stream.
+///
+/// This structure is produced by the [`StreamExt::with_handle`] method.
+pub struct WithHandle<S, H> {
+    stream: S,
+    _handle: H,
+}
+
 /// This `Stream` extension trait provides a `take_until` method that terminates the stream once
 /// the given future resolves.
 pub trait StreamExt: Stream {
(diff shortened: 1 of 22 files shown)


src/sources/util/tcp.rs
```diff
@@ -131,6 +134,7 @@ pub trait TcpSource: Clone + Send + 'static {
                     Ok(())
                 })
                 .inspect(|_| trigger.cancel());
+            // todo signal when shutdown is complete
```
stbrody: When I first read your [proposal](https://github.com/timberio/vector/issues/1091), @LucioFranco, I didn't understand why you suggested that the Sources would have "handle" that would notify the shutdown coordinator that their shutdown is complete by dropping the handle, so I went with the approach coded here of using a Trigger object directly, figuring the Sources could just signal the Trigger whenever their shutdown is complete.

Now, having waded into the code a bit deeper, I think I understand better.  The issue is that a single source can spawn multiple futures in parallel, so knowing once all child futures have been fully cleaned up is difficult.  If instead we had a reference-counted object passed into each spawned future, then as the various future chains resolve at shutdown we'd know all sub-tasks have completed when the refcount of the handle object goes to zero.  The handle object's destructor could then signal a Trigger object to notify the shutdown coordinator that this source's shutdown is complete.

So now I'm thinking I should replace the ShutdownSignals.shutdown_complete Trigger object with a Rc<TriggerHandle> that contains the equivalent of the current shutdown_complete Trigger object and overloads the Drop trait to call .cancel() on the Trigger.

Is this in line with what you were thinking?

LucioFranco:
Sorry for the delay on this, I'd like to take a peek at this but most likely will do so tomorrow.

src/shutdown.rs
```diff
@@ -0,0 +1,12 @@
...
+
+// Passed to each Source to coordinate the global shutdown process.
+pub struct ShutdownSignals {
+    // This will be triggered when global shutdown has begun, and is a sign to the Source to begin
+    // its shutdown process.
+    pub begin_shutdown: Tripwire,
+
+    // Triggered by an individual Source to inform the global shutdown coordinator that its local
+    // shutdown process is complete.
+    pub shutdown_complete: Trigger,
```
LucioFranco: This should be possible to do via just ending the future? Otherwise, this seems like an easy footgun to forget to do.

src/sources/socket/udp.rs
```diff
@@ -55,7 +65,12 @@ pub fn udp(address: SocketAddr, host_key: Atom, out: mpsc::Sender<Event>) -> Sou
                 .map_err(|error: io::Error| error!(message = "error reading datagram.", %error))
                 .forward(out)
                 // Done with listening and sending
-                .map(|_| ())
+                .map(|_| {
```
LucioFranco: For example, we could find a way to store this "oneshot" handle along side this returned future and when it exists we can then notify the main tripwire.

src/sources/util/tcp.rs
LucioFranco: Yes, this is some what along the lines of what I was thinking. This should be very similar to a waitgroup in go. I am not 100% sure on the API yet but we can find ways to for example wrap spawn so when you spawn a future it will get a new bumped ref counted trigger with it. This way we can even track how many subtasks per source are left.

Happy to talk more through this via slack as well :)

src/topology/builder.rs
```diff
@@ -118,27 +120,37 @@ pub fn build_pieces(
...
-        let (trigger, tripwire) = Tripwire::new();
-
         let (output, control) = Fanout::new();
         let pump = rx.forward(output).map(|_| ());
         let pump = Task::new(&name, &typetag, pump);
 
-        let server = server.select(tripwire.clone()).map(|_| ()).map_err(|_| ());
+        // TODO: Remove this line once all Sources do this internally.
+        let server = server
+            .select(global_shutdown_begun_tripwire)
```
LucioFranco: I think this select is related to this https://github.com/timberio/vector/pull/1917/files#r383582914

We need someway to force cancel the task if its taking too long. One trick is to wrap the source future with a select future that will drop the source future when it receives the force cancel command.

stbrody:
@LucioFranco, I have updated the POC with my improved understanding of how the 'force' shutdown works, and switched ShutdownSignals from using a Trigger object directly to using an Arc<Trigger> that'll get signaled when the handle gets dropped and the Arc refcount goes to 0.
This still does not include the work to actually clean up the code at all.  In the real version I plan to add a Shutdown object that'll own all the various Triggers and Tripwires from topology/{mod.rs, builder.rs} and encapsulate much of that logic.

The part I'm most concerned about with this review is how we ensure that Source implementations properly pass around the ShutdownSignals struct to all their child futures.  I added a take() method that can force the compiler to move the struct into the future it is called in, but I'm still not particularly satisfied with that approach.  You suggested there may be a way to override the tokio::spawn method so this happens automatically without Source implementations having to change any code at all.  That would be great but I'm not sure how to go about doing that.  Would love to connect with you on that if you have any ideas.

src/topology/builder.rs
stbrody: Thanks, digging into this a bit more I think I understand how this works better now.  Please check out the addition of the 'shutdown_force_triggers' I added in the new version which I believe better handles this now.

src/shutdown.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// Used to force moving this struct into a closure.
+    /// Every thread doing work on behalf of a Source must call this at least once.
+    pub fn take(&mut self) {
+        self.taken = true;
+    }
+}
+
+impl Drop for SourceShutdownHandle {
+    fn drop(&mut self) {
+        if !self.taken {
```
stbrody: I'd really like if this could be enforced at compile time rather than run time somehow.

LucioFranco:
@stbrody I think what we chatted about on slack mostly covers what you were asking. One added thing is there might be a way to `wrap` out `TaskExecutor` in a way to "decorate" spawning new futures. Let me know if you have anymore questions!

stbrody:
@LucioFranco, I just uploaded a new version with the changes we discussed.  I think now udp.rs is a decent example of a "normal" source and tcp.rs a decent example of one that has to do some extra work before signaling that its shutdown is done.  I do still use .take() in the tcp source to force moving the handle into the spawned futures, I'm curious if you think that can be avoided.

src/sources/util/tcp.rs
```diff
@@ -118,6 +123,7 @@ pub trait TcpSource: Clone + Send + 'static {
                         let events_in = FramedRead::new(socket, source.decoder())
                             .take_until(tripwire)
                             .filter_map(move |frame| {
+                                shutdown_complete_handle.take();
```
stbrody: I still use .take() here because I'm not sure how else to ensure that the future created here and passed into tokio::spawn a few lines later actually captures the shutdown_complete_handle.  I certainly could call clone() instead, but that's an additional clone where just moving the copy created on line 115 into this future would suffice.

LucioFranco: So this looks like a stream, might make sense here to write some sort of stream combinator that will hold onto the handle until the stream is completed?

LucioFranco:
So I think take_until should only drop the handle once the stream has been dropped? So I think that should kind of work? Would be good to test that.

LucioFranco:
We may want to implement a custom take_until

stbrody:
> We may want to implement a custom take_until

I was actually right in the process of writing a comment calling out the issue with take_until and the fact that the future it accepts isn't allowed to resolve to a value.  I think you're right and we may indeed have to write our own take_until that works a little more like hyper's with_graceful_shutdown.  I think I'll give that a shot next

LucioFranco:
@stbrody yeah, basically it needs to store the return value in the struct, so you could just augment it. Let me talk to jon to see if he'd accept that in stream cancel.

stbrody:
Okay, new version has been uploaded that creates our own version of TakeUntil that can take a future with a return value and save that value internally, as well as adding a new WithHandle stream combinator for tying the lifetime of the SourceShutdownHandle to that of a given stream.

I think this is starting to come together, thanks for all your help @LucioFranco!

src/combinator.rs
```diff
@@ -0,0 +1,129 @@
...
+{
+    type Item = S::Item;
+    type Error = S::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if !self.free {
+            match self.until.poll() {
+                Ok(Async::Ready(res)) => {
+                    // future resolved -- terminate stream
+                    self.until_res = Some(res);
```
stbrody: Ideally these changes to TakeUntil could be upstreamed into stream_cancel directly.

src/combinator.rs
```diff
@@ -0,0 +1,129 @@
...
+                }
+                Ok(Async::NotReady) => {}
+            }
+        }
+
+        self.stream.poll()
+    }
+}
+
+impl<S, H> Stream for WithHandle<S, H>
```
stbrody: This new combinator I would not imagine upstreaming, we could just add this locally for our own use.  Simply placed in this file for convenience sake for the POC

stbrody:
Closing in favor of https://github.com/timberio/vector/pull/1994

PR #1921, opened 2020-02-25
fix(cli): Relax the requirements for `no_outputs_from` unit test targets

Jeffail:
When verifying unit tests Vector checks that input targets are actually connected to output targets in the topology. However, for targets within `no_outputs_from` this check seems heavy handed as it's nice to explicitly confirm it in a test even if components aren't necessarily directly connected.

Also fixes an issue with validating the tested topology when using `no_outputs_from` that was blocking it from being used with `swimlanes`.

### DIFF
--- a/src/topology/unit_test.rs
+++ b/src/topology/unit_test.rs
@@ -218,7 +218,10 @@ fn reduce_transforms(
 
     if roots
         .iter()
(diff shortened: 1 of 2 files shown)


Hoverbear:
@Jeffail Please add a description and explanation of the PR before merging. :)

PR #1923, opened 2020-02-25
fix(config): Partial `LogSchema`

ktff:
As it stands, vector will fail to run with partial `LogSchema` configuration. Example:

```toml
[log_schema]
    message_key = "message"
    timestamp_key = "timestamp" 
```

would report "missing field `host_key`".

One of the consequences is that adding a new key like `kubernetes_key` would break all configurations that are configuring `log_schema`.

This PR adds `#[serde(default)]` to the `struct LogSchema` which enables partial configuration.

### DIFF
--- a/src/event/mod.rs
+++ b/src/event/mod.rs
@@ -762,7 +762,7 @@ mod test {
 
     #[test]
     fn partial_log_schema() {
-        let toml = r#"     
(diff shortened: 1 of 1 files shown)


Hoverbear:
Thanks @ktff :)

PR #1936, opened 2020-02-26
fix(merge transform): Support nested fields and arrays

MOZGIII:
This PR supersedes #1825. We couldn't advance with that approach, so we have to try again.

This is merely a rebase to `master` after #1902 has been merged.

So, the core problem is the `Map` values don't respect the invariants required by the `Discriminant`. This PR adds tests, that demonstrate the failure scenario. For things to work properly, all tests introduced in this PR have to pass.

This PR only adds tests that now fail. What to do with this PR is a question to me. We can merge it to master to force-break it, and then fix it later, or we can add fix here and then merge it.

/cc @a-rodin


a-rodin:
Let's keep this PR open until the tests are fixed.

a-rodin:
@MOZGIII I implemented `value_eq` for new value kinds (arrays, maps, and null values). On the way, #2062 was created because otherwise the tests didn't pass. I expect #2062 to be merged first, and then it would be possible to merge `master` to this branch and merge this PR after.

PR #1938, opened 2020-02-26
chore(topology): Allow transforms to implement `Stream`

### LINKED ISSUE #1598: Transforms as tasks

LucioFranco:
This adds a new `Transform::transform_stream` which allows consumers to
return their own type erased `Stream`. This should allow all various
types of advanced transforms, while still supporting simpler transforms
with a simpler API.

This fixes trait size issues by using arbitrary self types!

aka we were unable to call this type of trait fn:

```rust
fn transform_stream(self) -> Box<dyn Stream> { ... }
```

This is because the local `self` is required to be sized to be able to call it from a unsized trait object like `Box<dyn Transform>` which is what we use internally within topology.

To fix this we can use arbitrary self types to tell the compiler that we know what type of indirection we are using for the trait object. This allows us to call this code from a trait object backed by a boxed pointer.

```rust
fn transfrom_stream(self: Box<Self>) -> Box<dyn Stream> { ... }
```

Re-opening of #1909

Closes #1598

### DIFF
--- a/src/topology/builder.rs
+++ b/src/topology/builder.rs
@@ -152,7 +152,7 @@ pub fn build_pieces(
             exec: exec.clone(),
         };
 
-        let mut transform = match transform.inner.build(cx) {
+        let transform = match transform.inner.build(cx) {
             Err(error) => {
                 errors.push(format!("Transform \"{}\": {}", name, error));
                 continue;
@@ -165,13 +165,8 @@ pub fn build_pieces(
 
         let (output, control) = Fanout::new();
 
-        let transform = input_rx
-            .map(move |event| {
-                let mut output = Vec::with_capacity(1);
-                transform.transform_into(&mut output, event);
-                futures01::stream::iter_ok(output.into_iter())
-            })
-            .flatten()
+        let transform = transform
+            .transform_stream(input_rx)
             .forward(output)
             .map(|_| ());
         let task = Task::new(&name, &typetag, transform);
--- a/src/transforms/mod.rs
+++ b/src/transforms/mod.rs
@@ -25,6 +25,8 @@ pub mod split;
 pub mod swimlanes;
 pub mod tokenizer;
 
+use futures01::{sync::mpsc::Receiver, Stream};
+
 pub trait Transform: Send {
     fn transform(&mut self, event: Event) -> Option<Event>;
 
@@ -32,6 +34,25 @@ pub trait Transform: Send {
         if let Some(transformed) = self.transform(event) {
             output.push(transformed);
         }
+    }
+
+    fn transform_stream(
+        self: Box<Self>,
+        input_rx: Receiver<Event>,
+    ) -> Box<dyn Stream<Item = Event, Error = ()> + Send>
+    where
+        Self: 'static,
+    {
+        let mut me = self;
+        Box::new(
+            input_rx
+                .map(move |event| {
+                    let mut output = Vec::with_capacity(1);
+                    me.transform_into(&mut output, event);
+                    futures01::stream::iter_ok(output.into_iter())
+                })
+                .flatten(),
+        )
     }
 }
 


Hoverbear:
Nice type level trickery. :)

PR #1955, opened 2020-02-28
feat(cli): Allow names in generate subcmd expression

Jeffail:
This updates the `vector generate` sub command expression to allow custom names for components rather than needing to manually change generated names such as `transform0` to something useful. The syntax is an optional prefix to the component type of the format `<name>:`, which looks like this:

`vector generate stuff:stdin/parser:regex_parser,cool_stuff:lua/console`

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -26,15 +26,21 @@ pub struct Opts {
     ///
     /// `stdin//http` prints a `stdin` source and an `http` sink.
     ///
+    /// Generated components are given incremental names (`source1`, `source2`,
+    /// etc) which should be replaced in order to provide better context. You
+    /// can optionally specify the names of components by prefixing them with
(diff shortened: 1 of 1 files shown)


src/generate.rs
```diff
@@ -85,11 +91,26 @@ fn generate_example(expression: &str) -> Result<String, Vec<String>> {
     if let Some(source_types) = components.get(0) {
         let mut sources = IndexMap::new();
 
-        for (i, source_type) in source_types.iter().enumerate() {
-            let name = format!("source{}", i);
+        for (i, source_expr) in source_types.iter().enumerate() {
+            let (name, source_type) = if let Some(c_index) = source_expr.find(':') {
```
Hoverbear: I feel like this is a situation where a `match` is a bit cleaner, but c'este la vie. :)

PR #1988, opened 2020-03-05
chore(file sink): Upgrade file sink for tokio-compat

MOZGIII:
This is the file sink upgraded for `tokio-compat` support. It's being developed atop of the #1922 and will be rebased to `master` when then branch is merged. For now, the base of the PR is set to the `tokio-compat` branch.

It relies heavily on the solution proposed by @LucioFranco at #1945.

I tried to preserve the tests as much as possible.

This PR also introduces the `ExpiringHashMap` - a component that's designed to be reusable. I decided to introduce it because I saw this logic implemented multiple times across the codebase. This should make it easier to add expiration functionality to the existing code where we just use `HashMap`s - for instance in places where we need per stream buffers, sockets, files and so on.

To do:

- [x] tests for `ExpiringHashMap` (don't forget: https://github.com/timberio/vector/pull/1988#discussion_r388540062)
- [x] polish the `ExpiringHashMap` interface
- [x] add documentation to the `ExpiringHashMap` - explain how `continuing` at the `None` case may result in a spin loop, and that to prevent this users MUST do `is_empty` check


MOZGIII:
Aaand just after I created a PR noticed a problem.

The `sinks::file::tests::many_partitions` now randomly fails.
Actually, it's not even random: `RUST_BACKTRACE=1 TEST_LOG=trace cargo test sinks::file` passes reliably, but `cargo test` fails reliably on the `sinks::file::tests::many_partitions`.
I this must be a race condition somewhere, but I don't quite get where. Something tells me it could be at `streaming_sink.rs` - where we do `tokio02::spawn`...

src/sinks/file/mod.rs
```diff
@@ -1,19 +1,29 @@
...
+use std::time::{Duration, Instant};
+use tokio02::{
+    fs::{self, File},
+    io::AsyncWriteExt,
+};
+
+mod bytes_path;
+use bytes_path::BytesPath;
+
+mod streaming_sink;
```
LucioFranco: We could probably place this stuff in `sinks/mod.rs`

src/sinks/file/mod.rs
```diff
@@ -88,98 +98,125 @@ impl PartitionedFileSink {
...
-            let ready = match partition.poll_complete() {
-                Ok(Async::Ready(())) => true,
-                Ok(Async::NotReady) => false,
-                Err(error) => {
-                    let file = String::from_utf8_lossy(&file[..]);
-                    error!(message = "Unable to flush file.", %file, %error);
-                    true
+    async fn run(&mut self, mut input: Receiver<Event>) -> crate::Result<()> {
+        loop {
+            let what = select(input.next(), self.files.next()).await;
```
LucioFranco: Can we use the new tokio select here?

src/sinks/file/mod.rs
```diff
@@ -88,98 +98,125 @@ impl PartitionedFileSink {
...
-                    self.closing.insert(key.clone(), file);
+                Either::Left((Some(event), _)) => {
+                    let path = match self.partition_event(&event) {
+                        Some(path) => path,
+                        None => {
+                            // We weren't able to find the path to use for the
+                            // file.
+                            // This is already logged at `partition_event`, so
+                            // here we just skip the event.
+                            warn!("Unable to partition an event, dropping it");
```
LucioFranco: Most likely want to rate limit this like we do in other sinks.

src/sinks/file/mod.rs
```diff
@@ -88,98 +98,125 @@ impl PartitionedFileSink {
...
+                            continue;
+                        }
+                    };
+
+                    let next_deadline = self.deadline_at();
+                    trace!(next_deadline = ?next_deadline, "Computed next deadline");
```
LucioFranco: for tracing we usually use the `trace!(message = "..", key = value);` pattern.

src/sinks/file/mod.rs
```diff
@@ -88,98 +98,125 @@ impl PartitionedFileSink {
...
+                    };
+
+                    trace!(path = ?path, "Writing an event to file");
+                    if let Err(err) = write_event_to_file(file, event, &self.encoding).await {
+                        error!(path = ?path, "Failed to write file: {}", err);
+                    }
+                }
+                Either::Right((None, _)) => {
+                    // Expiration queue is empty, whatever.
+                    debug!("File expiration queue empty");
```
LucioFranco: this would mean no events have been fed through prob make this a rate limited trace?

src/expiring_hash_map.rs
```diff
@@ -0,0 +1,118 @@
...
+    K: Eq + Hash + Clone,
+{
+    pub fn new() -> Self {
+        Self {
+            map: HashMap::new(),
+            expiration_queue: DelayQueue::new(),
+        }
+    }
+
+    pub fn insert(&mut self, key: K, value_with_ttl: (V, Duration)) {
```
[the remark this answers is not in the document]
MOZGIII: Sure. I'm planning to reshape this API in a bit. And add some tests, otherwise, I don't think it's a good idea.

src/expiring_hash_map.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    pub fn insert_at(&mut self, key: K, value_with_deadline: (V, Instant)) {
+        let (value, deadline) = value_with_deadline;
+        let delay = self
+            .expiration_queue
+            .insert_at(key.clone(), deadline.into());
+        self.map.insert(key, (value, delay));
+    }
+
+    pub fn get<Q: ?Sized>(&self, k: &Q) -> Option<&V>
```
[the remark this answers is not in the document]
MOZGIII: Hm, I agree. I based this off of the `HashMap` version of the same call. Wonder why is it like that there.

src/expiring_hash_map.rs
```diff
@@ -0,0 +1,118 @@
...
+            Some(Ok(key)) => key,
+        };
+        let (value, _) = self.map.remove(key.get_ref()).unwrap();
+        Poll::Ready(Some(Ok((value, key))))
+    }
+}
+
+impl<K, V> Stream for ExpiringHashMap<K, V>
+where
+    K: Unpin + Eq + Hash + Clone,
```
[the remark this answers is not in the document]
MOZGIII: Interesting, I'll try it. Tbh, while I know what `Unpin` does, I didn't spend enough time with it yet to be comfortable with the design decisions around it. This is a really valuable feedback!

src/sinks/file/mod.rs
MOZGIII: If we end up going with it - definitely! I think we can extract `bytes_path` into global utils too. I'll do a final pass on the locations of things before the merge.

src/sinks/file/mod.rs
MOZGIII: I haven't used `tokio::select`, but I've used `futures::select` before, and I try to keep away from them until they're absolutely necessary. The reason for that is they're huge macros, and significantly complicate the debugging process and other kinds of troubleshooting. Using `select` fn, on the other hand, is simple, and is only a minor syntax annoyance compared to a headache of dealing with complicated macros.
I'd prefer if we leave it as-is :smile:

src/sinks/file/mod.rs
MOZGIII: Rate limit the dropping? Could you link some relevant code or describe the desired logic here?

src/sinks/file/mod.rs
MOZGIII: I thought this has different semantics, and it looks like that's the case from what I see in the code. I'll make sure to add tests for it at `ExpiringHashMap`.

src/sinks/file/mod.rs
LucioFranco: Futures select is hacky while tokio's is much better and isn't 100% a proc macro, I'd suggest trying it out because the Either's everywhere is super hard to read :)

src/sinks/file/mod.rs
LucioFranco: Oh we can rate limit how often a log is logged https://github.com/timberio/vector/search?q=rate_limit_secs&unscoped_q=rate_limit_secs

src/expiring_hash_map.rs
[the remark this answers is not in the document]
LucioFranco: Yeah, its prob not going to affect us because almost all the types we may put in here will be `Unpin` but I think its ugly to keep the bound :)

src/sinks/file/mod.rs
MOZGIII: This `tracing` crate is damn good. I really like how formatted messages (`trace!("a: {}", b)`) are sort of discouraged by being only allowed at the end. :+1:

src/sinks/file/mod.rs
MOZGIII: I gave it a try, hopefully it'll be better.

As a side effect, I had to switch to nested matches, and now the nesting is real deep.
Before: `loop` > `match` (two levels)
After: `loop` > `tokio::select!`(two levels) -> `match` (two levels)

It's not critical, but is there a way to work around this?

src/sinks/file/mod.rs
```diff
@@ -88,98 +95,143 @@ impl PartitionedFileSink {
...
+                    match event {
+                        None => {
+                            // If we got `None` - terminate the processing.
+                            debug!(message = "Receiver exausted, terminating the processing loop.");
+                            break;
+                        }
+                        Some(event) => self.process_event(event).await,
+                    }
+                }
+                result = self.files.next() => {
```
LucioFranco: I can't remember but can we match on the left side?

```
Some(Ok(..)) = self.files.next() => ...,
Some(Err(...)) = self.files.next() => ...,
```
etc?

src/sinks/file/mod.rs
```diff
@@ -88,98 +95,143 @@ impl PartitionedFileSink {
...
-            partition
-                .start_send(event)
-                .map_err(|error| error!(message = "Error writing to partition.", %error))
-        } else {
-            Ok(AsyncSink::Ready)
+    async fn run(&mut self, input: impl Stream<Item = Event> + Send + Sync) -> crate::Result<()> {
+        pin_mut!(input);
+        loop {
+            tokio02::select! {
+                event = input.next() => {
```
LucioFranco: we could prob drop the extra brackets surrounding this.

src/sinks/file/mod.rs
```diff
@@ -88,98 +95,143 @@ impl PartitionedFileSink {
...
-        if (self.partitions.is_empty() && self.closing.is_empty()) || all_files_ready {
-            Ok(Async::Ready(()))
-        } else {
-            Ok(Async::NotReady)
-        }
+#[async_trait]
+impl StreamingSink for FileSink {
+    async fn run<'a>(
+        &'a mut self,
+        input: impl Stream<Item = Event> + Send + Sync + 'a,
```
LucioFranco: this life time should be inferred? Not sure why you need to add this?

src/sinks/file/streaming_sink/compat.rs
```diff
@@ -0,0 +1,83 @@
...
+            .run(stream)
+            .await
+            .expect("streaming sink error")
+    });
+
+    Box::new(sink)
+}
+
+/// Wraps any [`futures01::Sink`] with `SinkError = ()`.
+pub struct UnitTypeErrorSink01<T: futures01::Sink>(T);
```
LucioFranco: should be able to achieve this with https://docs.rs/futures01/0.1.29/futures01/sink/trait.Sink.html#method.sink_map_err and `sink_map_err(drop)`

src/sinks/file/streaming_sink/compat.rs
```diff
@@ -0,0 +1,83 @@
+use super::StreamingSink;
```
LucioFranco: What is the reasoning for moving this out of being on the streamingsink type?

src/sinks/file/mod.rs
MOZGIII: The syntax, just from the looks of it, should be causing a multi-mut error. I can try, but it surely looks cuter intuitive, and I doubt it'll work.

src/sinks/file/mod.rs
MOZGIII: Yep. I wonder why doesn't autoformatter kick in. I guess it's another drawback of using macros.
I'll correct it.

src/sinks/file/mod.rs
MOZGIII: Oh, it does. I wonder how I ended up with this.

src/sinks/file/streaming_sink/compat.rs
MOZGIII: I was looking for this! It can be hard sometimes to navigate the rust ecosystem.

src/sinks/file/streaming_sink/compat.rs
MOZGIII: The rationale is this code is more of a compatibility layer to the pre-async topology. This is partially outlined in the comments to the units inside the mod already. The idea is we can support this pull-based operation mode as a first-class at the topology layer, and it then we will be able to simply remove this mod. The `StreamingSink` type will remain though. For this reason, I think it's a good idea to separate them from the start by design.

If you are asking in regards to why use separate files - I just feel like it's cleaner to have multiple smaller files rather than a big one, especially if they have different concerns, like in this case.

src/sinks/file/mod.rs
MOZGIII: You are right! Tests show that it does indeed return `None` if the queue is empty. I think it is not a good fit for our use case.

If would make more sense to make it so `ExpiringHashMap::poll_expired` returns just `Poll<Result<ExpiredItem<K, V>, Error>>`, and for it to be pending if the map - and inner queue - are empty.
This will require us to add a custom waker, We'll return it if the queue returns `None`, and wake it when an element is inserted at an empty queue. It's a bit of overhead, but at least it's simple and better than making the consumer spinning on `None` - as what would be the case currently.

So, what do you think?

LucioFranco: That may work, can we use something like https://docs.rs/tokio/0.2.13/tokio/sync/struct.Notify.html?

src/sinks/file/mod.rs
LucioFranco: rustfmt can't format macro_rules unfortunately

src/expiring_hash_map/mod.rs
```diff
@@ -1,5 +1,6 @@
 use futures::ready;
 use futures::stream::Stream;
+use futures::task::AtomicWaker;
```
LucioFranco: We should try to avoid sync/task types from the futures crate for tokio based ones since those are loom checked.

src/expiring_hash_map/tests.rs
```diff
@@ -0,0 +1,92 @@
+use super::*;
+use crate::runtime::Runtime;
+use futures::future::poll_fn;
+use futures_test::task::{new_count_waker, noop_context, panic_context};
```
LucioFranco: We should use `tokio-test` here plus we can upstream things much easier.

src/expiring_hash_map/tests.rs
```diff
@@ -0,0 +1,92 @@
...
+#[test]
+fn poll_does_not_return_ready_with_empty_map() {
+    let mut map = ExpiringHashMap::<String, String>::new();
+    let mut cx = noop_context();
+    assert!(map.poll_expired(&mut cx).is_pending());
+}
+
+#[test]
+fn poll_does_not_return_ready_with_empty_map_after_it_was_non_empty() {
+    let mut rt = Runtime::new().unwrap();
```
LucioFranco: so for these types of tests we can use this type of type https://docs.rs/tokio-test/0.2.0/tokio_test/task/struct.Spawn.html, happy to chat about testing these things in tokio 0.2 over slack :)

Also the tests dir in the tokio repo is a great place to reference for testing async code.

src/expiring_hash_map/tests.rs
```diff
@@ -0,0 +1,92 @@
...
+    });
+}
+
+#[test]
+fn it_does_not_call_waker_if_polled_and_ready() {
+    let mut rt = Runtime::new().unwrap();
+    rt.block_on_std(async {
+        let mut map = ExpiringHashMap::<String, String>::new();
+
+        let a_minute_ago = Instant::now() - Duration::from_secs(60);
```
LucioFranco: We should also be using the built in test-util functionality to test timers, refer to the tokio time tests where we pause the time and advance it.

src/sinks/file/mod.rs
MOZGIII: I've already implemented this, check the last few commits! I used `futures::task::AtomicWaker`. Looks like it's a very similar thing, but I like the API a bit better - feel like it's more suited for our use case.

LucioFranco: Which one is? I don't see why we need to use a low level type like an atomic waker? In reality we probably shouldn't even expose a pollfn!

MOZGIII: On a second, since we don't `Pin` when we poll, maybe it's better to use `Notify`... I'll add some tests to see if we'll have issues with the current implementation, or if we have to switch to `Notfiy`. The problematic scenario is when the future if polled from multiple call-sites - but I doubt it can realistically happen with the mutability rules we impose. I can't think of a way to accurately prove the possibility or impossibility of it atm, but I'll spend some time thinking about it. Anyhow, my opinion is that we should provide as good test coverage as possible here.

Btw, I don't think atomic waker is more low-level than notify. Fundamentally they solve the same issues but adopting `Notify` for the pollfn might be too much work.
In that light, getting rid of raw pollfn and switching to exposing a future *may be* justified. However, estimating the amount of code we'll need for that, and anticipated changes in the API - I doubt it'll be worth it. The benefit is that we would move the whole implementation into a relatively simpler (due to stricter interfaces) domain of futures. It will introduce some runtime overhead, but, ultimately, I think we can ignore that if the interface is significantly better. But, well, that's the problem - I don't think we can make it better.

Or did you mean that we should just hide the pollfn from the interface and only leave stream impl? That would somewhat improve the API. But based on the fact that `DelayQueue` itself exposes pollfn - I don't think we should try to do something fancier, especially if we can. That is, if we can - but I'll add some more tests to check edge cases.

LucioFranco: We should basically try to avoid as much poll based fn as possible. This is what notify enables, it gives you a waker like api that you can just use directly in an async fn.

MOZGIII: I mean, we implement `Stream` here, and it works atop of a pollfn internally. Should we hide the custom pollfn but leave `Stream` impl, or just get rid of both? If so, what to add instead - a `Future` impl, or an fn that returns `Future`? I was thinking about the latter, and this is what we'll probably have to use if we want to use `Notify`.
So, essentially, the significantly different cases here are:
- to have `ExpiringHashMap` implement polling on itself somehow (via `Stream`, `Future` or custom pollfn);
- to add an fn to `ExpiringHashMap` that would return `Future` or `Stream`.

What are the differences between the two? I guess, a function would be better, cause it would be able to lock the required for polling. But then - we won't possibly have multiple wakers for sure, and therefore we won't need `Notify`, as `AsyncWaker` would suffice. Otherwise, `Notify` is useful, but we need to put the handles somewhere.

Wait. If we add an fn that returns a `Future` or `Stream`, and locks the `ExpiringHashMap` under a mut ref, we won't be able to add elements into it in the other select branches, right? I think we explicitly need the `ExpiringHashMap` to be pollable, not some derived object.

src/expiring_hash_map/tests.rs
MOZGIII: I saw those at the `tokio` repo, but, unfortunately, everything is `pub(create)` there. It would be nice to use that indeed.

src/expiring_hash_map/tests.rs
MOZGIII: This `Spawn` thing looks very interesting, I'll definitely check it out. It doesn't have usage examples though.

src/sinks/file/mod.rs
MOZGIII: This is what's happening if I remove `'a`.

```
error[E0309]: the parameter type `impl Stream<Item = Event> + Send + Sync` may not live long enough
   --> src/sinks/file/mod.rs:233:28
    |
233 |       ) -> crate::Result<()> {
    |  ____________________________^
234 | |         FileSink::run(self, input).await
235 | |     }
    | |_____^
    |
note: ...so that the type `impl core::future::future::Future` will meet its required lifetime bounds
   --> src/sinks/file/mod.rs:233:28
    |
233 |       ) -> crate::Result<()> {
    |  ____________________________^
234 | |         FileSink::run(self, input).await
235 | |     }
    | |_____^
help: consider adding an explicit lifetime bound  `'async_trait` to `impl Stream<Item = Event> + Send + Sync`...
    |
232 |         input: impl Stream<Item = Event> + Send + Sync + 'async_trait,
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

So, I think it makes sense. The default rules don't work because `async_trait` macro injects it's own `'async_trait` lifetime. In fact, this is sort of ok: we want our lifetime to be of not the future or it's pinned `self`, but of the owning object, and real `self`.

src/sinks/file/mod.rs
MOZGIII: It's now allowed. `first mutable borrow occurs here`, `second mutable borrow occurs here` - this kind of errors.

src/expiring_hash_map/tests.rs
MOZGIII: `tokio-test` and `futures-test` don't have feature-parity and are not swappable. I'm not sure if it's possible to use `tokio-test` with the same elegance.

src/expiring_hash_map/mod.rs
MOZGIII: This is a really valuable advantage of the `tokio` types indeed. I'll switch to `Notify`.

src/expiring_hash_map/tests.rs
MOZGIII: Nvm, it works pretty well. The approach is different, but I can implement the same assertions. :+1:

src/expiring_hash_map/tests.rs
MOZGIII: I created https://github.com/tokio-rs/tokio/issues/2313

src/sinks/file/mod.rs
MOZGIII: This branch got removed with a whole wrapping layer, so this is now obsolete.

src/expiring_hash_map/tests.rs
LucioFranco: What API's from clock do you need?

MOZGIII: Ideally all of them. I haven't looked thoroughly, but I think I'll need a clock-based `Instance` override there, and the calls to freeze time (`pause`, `advance`, etc).

My idea is that instead of practically waiting, we should freeze time, then shift time into the future. This, in turn, should trigger timer - and then we poll again, and the test should pass.

There should be a better way than doing `delay`s in tests, cause it's a problematic design for a whole lot of reasons. I'm already used to time freezing in more mature test ecosystems of other languages, so it really catches my attention. 😄

MOZGIII:
We're also blocked by #1922.

In addition to that, there's a API change to the `ExpiringHashMap` that I forgot to push apparently. One part of that, that's still pending is the documentation for `ExpiringHashMap`. It's important because the new API lets easily shooting yourself in the foot and get a spinlock if not properly used.
We've discussed the approach to the API with @LucioFranco, and the consensus is that we want to offer this API for now - instead of a more limited but safer one - due to the fact that we can't anticipate all the possible use cases. Another reason to provide API in this particular form is that it allows us to implement `poll_expired` as the async fn: `async fn poll_expired(...) -> Option<Result<...>>` . One of the alternative signatures was `fn poll_expired(...) -> Option<impl Future<Output = Result<...>>>`, it was imo better communicating the API contract, but @LucioFranco found the arguments to convince me we should go with the `async fn` variant. I want to document this decision-making process too (considered design options, tradeoffs and the rationales that contributed to the final decision). To some degree.

src/expiring_hash_map/mod.rs
MOZGIII: We changed the API to get rid of the need to have an internal `AsyncWaker`/`Notify`.

src/sinks/file/mod.rs
MOZGIII: I've implemented the API we settled on!

src/sinks/file/mod.rs
MOZGIII: It works, so I'll leave it for now.

MOZGIII:
@LucioFranco I've implemented the `ExpiringHashMap` API we settled on during the discussion. 

Please re-review!

src/expiring_hash_map/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    ///             "key".to_owned(),
+    ///             "val".to_owned(),
+    ///             Duration::from_millis(30),
+    ///         ),
+    ///     }
+    /// }
+    /// # });
+    /// ```
+    pub async fn next_expired(&mut self) -> Option<Result<ExpiredItem<K, V>, Error>> {
+        let key = self.expiration_queue.next().await?;
```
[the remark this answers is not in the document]
MOZGIII: Yes, it's a https://docs.rs/tokio/0.2.13/tokio/time/struct.Error.html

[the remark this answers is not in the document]
lukesteensen: Makes sense. This is a little bit of a bummer to expose since it's something we'd usually unwrap, but I'm fine with either as long as we make the nature of the possible error clear.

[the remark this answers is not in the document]
MOZGIII: Right! I forgot to add the errors section :frowning:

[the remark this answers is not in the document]
LucioFranco: huh this is odd I thought we removed errors for timers from tokio 0.2 :thinking:

src/expiring_hash_map/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+use futures::stream::StreamExt;
+use std::borrow::Borrow;
+use std::collections::HashMap;
+use std::fmt;
+use std::hash::Hash;
+use std::time::{Duration, Instant};
+use tokio02::time::{delay_queue, DelayQueue, Error};
+
+#[cfg(test)]
+mod tests;
```
LucioFranco: not-minor nit: the file isn't big enough to split these can we just have the tests and the map in the same file?

src/expiring_hash_map/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Check whether the [`ExpiringHashMap`] is empty.
+    /// If it's empty, the `next_expired` function immediately resolves to
+    /// [`None`]. Be aware that this may cause a spinlock behaviour if the
+    /// `next_expired` is polled in a loop while [`ExpiringHashMap`] is empty.
+    /// See [`ExpiringHashMap::next_expired`] for more info.
+    pub fn is_empty(&self) -> bool {
+        self.expiration_queue.is_empty()
+    }
+
+    /// If the [`ExpiringHashMap`] is empty, immediately returns `None`.
```
LucioFranco: great write up 👍

src/expiring_hash_map/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// # let mut rt = tokio::runtime::Runtime::new().unwrap();
+    /// # rt.block_on(async {
+    /// use vector::expiring_hash_map::ExpiringHashMap;
+    /// use std::time::Duration;
+    ///
+    /// let mut map: ExpiringHashMap<String, String> = ExpiringHashMap::new();
+    ///
+    /// loop {
+    ///     tokio::select! {
+    ///         // Avoid spin lock!
```
LucioFranco: I'd change this to say, you need to ensure that this branch is disabled if the map is empty. That is the more important thing to understand. That this select branch should be disabled.

src/expiring_hash_map/tests.rs
```diff
@@ -0,0 +1,89 @@
+use super::*;
+use std::task::Poll;
+use tokio02 as tokio;
+use tokio_test::{assert_pending, assert_ready, task};
+
+fn unwrap_ready<T>(poll: Poll<T>) -> T {
```
LucioFranco: I feel like using https://docs.rs/tokio-test/0.2.0/tokio_test/macro.assert_ready_eq.html would work well here, happy to add more asserts to tokio-test.

src/sinks/file/mod.rs
```diff
@@ -88,98 +95,137 @@ impl PartitionedFileSink {
...
+                            // If we got `None` - terminate the processing.
+                            debug!(message = "Receiver exausted, terminating the processing loop.");
+                            break;
+                        }
+                        Some(event) => self.process_event(event).await,
+                    }
+                }
+                result = self.files.next_expired(), if !self.files.is_empty() => {
+                    match result {
+                        None => unreachable!(),
```
LucioFranco: we should add a message here why this is unreachable

src/sinks/file/mod.rs
```diff
@@ -88,98 +95,137 @@ impl PartitionedFileSink {
...
+                        Some(Ok((mut expired_file, path))) => {
+                            // We got an expired file. All we really want is to
+                            // flush and close it.
+                            if let Err(error) = expired_file.flush().await {
+                                error!(
+                                    message = "Failed to flush file.",
+                                    path = ?path.get_ref(),
```
LucioFranco: it should borrow the path anyways, I don't think you need the `get_ref`?

src/sinks/file/streaming_sink/compat.rs
```diff
@@ -0,0 +1,66 @@
...
```
LucioFranco: I think it makes sense to just expose a struct that handles all of this instead of a bunch of functions for now. Lets try to keep it simple.

src/sinks/file/streaming_sink/mod.rs
```diff
@@ -0,0 +1,13 @@
+use crate::Event;
+use async_trait::async_trait;
+use futures::Stream;
+
+pub mod compat;
+
+#[async_trait]
+pub trait StreamingSink: Send + Sync + 'static {
+    async fn run<'a>(
```
LucioFranco: no need for `'a`

MOZGIII: See https://github.com/timberio/vector/pull/1988#discussion_r390264118

LucioFranco: Make `impl Stream + Send + Sync + 'static` then.

src/expiring_hash_map/tests.rs
MOZGIII: Ready eq requires specifying a type for expected `Ok`. I don't know how to work around that, but I'm happy to correct it if you show me the way.

LucioFranco: It shouldn't be working with result types? You can peek the internal impl here https://docs.rs/tokio-test/0.2.0/src/tokio_test/macros.rs.html#174 it just does basically what this does.

src/expiring_hash_map/mod.rs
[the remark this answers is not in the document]
MOZGIII: Do timers just panic now?
It might make sense to return errors from timers IMO, but I'm not aware of how the development unfolds in `tokio`. :smile:  Either way - it doesn't make sense to go ahead of the `tokio` changes. We'll just remove errors here if/when relevant `tokio` update arrives.

src/expiring_hash_map/mod.rs
MOZGIII: Good idea! I changed the comment accordingly.

src/expiring_hash_map/tests.rs
MOZGIII: I do `assert_eq!(unwrap_ready(fut.poll()).unwrap().unwrap().0, "val");`
I can't find a way to do the same with `assert_ready_eq`.
For instance:
- `assert_ready_eq!(fut.unwrap().unwrap().0, "val");` - down't work, cause the first argument has to be the future;
- `assert_ready_ok!(fut.unwrap().0, "val");` - down't work either for the same reason;
- `assert_ready_ok!(fut, Some(("val", ?));` - I'd love to write it in this form, but I don't know the value of the `Key` here, so it doesn't work either.

I'd be happy to improve anything there. The solution I settled on works, but I'm not very pleased with it.

src/sinks/file/streaming_sink/mod.rs
MOZGIII: Right! :) Applied this change.

src/sinks/file/mod.rs
MOZGIII: Good catch, thanks!

src/expiring_hash_map/mod.rs
MOZGIII: I'll change it since you asked. However, I personally think this style of keeping tests mod in an adjacent file is easier to maintain.
I don't think I should have a strong opinion on how to organize files yet, since I'm still relatively new in this project, and haven't touched/seen the majority of the code. But these minor details usually matter to me. On this particular topic, I'll just say I like how I did it more. :smile:

src/sinks/file/streaming_sink/compat.rs
MOZGIII: What do you mean? Something like `struct Compat`? Sorry, I don't understand what you're proposing here.

src/sinks/file/mod.rs
MOZGIII: Moving it upper now.

MOZGIII: Aborting, leaving it as-is for now, we'll correct it later.

MOZGIII:
Merging this to unlock the hotfix for the `console` sink.

Pending questions to be solved after the merge.

PR #2014, opened 2020-03-09
chore(logplex source): refactor to use util/HttpSource

bill-bateman:
Refactoring the logplex source to reuse code from the http source.

Mainly, this moves the creation of the warp server out of the logplex code.

### DIFF
--- a/src/sources/logplex.rs
+++ b/src/sources/logplex.rs
@@ -22,7 +22,7 @@ pub struct LogplexConfig {
     tls: Option<TlsConfig>,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Default)]
 struct LogplexSource {}
 
 impl HttpSource for LogplexSource {
@@ -43,7 +43,7 @@ impl SourceConfig for LogplexConfig {
         _: &GlobalOptions,
         out: mpsc::Sender<Event>,
     ) -> crate::Result<super::Source> {
-        let source = LogplexSource {};
+        let source = LogplexSource::default();
         source.run(self.address, "events", &self.tls, out)
     }
 
@@ -57,7 +57,7 @@ impl SourceConfig for LogplexConfig {
 }
 
 fn decode_message(body: FullBody, header_map: HeaderMap) -> Result<Vec<Event>, ErrorMessage> {
-    //deal with headers
+    // Deal with headers
     let msg_count = match usize::from_str(get_header(&header_map, "Logplex-Msg-Count")?) {
         Ok(v) => v,
         Err(e) => return Err(header_error_message("Logplex-Msg-Count", &e.to_string())),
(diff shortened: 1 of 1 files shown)


src/sources/logplex.rs
```diff
@@ -1,26 +1,40 @@
...
+use warp::filters::body::FullBody;
+use warp::http::{HeaderMap, StatusCode};
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct LogplexConfig {
     address: SocketAddr,
     tls: Option<TlsConfig>,
 }
 
+#[derive(Clone)]
```
Hoverbear:
```suggestion
#[derive(Clone, Default)]
```

src/sources/logplex.rs
```diff
@@ -29,52 +43,8 @@ impl SourceConfig for LogplexConfig {
...
-
-        info!(message = "building logplex server", addr = %self.address);
-
-        let tls = TlsSettings::from_config(&self.tls, true)?;
-        let incoming = MaybeTlsIncoming::bind(&self.address, tls)?;
-
-        let server = warp::serve(routes).serve_incoming_with_graceful_shutdown(incoming, tripwire);
-
-        Ok(Box::new(server))
+        let source = LogplexSource {};
```
Hoverbear:
```suggestion
        let source = LogplexSource::default();
```

src/sources/logplex.rs
```diff
@@ -86,7 +56,56 @@ impl SourceConfig for LogplexConfig {
     }
 }
 
-fn body_to_events(body: impl Buf) -> Vec<Event> {
+fn decode_message(body: FullBody, header_map: HeaderMap) -> Result<Vec<Event>, ErrorMessage> {
+    //deal with headers
```
Hoverbear:
```suggestion
    // Deal with headers
```

src/sources/logplex.rs
```diff
@@ -86,7 +56,56 @@ impl SourceConfig for LogplexConfig {
...
+    let msg_count = match usize::from_str(get_header(&header_map, "Logplex-Msg-Count")?) {
+        Ok(v) => v,
+        Err(e) => return Err(header_error_message("Logplex-Msg-Count", &e.to_string())),
+    };
+    let frame_id = get_header(&header_map, "Logplex-Frame-Id")?;
+    let drain_token = get_header(&header_map, "Logplex-Drain-Token")?;
+    info!(message = "Handling logplex request", %msg_count, %frame_id, %drain_token);
+
+    //deal with body
```
Hoverbear:
```suggestion
    // Deal with body
```

src/sources/logplex.rs
```diff
@@ -86,7 +56,56 @@ impl SourceConfig for LogplexConfig {
...
+    let events = body_to_events(body);
+
+    if events.len() != msg_count {
+        if cfg!(test) {
+            panic!("Parsed event count does not match message count header");
```
Hoverbear: Can we add the count information here too? I see it below!

We could probably just do the format once and let these share the message.

bill-bateman: Did this and the other suggestions in the new commit below
Thanks for the feedback 😄

PR #2018, opened 2020-03-10
docs: Fix for "download source code" step in install from source

yvasiyarov:
If you directly copy and paste provided command it will not work
mkdir -p vector && \
  curl -sSfL --proto '=https' --tlsv1.2 https://api.github.com/repos/timberio/vector/tarball/v0.8.X | \
  tar xzf - -C vector --strip-components=1
You need to replace  https://api.github.com/repos/timberio/vector/tarball/v0.8.X with correct version number (v0.8.2) in my case
I'm not sure if proposed solution is correct, but I do not know how to automatically substitute v0.8.X with correct version number

### DIFF
--- a/website/docs/setup/installation/manual/from-source.md
+++ b/website/docs/setup/installation/manual/from-source.md
@@ -77,7 +77,7 @@ The following steps should be used to compile Vector directly on Linux based sys
 
     ```bash
     mkdir -p vector && \
-      curl -sSfL --proto '=https' --tlsv1.2 https://api.github.com/repos/timberio/vector/tarball/v0.8.X | \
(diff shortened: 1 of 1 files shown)


yvasiyarov:
Looks like I've updated generated file. Please, ignore this PR

Hoverbear:
The source for this is at https://github.com/timberio/vector/blob/f07a3d22c9bba01612cc330abaff928df7dce8b2/website/docs/setup/installation/manual/from-source.md.erb#L69-L71 :)


In this case, you can see it was deliberate.

Thanks so much for opening this PR though! Please feel welcome to reach out if you have other issues!

PR #2023, opened 2020-03-10
enhancement(regex_parser transform): Add support for `target_field`

### LINKED ISSUE #1812: `regex_parser` transform doesn't work with nested fields

bruceg:
Closes #1812


src/transforms/regex_parser.rs
```diff
@@ -12,28 +12,21 @@ use std::collections::HashMap;
...
 use string_cache::DefaultAtom as Atom;
 
-#[derive(Deserialize, Serialize, Debug)]
+#[derive(Debug, Derivative, Deserialize, Serialize)]
+#[derivative(Default)]
 #[serde(default, deny_unknown_fields)]
 pub struct RegexParserConfig {
     pub regex: String,
     pub field: Option<Atom>,
+    #[derivative(Default(value = "false"))]
```
Hoverbear:
```suggestion
    #[derivative(Default(value = "true"))]
```

.meta/transforms/regex_parser.toml
```diff
@@ -39,4 +39,23 @@ description = """\
 The Regular Expression to apply. Do not include the leading or trailing `/`.\
 """
 
+[transforms.regex_parser.options.overwrite_target]
```
binarylogic: I don't think we need this option. Other transforms simply overwrite existing values, which I think is a fine default.

binarylogic:
Nice work. Everything else, docs wise, looks good to me.

src/transforms/regex_parser.rs
```diff
@@ -204,38 +224,29 @@ mod tests {
...
 
-    fn do_transform(
-        event: &str,
-        regex: &str,
-        field: Option<&str>,
-        drop_field: bool,
-        drop_failed: bool,
-        types: &[(&str, &str)],
-    ) -> Option<LogEvent> {
+    fn do_transform(event: &str, regex: &str, config: &str) -> Option<LogEvent> {
```
Hoverbear: 🎉

src/transforms/regex_parser.rs
bruceg: This follows the lead of other parser transforms like `json_parser` that default to `false`.

.meta/transforms/regex_parser.toml
bruceg: As indicated to @Hoverbear, this mirrors the `json_parser` transform. Since it is the only other parser transform that includes the `target_field` capability, I thought it made sense to copy its behavior as well.

src/transforms/regex_parser.rs
Hoverbear: Please update https://github.com/timberio/vector/blob/8c252a2c32fda4317284be46b70cf2e64e13d651/.meta/transforms/regex_parser.toml#L18 then! :)

.meta/transforms/regex_parser.toml
binarylogic: Ah, I wasn't aware we had an `overwrite_target` option. Ignore my comment then.

src/transforms/regex_parser.rs
bruceg: That would be the default for `drop_field`, which is a different setting.

Hoverbear: Yes, but in your branch:

image

image
In master:

image

https://github.com/timberio/vector/blob/9cd0ba6fca6b577c4c76956fc550496f5031175f/.meta/transforms/regex_parser.toml#L18

https://github.com/timberio/vector/blob/9cd0ba6fca6b577c4c76956fc550496f5031175f/src/transforms/regex_parser.rs#L30

lukesteensen (commented):
I'm curious how people feel about the `overwrite_target` option. I'm not sure it's useful enough to warrant the additional implementation complexity, but there may be use cases I'm not thinking about.

(I could have sworn we talked about this somewhere but now I can't find it)

binarylogic:
I'm indifferent, but I tend to agree it adds confusion to the UX as well. My opinion is not strong.

src/transforms/regex_parser.rs
bruceg: Ah I see it now, sorry. That was part of a lead up patch that I had forgotten about. Will fix.

bruceg:
So where are we with `overwrite_target`? Should I leave it the same as `json_parser`, make it default to `true`, or drop it entirely?

binarylogic:
Leave it and default it to `true`. It's fine as an advanced option. We should follow up by changing the default to `true` in the `json_parser` as well.

Hoverbear (approved):

binarylogic (approved):
Looks good!

binarylogic:
Merging so that we can get this in `0.9.0`.

PR #2025, opened 2020-03-10
enhancement(vector sink): Add support for TLS

### LINKED ISSUE #1993: Reorganize and centralize TLS implementations
There are many pieces of TLS implementations sprinkled throughout the sinks and sources. In particular, any sink or source that can handle TLS communication (or HTTPS) needs to be aware of this and have at least some difference in the control flow for TLS versus non-TLS connections. This should be formalized and centralized into `src/tls.rs` as a set of `MaybeTls` types and trait implementations that can operate in both modes. This will reduce the complexity of sinks and sources that may handle both types of connections.

### LINKED ISSUE #2024: Vector sink does not support TLS
In #1892 TLS support was added to the Vector source! This is awesome, but the Vector sink does not yet support TLS.

Fixing this issue means adding a TLS option to the Vector sink and making sure it works.

The options should match alongside https://vector.dev/docs/reference/sources/vector/ where possible.

bruceg:
Closes #2024

### DIFF
--- a/src/sources/vector.rs
+++ b/src/sources/vector.rs
@@ -25,12 +25,13 @@ fn default_shutdown_timeout_secs() -> u64 {
     30
 }
 
+#[cfg(test)]
 impl VectorConfig {
-    pub fn new(address: SocketListenAddr) -> Self {
+    pub fn new(address: SocketListenAddr, tls: Option<TlsConfig>) -> Self {
         Self {
             address,
             shutdown_timeout_secs: default_shutdown_timeout_secs(),
-            tls: None,
+            tls,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/sources/vector.rs
```diff
@@ -25,12 +25,13 @@ fn default_shutdown_timeout_secs() -> u64 {
     30
 }
 
+#[cfg(test)]
```
Hoverbear: This seems like it might have slipped in, is this intentional?

bruceg: Very intentional. The tests for both the sink and source are in this file, so this is required to test TLS support.

Hoverbear (approved):
Rest looks great!

src/sources/vector.rs
Hoverbear: Ok!

binarylogic (approved):
:shipit:

lukesteensen (approved):

PR #2027, opened 2020-03-10
docs: Fix for "download source code" step in install from source

yvasiyarov:
I've made some minor changes in "install from source code" document:
1. Remove 3-d step because its already included in 4-th step (mkdir -p vector)
2. Fixed download link used by curl, so you can paste it in console and its just work

### DIFF
--- a/website/docs/setup/installation/manual/from-source.md.erb
+++ b/website/docs/setup/installation/manual/from-source.md.erb
@@ -61,7 +61,7 @@ The following steps should be used to compile Vector directly on Linux based sys
 
     ```bash
     mkdir -p vector && \
-      curl -sSfL --proto '=https' --tlsv1.2 https://api.github.com/repos/timberio/vector/tarball/v<%= metadata.latest_version %> | \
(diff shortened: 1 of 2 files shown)


website/docs/setup/installation/manual/from-source.md.erb
```diff
@@ -67,7 +61,7 @@ The following steps should be used to compile Vector directly on Linux based sys
 
     ```bash
     mkdir -p vector && \
-      curl -sSfL --proto '=https' --tlsv1.2 https://api.github.com/repos/timberio/vector/tarball/v<%= metadata.latest_version.minor_x %> | \
```
Hoverbear: This was done as part of https://github.com/timberio/vector/issues/1255, @a-rodin was #1255 partially completed?

a-rodin: #1255 added `.X` versions for pre-compiled packages from packages.timber.io. However, the source archive is served from GitHub, so it doesn't seem to support arbitrary redirects. However, it seems like GitHub assets support URLs like https://api.github.com/repos/timberio/vector/tarball/v0.8 which return the most recent release. I think we need to use them instead.

a-rodin:
Thanks! I slightly modified the URL, so that now it always points to the latest patch release for the given major/minor version.

PR #2026, opened 2020-03-10
chore(statsd sink): Fix race condition in test

LucioFranco:
Closes #2016

PR #2035, opened 2020-03-11
chore(statsd sink): Reshuffle timeout in test

LucioFranco:
This reshuffles our timeout that was added in #2026 to correctly delay the write side of the test. This should give enough time to let the read side spawned task to poll for read interest. This should avoid a race condition where the read side has not gotten enough time to bind and poll for read interest by the time the write task has sent the packet. This is mostly an internal detail to this test and does not affect the actual sink.

### DIFF
--- a/src/sinks/statsd.rs
+++ b/src/sinks/statsd.rs
@@ -354,6 +354,14 @@ mod test {
 
         let stream = stream::iter_ok(events.clone().into_iter());
         let sender = sink.send_all(stream);
+        let deadline = Instant::now() + Duration::from_millis(100);
+
+        // Add a delay to the write side to let the read side
+        // poll for read interest. Otherwise, this could cause
+        // a race condition in noisy environments.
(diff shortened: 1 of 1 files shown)


MOZGIII:
I get it now. I was attempted a similar fix at my experimental branch: https://github.com/timberio/vector/commit/d47f5a54cb6298622699f27ad09f29664f2b8a70

It has a similar idea but attempts an explicit synchronization instead.

Still, I wasn't able to reproduce the bug with and without a fix. When I run the test in a loop it can run passing for many hours. But it has failed once or twice after a code recompilation.

@LucioFranco were you able to reproduce the crash on your end with any kind of reliability?

LucioFranco:
@MOZGIII I was not able to reproduce it at all, though my system wasn't that busy so I assume that was why. I chatted with some of the mio people and this is what they suggested. So we can try this and if it breaks again we can take a look again. Thanks for digging in though, this is one nuanced race condition!

PR #2068, opened 2020-03-14
fix(rename_fields transform): Improve rename_fields table handling

Hoverbear:
We discovered some corner cases around table handling in rename fields.

As you can see our docs show you can use tables instead of quoted strings, but if you did this you might find errors or even dropped data.

image

This PR adds more extensive support for this feature.

### DIFF
--- a/src/transforms/rename_fields.rs
+++ b/src/transforms/rename_fields.rs
@@ -1,18 +1,17 @@
 use super::Transform;
 use crate::{
     event::Event,
+    serde::Fields,
     topology::config::{DataType, TransformConfig, TransformContext, TransformDescription},
 };
 use indexmap::map::IndexMap;
 use serde::{Deserialize, Serialize};
-use snafu::Snafu;
 use string_cache::DefaultAtom as Atom;
-use toml::value::Value as TomlValue;
 
 #[derive(Deserialize, Serialize, Debug)]
 #[serde(deny_unknown_fields)]
 pub struct RenameFieldsConfig {
-    pub fields: IndexMap<String, TomlValue>,
+    pub fields: Fields<String>,
 }
 
 pub struct RenameFields {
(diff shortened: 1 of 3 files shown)


src/transforms/rename_fields.rs
```diff
@@ -44,11 +44,16 @@ impl TransformConfig for RenameFieldsConfig {
 
 impl RenameFields {
     pub fn new(fields: IndexMap<String, TomlValue>) -> crate::Result<Self> {
+        let fields =  fields
+            .into_iter()
+            .map(|kv| flatten(kv, None))
+            .collect::<crate::Result<Vec<_>>>()?;
+        let fields = fields.into_iter().flat_map(|v| v).map(|v| match v {
```
Hoverbear: Look at this expert double collect! 



This is mostly so we can emit errors properly (since `Result` implements `Iterator` it kind of gets janky otherwise)

src/transforms/rename_fields.rs
```diff
@@ -44,12 +44,19 @@ impl TransformConfig for RenameFieldsConfig {
...
+        let fields = fields
+            .into_iter()
+            .map(|kv| flatten(kv, None))
+            .collect::<crate::Result<Vec<_>>>()?;
+        let fields = fields
+            .into_iter()
+            .flat_map(|v| v)
+            .map(|v| match v {
+                (origin, TomlValue::String(new)) => Ok((origin.into(), new.into())),
+                (origin, _) => Err(FlattenError::CannotFlatten { key: origin }.into()),
```
a-rodin: Could the error handling be offloaded on Serde with assistance of the type system?

For example, it might be possible to set the type of `fields` in the config struct to a newly defined `Fields` type:

```rust
#[derive(Deserialize)]
struct Fields(IndexMap<String, FieldsOrValue<String>>);

#[derive(Deserialize)]
#[serde(untagged)]
enum FieldsOrValue {
    Fields(Fields),
    Value(String),
}
```

And implement `all_fields` method for `Fields` which would return a list of flattened fields and their corresponding values, without the need to check for errors.

Also, could this part of code be shared between `rename_fields` and `add_fields`, for example by placing it into a separate file such in `src/transforms/util`? In that case further changes such as adding support for escaping in the field path notation would require to change only a single place of code.

Here is a full example of the auxiliary types approach which is generic over the target value type: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9feba91b1777279144dc6c0630b13c74.

Hoverbear: I think that's a great idea, Alex!

Hoverbear:
@binarylogic do you know about this broken checklist?

binarylogic:
No, just ignore it for now.

Hoverbear:
@a-rodin can you have another look?

ghost (approved):

binarylogic:
Are there any tests, unit or behavior, for these changes?

binarylogic:
Mostly asking so I understand what documentation changes need to be made.

Hoverbear:
This just brought us inline with documentation, no updates needed.

PR #2082, opened 2020-03-17
chore: Add `*_flat` methods for log events

a-rodin:
Together with #2081 it is supposed to solve the issue of supporting fields with `.` and other special characters.

Not sure if `_flat` suffix is clear enough, or is there a better name for this. I considered `*_top_level` suffix, but discarded it as too long.

It would be interesting to @ktff and @bruceg.


src/event/mod.rs
```diff
@@ -99,14 +99,26 @@ impl LogEvent {
...
 
+    pub fn get_flat(&self, key: &Atom) -> Option<&Value> {
+        self.fields.get(key)
+    }
+
     pub fn get_mut(&mut self, key: &Atom) -> Option<&mut Value> {
         util::log::get_mut(&mut self.fields, key)
     }
 
+    pub fn get_mut_flat(&mut self, key: &Atom) -> Option<&mut Value> {
```
bruceg: I'm unsure what the right order would be, this or `get_flat_mut`

ktff:
As @bruceg said, it's not clear that this is useful. It at least can't be used in Kubernetes part because only the end of the key is flat, and not all.

ktff:
>  it's not clear that this is useful

Actually forget that, it is useful. Value just needs to be constructed by hand.

a-rodin:
@ktff 

> Value just needs to be constructed by hand.

Note that `Value` implements `From<JsonValue>`, so it might be possible to construct values directly from JSON, if it helps.

a-rodin:
@bruceg 
> I'm unclear where it is meant to be useful.

It is supposed to be used in cases when it is necessary to insert the field at the top level and ignore the special characters like `.` or `[`/`]`. For example, it seems to be userful in `journald` source where the event structure is flat. Currently `journald` source uses `from_records`, which uses `insert`, but because the records in `journald` are always flat, the implementation could have used `insert_flat` instead to improve performance and support names with dots.

a-rodin:
It is going to be used in the `lua` transform, so I'm merging this.

PR #2096, opened 2020-03-19
fix(console sink): Upgrade console sink for tokio 0.2

MOZGIII:
Signed-off-by: MOZGIII <mike-n@narod.ru>


src/sinks/console.rs
```diff
@@ -85,7 +88,44 @@ fn encode_event(mut event: Event, encoding: &EncodingConfig<Encoding>) -> Result
...
+        let output = &mut self.output;
+        pin_mut!(output);
+        pin_mut!(input);
+        loop {
+            let event = match input.next().await {
+                Some(v) => v,
+                None => break,
+            };
+            write_event_to_output(&mut output, event, &self.encoding).await?;
+        }
```
LucioFranco:
```rust
while let Some(event) = input.next().await {
	write_event_to_output(&mut output, event, &self.encoding).await?
}
```

Would look much better :)

MOZGIII:
> `FramedWrite` gave us some buffering on the output. Did we lose that here? Might be worth checking if throughput changed at all.

This is a good idea, I'll create an issue.

PR #2098, opened 2020-03-19
chore(topology): Make RunningTopology::stop() signal all sources to shut down using the new ShutdownCoordinator (1091)

stbrody:


tests/topology.rs
```diff
@@ -39,6 +43,63 @@ fn into_message(event: Event) -> String {
         .to_string_lossy()
 }
 
+#[test]
+fn topology_shutdown_while_active() {
```
stbrody: So turns out this test actually passes even without my changes, so it's not actually testing this change really.  Nevertheless I already had it written and it seemed worthwhile so I left it in to just add to our overall coverage.

stbrody:
Question: Since `ShutdownCoordinator` is only used for shutting down sources, should I rename it `SourceShutdownCoordinator`?  I think probably yes.

src/topology/mod.rs
```diff
@@ -88,6 +88,12 @@ pub fn validate(config: &Config, exec: runtime::TaskExecutor) -> Option<Pieces>
 }
 
 impl RunningTopology {
+    /// Sends the shutdown signal to all sources and returns a future that resolves
+    /// once all components (sources, transforms, and sinks) have finished shutting down.
+    /// Transforms and sinks should shut down automatically once their input tasks finish.
+    /// Note that this takes ownership of `self`, so once this function returns everything in the
+    /// RunningTopology instance has been dropped except for the `tasks` map, which gets moved
+    /// into the returned future and is used to poll for when the tasks have completed.
```
LucioFranco: Would be good to note here that once the returned future is dropped everything then is fully dropped.

src/topology/mod.rs
```diff
@@ -228,32 +248,40 @@ impl RunningTopology {
             to_remove_change_add(&self.config.sources, &new_config.sources);
 
         // First pass to tell the sources to shut down.
+        let mut source_shutdown_complete_futures = Vec::new();
+        // TODO: Once all Sources properly look for the ShutdownSignal, up this time limit to something
```
LucioFranco: Lets make sure we have issues for some of these TODOs

stbrody: Created https://github.com/timberio/vector/issues/2108 and https://github.com/timberio/vector/issues/2109

src/topology/mod.rs
stbrody: Done

stbrody:
> Question: Since `ShutdownCoordinator` is only used for shutting down sources, should I rename it `SourceShutdownCoordinator`? I think probably yes.

Not sure if you saw this question I posed above, but just FYI @LucioFranco, I went ahead and made this change.

PR #2101, opened 2020-03-20
chore: TLS reorganization

bruceg:
The `src/tls.rs` file was getting over large (more lines than `src/main.rs`) and hard to navigate. This reorganization has several parts:

1. Break up the source into multiple modules
1. Introduce a generic `MaybeTls` type, trait implementations, and type aliases using it.
1. In particular, adds a `MaybeTlsSettings` to replace the somewhat clunky `Option<TlsSettings>`.
1. Move in TLS related code from `sinks::util::tls`.

@LucioFranco, we had discussed this a while back, is this the right direction to be going? We also talked about using the `MaybeTls` framework to create a generic TCP connector. However, some of that has been subsumed by the `HttpSink` framework work. Do you think that's still relevant?

Ref #1993


src/sinks/util/tcp.rs
```diff
@@ -81,13 +81,11 @@ enum TcpSinkState {
     Backoff(Delay),
 }
 
-type TcpOrTlsStream = MaybeTlsStream<
-    FramedWrite<TcpStream, BytesCodec>,
-    FramedWrite<SslStream<TcpStream>, BytesCodec>,
->;
+type TcpOrTlsStream =
+    MaybeTls<FramedWrite<TcpStream, BytesCodec>, FramedWrite<SslStream<TcpStream>, BytesCodec>>;
```
LucioFranco: I believe we want this to look like:

`FramedWrite<MaybeTlsStream, BytseCodec>` where `MaybeTlsStream` is an enum abstracted over a Tcp or Tls stream, but `FramedWrite` shouldn't care since they expose the same `AsyncWrite/AsyncRead`

src/tls/incoming.rs
```diff
@@ -0,0 +1,108 @@
+use super::*;
```
LucioFranco: we should use full imports rather than globs so its a bit easier to track down types.

src/tls/maybe_tls.rs
```diff
@@ -0,0 +1,123 @@
...
+impl<R: AsyncWrite, T: AsyncWrite> AsyncWrite for MaybeTls<R, T> {
+    fn shutdown(&mut self) -> Poll<(), io::Error> {
+        match self {
+            Self::Tls(s) => s.shutdown(),
+            Self::Raw(s) => s.shutdown(),
+        }
+    }
+}
+
+impl<R, T, I, E> Sink for MaybeTls<R, T>
```
LucioFranco: Where is this used?

src/sinks/util/tcp.rs
bruceg: Ok, I can see how a `MaybeTlsStream` would be a more useful generic.

src/tls/maybe_tls.rs
bruceg: It's used in `sinks::util::tcp::TcpSink`, which uses the `TcpOrTlsStream` above.

src/sinks/util/tcp.rs
bruceg: Heh, actually, we already have that abstraction. Fixed.

src/tls/incoming.rs
bruceg: Done

src/tls/maybe_tls.rs
LucioFranco: Yeah, if we have a MaybeTlsStream then we can pass that to `FramedWrite` and that then exposes the sink api. 

So I think the most valuable thing here is to provide abstracted tcp connectors/acceptors that are wrappers around concrete tcp connect/listen/accept. That way we have only one path. We can provide a MaybeTlsAccept future as well that wraps the internals so our tcp source doesn't need any extra state for tls. There is no reason beyond configuring the tcp listener any of our sources/sinks should be concerned with handling tcp/tls state.

bruceg: I think this is done, with the exception of `test_util.rs` and our use of `hyper_openssl::HttpsConnector`. The former requires direct use of openssl, and the latter appears to be more complicated to internalize. All other uses of `openssl` and `tokio_openssl` have been consolidated in this tls lib.

lukesteensen:
> only thing that is bugging me is that sink impl that doesn't quite make sense to me but happy to move forward.

Could you be more specific? Do you think we should not need to implement `Sink` for `MaybeTls`?

LucioFranco:
> Could you be more specific? Do you think we should not need to implement Sink for MaybeTls?

TLS should be handled at the `AsyncRead`/`AsyncWrite` level instead of at the `Stream`/`Sink` level. This is because in reality we are only wrapping a `TcpStream` on either end (source or sink). Therefore we only need a `TcpOrTlsStream` type (which I call `MaybeTlsStream`). This type then gets fed into an adapter like `FramedWrite`/`FramedRead` which _then_ provides a `Stream`/`Sink` interface. This is why I don't think Tls should even implement `Stream` or `Sink`.

bruceg:
It turns out that with the most recent changes, the `Sink` implementation did become redundant. It was needed previously, but no more.

LucioFranco:
@bruceg thanks for pushing this all through!

PR #2106, opened 2020-03-20
chore: Racy buffering tests correction

### LINKED ISSUE #2102: `test_max_size` test is failing
`test_max_size` test is failing, we should investigate. This started happening more frequently with `tokio-compat` migration, however this migration is suspected not to be root cause, as it seems like this issue was happening before.
We need to investigate the problem with `test_max_size` and surrounding tests and fix it.

A sample failed CI run: https://circleci.com/gh/timberio/vector/90109

MOZGIII:
Closes #2102.

This solved the issue with racy buffer checking tests.
This, in turn, should fix our nightly builds! :tada:

### DIFF
--- a/tests/support/mod.rs
+++ b/tests/support/mod.rs
@@ -238,8 +238,10 @@ impl SinkConfig for MockSinkConfig {
             .sender
             .clone()
             .unwrap()
-            .stream_ack(cx.acker())
-            .sink_map_err(|e| error!("Error sending in sink {}", e));
+            .sink_map_err(
+                |error| error!(message = "Ingesting an event failed at mock sink", %error),
+            )
+            .stream_ack(cx.acker());
         let healthcheck = match self.healthy {
             true => future::ok(()),
             false => future::err(HealthcheckError::Unhealthy.into()),
--- a/tests/support/mod.rs
+++ b/tests/support/mod.rs
@@ -27,18 +27,16 @@ use vector::topology::config::{
 };
 use vector::transforms::Transform;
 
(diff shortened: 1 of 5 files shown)


tests/support/mod.rs
```diff
@@ -1,3 +1,8 @@
+// Using a shared mod like this is probably not the best idea, since we have to
+// disable the `dead_code` lint, as we don't need all of the helpers from here
+// all over the place.
+#![allow(dead_code)]
```
Hoverbear: I'm a bit confused. If we have dead code why do we not just remove it?

MOZGIII: The problem is this mod is used in tests, and each test is an individual compile unit. So, `tests/buffering.rs` complains about one set of fns as dead code, and `tests/topology.rs` doesn’t like the other. This is kind of a bad pattern to organize test helpers like that, but here we are.

MOZGIII: The problem is this mod is used in tests, and each test is an individual compile unit. So, `tests/buffering.rs` complains about one set of fns as dead code, and `tests/topology.rs` doesn’t like the other. This is kind of a bad pattern to organize test helpers like that, but here we are.

lukesteensen (approved):

LucioFranco (approved):
much better :+1:

tests/support/mod.rs
MOZGIII: @Hoverbear we'll need to refactor it at some point. I don't want to do it as part of this PR though since it's sort of a hotfix, and we better merge is asap.

Hoverbear: Agreed. :)

PR #2146, opened 2020-03-25
fix(networking): Fix handling of incoming TLS connections

### LINKED ISSUE #2123: bug(networking): TLS reorganization work broke TCP source
The TLS reorganization work in #2101 ended up breaking the TCP source tests, and I missed it. Reverting 67a75445 (Use MaybeTlsListener in TcpSource) fixes the problem, but undoes some of why we wanted the TLS reorganization in the first place. I have an idea what's going on and I'd like a second opinion if I have thought it through properly, because the fix isn't straightforward.

The `MaybeTlsIncoming` type (introduced in b22e4d39a60b3e35e605d1f3997c32e14886853d) is intended to be a replacement for the `TcpListener` class, returning a stream of connected and negotiated `MaybeTlsStream`. It does this, but only for one stream at a time. If a second incoming connection comes while it's negotiating the first, it won't accept it, as it is in the wrong state.

Am I on the right track, or have I missed something in the type layers? If I'm right, this also implies the `splunk_hec` and `logplex` sources also have issues, as they were the first users.

I see two ways of handling this, neither particularly appealing:

1. When working as a TLS acceptor, maintain a set of accepted streams within `MaybeTlsIncoming` and handle their negotiation in `poll` before returning any. This keeps the code outside of `src/tls` clean but adds a lot of state management to this class.

2. Force all other code to do the TLS negotiation step by turning `MaybeTlsIncoming::poll` into a trivial wrapper for `incoming.poll` and then handle the negotiation in the returned struct, with non-TLS negotiation being a straight return. This requires extra steps in everything that uses this, but keeps the TLS lib saner.

As this is a regression on master, I am inclined to revert first and ask questions later, but if there is a simple fix I'd rather do that instead.

bruceg:
Closes #2123


src/sources/util/tcp.rs
```diff
@@ -121,15 +135,8 @@ pub trait TcpSource: Clone + Send + 'static {
...
-                        accept_socket(
-                            span.clone(),
-                            socket,
-                            source,
-                            tripwire,
-                            host,
-                            out.clone(),
-                            tls.clone(),
-                        )
+                        debug!("accepted a new connection");
```
LucioFranco: we could include the remote addr here.

LucioFranco (approved):
LGTM!

src/sources/util/tcp.rs
bruceg: Good idea. The old code didn't report the remote address neither, but it's still a good idea.

LucioFranco: We might be able to append it to the span too via https://docs.rs/tracing/0.1.13/tracing/struct.Span.html#method.record

src/tls/incoming.rs
```diff
@@ -98,3 +105,101 @@ impl From<TcpListener> for MaybeTlsListener {
         }
     }
 }
+
+pub struct MaybeTlsIncomingStream<S> {
```
LucioFranco: Why is this called with `Incoming` on it? This should just be a `MaybeTlsStream`? It can store an Option<PeerAddr> to make it more flexible.

src/sources/socket/tcp.rs
```diff
@@ -62,9 +62,7 @@ impl TcpSource for RawTcpSource {
             &event::log_schema().host_key()
         };
 
-        if let Some(host) = host {
-            event.as_mut_log().insert(host_key.clone(), host);
-        }
+        event.as_mut_log().insert(host_key.clone(), host);
```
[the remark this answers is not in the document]
bruceg: This is just propagating the now effectively unconditional `peer_addr` down into callers. I'm not clear what should be optional here.

src/tls/incoming.rs
bruceg: So I initially rewrote the existing trivial `MaybeTlsStream` to handle the accept handshake case. However, this ended up complicating the sources with a lot (a dozen?) of `#[cfg]` flags due to the `Accepting` state variant never being constructed when TLS sources are compiled out. So, I restored the original `MaybeTlsStream` which handles the both a raw stream and a successfully negotiated TLS stream, and split off this incoming stream, which handles the negotiation and then drops into a stream state. It reduced the number of lines of code, so I figured it was a win.

src/sources/socket/tcp.rs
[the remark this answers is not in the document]
LucioFranco: Lets say you are proxing requests through something that would be the peer addr right? That isn't 100% the host that the log originated from and that could be confusing.

src/tls/incoming.rs
LucioFranco: Is this an issue to do with feature flags? if so I see that as a reason to remove them. Yeah, just seems odd that we need two types for this. It should be quite simple :/

LucioFranco (approved):
Happy to move forward and get this overwith.

src/tls/incoming.rs
bruceg: Even discounting the feature flags issue, splitting this reduced the lines of code a bit. The diff from before I added all the feature flag gates (so it nulls out adding and removing them):

```
 src/tls/incoming.rs | 123 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------
 src/tls/mod.rs      |  16 ++++++++--
 src/tls/outgoing.rs |   4 +--
 src/tls/stream.rs   | 135 -------------------------------------------------------------------------------------
 4 files changed, 130 insertions(+), 148 deletions(-)
```

LucioFranco: Yeah, mostly just find it confusing since for me my mental model is tcpstream wrapped by maybetlstream but with the incoming part it gets a bit fuzzy! but as long as it works I am happy.

lukesteensen (approved):

PR #2148, opened 2020-03-25
chore: add_fields, remove_fields, or rename_fields debug log on replaced or non-existing fields

### LINKED ISSUE #2044: Add rate-limited logs for the rename_fields transform
The `rename_fields` transform could use some _rate-limited_ logging to shed light on behavior:

- [x] Log a _rate-limited_ message, on the `INFO` level, when a field specified does not exist.
- [x] Log a _rate-limited_ message, on the `DEBUG` level, when a field overwrites an existing field.

### LINKED ISSUE #2045: Add rate-limited logs for the add_fields transform
The `add_fields` transform should include log messages that shed light on its behavior:

- [x] Add a _rate-limited_ log message, on the `DEBUG` level, if a field overwrites an existing field.

### LINKED ISSUE #2046: Add rate-limited logs for the remove_fields transform
The `remove_fields` transform should include log messages that shed light on its behavior:

- [x] Add a _rate-limited_ log message, on the `DEBUG` level, if a specified field does not exist.

Hoverbear:
Closes #2045, Closes #2046, Closes #2044 

Output a rate limited log on replaced or non-existing fields operated on by `add_fields`, `remove_fields`, or `rename_fields`

image

### DIFF
--- a/src/event/mod.rs
+++ b/src/event/mod.rs
@@ -119,7 +119,7 @@ impl LogEvent {
         self.fields.contains_key(key)
     }
 
-    pub fn insert<K, V>(&mut self, key: K, value: V)
+    pub fn insert<K, V>(&mut self, key: K, value: V) -> Option<Value>
     where
         K: AsRef<str>,
         V: Into<Value>,
--- a/src/event/util/log/insert.rs
+++ b/src/event/util/log/insert.rs
@@ -34,7 +34,7 @@ where
                 fields.insert(current, Value::Array(array))
             }
         }
-        _ => return None,
+        _ => None,
     }
 }
 
@@ -51,9 +51,7 @@ where
             while values.len() <= current {
                 values.push(Value::Null);
             }
-            let mut swapper = value;
-            std::mem::swap(&mut values[current], &mut swapper);
-            Some(swapper)
+            Some(std::mem::replace(&mut values[current], value))
         }
         (Some(PathComponent::Index(current)), Some(PathComponent::Key(_))) => {
             if let Some(Value::Map(map)) = values.get_mut(current) {
@@ -64,9 +62,7 @@ where
                 while values.len() <= current {
                     values.push(Value::Null);
                 }
-                let mut swapper = Value::Map(map);
-                std::mem::swap(&mut values[current], &mut swapper);
-                Some(swapper)
+                Some(std::mem::replace(&mut values[current], Value::Map(map)))
             }
         }
         (Some(PathComponent::Index(current)), Some(PathComponent::Index(next))) => {
@@ -78,12 +74,10 @@ where
                 while values.len() <= current {
                     values.push(Value::Null);
                 }
-                let mut swapper = Value::Array(array);
-                std::mem::swap(&mut values[current], &mut swapper);
-                Some(swapper)
+                Some(std::mem::replace(&mut values[current], Value::Array(array)))
             }
         }
-        _ => return None,
+        _ => None,
     }
 }
 
--- a/src/event/util/log/insert.rs
+++ b/src/event/util/log/insert.rs
@@ -2,41 +2,47 @@ use super::{Atom, PathComponent, PathIter, Value};
 use std::{collections::BTreeMap, iter::Peekable};
 
 /// Inserts field value using a path specified using `a.b[1].c` notation.
-pub fn insert(fields: &mut BTreeMap<Atom, Value>, path: &str, value: Value) {
-    map_insert(fields, PathIter::new(path).peekable(), value);
+pub fn insert(fields: &mut BTreeMap<Atom, Value>, path: &str, value: Value) -> Option<Value> {
+    map_insert(fields, PathIter::new(path).peekable(), value)
 }
 
-fn map_insert<I>(fields: &mut BTreeMap<Atom, Value>, mut path_iter: Peekable<I>, value: Value)
+fn map_insert<I>(
+    fields: &mut BTreeMap<Atom, Value>,
+    mut path_iter: Peekable<I>,
+    value: Value,
+) -> Option<Value>
 where
     I: Iterator<Item = PathComponent>,
 {
     match (path_iter.next(), path_iter.peek()) {
-        (Some(PathComponent::Key(current)), None) => {
-            fields.insert(current, value);
-        }
+        (Some(PathComponent::Key(current)), None) => fields.insert(current, value),
         (Some(PathComponent::Key(current)), Some(PathComponent::Key(_))) => {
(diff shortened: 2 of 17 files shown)


src/event/mod.rs
```diff
@@ -119,7 +119,7 @@ impl LogEvent {
         self.fields.contains_key(key)
     }
 
-    pub fn insert<K, V>(&mut self, key: K, value: V)
+    pub fn insert<K, V>(&mut self, key: K, value: V) -> Option<Value>
```
a-rodin: That's something that I considered to add when implementing nested fields! I decided against back then on the grounds that no component used it at the moment.

I like that the resulting API becomes more and more aligned with the APIs of the collections from the Rust's standard library, I think it should reduce cognitive overhead from using them.

src/sources/splunk_hec.rs
```diff
@@ -398,8 +398,12 @@ impl<R: Read> Stream for EventStream<R> {
...
                         match line {
                             // This don't quite fit the meaning of a event::schema().message_key
-                            JsonValue::Array(_) | JsonValue::Object(_) => log.insert("line", line),
-                            _ => log.insert(event::log_schema().message_key(), line),
+                            JsonValue::Array(_) | JsonValue::Object(_) => {
+                                log.insert("line", line);
+                            }
+                            _ => {
+                                log.insert(event::log_schema().message_key(), line);
+                            }
```
a-rodin: Just a matter of style, but it seems like the number of lines of code could have been reduced by putting `;` after the `match` statement rather than after each of the branches.

src/event/util/log/insert.rs
```diff
@@ -2,41 +2,47 @@ use super::{Atom, PathComponent, PathIter, Value};
...
+                array_insert(array, path_iter, value)
             } else {
                 let mut array = Vec::with_capacity(next + 1);
                 array_insert(&mut array, path_iter, value);
-                fields.insert(current, Value::Array(array));
+                fields.insert(current, Value::Array(array))
             }
         }
-        _ => return,
+        _ => return None,
```
bruceg: Is the `return` needed here? None of the other arms use `return`.

src/event/util/log/insert.rs
```diff
@@ -45,33 +51,39 @@ where
             while values.len() <= current {
                 values.push(Value::Null);
             }
-            values[current] = value;
+            let mut swapper = value;
+            std::mem::swap(&mut values[current], &mut swapper);
+            Some(swapper)
```
bruceg: Would this be better as `Some(std::mem::replace(&mut values[current], value))`?

src/event/util/log/insert.rs
```diff
@@ -45,33 +51,39 @@ where
...
             } else {
                 let mut map = BTreeMap::new();
                 map_insert(&mut map, path_iter, value);
                 while values.len() <= current {
                     values.push(Value::Null);
                 }
-                values[current] = Value::Map(map);
+                let mut swapper = Value::Map(map);
+                std::mem::swap(&mut values[current], &mut swapper);
+                Some(swapper)
```
bruceg: Ditto use of `replace` above

src/event/util/log/insert.rs
```diff
@@ -45,33 +51,39 @@ where
...
             } else {
                 let mut array = Vec::with_capacity(next + 1);
                 array_insert(&mut array, path_iter, value);
                 while values.len() <= current {
                     values.push(Value::Null);
                 }
-                values[current] = Value::Array(array);
+                let mut swapper = Value::Array(array);
+                std::mem::swap(&mut values[current], &mut swapper);
+                Some(swapper)
```
bruceg: Ditto use of `replace` above

src/event/util/log/insert.rs
```diff
@@ -45,33 +51,39 @@ where
...
             }
         }
-        _ => return,
+        _ => return None,
```
bruceg: Ditto drop the `return`?

src/transforms/add_fields.rs
```diff
@@ -97,7 +97,14 @@ impl Transform for AddFields {
                 .into(),
                 TemplateOrValue::Value(v) => v,
             };
-            event.as_mut_log().insert(key, value);
+            let old_val = event.as_mut_log().insert(&key, value);
+            if old_val.is_some() {
```
bruceg: Since you don't actually use the old value, this seems simpler:
`if let Some(_) = event.as_mut_log().insert(&key, value) {`
or if you prefer the `is_some`:
`if event.as_mut_log().insert(&key, value).is_some() {`

src/transforms/coercer.rs
```diff
@@ -60,7 +60,9 @@ impl Transform for Coercer {
             for (field, conv) in &self.types {
                 if let Some(value) = log.remove(field) {
                     match conv.convert(value) {
-                        Ok(converted) => new_log.insert(field, converted),
+                        Ok(converted) => {
+                            new_log.insert(field, converted);
+                        }
```
bruceg: These (and other related changes in your patch) could be reduced to a single line with `drop(new_log.insert(field, converted))` which avoids the need for a block statement. This may not be recommended though due to the confusion that the `drop` may add.

src/transforms/rename_fields.rs
```diff
@@ -60,8 +60,24 @@ impl Transform for RenameFields {
     fn transform(&mut self, mut event: Event) -> Option<Event> {
         for (old_key, new_key) in &self.fields {
             let log = event.as_mut_log();
-            if let Some(v) = log.remove_prune(&old_key, self.drop_empty) {
-                log.insert(new_key.clone(), v)
+            match log.remove_prune(&old_key, self.drop_empty) {
+                Some(v) => {
+                    let old_val = log.insert(new_key.clone(), v);
+                    if old_val.is_some() {
```
bruceg: See comment above about `if let Some...`

src/sources/splunk_hec.rs
Hoverbear: You're right!

src/event/mod.rs
Hoverbear: It's my secret goal to make `Event` just a serde_json, bytes, or a simdjson object, so that sounds about right. :)

src/event/util/log/insert.rs
Hoverbear: I think you're right

src/transforms/coercer.rs
Hoverbear: Kind of how I felt too. :S

Hoverbear:
Thanks @bruceg , reflected those. I'll merge after a green CI.

PR #2153, opened 2020-03-26
chore: Encoding config accessor visibility reduction

Hoverbear:
Fixes https://github.com/timberio/vector/issues/2008.

Reduces the accessor levels of the `EncodingConfig` struct to private and exposes public (crate) accessors.

This also structures the module hierarchy a bit more in line with convention.


src/sinks/util/encoding/encoding_config.rs
```diff
@@ -0,0 +1,142 @@
+use crate::sinks::util::encoding::{
+    encoding_config_with_default::EncodingConfigWithDefault, inner::Inner, EncodingConfiguration,
+    TimestampFormat,
+};
+use serde::de::{DeserializeOwned, IntoDeserializer, MapAccess, Visitor};
+use serde::{de, Deserialize, Deserializer, Serialize};
+use std::fmt;
+use std::fmt::Debug;
```
bruceg: Just a nit, but these can be combined: `use std::fmt::{self, Debug};` So could the `serde` lines, but that's will end up making it a larger block and is something we've done both ways. It would be nice if `rustfmt` would standardize that for us.

src/sinks/util/encoding/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// If `Some(_)` this configuration will filter out any field not listed.
+    fn only_fields(&self) -> &Option<Vec<Atom>>;
+    /// If `Some(_)` this configuration will filter out any field listed.
+    fn except_fields(&self) -> &Option<Vec<Atom>>;
+    /// If `Some(_)` this configuration will configure the timestamp output.
+    fn timestamp_format(&self) -> &Option<TimestampFormat>;
+
+    fn set_only_fields(&mut self, fields: Option<Vec<Atom>>) -> Option<Vec<Atom>>;
+    fn set_except_fields(&mut self, fields: Option<Vec<Atom>>) -> Option<Vec<Atom>>;
+    fn set_timestamp_format(&mut self, format: Option<TimestampFormat>) -> Option<TimestampFormat>;
```
bruceg: I see a single caller of the `.set_*` functions, and it doesn't make use of the return value. Would a builder pattern be better, or even an init struct pattern? I don't like the use of setter functions on something that will otherwise be constant during its lifetime.

src/sinks/util/encoding/inner.rs
```diff
@@ -0,0 +1,14 @@
...
+#[derive(Deserialize, Serialize, Debug, Eq, PartialEq, Clone)]
+pub struct Inner<E> {
+    pub(crate) codec: E,
+    #[serde(default)]
+    pub(crate) only_fields: Option<Vec<Atom>>,
+    #[serde(default)]
+    pub(crate) except_fields: Option<Vec<Atom>>,
+    #[serde(default)]
+    pub(crate) timestamp_format: Option<TimestampFormat>,
+}
```
bruceg: Since these are so tiny and only used by the corresponding config, I recommend moving these into the config sources. If I'm reading right that should eliminate the `pub(crate)` here and drop some `use` clauses.

src/sinks/util/encoding/encoding_config.rs
```diff
@@ -0,0 +1,142 @@
+use crate::sinks::util::encoding::{
```
bruceg: Our pattern has been to avoid the repetition of module names in filenames. I recommend `encoding/with_default.rs` and something like `encoding/without_default.rs` or `encoding/no_default.rs`.

src/sinks/util/encoding/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+        assert_eq!(config.encoding.codec(), &TestEncoding::Snoot);
+        assert_eq!(config.encoding.except_fields(), &Some(vec!["Doop".into()]));
+        assert_eq!(config.encoding.only_fields(), &Some(vec!["Boop".into()]));
+    }
+
+    const TOML_EXCLUSIVITY_VIOLATION: &str = "
+        encoding.codec = \"Snoot\"
+        encoding.except_fields = [\"Doop\"]
+        encoding.only_fields = [\"Doop\"]
+    ";
```
bruceg: You can simplify the quoting using a raw string literal:
```
const TOML_EXCLUSIVITY_VIOLATION: &str = r#"
        encoding.codec = "Snoot"
        encoding.except_fields = ["Doop"]
        encoding.only_fields = ["Doop"]
"#;
```

src/sinks/humio_logs.rs
```diff
@@ -36,29 +39,26 @@ pub enum Encoding {
...
     }
 }
 
 #[typetag::serde(name = "humio_logs")]
 impl SinkConfig for HumioLogsConfig {
     fn build(&self, cx: SinkContext) -> crate::Result<(super::RouterSink, super::Healthcheck)> {
-        if self.encoding.codec != Encoding::Json {
+        if self.encoding.codec() != &Encoding::Json {
             error!("Using an unsupported encoding for Humio");
         }
```
bruceg: I know this isn't directly relevant to this change set, but is this a useful test to be doing? Serde should have already ensured that the encoding can only be one of the valid variants in the enum.
 
I'm also a bit surprised Rust doesn't warn about this comparison being unconditionally false. Probably just a bit beyond its const checking.

src/sinks/util/encoding/encoding_config_with_default.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn set_timestamp_format(&mut self, format: Option<TimestampFormat>) -> Option<TimestampFormat> {
+        std::mem::replace(&mut self.timestamp_format, format)
+    }
+}
+
+impl<E> EncodingConfigWithDefault<E>
+where
+    E: Default + PartialEq,
+{
+    #[allow(dead_code)] // Required for `make check-component-features`
```
bruceg: I think this could be `#[cfg(feature = "sinks-humio_logs")]` since that's the only user I can see. That's bound to explode though if others need it.

bruceg:
Ref [init struct pattern](https://xaeroxe.github.io/init-struct-pattern/)

src/sinks/util/encoding/encoding_config.rs
Hoverbear: You're right!

src/sinks/util/encoding/mod.rs
Hoverbear: Right, that one was basically added for API consistency. I think a builder is better indeed!

src/sinks/humio_logs.rs
Hoverbear: Absolutely.

src/sinks/util/encoding/encoding_config_with_default.rs
Hoverbear: I don't want to spend more time waiting for ``make check-component-features` than I have to so lets leave it.

Hoverbear:
@bruceg I used the preferred builder pattern in the [book](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html#non-consuming-builders-(preferred))

src/sinks/util/encoding/with_default.rs
```diff
@@ -0,0 +1,217 @@
...
+        self
+    }
+}
+
+impl<E> EncodingConfigWithDefault<E>
+where
+    E: Default + PartialEq,
+{
+    #[allow(dead_code)] // Required for `make check-component-features`
+    pub(crate) fn transmute<X>(self) -> EncodingConfigWithDefault<X>
```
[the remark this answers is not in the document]
Hoverbear: Let's not spend more time fixing this again to make the feature test pass.

src/sinks/util/encoding/mod.rs
```diff
@@ -469,20 +170,19 @@ mod tests {
...
-        encoding = "Snoot"
-    "#;
-
     // TODO(2410): Using PathComponents here is a hack for #2407, #2410 should fix this fully.
     fn as_path_components(a: &str) -> Vec<PathComponent> {
         PathIter::new(a).collect()
     }
 
+    const TOML_SIMPLE_STRING: &str = r#"
+        encoding = "Snoot"
```
LucioFranco, line 179: boop

PR #2154, opened 2020-03-26
chore: Add LogEvent::new

MOZGIII:
This is an internal life quality improvement, allows us to get rid of doing `Event::new_empty_log().into_log()`.

### DIFF
--- a/src/event/discriminant.rs
+++ b/src/event/discriminant.rs
@@ -157,12 +157,8 @@ fn hash_null<H: Hasher>(hasher: &mut H) {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::event::Event;
+    use crate::event::LogEvent;
     use std::collections::{hash_map::DefaultHasher, HashMap};
-
-    fn new_log_event() -> LogEvent {
-        Event::new_empty_log().into_log()
-    }
 
     fn hash<H: Hash>(hash: H) -> u64 {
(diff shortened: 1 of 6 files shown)


a-rodin:
Could it make sense to also replace `Event::new_empty_log().into_log()` by `LogEvent::new()` throughout the codebase?

MOZGIII:
Definitely! But let's do it in another PR, there might be non-trivial transformations required, and it'll slow down the adoption. While if we just merge this and convert the rest later - it'll be already usable in the new code.
Plus, the concurrent PRs can't use this yet, and this is another reason why it'd be good to delay codebase-wide change a bit.

Hoverbear:
Would it not still be `Log(LogEvent::new())`? This PR would ideally migrate our existing code, or at least demonstrate usage once.

MOZGIII:
`Event::new_empty_log().into_log()` => `LogEvent::new()`

`Event::new_empty_log()` is still there, and you can now additionally do `Event::Log(LogEvent::new())` if you want to...

MOZGIII:
Converted some trivial code to use new notation to provide usage examples! :smile:

MOZGIII:
```rust
$ cargo check
    Finished dev [unoptimized + debuginfo] target(s) in 0.32s
```
Hmm, apparently `cargo check` without args is not good for doing what I expect it'd do...

PR #2162, opened 2020-03-26
feat(socket sink): Add support for UDP mode

### LINKED ISSUE #2090: Add `udp` mode to `socket` sink
Currently the `socket` sink only supports `tcp` and `unix` mode.

However the `socket` source supports `tcp`, `unix` and `udp`.

For symmetry I would have expected there to be a `udp` mode in the `socket` sink aswell.

The implementation could be similar to the udp implemenation already present in the `statsd` sink, but it would handle log events instead of metric events.

bruceg:
Closes #2090


binarylogic (approved):
Docs look great as always.

lukesteensen (approved):
Looks good!

PR #2220, opened 2020-04-04
chore(operations): RPM for aarch64 was using armv7

bfritz:
Please double-check me here.  I'm still getting my head around the build process and relying on naive pattern matching at this point.


a-rodin:
@bfritz Thanks for fixing this!

This does not affect the release binaries published on [vector.dev](https://vector.dev) and as GitHub assets though, because `tests/Makefile` is currently used only for local tests.

PR #2231, opened 2020-04-05
enhancement(ux)!: Use external tagging for metrics serialization

### LINKED ISSUE #2021: Use external tagging for metrics serialization
Currently metrics are serialized to JSON using [internal tagging](https://serde.rs/enum-representations.html#internally-tagged) for the metric type, like this:

```js
{
  "name": "login.count",
  "timestamp": "2019-11-01T21:15:47+00:00",
  "kind": "absolute",
  "tags": {
    "host": "my.host.com"
  },
  "value": {
    "type": "counter", // <-- metric type
    "value": 24.2
  }
}
```

where `value.value` is the metric type. I would argue that [external tagging](https://serde.rs/enum-representations.html#externally-tagged) might be easier to grasp and, in case of scripted transforms, easier to work with.

With external tagging the event above could have looked like this:

```js
{
  "name": "login.count",
  "timestamp": "2019-11-01T21:15:47+00:00",
  "kind": "absolute",
  "tags": {
    "host": "my.host.com"
  },
  "counter": { // <-- metric type
    "value": 24.2
  }
}
```

ghost:
This is a small change that closes https://github.com/timberio/vector/issues/2021.

One reason to have it is that it would make metrics representation in `stdout` sink consistent with it in Lua transform. However, it might break compatibility for those who rely on this serialization.


lukesteensen (approved):

binarylogic (approved):

a-rodin:
I've updated the "Schema" section in the metrics data model description, so now all kinds of metrics are documented as top-level fields. The fact that only one object of each kind can be present in a metric now mentioned in the beginning of the section.

Not sure is it possible or necessary to separate "one of" fields (`aggregated_histogram`, `aggregated_summary`, `counter`, `distribution`, `gauge`, `set`) from the rest of the fields (`name`, `timestamp`, `kind`, `tags`).

binarylogic (approved):
Looks good!

binarylogic:
> Not sure is it possible or necessary to separate "one of" fields (aggregated_histogram, aggregated_summary, counter, distribution, gauge, set) from the rest of the fields (name, timestamp, kind, tags).

I don't think so. Looks good to me.

PR #2234, opened 2020-04-06
fix(loki sink): Pass tls settings to healthcheck

LucioFranco:


lukesteensen (approved):
Is there a quick test we could add for this?

LucioFranco:
> Is there a quick test we could add for this?

I am not totally sure, we probably have this issue in other sinks. I think the best option here is to use the type system to ensure the only way we can generate TlsSettings is from our config? I think that would be much better than writing a test for each sink.

lukesteensen:
I mean the best way would be to stop creating http clients twice, but we still should be verifying things work properly no matter how we solve it.

I'd go ahead and merge this if you don't have an idea for a quick fix. I'll add an issue to audit other sinks for this issue.

PR #2238, opened 2020-04-06
fix(influxdb_metrics sink): Pass tls to healthcheck

### LINKED ISSUE #2237: Healthchecks should use the same http client
Ref #2234 

Currently, many of our healthchecks instantiate their own http client. This opens up the possibility for mismatched configs, etc.

We should go through each sink and ensure that, as much as possible, things like http clients are created in `build` and used for both the healthcheck and the sink itself.
For non-http-based sinks, we should at least double check that all connection settings are propogated correctly and look for ways to consolidate them.

### Todo

- [x] Construct `tcp healthcheck` with TLS.
  - Affects sinks: `vector`, `datadog`, `papertrail`, `tcp`, `socket`.
<p>

- [x] Build `HttpClient` once.
  - All of our http sinks now depend on our `util` so this is actually quite straight forward.
  - Affects sinks: `splunk_hec`, `loki`, `logdna`, `http`, ...
<p>

- [x] Build `aws` clients once.
  - Affects all `aws` sinks.

LucioFranco:
Closes #2237

### DIFF
--- a/src/sinks/influxdb_metrics.rs
+++ b/src/sinks/influxdb_metrics.rs
@@ -1235,6 +1235,7 @@ mod integration_tests {
         onboarding_v2();
 
         let mut rt = Runtime::new().unwrap();
+        let cx = SinkContext::new_test(rt.executor());
         let config = InfluxDBConfig {
             namespace: "ns".to_string(),
             endpoint: "http://localhost:9999".to_string(),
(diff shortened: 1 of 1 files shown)


src/sinks/influxdb_metrics.rs
```diff
@@ -219,11 +225,10 @@ impl InfluxDBSvc {
 
         let request = hyper::Request::get(uri).body(hyper::Body::empty()).unwrap();
 
-        let https = HttpsConnector::new(4).expect("TLS initialization failed");
-        let client = hyper::Client::builder().build(https);
+        let mut client = HttpClient::new(resolver, None)?;
```
lukesteensen, line 228: Do we not support the usual TLS options here?

LucioFranco: Looks like it, I didn't want to change the behavior this just consolidates all tls building to one part of the codebase.

PR #2259, opened 2020-04-08
fix(gcp_cloud_storage sink): Use kebab case for ACL encoding

a-rodin:
@alathon pointed out in the chat that GCS XML API, unlike the JSON one, uses kebab case instead of camel case.

This PR adds that change, although I haven't tested it myself.


alathon:
Is there any chance of this triggering a hotfix release, since the GCP sink doesn't work at all without this? Alternately, what's the timeline on releases?

PR #2261, opened 2020-04-08
chore(file source): Wire `ShutdownSignal`

ktff:
ref #2108.

- [x] Is there some nicer way in covering all possible ways of shutdown?

### DIFF
--- a/lib/file-source/Cargo.toml
+++ b/lib/file-source/Cargo.toml
@@ -16,7 +16,7 @@ indexmap = {version = "1.0.2", features = ["serde-1"]}
 flate2 = "1.0.6"
 winapi = { version = "0.3", features = ["winioctl"] }
 libc =  "0.2"
-futures-timer = "3.0.2"
+tokio = { version = "0.2.13", features = ["time"] }
 
 [dev-dependencies]
 quickcheck = "0.6"
--- a/lib/file-source/Cargo.toml
+++ b/lib/file-source/Cargo.toml
@@ -16,6 +16,7 @@ indexmap = {version = "1.0.2", features = ["serde-1"]}
 flate2 = "1.0.6"
 winapi = { version = "0.3", features = ["winioctl"] }
 libc =  "0.2"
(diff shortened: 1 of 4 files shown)


lib/file-source/Cargo.toml
```diff
@@ -16,6 +16,7 @@ indexmap = {version = "1.0.2", features = ["serde-1"]}
 flate2 = "1.0.6"
 winapi = { version = "0.3", features = ["winioctl"] }
 libc =  "0.2"
+futures-timer = "3.0.2"
```
LucioFranco, line 19: Please use tokio types.

ktff: Done

ktff:
@LucioFranco this PR has been updated and is ready for review.

lib/file-source/src/file_server.rs
```diff
@@ -282,10 +288,12 @@ impl FileServer {
             }
             let backoff = backoff_cap.saturating_sub(global_bytes_read);
 
-            match shutdown.recv_timeout(time::Duration::from_millis(backoff as u64)) {
-                Ok(()) => unreachable!(), // The sender should never actually send
-                Err(RecvTimeoutError::Timeout) => {}
-                Err(RecvTimeoutError::Disconnected) => return,
+            match block_on(select(
```
LucioFranco, line 291: This should panic, so I don't see how this works?

ktff: You are right it was panicking before for that reason, that's why the whole `file_server` is now executed with `spawn_blocking` instead of `thread_spawn`. https://github.com/timberio/vector/blob/bc2a2c1cfbab9ea45bd9c67f513c81ad85d39d14/src/sources/file/mod.rs#L304

LucioFranco: Ok, can we add a comment above then why this does work and something that should make it easy to understand why this might break in the future if we no longer use spawn_blocking.

PR #2283, opened 2020-04-09
enhancement(gcp_cloud_storage sink)!: Make ACL optional

### LINKED ISSUE #2253: Support not sending ACL for `gcl_cloud_storage` sink
Currently Vector always sets ACL for newly created objects in `gcp_cloud_storage` sink:

https://github.com/timberio/vector/blob/e4c383881b7727289a87327fcd081b1f366952a6/src/sinks/gcp/cloud_storage.rs#L302

If the [`acl`](https://vector.dev/docs/reference/sinks/gcp_cloud_storage/#acl) configuration option is not specified, the default value `projectPrivate` is used.

However, buckets with [uniform bucket-level access](https://cloud.google.com/storage/docs/uniform-bucket-level-access) don't support setting ACL for objects inside them (HTTP 400 error is returned).

I think the default behavior should be changed to not send any ACL at all. It is not clear would it breaking things for buckets without uniform bucket-level access though.

bruceg:
Closes #2253 

### DIFF
--- a/.meta/sinks/gcp_cloud_storage.toml
+++ b/.meta/sinks/gcp_cloud_storage.toml
@@ -67,8 +67,11 @@ type = "string"
 category = "Object Attributes"
 common = false
 required = false
-description = "Predefined ACL to apply to the created objects. For more information, see [Predefined ACLs][urls.gcs_predefined_acl]."
-default = "projectPrivate"
+description = """
+Predefined ACL to apply to the created objects. \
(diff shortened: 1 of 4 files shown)


a-rodin:
I think `!` should be added to the PR title to identify a breaking change.

ghost (approved):

binarylogic (approved):

PR #2290, opened 2020-04-10
chore(sources): Wire ShutdownSignal in `trait HttpSource`

ktff:
ref #2108 

Wires ShutdownSignal for:

- `http` source
- `logplex` source

### DIFF
--- a/src/sources/util/http.rs
+++ b/src/sources/util/http.rs
@@ -1,12 +1,13 @@
 use crate::event::Event;
-use crate::tls::{MaybeTlsSettings, TlsConfig};
+use crate::{
+    shutdown::ShutdownSignal,
+    tls::{MaybeTlsSettings, TlsConfig},
+};
 use futures01::{sync::mpsc, Future, IntoFuture, Sink};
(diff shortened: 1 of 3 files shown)


src/sources/util/http.rs
```diff
@@ -102,8 +99,10 @@ pub trait HttpSource: Clone + Send + Sync + 'static {
         let tls = MaybeTlsSettings::from_config(tls, true)?;
         let incoming = tls.bind(&address)?.incoming();
 
-        let server = warp::serve(routes).serve_incoming_with_graceful_shutdown(incoming, tripwire);
+        let server = warp::serve(routes)
+            .serve_incoming_with_graceful_shutdown(incoming, shutdown.clone().map(|_| ()));
 
-        Ok(Box::new(server))
+        // We need to drop the last copy of ShutdownSignalToken only after server has shutted down.
```
LucioFranco, line 105:
```suggestion
        // We need to drop the last copy of ShutdownSignalToken only after server has shut down.
```

PR #2294, opened 2020-04-10
chore(networking): Upgrade http client to hyper 0.13

LucioFranco:
This is the first step in our upgrade campaign. We are going to do things by breaking out upgrades into small components into files that have `2` appended at the end. Once, we have enough components upgraded to start upgrading sinks we can do this one by one and this work should be easy to parallelize. 

Commits are broken out into a copy commit and an upgrade commit.


src/dns.rs
```diff
@@ -118,6 +122,23 @@ impl Resolve for Resolver {
     }
 }
 
+impl Service<Name13> for Resolver {
```
lukesteensen, line 125: `Name13` sounds like a dystopian sci-fi movie.

src/sinks/util/http2.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused_imports)]
```
lukesteensen, line 1: Why do we need this?

Issue #1891, opened 2020-02-21
Benefits and downsides of `Atom`s


a-rodin:
In this issue I want to discuss are there still any real benefits in using [`Atom`](https://docs.rs/string_cache/0.8.0/string_cache/type.DefaultAtom.html)s as key values in [`LogEvent`](https://github.com/timberio/vector/blob/650ce582e2eab05f585505e5b66197eb672c556c/src/event/mod.rs#L41) instead of `String`s.

It seems to me that switching to `BTreeMap` (which uses actual strings comparisons instead of comparing hashes) and introduction of default log schemas (https://github.com/timberio/vector/pull/1769, which reduced use of static atoms) greately reduced benefits of using `Atom`s instead of plain `String`s. On the other hand, maintaining code working with `String`s directly is easier and some invocations of `clone` over the codebase could potentially be avoided with them.

lukesteensen:
This is a great question! They definitely add a bit of friction to the development process. Their `clone` is very cheap, so I wouldn't worry about that too much.

I'd be curious to see the effects on throughput due to memory use (duplicate string keys on the heap) and CPU (interning and reusing atoms vs allocating and deallocating strings). With some numbers, we'd be able to weigh ergonomics in the code vs any performance differences.

Hoverbear:
@a-rodin Would using Strings ease integration with Lua and Javascript transforms? Seems like it would for WASM/FFI in general

lukesteensen:
Some simple benchmarks show that we get about 6-7% more throughput by removing `Atom` in favor of `std::string::String`.

Better performance and better ergonomics make this a clear win, so we should do it.

jszwedko:
Closing since we dropped the use of `Atom`s due to poor performance.

PR #2295, opened 2020-04-10
perf: start removing Atom from the hot path

lukesteensen:
Ref #1891 

This removes `Atom` from our data model and the `grok_parser` transform. We can follow up with similar changes to other components, but this was sufficient for a ~15% throughput increase in a simple grok-heavy pipeline.

I also removed some unused methods `*_flat` that kept catching my eye for looking slow. Let me know if we should keep these around for any reason @a-rodin.


LucioFranco (approved):

lukesteensen:
I'm going to merge with the one approval for now, but please do still review when you get a chance :smile:

Hoverbear (commented):
Looks good!

PR #2297, opened 2020-04-11
chore(config): Remove `log_schema.source_key`

ktff:
ref. #2147 

Removes recently added `log_schema.source_key` because in the recent changes to #2147 it became unneeded.


ktff:
`make generate` has been applied, although it's modifying `website/guides/getting-started/deploying.md` without any reason that I can see.

PR #2298, opened 2020-04-11
enhancement(sources): Add `source_type` field

ktff:
ref. #2147 

Adds `source_type` field to sources:
- `docker`
- `http`
- `journald`
- `kafka`
- `kubernetes`
- `logplex`
- `socket`
- `splunk_hec`
- `stdin`
- `syslog`
- `file`


ktff:
This PR isn't changing anything in the sources, just adding simple lines, so I'll only leave one code owner to review it.

src/event/mod.rs
```diff
@@ -130,6 +130,18 @@ impl LogEvent {
...
+    where
+        V: Into<Value>,
+    {
+        if self.contains(key) {
+            Some(value)
+        } else {
+            self.insert(key.clone(), value);
+            None
+        }
+    }
```
bruceg, line 143: I don't see any of the callers of this method using the return value. If true, I think it would be better to drop both result values until they are needed.

Also, would using the `Entry` interface make sense? That is: ```self.entry(key.clone()).or_insert(value);``` I don't know what is more efficient here.

src/sources/file/mod.rs
```diff
@@ -327,6 +327,11 @@ fn create_event(
         event.as_mut_log().insert(host_key, hostname.clone());
     }
 
+    // Add source type
+    event
+        .as_mut_log()
+        .try_insert(event::log_schema().source_type_key(), "file");
+
```
bruceg, line 334: Would it make sense to add methods beside `Event::from` that add the source type right there? That would eliminate these repeated calls to insert the source type later.

src/sources/http.rs
```diff
@@ -54,6 +54,15 @@ impl HttpSource for SimpleHttpSource {
...
             .map(|events| add_headers(events, &self.headers, header_map))
+            .map(|mut events| {
+                // Add source type
+                for event in events.iter_mut() {
+                    event
+                        .as_mut_log()
+                        .try_insert(event::log_schema().source_type_key(), "http");
+                }
+                events
+            })
```
bruceg, line 65: I suggest moving the `event::log_schema().source_type_key()` lookup out of the loop here.

src/sources/logplex.rs
```diff
@@ -147,7 +147,14 @@ fn line_to_event(line: String) -> Event {
             fields = parts.len()
         );
         Event::from(line)
-    }
+    };
+
+    // Add source type
+    event
+        .as_mut_log()
+        .try_insert(event::log_schema().source_type_key(), "logplex");
```
bruceg, line 155: Having a two-parameter `Event::from` would eliminate this need for a `mut event`.

src/event/mod.rs
ktff: True, and regarding `Entry` we can't use `HashMap`'s `Entry` since the key can be composed of multiple parts so we need to go through `util::log::` methods.

src/sources/file/mod.rs
ktff: It would require more modifications/methods for a benefit of removing a function call at best. Although we could insert, instead of try_insert, immediately after that since we know what fields are present.

src/sources/logplex.rs
ktff: It is two parameter at the moment, but we would need to add a third one for `source_type_key` in the near future, so I don't think having a specialized `from` would save us a lot.

ktff: The same goes for specializing `LogEvent::new` and `Event::new_empty_log`.

src/event/mod.rs
bruceg: Right, of course.

src/sources/file/mod.rs
bruceg: Hmmm, ok, I had forgotten that the keys could/will be per-source. That makes it more complicated.

ktff:
> I reran the tests and eventually they succeeded.

Yep, `test_source_error`  and  `test_source_panic` have been failing for some time now in the master.

PR #2311, opened 2020-04-13
perf(operations): Set `-C codegen-units=1` for release builds

### LINKED ISSUE #1798: Set Rust flag `codegen-units` to 1 for release Vector builds
The results of tests performed by @bruceg in https://github.com/timberio/vector/issues/309#issuecomment-586505606 show that `codegen-units=1` Rust flag can reduce Vector binary size by 13%.

I think it is worth to add support for setting it to `scripts/build-archive.sh` script and enable it by default for release archives.

ghost:
Closes https://github.com/timberio/vector/issues/1798.

### DIFF
--- a/scripts/build-archive.sh
+++ b/scripts/build-archive.sh
@@ -20,6 +20,7 @@ RUST_LTO=${RUST_LTO:-}
 STRIP=${STRIP:-}
 FEATURES=${FEATURES:-}
 ARCHIVE_TYPE=${ARCHIVE_TYPE:-tar.gz}
+RUSTFLAGS=${RUSTFLAGS:--C codegen-units=1}
 
 if [ -z "$FEATURES" ]; then
     FEATURES="default"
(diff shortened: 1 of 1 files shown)


Hoverbear (approved):

bruceg (commented):
Should this go into `Cargo.toml` so that `make build` also gets this set?

a-rodin:
> Should this go into `Cargo.toml` so that `make build` also gets this set?

This is something worth to consider, as it makes sense to ensure that the same level optimizations is enabled for all builds, not only those we build using `make build-{archive,deb,rpm}`. But then it should also be applicable to Rust codegen LTO, as it is another "slow" optimization which improves performance and is enabled only in `build-archive.sh` script. Because of that I've opened  a separate issue https://github.com/timberio/vector/issues/2315.

a-rodin:
Superseded by #2321.

PR #2319, opened 2020-04-14
perf: batch disk buffer deletes

lukesteensen:
I'm having trouble getting consistent benchmark results for a "before" comparison, but this does seem to improve disk buffer throughput a bit.

### DIFF
--- a/src/buffers/disk.rs
+++ b/src/buffers/disk.rs
@@ -64,6 +64,7 @@ pub struct Writer {
     offset: Arc<AtomicUsize>,
     write_notifier: Arc<AtomicTask>,
     blocked_write_tasks: Arc<Mutex<Vec<Task>>>,
+    blocked_write_task_count: Arc<AtomicUsize>,
     writebatch: Writebatch<Key>,
     batch_size: usize,
     max_size: usize,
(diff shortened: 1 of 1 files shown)


src/buffers/disk.rs
```diff
@@ -227,9 +232,12 @@ impl Drop for Reader {
 
 impl Reader {
     fn delete_acked(&mut self) {
-        let num_to_delete = self.ack_counter.swap(0, Ordering::Relaxed);
+        let num_to_delete = self.ack_counter.load(Ordering::Relaxed);
+        let blocked_writers = self.blocked_write_task_count.load(Ordering::Relaxed);
```
LucioFranco, line 236: why relaxed?

lukesteensen: Because that's what we're already using. I'm concerned more with the overall perf impact right now than the specifics of the atomics ordering.

Hoverbear (approved):

lukesteensen:
The last commit fixed one correctness issue, but I fear more remain.

Hoverbear:
Yeah, I can definitely see that. :( Maybe work making this an opt-in beta knob?

lukesteensen:
This does not seem to be a good path forward. In addition to the correctness issue, it turns out there is very little difference when taking into account the way that our sinks do acking (they're already heavily batched).

PR #2341, opened 2020-04-15
chore: RFC #2341 - 2020-04-15 - WASM Plugins Support

### LINKED ISSUE #2280: WASM plugin architecture RFC

Hoverbear:
The RFC component of #2280, related to #2006 

Rendered: https://github.com/timberio/vector/blob/41d90b7df1995429d179c89a4f1c176d72a263f1/rfcs/2020-04-15-wasm-plugins.md


rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+
+Introduce Vector's foreign module support, a way for users to add custom functionality to Vector. This RFC introduces
+WASM modules specifically, but we envision the possibility of other implementations (these will be a different RFC).
+This RFC introduces Transforms and will be later amended with Source and Sink functionality.
+
+
+## Motivation
+
+There is a plurality of reasons for us to wish for flexible, practical foreign module support to Vector. These may also
+ be regarded as 'Plugins' or 'Extensions.' WASM is the most versatile and obvious choice for a first implementation as
```
a-rodin, line 11: > for a first implementation

Is this proviso necessary? It implies that there are some other implementations of modules in plans, but doesn't mention it.

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+
+Terraform typically runs on CIs and developer machines, making portability valuable. For Terraform, users can follow the
+[guide](https://www.terraform.io/docs/extend/writing-custom-providers.html) to write a Go-based provider, which they can
+then distribute as a portable binary. Notably: This means either distributing an unoptimized binary, distributing a lot
+of optimized binaries, or requiring folks build their own optimized binaries. Terraform doesn't care much about speed,
+so unoptimized binaries are acceptable.
+
+Vector is in a slightly different position than Terraform though! Vector runs primarily in servers and even end user
+machines. We can't expect users to have build tooling installed on their servers (it's a security risk!) and we
+definitely can't expect it on an end-user machines. Most people just aren't that interested in computers. Vector has
```
a-rodin, line 80: > Most people just aren't that interested in computers.

That's a curious observation!

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+
+* `Transform`: Modules of this role can be used as a transform `type`.
+
+  ```rust
+  use foreign_modules::hostcall;
+
+  // TODO: Add better FFI error handling!
+  #[no_mangle]
+  pub extern "C" fn process() -> usize {
+      if let Some(value) = hostcall::get("field").unwrap() {
```
a-rodin, line 239: Should manipulations on events, such as dropping events or emitting multiple events out of a single one, be supported by transforms too?

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+toolchain's `nightly` version:
+
+```bash
+rustup target add wasm32-wasi --toolchain nightly
+```
+
+Then, create a module:
+
+```bash
+cargo +nightly init --lib echo
```
a-rodin, line 300: Is it known when it would be possible to use `stable` instead of `nightly` for this? Although if Vector's SDK was "packaged" somehow, so that installation of the toolchain happened behind the curtain, then using nightly could have been less an issue, as it would be possible to just pin a particular version of the nightly in the corresponding scripts.

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+  #[no_mangle]
+  // TODO: Add better FFI error handling!
+  pub extern "C" fn process() {
+      if let Some(value) = hostcall::get("field").unwrap() {
+          hostcall::insert("field", value).unwrap();
+      }
+  }
+  ```
+
+* `Codec`:
```
a-rodin, line 274: The concept of codecs (as an entity which is separate from transforms) is really interesting. I feel like it might be worth to focus on it at first, as it is something that is easy to "sell" to a large market.

For example, it is possible to invest in providing a nice UX for people who just want to sent Protobufs with custom schema through Vector, but are not familiar with Rust/WASM/etc. If there was an easy way for them to put `.proto` file on the input and get a `.wasm` file on the output, and then plug this `.wasm` file into `vector.toml`, then they would get all the benefits of using protobufs, including native-level performance, without the need to know anything about compilers, toolchains, and APIs.

Furthermore, if we start from exposing publicly only this particular use case (`.proto` in, `.wasm` out), it would be possible to delay the stablization of the FFI APIs, as they could be kept internal while still being useful.

However, it seems like the codecs implementation should be somewhat different from transforms in order to make them fast. I think they should take batches of raw bytes on the input and yield Vector events on the output for the decoder case (or, correspondingly, Vector events on the input and batches of raw bytes on the output for the encoder case).

ghost (approved):
@Hoverbear This is a great RFC!

rfcs/2020-04-15-wasm-foreign-modules.md
Hoverbear: I think you're right. :)

rfcs/2020-04-15-wasm-foreign-modules.md
Hoverbear: I'd like to find API parity with the LUA transform, so yes. I think this can be reworked a bit. If we can make `Event` a `repr(C)` type that would help a lot. Let me experiment with this a bit?

rfcs/2020-04-15-wasm-foreign-modules.md
Hoverbear: I think we can use stable! :)

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+We should support modules so broadly?
+
+* As sinks? For this we should consider how we handle batching, partitioning, etc, more.
+  * **Temporary Conclusion:** Held off on deciding. Some interest.
+* As sources? This requires thinking about how we can do event sourcing.
+  * **Temporary Conclusion:** Held off on deciding. Some interest.
+* What about the idea of codecs being separate?
+  * **Temporary Conclusion:** Held off on deciding. Some interest.
+* This RFC makes some provisions for a future Event refactor, is it possible that this might happen?
+  * **Temporary Conclusion:** Held off on deciding. Some interest.
```
Jeffail, line 457: :+1: to codecs being separate plugin types. I'd prioritize as follows:

- A super-hot transform API, in the short term this can also be used to build codecs
- A codec API, and maybe a bit of cleanup with regard to how codecs are shared across the codebase

Then I would stop there and see how these features land before expanding into sources and sinks.

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+transform. Indeed, our initial implementation should include just a transform, as it's by far the easiest place to
+introduce this feature.
+
+During investigation of the initial POC we determined that our most desirable "blessed" language would be
+AssemblyScript, unfortunately, it's currently not able to run inside Lucet. Work is ongoing and we expect it to be
+possible soon.
+
+**Temporary Conclusion:** AssemblyScript and other languages we evaluated did not provide a way to package themselves as
+WASM modules, making it hard to integrate with them. We are planning to revisit this at a later date when the ecosystem
+is more mature.
```
Jeffail, line 472: I think this should be considered critical for v1. I doubt many users would be thrilled over the prospect of writing plugins in c, and writing them in rust means we're competing against simply forking and writing components natively, which given it has benefits in terms of portability might end up winning for some use cases.

I think the teams that need wasm plugins the most are also going to need a more approachable language.

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+  * https://github.com/bytecodealliance/wasmtime/pull/1216
+  * https://github.com/bytecodealliance/lucet/issues/442
+  * https://github.com/bytecodealliance/lucet/pull/437
+* [ ] FreeBSD (x86_64)
+  * https://github.com/bytecodealliance/lucet/pull/419
+  * https://github.com/bytecodealliance/lucet/pull/437
+
+**Temporary Conclusion:** The Lucet team desires to support other platforms in the near future. Since the lion's share
+of Vector usage is on X86_64 Linux, and this platform is already supported, we decided to adopt Lucet for now. We could
+consider adopting `wasmtime` in the future if lucet does not eventually reach it in terms of platform parity.
```
Jeffail, line 539: Worth noting that even though most deployment targets are linux it's possible that a lot of developer environments are macos, and this means only being able to test your local configs via virtualization.

Jeffail:
Looking good :+1:

rfcs/2020-04-15-wasm-foreign-modules.md
Hoverbear: Agreed! We can definitely support those through hostcalls, it's just a matter of figuring out an ergonomic API that works!

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+
+A guest module is not able to access the memory of the Vector instance, but Vector is free to modify the guest memory.
+For example, behind the scenes of a `get` call, the Guest must allocate some memory for Vector to write the result into.
+
+Due to the current semantics of the `vector::Event` type, passing data between WASM modules and the Vector host involves
+serialization to JSON. We're investigating ways to speed this up.
+
+* It may be possible to add a `repr(C)` flag to Event.
+* It may be worthwhile to explore an `Event` refactor.
+* Other serialization options may be much faster but less compatible.
```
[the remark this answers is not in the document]
Hoverbear: We could have `get::<u64>("tmp.foo")` but we might be able to send `Events` over FFI

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+
+
+### Initialization
```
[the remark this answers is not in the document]
Hoverbear: So the current definition (0.1) doesn't include one, but this is definitely something I'd like to support in a later version.

rfcs/2020-04-15-wasm-foreign-modules.md
```diff
@@ -0,0 +1,559 @@
...
+      if let Some(value) = hostcall::get("field").unwrap() {
+          hostcall::insert("field", value).unwrap();
+      }
+      0
+  }
+  ```
+
+> **TODO:** Sources, sinks, and codecs have no POC or formal specification, below is a work in progress.
+
+* `Source`: Modules of this role can be used as a source `type`.
```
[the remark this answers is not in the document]
Hoverbear: If we go with the option of having first class codecs, yes we should! But without the idea of first class codecs this would be hard.

rfcs/2020-04-15-wasm-foreign-modules.md
Hoverbear: Agree! Do you think a first step would be a `codec` transform?

rfcs/2020-04-15-wasm-foreign-modules.md
Hoverbear: So folks can always write in assemblyscript outside vector and build it into a `.wasm` then give it to vector. We just can't support inlined assemblyscript etc. So we can't support this immediately:

```toml
[transforms.foo]
type = "assemblyscript"
inline = """
foo = "bar"
"""
```

That'll need more work and be part of a later version.

rfcs/2020-04-15-wasm-foreign-modules.md
Hoverbear: While I don't disagree with you, I don't think "developers who code exclusively on Mac OS and deploy on Linux, but don't use a virtualization solution" is that huge a market to trade off the gains we get from lucet in performance and security.

From https://blog.rust-lang.org/2020/04/17/Rust-survey-2019.html:

image

binarylogic: Another reference point: https://insights.stackoverflow.com/survey/2019#technology-_-developers-primary-operating-systems

Image 2020-04-21 at 2 40 01 PM

rfcs/2020-04-15-wasm-foreign-modules.md
binarylogic: Can we use the new lua changes as inspiration? I feel like the `emit` function used there works well.

rfcs/2020-04-15-wasm-foreign-modules.md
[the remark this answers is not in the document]
binarylogic: I'm 👍 on first-class codecs. It makes a lot of sense.

rfcs/2020-04-15-wasm-foreign-modules.md
binarylogic: I think that's a reasonable first step, especially given that protobuf use case we're solving for.

rfcs/2020-04-15-wasm-foreign-modules.md
binarylogic: I think this is fine for v1.

rfcs/2020-04-15-wasm-foreign-modules.md
Hoverbear: Yeah that's basically what I want to copy. :)

Hoverbear: It works!

rfcs/2020-04-15-wasm-foreign-modules.md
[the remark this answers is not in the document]
Hoverbear: We refactored to ship raw JSON. :)

Hoverbear:
@lukesteensen @binarylogic this is ready for merge. :)

lukesteensen (approved):
Everything here sounds reasonable. Given the amount of engineering complexity and API surface area, I expect this feature to take a long time to settle. For the time being, I think we continue to focus on completing the thinnest possible vertical slice (parsing protobufs) and gain confidence in that before expanding the API at all. We have a lot of learning to do and getting out something super minimal to start playing with will help us do that.

Hoverbear:
This is still waiting for one approval :)

Hoverbear:
A little bird told me this is mergable so I'm merging it. :tada:

PR #2362, opened 2020-04-17
chore(networking): Upgrade `service` and `retries` in sink util

LucioFranco:
This is my second attempt at upgrading tower services internally. We now no longer vendor `Map` since that is now included in tower as `map_request`. We also no longer need to vendor timeout.

This is broken up into three commits 1) copy 2) upgrade 3) compat layer and re-enabling batch. I also had to disable some tests because of a fun rustc bug on stable that is fixed on nightly.


Cargo.toml
```diff
@@ -75,7 +75,7 @@ rusoto_sts = { version = "0.41.0", optional = true }
 
 # Tower
 tower = "0.1.1"
-tower03 = { package = "tower", version = "0.3" }
+tower03 = { package = "tower", version = "0.3", git = "https://github.com/tower-rs/tower", features = ["limit", "retry", "timeout", "util"] }
```
lukesteensen, line 78: Why the git dep?

LucioFranco: current release of 0.3 has some bounds issues that are fixed on master, we plan to release 0.4 soon with proper docs etc. So this is a place holder until then. The master version contains the layers we need as well that I up streamed.

LucioFranco: https://github.com/tower-rs/tower/issues/431

PR #2383, opened 2020-04-21
chore(operations): Template the CI file

Hoverbear:
Template out the CI file with data from our meta.


Hoverbear:
It seems we have some inconsistencies with our features, if I read the transforms etc we list, we see errors: 

* First one is that the feature is `sinks-datadog`
* Second one is that the feature is `sinks-gcp`

image

Hoverbear:
Need to test out tradeoffs around incremental and the cache.

Hoverbear:
Bumped into #2393

binarylogic:
Going to close this in favor of https://github.com/timberio/vector/pull/2440#issuecomment-619206191

PR #2500, opened 2020-04-29
enhancement(aws_ec2_metadata transform): Add `instance-type` field

szibis:
enhancement(aws_ec2_metadata transforms): add instance-type as one more element collected from ec2metadata

Instance-type missing and it is one of the minimal default metadata needed from EC2 instance-level via aws_ec2_metadata transformation.


LucioFranco:
@szibis looks like there are some compiler failures? Once these get fixed we can merge, if you need any help feel free to ping me :)

PR #2501, opened 2020-04-29
fix(kafka sink): Run entire healthcheck lazily

LucioFranco:

### DIFF
--- a/src/sinks/kafka.rs
+++ b/src/sinks/kafka.rs
@@ -209,9 +209,11 @@ impl Sink for KafkaSink {
 }
 
 fn healthcheck(config: KafkaSinkConfig) -> super::Healthcheck {
-    let consumer: BaseConsumer = config.to_rdkafka().unwrap().create().unwrap();
+    let client = config.to_rdkafka().unwrap();
 
     let check = future::lazy(move || {
+        let consumer: BaseConsumer = client.create().unwrap();
+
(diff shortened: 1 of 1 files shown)


src/sinks/kafka.rs
```diff
@@ -209,9 +209,11 @@ impl Sink for KafkaSink {
 }
 
 fn healthcheck(config: KafkaSinkConfig) -> super::Healthcheck {
-    let consumer: BaseConsumer = config.to_rdkafka().unwrap().create().unwrap();
+    let client = config.to_rdkafka().unwrap();
```
lukesteensen, line 212: This is technically a config, not a client, but probably not a big deal.

lukesteensen (approved):

src/sinks/kafka.rs
LucioFranco: Yeah, probably not but doesn't hurt. I think its better to just enforce that we are doing it lazily.

bruceg: I think @lukesteensen was commenting on the variable name.

lukesteensen: Correct :smile:

bruceg (approved):

src/sinks/kafka.rs
LucioFranco: hahah oops yeah

PR #2532, opened 2020-05-03
enhancement(elasticsearch sink): Add explicit AWS region

bruceg:
Ref https://github.com/timberio/vector/pull/2412#issuecomment-618470561

This is the final part of the Elasticsearch sink AWS fix. It adds support for an explicit AWS region, and defaults (via the rusoto crate) to `$AWS_REGION`, `$AWS_DEFAULT_REGION`, or `"us-east-1"`.

The default region behavior is also extended to all other AWS sinks when configured using an endpoint. I don't think this should be a breaking change, but I'd like confirmation.

### DIFF
--- a/src/region.rs
+++ b/src/region.rs
@@ -65,7 +65,9 @@ impl TryFrom<RegionOrEndpoint> for Region {
 /// Translate an endpoint URL into a Region
 pub fn region_from_endpoint(endpoint: &str) -> Result<Region, ParseError> {
     let uri = endpoint.parse::<Uri>().context(EndpointParseError)?;
-    let name = region_name_from_host(uri.host().unwrap_or(""))
+    let name = uri
+        .host()
+        .and_then(region_name_from_host)
         .unwrap_or_else(|| Region::default().name().into());
     let endpoint = strip_endpoint(&uri);
     Ok(Region::Custom { name, endpoint })
--- a/src/region.rs
+++ b/src/region.rs
@@ -65,29 +65,29 @@ impl TryFrom<RegionOrEndpoint> for Region {
 /// Translate an endpoint URL into a Region
 pub fn region_from_endpoint(endpoint: &str) -> Result<Region, ParseError> {
     let uri = endpoint.parse::<Uri>().context(EndpointParseError)?;
-    let name = region_name_from_host(uri.host().unwrap_or(""));
+    let name = region_name_from_host(uri.host().unwrap_or(""))
+        .unwrap_or_else(|| Region::default().name().into());
+    let endpoint = strip_endpoint(&uri);
+    Ok(Region::Custom { name, endpoint })
+}
 
-    // Reconstitute the endpoint from the URI, but strip off all path components
+/// Reconstitute the endpoint from the URI, but strip off all path components
+fn strip_endpoint(uri: &Uri) -> String {
     let pq_len = uri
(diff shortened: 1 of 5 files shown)


binarylogic (approved):
Docs and options look good to me.

src/region.rs
```diff
@@ -65,29 +65,29 @@ impl TryFrom<RegionOrEndpoint> for Region {
 /// Translate an endpoint URL into a Region
 pub fn region_from_endpoint(endpoint: &str) -> Result<Region, ParseError> {
     let uri = endpoint.parse::<Uri>().context(EndpointParseError)?;
-    let name = region_name_from_host(uri.host().unwrap_or(""));
+    let name = region_name_from_host(uri.host().unwrap_or(""))
+        .unwrap_or_else(|| Region::default().name().into());
```
lukesteensen, line 69: **nit**: Could this be something like
```rust
let name = uri.host()
    .map(region_name_from_host)
    .unwrap_or_else(|| Region::default().name().into());
```
The flow is a little bit funky right now with the two unwraps and passing `""` to `region_name_from_host`.

src/region.rs
```diff
@@ -65,29 +65,29 @@ impl TryFrom<RegionOrEndpoint> for Region {
...
+fn strip_endpoint(uri: &Uri) -> String {
     let pq_len = uri
         .path_and_query()
         .map(|pq| pq.as_str().len())
         .unwrap_or(1);
     let endpoint = uri.to_string();
-    let endpoint = endpoint[..endpoint.len() - pq_len].to_string();
-
-    Ok(Region::Custom { name, endpoint })
+    endpoint[..endpoint.len() - pq_len].to_string()
```
lukesteensen, line 81: Probably not a problem, but do we need to worry about multi-byte characters with the string indexing here?

src/sinks/elasticsearch.rs
```diff
@@ -291,7 +293,10 @@ impl ElasticSearchCommon {
         };
 
         let base_url = config.host.clone();
-        let region = region_from_endpoint(&config.host)?;
+        let region = match &config.aws {
+            Some(region) => Region::try_from(region)?,
+            None => region_from_endpoint(&config.host)?,
```
lukesteensen, line 298: Are there cases where this would incorrectly assume you're talking to the AWS service based on the hostname? E.g. would it try to use AWS authentication when talking to `us-east-1.lukes-es-service.com`?

lukesteensen (approved):
Looks good! A couple of notes and questions but I don't think anything that needs to block merging.

src/region.rs
bruceg: Oh, yeah, definitely, although it needs to be `and_then` instead of `map`. The logic above grew out of a simpler expression in a couple of steps.

src/region.rs
bruceg: I don't see how. Both lengths are already proper indexes if the substrings contain multi-byte characters. This is just getting the byte index at which the hostname ends.

src/sinks/elasticsearch.rs
bruceg: No. The choice of using AWS authentication is guarded by the `auth.strategy` configuration setting. This is just figuring out what region to use if it gets there.

lukesteensen (approved):

PR #2541, opened 2020-05-05
chore: Generalize file server around paths provider

MOZGIII:
This PR generalizes the `file_source::FileServer` around the ways it obtains the lists of paths to work with.

The rationale for this change is we'll be using an alternative way of obtaining the lists of paths at the k8s implementation - it'll be based on the continuously updated data obtained from the k8s API.

An alternative to this change was to implement a new solution for file reading from scratch. I discouraged that because we already have an excellent implementation that can be generalized to fit the needs.

I was cautious to limit the scope of the change to the minimum, to decrease the effect on the other components.

### DIFF
--- a/lib/file-source/src/paths_provider/glob.rs
+++ b/lib/file-source/src/paths_provider/glob.rs
@@ -1,0 +1,123 @@
+//! [`Glob`] paths provider.
+
+use super::PathsProvider;
+
+use glob::Pattern;
+use std::path::PathBuf;
+
+pub use glob::MatchOptions;
+
+/// A glob-based path provider.
+///
+/// Provides the paths to the files on the file system that match include
+/// patterns and don't match the exclude patterns.
+pub struct Glob {
+    include_patterns: Vec<String>,
+    exclude_patterns: Vec<Pattern>,
+    glob_match_options: MatchOptions,
+}
+
+impl Glob {
+    /// Create a new [`Glob`].
+    ///
+    /// Returns `None` if patterns aren't valid.
+    pub fn new(
+        include_patterns: &[PathBuf],
+        exclude_patterns: &[PathBuf],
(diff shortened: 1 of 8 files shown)


lib/file-source/src/paths_provider/glob.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Iterator for [`Glob`].
+pub struct Iter<'a> {
+    include_patterns_iter: std::slice::Iter<'a, String>,
+    exclude_patterns: &'a [Pattern],
+    glob_match_options: &'a glob::MatchOptions,
+    current_glob_iter: Option<glob::Paths>,
+}
+
+impl<'a> Iterator for Iter<'a> {
```
lukesteensen, line 66: It seems like we're taking on a decent amount of additional complexity to implement `Iterator` instead of just building a `Vec`. Given that we can only expose `Vec` via the API, why not go for a simpler implementation?

MOZGIII: I've pushed an alternative implementation that's more efficient than `Vec`. It has some issues though: it's definitely more complex and verbose. But the good thing is it doesn't allocate the files list in memory.
I'm considering ditching that implementation and switching back to `Vec`, but simpler, to trade-off the code simplicity for the runtime performance.
What do you think?

lukesteensen: While this is a very cool implementation, I don't think this functionality is performance-sensitive enough to justify the additional complexity. I would rather start with something that minimizes our maintenance overhead and then optimize if/when this code is determined to be a bottleneck.

MOZGIII: Sounds good! I’m going to replace the custom iter impl with a composition of combinators.

PR #2558, opened 2020-05-07
chore(topology): Test `vector` shutdown

### LINKED ISSUE #1089: Test shutdown works via `topology::stop`

### LINKED ISSUE #2551: Enable `ShutdownSignal`
Our stop procedure is still dropping source handles, so `ShutdownSignal` has no effect. 

The solution is a simple one line, but I would propose to enable this after we have #1089 for safe measure. 

Blocked on #1089.

ktff:
Closes #1089
Closes #2551

Adds tests for shutdown triggered by:
 - all sources have finished
 - process signal

Also adds a fix for `journald` source. `journald` source wasn't shutting down in a timely matter. It was caused by the internal `blocking_future` which has a blocking read on an output channel of a child process. The fix sends a `SIGTERM` signal to the child process so to unblock the future. One additional consequence is that `journald` is now behind `unix` feature, which shouldn't change anything since the source only works on `Linux` already.

Also for the above fix to work, we've stopped dropping the source future so `ShutdownSignal` now has a full effect.

###  Todo
 - [x] fix shutdown bug for `journald` 
 - [x] Enable tests for sources that require docker services


ktff:
>May 08 11:05:28.428  WARN vector::sources::journald: The `units` setting is deprecated, use `include_units` instead

Just checked this, it's a bug in `journald`. I'll add a separate PR for that.

tests/shutdown.rs
```diff
@@ -0,0 +1,290 @@
+#![cfg(all(feature = "shutdown-tests"))]
+
+extern crate assert_cmd;
```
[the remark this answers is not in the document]
ktff: True, force of habit.

tests/shutdown.rs
```diff
@@ -0,0 +1,290 @@
...
+        .env("VECTOR_TEST_ADDRESS", format!("{}", next_addr()));
+
+    cmd
+}
+
+fn test_timely_shutdown(mut cmd: Command) {
+    let mut vector = cmd.spawn().unwrap();
+
+    // Give vector time to start.
+    sleep(Duration::from_secs(2));
```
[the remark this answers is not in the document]
ktff: We increase the chance of exiting vector before it gets a chance to capture `SIGTERM` and so we wouldn't be testing `topology.stop. But yes, 2 sec is too long for a single source. 2sec remained from the first iteration of tests which tested `vector` with one config with all of the source. 1sec should be also fine.

tests/shutdown.rs
```diff
@@ -0,0 +1,290 @@
...
+    // Signal shutdown
+    kill(Pid::from_raw(vector.id() as i32), Signal::SIGTERM).unwrap();
+
+    // Time shutdown.
+    let now = Instant::now();
+
+    // Wait for shutdown
+    assert!(vector.wait().unwrap().success());
+
+    assert!(now.elapsed() < Duration::from_secs(3));
```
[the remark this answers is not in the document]
ktff: Something like this?

src/topology/mod.rs
```diff
@@ -118,24 +118,27 @@ impl RunningTopology {
         let mut running_tasks = self.tasks;
 
         let mut wait_handles = Vec::new();
-        let mut check_handles = HashMap::new();
+        let mut check_handles = HashMap::<String, Vec<_>>::new();
 
-        for (name, task) in running_tasks.drain() {
+        for (name, task) in running_tasks.drain().chain(self.source_tasks.into_iter()) {
```
lukesteensen, line 123: Could you explain this change? I'm curious why source tasks aren't included in `running_tasks` already and why we now need to track multiple handles per task.

src/topology/mod.rs
```diff
@@ -664,6 +670,18 @@ fn handle_errors(
...
+fn retain<T>(vec: &mut Vec<T>, mut retain_filter: impl FnMut(&mut T) -> bool) {
+    let mut i = 0;
+    while let Some(data) = vec.get_mut(i) {
+        if retain_filter(data) {
+            i += 1;
+        } else {
+            let _ = vec.remove(i);
+        }
+    }
+}
```
lukesteensen, line 683: This looks just like `Vec::retain` to me. Am I missing something or can we just use the standard library version?

tests/shutdown.rs
```diff
@@ -0,0 +1,298 @@
...
+}
+
+fn test_timely_shutdown(mut cmd: Command) {
+    let mut vector = cmd.stdin(std::process::Stdio::piped()).spawn().unwrap();
+
+    // Give vector time to start.
+    sleep(Duration::from_secs(1));
+
+    // Check if vector is still running
+    assert_eq!(None, vector.try_wait().unwrap(), "Vector exited to early.");
```
lukesteensen, line 104:
```suggestion
    assert_eq!(None, vector.try_wait().unwrap(), "Vector exited too early.");
```

tests/shutdown.rs
```diff
@@ -0,0 +1,298 @@
...
+    // Signal shutdown
+    kill(Pid::from_raw(vector.id() as i32), Signal::SIGTERM).unwrap();
+
+    // Time shutdown.
+    let now = Instant::now();
+
+    // Wait for shutdown
+    assert!(
+        vector.wait().unwrap().success(),
+        "Vector didn't exited successfully."
```
lukesteensen, line 115:
```suggestion
        "Vector didn't exit successfully."
```

src/topology/mod.rs
LucioFranco: Also I'd like to see some comments explaining some of this logic if we make it a bit more complex its starting to get harder to follow 😄

LucioFranco:
Should we open an issue around adding more sources/sinks to this test? I ideally would like us to cover all our sources/sinks.

src/sources/journald.rs
```diff
@@ -174,6 +178,16 @@ impl JournaldConfig {
             .boxed()
             .compat()
             .map_err(|error| error!(message="Journald server unexpectedly stopped.",%error))
+            .select(shutdown.map(|_| ()))
+            .map(move |_| {
+                if let Some(pid) = pid {
+                    // Signal the child process to terminate so that the
+                    // blocking future can be unblocked earlier rather
+                    // than later.
+                    let _ = kill(pid, Signal::SIGTERM);
```
bruceg, line 187: Good thinking, definitely have to deal with this. However, could this be part of the `JournalSource` trait, to let the source itself handle the kill instead of handling it externally? That would eliminate the need to handle the PID outside of `struct Journalctl`.

src/topology/mod.rs
ktff: We need mutable access to the items, but `Vec::retain` gives us immutable. We could use `Vec::drain_filter` but it's a nightly feature at the moment.

src/topology/mod.rs
ktff: As far as I can see, `source_tasks` is needed because we treat their tasks differently. For example when we are shutting down sources we are [waiting](https://github.com/timberio/vector/blob/1c01af7bd1ac7fd0b27e5517907480995104f328/src/topology/mod.rs#L339) on their tasks, while we are just [forgetting](https://github.com/timberio/vector/blob/1c01af7bd1ac7fd0b27e5517907480995104f328/src/topology/mod.rs#L349) other ones. 

With this change we stop dropping their tasks, and instead we started treating them as regular tasks. This is necessary so to give them time to shutdown gracefully through `ShutdownSignal`. 

Regarding multiple handles per task, we actually aren't doing that, there is only one handle per task, but we have two tasks under the same name, [one](https://github.com/timberio/vector/blob/1c01af7bd1ac7fd0b27e5517907480995104f328/src/topology/builder.rs#L154) for `source`  and [the other](https://github.com/timberio/vector/blob/1c01af7bd1ac7fd0b27e5517907480995104f328/src/topology/builder.rs#L153) for `pump`, hence the need for `Vec`.

ktff:
> Should we open an issue around adding more sources/sinks to this test? I ideally would like us to cover all our sources/sinks.

That would be good, also documenting/enforcing somewhere that all sources/sinks have this test.

ktff:
Also added  test for `internal_metrics` which I somehow missed.

src/sources/journald.rs
ktff: PID can be moved, but something still needs to be passed back to trigger the kill. So what about this with passing `FnOnce()`? It is nicer.

bruceg: I see the `journald_server` ends up being moved into the spawned block, and has a `&mut self` method, which makes sharing it overly difficult. That is probably the best compromise.

src/sources/mod.rs
```diff
@@ -11,7 +11,7 @@ pub mod generator;
 pub mod http;
 #[cfg(feature = "sources-internal_metrics")]
 pub mod internal_metrics;
-#[cfg(feature = "sources-journald")]
+#[cfg(all(feature = "sources-journald", feature = "unix"))]
```
fanatid, line 14: Unfortunately, this not works :disappointed:

ktff: How not? Is it because of `mac`?

fanatid: I'm not sure why, but `cargo check --tests --no-default-features --features sources-journald` not react on `sources/journald.rs` changes. I have the same issue in #3298.

`cfg` in two lines works, I'll fix it in #3298.
```rust
#[cfg(unix)]
#[cfg(feature = "sources-journald")]
```

fanatid: Actually, this can be written in 1 line:
```rust
#[cfg(unix, feature = "sources-journald")]
```
[examples in reference](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg-attribute)

ktff: That's right, we have a `unix` feature but it's not used like this, so just using `unix` is correct. Good catch. This is also used in `internal_events/mod.rs` so I would suggest you open a PR that fixes this.

PR #2567, opened 2020-05-10
fix(journald source): Fix `units`/`include_units` switch 

ktff:
Fixes the bug of getting
>May 08 11:05:28.428 WARN vector::sources::journald: The units setting is deprecated, use include_units instead

and using `units` even if `include_units` is used. 

ref #2558.

### DIFF
--- a/src/sources/journald.rs
+++ b/src/sources/journald.rs
@@ -87,7 +87,9 @@ impl SourceConfig for JournaldConfig {
                 warn!("The `units` setting is deprecated, use `include_units` instead");
                 &self.units
             }
-            (_, false) => &self.include_units,
+            // include_units is either empty or not, both cases are fine since units is 
+            // certainly empty.
+            _ => &self.include_units,
         };
 
         let include_units: HashSet<String> = include_units.iter().map(fixup_unit).collect();
--- a/src/sources/journald.rs
+++ b/src/sources/journald.rs
@@ -87,7 +87,7 @@ impl SourceConfig for JournaldConfig {
                 warn!("The `units` setting is deprecated, use `include_units` instead");
                 &self.units
             }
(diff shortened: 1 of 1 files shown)


src/sources/journald.rs
```diff
@@ -83,11 +83,13 @@ impl SourceConfig for JournaldConfig {
 
         let include_units = match (self.units.is_empty(), self.include_units.is_empty()) {
             (false, false) => return Err(BuildError::BothUnitsAndIncludeUnits.into()),
-            (true, false) => {
+            (false, true) => {
```
bruceg, line 86: Good catch, thanks!

src/sources/journald.rs
```diff
@@ -83,11 +83,13 @@ impl SourceConfig for JournaldConfig {
...
                 warn!("The `units` setting is deprecated, use `include_units` instead");
                 &self.units
             }
-            (_, true) => &self.include_units,
+            // include_units is either empty or not, both cases are fine since units is
+            // certainly empty.
+            _ => &self.include_units,
```
bruceg, line 92: I don't see the point in changing this from a more explicit pattern to a catch-all. It doesn't improve clarity (as demonstrated by the need for a comment), nor would I expect it to improve code generation.

bruceg:
After seeing this logic mistake I left in, I wonder if it wouldn't be better to match directly on `(self.units, self.include_units)` so the arms could contain `(Some(_), None)` for example. It would require four matches instead of three and extra `Some`-wrapping, but the logic would be much more obvious. That's not necessary for this PR to pass.

ktff:
That's true, it would be much clearer. I'll change it that.

src/sources/journald.rs
ktff: True

PR #2572, opened 2020-05-10
enhancement(humio_logs sink): Enable gzip compression

geekflyer:
This enables gzip compression for the humio logs sink.
I believe it's kinda best-practice to have to this always enabled and humio supports it always, so let's just enable it always without exposing a config option.


geekflyer:
ok, I hear ya.
I added the gzip compression as config param now but enable it by default.
Does this address your concerns?

fanatid:
@geekflyer I added compression option for `humio_logs` in https://github.com/timberio/vector/pull/2637/commits/7e2b544188603c5c094fc76013f4b0071d8898f5

geekflyer:
@fanatid cool thanks! your PR looks like a superset of mine, so I'm just closing this.

PR #2573, opened 2020-05-11
chore(sinks): Add `BatchedHttpSink` back to `http2`

LucioFranco:

### DIFF
--- a/src/sinks/util/http2.rs
+++ b/src/sinks/util/http2.rs
@@ -1,10 +1,17 @@
-use super::retries::{RetryAction, RetryLogic};
+use super::{
+    retries2::{RetryAction, RetryLogic},
+    service2::{TowerBatchedSink, TowerRequestSettings},
+    Batch, BatchSettings,
+};
 use crate::{
     dns::Resolver,
+    event::Event,
     tls::{tls_connector_builder, MaybeTlsSettings},
+    topology::config::SinkContext,
 };
 use bytes05::{Buf, Bytes};
 use futures::future::BoxFuture;
+use futures01::{Async, AsyncSink, Poll as Poll01, Sink, StartSend};
 use http02::header::HeaderValue;
 use http02::{Request, StatusCode};
 use http_body::Body as HttpBody;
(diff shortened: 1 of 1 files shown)


src/sinks/util/http2.rs
```diff
@@ -26,6 +33,136 @@ pub type Response = http02::Response<Bytes>;
...
+    T: HttpSink<Input = B::Input, Output = B::Output>,
+    L: RetryLogic<Response = http02::Response<Bytes>> + Send + 'static,
+{
+    type SinkItem = crate::Event;
+    type SinkError = crate::Error;
+
+    fn start_send(&mut self, item: Self::SinkItem) -> StartSend<Self::SinkItem, Self::SinkError> {
+        if self.slot.is_some() {
+            self.poll_complete()?;
+            return Ok(AsyncSink::NotReady(item));
```
lukesteensen, line 141: If `poll_complete` succeeds we should be able to make progress here.

src/sinks/util/http2.rs
```diff
@@ -26,6 +33,136 @@ pub type Response = http02::Response<Bytes>;
...
+        }
+
+        if let Some(item) = self.sink.encode_event(item) {
+            if let AsyncSink::NotReady(item) = self.inner.start_send(item)? {
+                self.poll_complete()?;
```
lukesteensen, line 146: Same here.

src/sinks/util/http2.rs
LucioFranco: This is the same implementation we had before, so I'd like to defer to that instead of changing behavior here? What do you think?

LucioFranco: https://github.com/timberio/vector/blob/master/src/sinks/util/http.rs#L136

lukesteensen: Yeah if it's just copied that's fine. Probably still worth taking a look later though.

LucioFranco: https://github.com/timberio/vector/issues/2577

PR #2576, opened 2020-05-11
chore: Add clippy to CI

fanatid:
Not sure that this is wanted, but found issue https://github.com/timberio/vector/pull/1034
Currently clippy gives a loooot of warnings (some of them looks really reasonable).

New CI job not fails on warnings right now (I think that fixes should came in other PR after discussion).

For posterity, code for generating rules from clippy output:
```js
const text = require('fs').readFileSync('./clippy.log', 'utf8')

const rules = new Set()
for (const line of text.split('\n')) {
    const match = line.match(/\.html#(\w+)$/)
    if (match) rules.add(match[1])
}

for (const rule of Array.from(rules.keys()).sort()) {
  console.log(`#![allow(clippy::${rule})]`)
}
```


LucioFranco:
There is this https://github.com/actions-rs/clippy

and also you may be able to use https://github.com/rust-lang/rust-clippy#automatically-applying-clippy-suggestions

fanatid:
Wow, https://github.com/actions-rs/clippy-check looks really great. Problem is that `actions-rs/clippy` still under WIP and `actions-rs/clippy-check` not work for Pull Requests from forked repositories (see [limitations](https://github.com/actions-rs/clippy-check#limitations)). So probably they can not be used right now =(

Need rebase after https://github.com/timberio/vector/pull/2579 and test again

Hoverbear:
@lukesteensen We can start but allowing all our current warnings then making issues to turn them back to disallowed.

fanatid:
> @lukesteensen We can start but allowing all our current warnings then making issues to turn them back to disallowed.

I like this. Should I change PR for following in this direction?

lukesteensen:
Yep, that sounds good to me.

fanatid:
Unfortunately it's not possible have one shared clippy config, so I was need put rules to each entry point =(

Can be removed after #2579:
```yaml
        env:
          USE_CONTAINER: docker
```

Hoverbear:
@fanatid You are so awesome :)

Hoverbear:
@fanatid Can I ask you to clean up the `make fmt` error? :) Rest looks good.

fanatid:
Need fix new lint errors :grimacing:

Hoverbear:
That's like the most clippy problem ever.

MOZGIII:
We probably want to add `#![deny(clippy::all)]` somewhere. What do you think?

fanatid:
@MOZGIII I do not have years of experience in Rust so can be wrong, but `#![deny(clippy::all)]` will react with errors on allowed rules? I'm sure we do not everything from that list. Maybe I'm wrong. @bruceg @Hoverbear @lukesteensen thoughts?

I propose merge it and if we need change rules make it in another PR.

MOZGIII:
I’d be fine with posponing this for later. The PR looks solid, and it’s in our best interest to speed up the integration. The longer we wait, the more dusty it becomes.

Hoverbear:
There used to be some sort of config file you could use, but it seems to be gone now :(

PR #2579, opened 2020-05-12
chore: Build docker compose images nightly and use them in CI

MOZGIII:
This is an attempt to finally streamline the CI and ensure we have quick builds in CI and a unified environment configs across local and CI tasks.

To make things work, I had to introduce the following changes:

- configure images for all the services at `docker-compose.yml`
- refresh the CI images build script
- set up automated nightly builds for CI images (the auto builds at the docker hub end are to be disabled)
- made the per-push CI flow use pre-built docker images - this allows us to properly control build environment at CI, but eliminates the image build step, dramatically speeding up the flow
- solve the issue with tmp files being on overlayfs (cause of issues with tests failing in docker)

While working on this, I uncovered a few issues that I'm not planning on solving in this PR:
- build image `Dockerfile`s are not cache-friendly, we should correct this
- `docker-compose.yml`, `Makefile` and `.github/workflow/code.yml` each contains configuration for the same "unit"s, but it's all coded separately; there's way too much attention required to implement keep things in order and consistent, and it's hard to maintain; we should revisit the approach, and root all those places in a single source of truth

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -385,9 +385,9 @@ jobs:
           name: Verify installation
           command: vector --version
 
-  verify-deb-artifact-on-ubuntu-19-04:
+  verify-deb-artifact-on-ubuntu-20-04:
     docker:
-      - image: timberiodev/vector-verifier-ubuntu-19-04:latest
+      - image: timberiodev/vector-verifier-ubuntu-20-04:latest
     steps:
       - *restore-artifacts-from-workspace
       - run:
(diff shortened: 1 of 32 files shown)


MOZGIII:
The rest of the CI failures are actual errors that went undetected with the current CI setup. Let's merge this and fix the errors in other PRs.

MOZGIII:
This PR still has one failing job - the `test-unit`. The interesting part is that it only consistently fails if we’re running the job within docker.
@LucioFranco you mentioned you were working on tests troubleshooting - is this among the things you’re already looking into?

MOZGIII:
This is blocked by https://github.com/timberio/vector/pull/2618

Hoverbear:
Is this still relevant @MOZGIII ?

MOZGIII:
The PR - definitely not. We might still want to build an environment image every night though.

PR #2616, opened 2020-05-15
enhancement(security): Add support for loading multiple CAs

### LINKED ISSUE #2612: TLS PEM cert chain is not correctly being used
Hi! I have this `ca.pem` file that has 6 CA certs belonging to two chains (so there's a chain with a root CA and an intermediate CA, then a second chain with a root CA and two intermediate CAs) and when I try validating my client cert with it I get:

```
❯ openssl verify -CAfile cas/foo-certs.pem certs/cert.pem
certs/cert.pem: OK
```

but then when I set this at the config:

```toml
    [sources.tls]
      type = "socket"
      address = "0.0.0.0:6514"
      mode = "tcp"
      tls.enabled = true
      tls.crt_path = "/workdir/cert.pfx"
      tls.ca_path = "/crt/ca.pem"
      tls.verify_certificate = true
```

I keep getting `"remote error: tls: unknown certificate authority"` but these are the same files. 'Tis a bug, perhaps? Ofc setting `verify_certificate = false` works.

bruceg:
This ends up reworking some aspects of certificate loading, hopefully making it easier to reason about.

Closes #2612


src/tls/settings.rs
```diff
@@ -187,6 +142,72 @@ impl TlsSettings {
...
+        match &self.ca_path {
+            None => Ok(vec![]),
+            Some(filename) => {
+                let data = open_read(filename, "certificate")?;
+                der_or_pem(
+                    data,
+                    |der| X509::from_der(&der).map(|x509| vec![x509]),
+                    |pem| {
+                        pem.match_indices(PEM_START_MARKER)
+                            .map(|(start, _)| X509::from_pem(pem[start..].as_bytes()))
```
lukesteensen, line 156: I assume from this that the parser knows when to stop on its own and we don't need to limit its view of the input data?

bruceg: As far as I am aware, yes, and evidence says yes too, otherwise parsing files containing multiple certificates would have failed completely. Instead, it used to load just the first certificate.

PR #2637, opened 2020-05-19
chore: Improve compression option

fanatid:
- Replace `Option<Compression>` to `Compression` (`Compression` enum already have `None` variant).
- Error on Gzip compression in Elasticsearch on AWS.
- Add compression option to `new_relic_logs`, `humio_logs`.

Questions:
`sematext_logs` -- Is compression supported by their Elasticsearch? (no found info about this)
`aws_s3` -- Allow compression by default as `none`?
`gcp_cloud_storage` -- Allow compression by default? `none` or `gzip`?


binarylogic:
Thanks!

> Replace Option<Compression> to Compression (Compression enum already have None variant).

This won't break backward compatibility for users that did not specify the option, right?

> Error on Gzip compression in Elasticsearch on AWS.

👍 

> Add compression option to new_relic_logs, humio_logs.

👍 

> Allow compression by default as none?

I think so. Compression, in my opion, should be opt-in unless it is transparent to the user. In the S3 case it is not transparent, so I think it should default to `none`.

> Allow compression by default? none or gzip?

Same as above. It should default to `none`.

fanatid:
> > Replace Option to Compression (Compression enum already have None variant).
> 
> This won't break backward compatibility for users that did not specify the option, right?

Yes. `Option<Compression>` for variant `none` equal to `Compression::None` in config. With `Option`, `Compression::None` not used at all.

Will add `none` to `aws_s3` and `gcp_cloud_storage`.

.meta/sinks/humio_logs.toml.erb
```diff
@@ -60,6 +60,20 @@ examples = ["http://myhumiohost.com"]
...
 description = "The optional host to send Humio logs to."
 
+[sinks.humio_logs.options.compression]
+type = "string"
+category = "Requests"
+common = true
+default = "none"
+description = """\
+The compression strategy used to compress the encoded event data before \
+outputting.\
```
bruceg, line 70: I'd suggest "transmission" rather than "outputting". Ditto for all the others.

.meta/sinks/gcp_cloud_storage.toml.erb
```diff
@@ -148,12 +148,12 @@ description = "The filename extension to use in the object name."
...
 type = "string"
 common = true
-required = true
+default = "none"
 description = "The compression mechanism to use."
 
 [sinks.gcp_cloud_storage.options.compression.enum]
-gzip = "GZIP compression"
 none = "No compression"
+gzip = "GZIP compression"
```
bruceg, line 156: I know it was all-caps before, but "gzip" should probably not be all-caps as it is not an acronym.

src/sinks/clickhouse.rs
```diff
@@ -116,15 +113,13 @@ impl HttpSink for ClickhouseConfig {
...
             .method(Method::POST)
             .uri(uri.clone())
             .header("Content-Type", "application/x-ndjson");
 
-        let builder = if let Compression::Gzip = self.compression.unwrap_or(Compression::Gzip) {
-            builder.header("Content-Encoding", "gzip")
-        } else {
-            builder
+        if self.compression == Compression::Gzip {
+            builder = builder.header("Content-Encoding", "gzip")
```
bruceg, line 122: Suggestion: Add a method to `impl Compression` that returns the right (optional) content encoding string.

src/sinks/gcp/cloud_storage.rs
```diff
@@ -145,8 +146,8 @@ impl Encoding {
 #[derivative(Default)]
 enum Compression {
     #[derivative(Default)]
-    Gzip,
     None,
+    Gzip,
```
bruceg, line 150: Why does this change the default?

src/sinks/util/buffer/mod.rs
```diff
@@ -9,9 +9,11 @@ pub mod partition;
 
 pub use partition::{Partition, PartitionBuffer, PartitionInnerBuffer};
 
-#[derive(Serialize, Deserialize, Debug, Copy, Clone, Eq, PartialEq)]
-#[serde(rename_all = "lowercase")]
+#[derive(Serialize, Deserialize, Debug, Derivative, Copy, Clone, Eq, PartialEq)]
+#[derivative(Default)]
+#[serde(rename_all = "snake_case")]
```
bruceg, line 14: Do we really want `snake_case` here? That seems to unnecessarily make this a breaking change.

bruceg:
I wonder if this might be better broken up into two PRs, as this is half "chore" and half "enhancement" to the `new_relic_logs` and `humio_logs` sinks that now have compression support added. In any case, the enhancements make the title wrong, and I noted a breaking change in the review.

If you want to continue work in this direction, consider passing `Compression` directly to the buffer creation instead of `gzip: bool`. This would allow for the representation of new compression types. However, that may require making `Compression` a trait implemented by multiple enums, as different sinks have different allowed compression types, and I hadn't yet figured out a good way of representing that so that serde forbids the bad options only for those sinks.

.meta/sinks/humio_logs.toml.erb
fanatid: Thanks. I'm thinking add compression to `_partials/fields`, because it's same words and data used in few files. Will fix it there.

src/sinks/util/buffer/mod.rs
fanatid: `snake_case` used in `gcp_cloud_storage` sink: https://github.com/timberio/vector/blob/6281932469d9bcef34edaf672b52edc804d49732/src/sinks/gcp/cloud_storage.rs#L144
Should both be `lowercase`?

src/sinks/gcp/cloud_storage.rs
fanatid: I though if config now will have default option as `none` it would be good have default variant `none` too.
Should I change it back?

bruceg: I would not change the default in this PR. If you think it should be changed, handle that separately.

src/sinks/util/buffer/mod.rs
bruceg: Snake case is used in the GCP sinks because we are encoding to a GCP REST API that requires the use of snake case. They use their own enums though. This is used by other sinks that don't have the same requirements.

fanatid:
@bruceg can you review again?

- Revert `humio_logs` and `new_relic_logs` changes.
- Change buffer for accepting `Compression` directly instead `gzip: bool`.
- Add `content_encoding` function to `Compression`.
- Update compression in docs by moving things to `_partials/fields/_compressions.toml.erb`

src/sinks/gcp/cloud_storage.rs
```diff
@@ -62,6 +63,7 @@ pub struct GcsSinkConfig {
     filename_append_uuid: Option<bool>,
     filename_extension: Option<String>,
     encoding: EncodingConfig<Encoding>,
+    #[serde(default = "Compression::config_default")]
     compression: Compression,
```
bruceg, line 67: Could this use the generic `util::Compression` here with an alternate default? I see that it uses a different case convention, but that could be handled in `util` with serde alias attributes.

src/sinks/clickhouse.rs
```diff
@@ -25,7 +25,7 @@ pub struct ClickhouseConfig {
     pub host: String,
     pub table: String,
     pub database: Option<String>,
-    pub compression: Option<Compression>,
+    pub compression: Compression,
```
bruceg, line 28: I think this needs a specific serde default here so that it doesn't accidentally become required (breaking existing configs) and defaults to gzip (also kinda breaking).

fanatid: Oh, skipped this. Why defaults to gzip? Currently default is `none` and `#[serde(default)]` should be enough, because `Compression::None` is default.

src/sinks/gcp/cloud_storage.rs
fanatid: I just removed internal `Compression` and proceed with `uitl::Compression`. Only difference is `snake_case` vs `lowercase`, but for `None` / `Gzip` no difference between this rename attributes.

src/sinks/clickhouse.rs
bruceg: The current default is `None`, because the field is an `Option<Compression>`. When it is unwrapped in `fn build`, the gzip default is applied. I do note that the existing documentation does not reflect this default. @lukesteensen you are the original author of this sink, maybe you could comment why it defaults to gzip compression.

src/sinks/gcp/cloud_storage.rs
bruceg: Ah, yes, of course. I was thinking of one of the other GCS attributes where we do encode it when building the request, so a specific rename convention is required.

src/sinks/clickhouse.rs
fanatid: Ohhh, I missed this! Messed up between changes. I'll change it then, fix docs and it will be possible switch to `none` if required in another PR.

fanatid: Fixed.

src/sinks/aws_s3.rs
```diff
@@ -51,6 +51,7 @@ pub struct S3SinkConfig {
         default
     )]
     pub encoding: EncodingConfigWithDefault<Encoding>,
+    #[serde(default)]
     pub compression: Compression,
```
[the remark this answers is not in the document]
fanatid: I see, in original code default enum variant was `Gzip`, but in same time `S3SinkConfig` was require compression option, so default value was not used anywhere.
Can we change to `None` seamlessly or `aws_s3` should have `Gzip` by default by some reason?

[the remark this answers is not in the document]
binarylogic: It’s more about now breaking backward compatibility. But I am fine defaulting to none in general. Typically, though, you’d want to enable gzip for S3.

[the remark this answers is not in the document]
fanatid: ok, changed to gzip default value

bruceg:
Thanks for this good work @fanatid

PR #2650, opened 2020-05-20
chore: A script to install git hooks

MOZGIII:
This significantly simplifies the use of git worktrees.

### DIFF
--- a/scripts/install-git-hooks.sh
+++ b/scripts/install-git-hooks.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 MODE="${1:-"all"}"
--- a/scripts/install-git-hooks.sh
+++ b/scripts/install-git-hooks.sh
@@ -3,8 +3,10 @@ set -euo pipefail
 
 MODE="${1:-"all"}"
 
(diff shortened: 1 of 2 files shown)


scripts/install-git-hooks.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -euo pipefail
+
+MODE="${1:-"all"}"
+
+mkdir -p "$(git rev-parse --git-dir)/hooks"
+
+if [[ "$MODE" == "all" || "$MODE" == "signoff" ]]; then
+  cp scripts/signoff-git-hook.sh "$(git rev-parse --git-dir)/hooks/commit-msg"
```
Hoverbear, line 9: Thanks for choosing `git rev-parse --git-dir` instead of `.git`!

scripts/install-git-hooks.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
Hoverbear, line 1:
```suggestion
#!/usr/bin/env bash
```

scripts/install-git-hooks.sh
MOZGIII: That's the only thing that correctly works with workspaces :smile:

PR #2656, opened 2020-05-20
chore(gcp_cloud_storage sink): Change default variant in enum Compression

fanatid:
Extracted from https://github.com/timberio/vector/pull/2637#discussion_r427447512

Currently default value used only in tests.


fanatid:
Closed in favour of #2637

PR #2659, opened 2020-05-21
docs: Add uint type

fanatid:
Closes #2592

PR #2662, opened 2020-05-21
docs: Add missed option

fanatid:
During #2659 noticed few small not right things.

Changes:
- Fix `max_length` for socket source (only `tcp` and `unix`, not `udp`)
- Adjust link content in socket source
- Add `commit_interval_ms` to docs from kafka source

Questions:
- In kafka source `host_key` and `compression` not used at all, is this ok or should be fixed?
- Lua transform describe only version 2 in docs, version 1 is outdated or just missed?


.meta/sources/socket.toml.erb
```diff
@@ -67,14 +67,15 @@ default = "host"
 description = """\
 The key name added to each event representing the current host. This can also \
 be globally set via the \
-[global `host_key` option][docs.reference.global-options#host_key].\
+[global 'host_key' option][docs.reference.global-options#host_key].\
```
bruceg, line 70: This seems to be a spurious change, is that intentional?

fanatid: Yes, this is intentional. `OptionLinker` change \`host_key`, so generated version looks incorrect.
https://github.com/timberio/vector/blob/1d8e88057f68d9cf9292ddc9edb69a7f8d3b3f92/scripts/generate/post_processors/option_linker.rb#L16
Screenshot from 2020-05-24 08-12-51
Replace to single quote was quick solution, but probably I need fix regex in rb file :thinking:

binarylogic:
> In kafka source host_key and compression not used at all, is this ok or should be fixed?

Yes, we should fix that here or open an issue. What do you think?

> Lua transform describe only version 2 in docs, version 1 is outdated or just missed?

Nope, we do not want to advertise v1, it's there for backward compatibility.

fanatid:
- revert commit with `host_key` change (create issue https://github.com/timberio/vector/issues/2822)
- remove `compression` and `host_key` from kafka source

binarylogic:
Looks like the windows lint test is failing.

fanatid:
Sorry, forgot remove extra import =(

amruta-square:
@bruceg @fanatid I have a question about the compression option that is removed. If I am compressing messages on producer end, I will need to decompress them, so how do I specify that option?

jszwedko:
> @bruceg @fanatid I have a question about the compression option that is removed. If I am compressing messages on producer end, I will need to decompress them, so how do I specify that option?

You are right, the `kafka` source doesn't support the compression option. Would you want to open an issue for that if you'd find it useful? It seems like compression was added to support other Kafka clients, but not Vector <-> Vector Kafka communication.

jszwedko:
You might be able to set it under the librdkafka options.

PR #2663, opened 2020-05-21
chore(syslog source): Add `hostname` and `source_ip`

Hoverbear:
Resolves https://github.com/timberio/vector/issues/2655 by adding `source_ip` and `hostname` fields that are always set.

### DIFF
--- a/src/sources/syslog.rs
+++ b/src/sources/syslog.rs
@@ -216,19 +216,12 @@ fn event_from_str(host_key: &str, default_host: Option<Bytes>, line: &str) -> Op
         .as_mut_log()
         .insert(event::log_schema().source_type_key(), "syslog");
 
-    match (parsed.hostname, default_host) {
-        (Some(parsed_hostname), Some(default_host)) => {
-            event.as_mut_log().insert(host_key, parsed_hostname.clone());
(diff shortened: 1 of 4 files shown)


src/sources/syslog.rs
```diff
@@ -215,10 +216,19 @@ fn event_from_str(host_key: &str, default_host: Option<Bytes>, line: &str) -> Op
...
+        },
+        (Some(parsed_hostname), None) => {
+            event.as_mut_log().insert(host_key, parsed_hostname.clone());
+        },
+        (None, Some(default_host)) => {
+            event.as_mut_log().insert(host_key, default_host.clone());
+            event.as_mut_log().insert("source_ip", default_host);
+        },
+        (None, None) => (),
     }
```
lukesteensen, line 232: Could this be something like the following?

```rust
if let Some(default_host) = default_host {
    event.as_mut_log().insert("source_ip", default_host);
}

if let Some(parsed_host) = parsed.hostname.or(default_host) {
    event.as_mut_log().insert(host_key, default_host.clone());
}
```

The combined match implies to me that there's some dependency between the two set fields, but I don't think there is.

PR #2671, opened 2020-05-22
fix(pulsar sink): Add new EncodingConfig

### LINKED ISSUE #2648: Pulsar Sink limited encoding configuration
As found [here](https://vector.dev/docs/reference/sinks/pulsar/#encoding), the documentation says that the encoding configuration has multiple sub elements (e.g.: codec, only_fields). However, trying to set up ```encoding.codec = "json"``` generates the following error:

```bash
... ERROR vector: Configuration error: "/etc/vector/vector.toml": unknown variant `codec`, expected `text` or `json` for key `sinks.pulsar`
```

Additionally, adding any other options such as ```encoding.timestamp_format = "rfc3339"``` produces the following error:

```bash
... ERROR vector: Configuration error: "/etc/vector/vector.toml": invalid value: map, expected map with a single key for key `sinks.pulsar`
```

So the only way I was able to run the sink without any issues was when I set the value as ```encoding = "json"```.

This makes me wonder not only if there's a mistake in the docs but also if there's no way to setup the additional configuration options such as ```only_fields``` or ```timestamp_format```?

leshow:
closes #2648 

Adds the new `EncodingConfig` to the pulsar sink. I believe this is all that's required to get everything going again. It passes all the tests on my machine.

### DIFF
--- a/src/sinks/pulsar.rs
+++ b/src/sinks/pulsar.rs
@@ -196,7 +196,7 @@ mod tests {
         let msg = "hello_world".to_owned();
         let mut evt = Event::from(msg.clone());
         evt.as_mut_log().insert("key", "value");
-        let result = encode_event(evt, Encoding::Json).unwrap();
+        let result = encode_event(evt, &EncodingConfig::from(Encoding::Json)).unwrap();
         let map: HashMap<String, String> = serde_json::from_slice(&result[..]).unwrap();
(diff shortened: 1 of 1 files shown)


fanatid (approved):

Hoverbear (approved):

Hoverbear:
Thanks so much @leshow !

PR #2685, opened 2020-05-25
feat(platforms): Add dev workflow simplification rfc

Hoverbear:
Adds the Dev workflow simplification RFC from https://www.notion.so/timber/Dev-Workflow-Simplification-c0990440bb494a71bc1150a52a4494d2

### DIFF
--- a/rfcs/2020-05-25-2685-dev-workflow-simplification.md
+++ b/rfcs/2020-05-25-2685-dev-workflow-simplification.md
@@ -1,4 +1,4 @@
-# RFC 2685 - 2020-04-04 - Kubernetes Integration
+# RFC 2685 - 2020-05-28 - Dev Workflow Simplification
 
 Vector's `Makefile` serves a variety of purposes, and this RFC attempts to tame the complexity of common dev tasks, improving contributor and team member experience.
 
@@ -34,7 +34,7 @@ This is ideal for users who want it to "Just work" and just want to start contri
 
 ```bash
 # Optional: Only if you use `podman`
-export container-tool="podman"
+export CONTAINER_TOOL="podman"
 ```
 
 By default, `make environment` style tasks will do a `docker pull` from Github's container repository, you can **optionally** build your own environment while you make your morning coffee ☕:
@@ -44,41 +44,49 @@ By default, `make environment` style tasks will do a `docker pull` from Github's
 make environment-prepare
 ```
 
-Now that you have your coffee, you can run any of the following:
+Now that you have your coffee, you can enter the shell!
 
 ```bash
 # Enter a shell with optimized mounts for interactive processes.
 # Inside here, you can use Vector like you have full toolchain (See below!)
 make environment
-# Add extra cli opts
-make enviroment cli-opts="--publish 3000:2000"
-# Validate your code can compile
-make enviroment-check
-# Validate your code actually does compile (in dev mode)
-make environment-build-dev
-# Validate your test pass
-make environment-test scope="sources::example"
-# Validate tests (that do not require other services) pass
-make environment-test
-# Validate your tests pass (starting required services in Docker)
-make environment-test-integration scope="sources::example"
-# Validate your tests pass against a live service.
-make environment-test-integration scope="sources::example" autospawn=false
-# Validate all tests pass (starting required services in Docker)
-make environment-test-integration
-# Run your benchmarks
-make environment-bench scope="transforms::example"
-# Rebuild Vector's metadata
-make environment-generate
-# Serve the website on port 3000
-make environment-website
-# Format your code before pushing!
-make environment-fmt
 # Try out a specific container tool. (Docker/Podman)
-make environment container-tool="podman"
+make environment CONTAINER_TOOL="podman"
+# Add extra cli opts
+make environment CLI_OPTS="--publish 3000:2000"
 ```
 
-We use explicit environment opt-in as many contributors choose to keep their Rust toolchain local, and use `make environment-generate` etc.
+Now you can use the jobs detailed in **"Bring your own toolbox"** below. 
+
+Want to run from outside of the environment? Clever. You can run any of the following:
+
+```bash
+
+# Validate your code can compile
+make check ENVIRONMENT=true
+# Validate your code actually does compile (in dev mode)
+make build-dev ENVIRONMENT=true
+# Validate your test pass
+make test SCOPE="sources::example" ENVIRONMENT=true
+# Validate tests (that do not require other services) pass
+make test ENVIRONMENT=true
+# Validate your tests pass (starting required services in Docker)
+make test-integration SCOPE="sources::example" ENVIRONMENT=true
+# Validate your tests pass against a live service.
+make test-integration SCOPE="sources::example" autospawn=false ENVIRONMENT=true
+# Validate all tests pass (starting required services in Docker)
+make test-integration ENVIRONMENT=true
+# Run your benchmarks
+make bench SCOPE="transforms::example" ENVIRONMENT=true
+# Rebuild Vector's metadata
+make generate ENVIRONMENT=true
+# Serve the website on port 3000
+make website ENVIRONMENT=true
+# Format your code before pushing!
+make fmt ENVIRONMENT=true
+```
+
+We use explicit environment opt-in as many contributors choose to keep their Rust toolchain local, and use `make generate ENVIRONMENT=true` etc.
 
 ### Using Nix
 
@@ -90,7 +98,7 @@ If you don't have Nix yet, [install it](https://nixos.org/download.html):
 
 ```bash
 curl -L [https://nixos.org/nix/install](https://nixos.org/nix/install) | sh
-# Hesitating? Uninstalling is just `rm -rf /nix`
+# Hesitating? Uninstalling is just `rm -rf /nix && rm -rf ~/.nix-*` then remove the import from `.bash_profile`
 ```
 
 Next, run `nix-shell` from the Vector directory.
@@ -132,31 +140,38 @@ If you find yourself needing to run something (such as `make generate`) inside t
 
 We're interested in reducing our dependencies if simple options exist. Got an idea? Try it out, we'd to hear of your successes and failures!
 
-In order to do your development on Vector, you'll primarily use a few commands, ordered from most to least frequently run:
+In order to do your development on Vector, you'll primarily use a few commands, such as `cargo` and `make` tasks you can use ordered from most to least frequently run:
 
 ```bash
 # Validate your code can compile
+cargo check
 make check
 # Validate your code actually does compile (in dev mode)
+cargo build
 make build-dev
 # Validate your test pass
+cargo test sources::example
 make test scope="sources::example"
 # Validate tests (that do not require other services) pass
+cargo test
 make test
 # Validate your tests pass (starting required services in Docker)
 make test-integration scope="sources::example" autospawn=false
 # Validate your tests pass against a live service.
 make test-integration scope="sources::example" autospawn=false
+cargo test --features docker sources::example
 # Validate all tests pass (starting required services in Docker)
 make test-integration
 # Run your benchmarks
 make bench scope="transforms::example"
+cargo bench transforms::example
 # Rebuild Vector's metadata
 make generate
 # Serve the website on port 3000
 make website
 # Format your code before pushing!
 make fmt
+cargo fmt
 ```
 
 If you run `make` you'll see a full list of all our tasks. Some of these will start Docker containers, sign commits, or even make releases. These are not common development commands and your mileage may vary.
@@ -206,12 +221,6 @@ There are tools like `hab` (from the Habitat project) and `packer` that can be u
 - It's **not backwards compatible**, meaning users who depend on the current `docker-compose` system might experience frustration.
 
 ## Outstanding Questions
-
-- Nix has support for Carnix/Bundix etc that allow us to manage some or all of our package dependencies with Nix. We may wish to evaluate using these more in the future when we are more comfortable with Nix.
-- Should we alter our exiting podman support? Right now it's behind the `USE_CONTAINERS` , should we have a `container_technology` or something?
-- Mac support is untested. (But should be improved)
-
- **TODO:** Ana has asked for a test
 
 - Windows/Mac/FreeBSD builds via `make build` et all will produce native binaries natively, we should be review those docs.
 - This RFC does not scope in integration tests beyond letting the `environment` run them. We may find motivation to explore a more ***slick*** solution in the future.
--- a/rfcs/2020-05-25-2685-dev-workflow-simplification.md
+++ b/rfcs/2020-05-25-2685-dev-workflow-simplification.md
@@ -1,4 +1,4 @@
-# RFC 2221 - 2020-04-04 - Kubernetes Integration
+# RFC 2685 - 2020-04-04 - Kubernetes Integration
 
 Vector's `Makefile` serves a variety of purposes, and this RFC attempts to tame the complexity of common dev tasks, improving contributor and team member experience.
 
--- a/rfcs/2020-05-25-2064-dev-workflow-simplification.md
+++ b/rfcs/2020-05-25-2064-dev-workflow-simplification.md
@@ -1,0 +1,244 @@
+# RFC 2221 - 2020-04-04 - Kubernetes Integration
+
+Vector's `Makefile` serves a variety of purposes, and this RFC attempts to tame the complexity of common dev tasks, improving contributor and team member experience.
+
+It proposes a practical base `environment` container/nix-env that merges the functionality of our non-integration test containers into one. It then suggests making common dev `make` tasks to rely on the caller environment having all dependencies, done at the same time it suggests adding `make` tasks to run common `make` tasks inside the environment. Finally, it suggests updating documentation to suggest users can use their native toolchains, `docker`, or `nix`.
+
+It proposes these changes such that our team and contributors will have to explicitly **opt in** to virtualization/containerization/shell-manipulation, and attempts to do it in a convienent, unsurprising way.
+
+## Motivation
+
+Currently, Vector's build system is a wild, powerful beast with a lot of amazing features and mechanisms. Unfortunately, complexity has kind of gotten away from us, and we need to reclaim ownership of the automations.
+
+Notably, our system works right now, as is, and we are not strongly motivated to replace or rebuild things just for the fun of it. We are motivated to make small, forward looking changes that improve the maintainability of our product.
+
+**Before we do this, we should make sure we agree this offers realistic benefits to our future maintenance.**
+
+## Guide-level Proposal
+
+> **Note:** Replaces the **Development → Setup** section of the `[CONTRIBUTING.md](http://contributing.md)` file.
+
+We're super excited to have you interested in working on Vector! Before you start you should pick how you want to develop.
+
+For small or first-time contributions, we recommend the Docker method. If you do a lot of contributing, try adopting the Nix method! It'll be way faster and feel more smooth. Prefer to do it yourself? That's fine too!
+
+### Use a Docker or Podman environment
+
+> **Targets:** You can use this method to produce AARCH64, Arm6/7, as well as x86/64 Linux builds.
+
+Since not everyone has a full working native environment, or can use Nix, we took our Nix environment and stuffed it into a Docker (or Podman) container!
+
+This is ideal for users who want it to "Just work" and just want to start contributing. It's also what we use for our CI, so you know if it breaks we can't do anything else until we fix it. 😉
+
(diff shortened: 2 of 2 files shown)


Hoverbear:
I've asked @zsherman if he can try this on his Mac for me. :) If anyone else wants to try I'd love that.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
```diff
@@ -0,0 +1,244 @@
...
+### Bring your own toolbox
+
+> **Targets:** This option is required for MSVC/Mac/FreeBSD toolchains.
+
+To build Vector on your own host will require a fairly complete development environment!
+
+We keep an up to date list of all dependencies used in our CI environment inside our `default.nix` file. Loosely, you'll need the following:
+
+- **To build Vector:** Have working Rustup, C++/C build tools (LLVM, GCC, or MSVC), Python, and Perl, `make` (the GNU one preferably), `bash`, `cmake`, and `autotools`
+- **To run integration tests:** Have `docker` available, or a real live version of that service.
```
MOZGIII, line 126: For k8s we'll also need `minikube` (alongside `docker`) to be available on the host for integration tests.
For dev flow, the `minikube` *or something like Docker Desktop* plus `skaffold` is required.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
```diff
@@ -0,0 +1,244 @@
...
+Now that you have your coffee, you can run any of the following:
+
+```bash
+# Enter a shell with optimized mounts for interactive processes.
+# Inside here, you can use Vector like you have full toolchain (See below!)
+make environment
+# Add extra cli opts
+make enviroment cli-opts="--publish 3000:2000"
+# Validate your code can compile
+make enviroment-check
```
MOZGIII, line 56: So, we'll have both `make enviroment-check` and `make check`? Can't we further reduce the duplication and entry points?

I don't know if you considered this option already, but can we decouple the environment and commands completely, and handle it via composition at the user level:

```shell
host$ make environment # drops the into the environment shell, possibly in a container
envshell$ make check # performs the check inside the environment
```

rfcs/2020-05-25-2685-dev-workflow-simplification.md
```diff
@@ -0,0 +1,244 @@
...
+
+> **Note:** We're still new at this Nix stuff, so if you're a Nix expert do feel free to share knowledge.
+
+If you're a Nix user, or you're open to trying out a new tool, you can use our Nix expressions!
+
+If you don't have Nix yet, [install it](https://nixos.org/download.html):
+
+```bash
+curl -L [https://nixos.org/nix/install](https://nixos.org/nix/install) | sh
+# Hesitating? Uninstalling is just `rm -rf /nix`
```
bruceg, line 93: As an aside, and not related to this RFC (this line is a quote from the Nix home page), this isn't entirely true. The Nix installer drops some `.nix-*` files into the installing user's home directory and modifies `.bash_profile` for bash users (presumably other files for different shells). As such, uninstalling is a little more complicated, but only a little.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
```diff
@@ -0,0 +1,244 @@
...
+
+- Let's face it, **Nix isn't awesome from a usability side**. `nix-env` and `nix-shell` have lots of warts, but so does every 20+ year old tech. It's a good tool to create and share reproducable environments, and that's what we're using it for.
+- This adds **another 2 official dev envs, Nix and Native, instead of just Docker.** This means we may need to provide support.
+- **Only Ana knows Nix right now**. But others can and have shown willingness to learn.
+- It's **not backwards compatible**, meaning users who depend on the current `docker-compose` system might experience frustration.
+
+## Outstanding Questions
+
+- Nix has support for Carnix/Bundix etc that allow us to manage some or all of our package dependencies with Nix. We may wish to evaluate using these more in the future when we are more comfortable with Nix.
+- Should we alter our exiting podman support? Right now it's behind the `USE_CONTAINERS` , should we have a `container_technology` or something?
```
bruceg, line 211: Nit, it's currently `USE_CONTAINER` (no S).

rfcs/2020-05-25-2685-dev-workflow-simplification.md
```diff
@@ -0,0 +1,244 @@
...
+
+If you find yourself needing to run something (such as `make generate`) inside the Docker environment described above, that's totally fine, they won't collide or hurt each other. In this case, you'd just run `make environment-generate`.
+
+We're interested in reducing our dependencies if simple options exist. Got an idea? Try it out, we'd to hear of your successes and failures!
+
+In order to do your development on Vector, you'll primarily use a few commands, ordered from most to least frequently run:
+
+```bash
+# Validate your code can compile
+make check
```
lukesteensen, line 139: Do we still want to use `make` in these situations? Especially when you've built your own dev env, it'd be nice to get rid of this extra layer if it's not adding anything over `cargo`.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
```diff
@@ -0,0 +1,244 @@
...
+# Validate your code actually does compile (in dev mode)
+make build-dev
+# Validate your test pass
+make test scope="sources::example"
+# Validate tests (that do not require other services) pass
+make test
+# Validate your tests pass (starting required services in Docker)
+make test-integration scope="sources::example" autospawn=false
```
lukesteensen, line 147: Is `autospawn=false` a typo? The comment says it will start services.

I've also not seen this style of passing arguments to `make` before. I'm a little concerned about it being foreign both to users and to `make` itself.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
MOZGIII: I think we don't need targets that are direct counterparts of the *simple* cargo commands, like `check`, `build` and `test`.

But then, if we omit them - `make` is no longer "the single entry point". Frankly, I'm not a big fan of `make` being used like this either. I prefer executable like `bin/test-integration`. With that approach, there's not so much contradiction, cause it feels more natural to use `cargo test` alongside `bin/test-integration ...`.

MOZGIII:
I've touched on this at https://github.com/timberio/vector/pull/2685/files#r431876203, but here is a better place to non-diff related discussion. How do you feel about getting rid of `make`? I know I've brought this up before is some issue or PR, but maybe we want to revisit this.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Yeah -- I think the k8s jobs will have their own special deps :(

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Right, we have `make environment-check`, to run in the env, and `make check` to run on the host.

What you wrote works exactly fine, `make environment-check` just does the same thing as a shorthand since it's mean to be called from the host directly. Primarily, I suspect we'll all use `make check`, but all also use `make generate`.

Ideally, you'd install Nix and just use that though. :)

rfcs/2020-05-25-2685-dev-workflow-simplification.md
```diff
@@ -0,0 +1,244 @@
...
+
+Since not everyone has a full working native environment, or can use Nix, we took our Nix environment and stuffed it into a Docker (or Podman) container!
+
+This is ideal for users who want it to "Just work" and just want to start contributing. It's also what we use for our CI, so you know if it breaks we can't do anything else until we fix it. 😉
+
+**Before you go farther, install Docker or Podman through your official package manager, or from the [Docker](https://docs.docker.com/get-docker/) or [Podman](https://podman.io/) sites.**
+
+```bash
+# Optional: Only if you use `podman`
+export container-tool="podman"
```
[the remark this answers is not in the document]
Hoverbear: Agreed, are you ok with having a small-case option so the invocations can look more shell-command like?

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Whoops! Thanks so much for catching that!

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Yeah I had basically the same conflict. Should we get rid of these jobs or just document that cargo works fine too?

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Really? I had no idea. WHat way is familiar? THis is just env vars?

rfcs/2020-05-25-2685-dev-workflow-simplification.md
MOZGIII: Hmm, got it. It's quite obvious, but did you think about doing something like `make environment check` instead of `make environment check`? I'd try to avoid having multiple definitions for `check` and `environment-check`. Or do you have something else in mind to eliminate that duplication?

rfcs/2020-05-25-2685-dev-workflow-simplification.md
MOZGIII: Would be cool if we pull them in via nix too - at least `skaffold` and `minikube`.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+Now that you have your coffee, you can run any of the following:
+
+```bash
+# Enter a shell with optimized mounts for interactive processes.
+# Inside here, you can use Vector like you have full toolchain (See below!)
+make environment
+# Add extra cli opts
+make enviroment cli-opts="--publish 3000:2000"
```
MOZGIII, line 54: How do you feel about using `net=host` for environment by default?

Hoverbear:
@MOZGIII So I'd be fine with removing `make` but I don't see an immediate need to remove it. I didn't really want that to be the scope of this RFC since that would impact our release processes etc. I was just trying to bite off the dev/test/website flows to give newer contributors a better experience.

You can still use `cargo` after this, you don't need to change anything. If you accidentally run make it works with your local tools, or you don't feel  like maintaining Rubies, you can run `make environment-generate` etc. :) If you are super awesome you might try nix, and if you do cargo etc will all still work still! :) You can try it on `nix-env-test` branch if you're curious.

My goal is to reduce the amount of docker-compose/bash magic in our builds and try to merge some containers we use.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: I would love to do `make environment check`!!! Sadly I'm not sure how'd that work with how make does tasks?

I also considered keeping the current ENV var method, but I think the explicit, per command opt in is useful so you don't accidently start container builds etc.

MOZGIII:
> I didn't really want that to be the scope of this RFC since that would impact our release processes etc

Ok, makes sense, let's omit this for now.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: We can! https://nixos.org/nixos/packages.html?channel=nixpkgs-unstable&query=skaffold  https://nixos.org/nixos/packages.html?channel=nixpkgs-unstable&query=minikube

Hoverbear: We can pass the docker socket into the environment, no big deal. And that won't be an issue in the nix shell either! Only in docker :)

rfcs/2020-05-25-2685-dev-workflow-simplification.md
MOZGIII: I'm oversimplifying, but `environment` target could just set `RUN` to `scripts/environment.sh`. And if we have `RUN`  empty by default, and keep using is as prefix - we'd, I think, get what we want.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
MOZGIII: I'm afraid passing docker socket is not enough, we'll also need to pass socket to the docker-in-docker that `minikube` starts on the node. It a level 2 inception, and the issue there might be not the docker socket itself (which is indeed trivial), but the overlay network address of minikube docker.
We can try to figure this out, but I don't have high hopes in this. Should work fine with nix on the host though, so "just use nix".

Hoverbear: It might be the best option for kube :( I wonder if we can run inside minikube like `podman` or `docker` and solve it that way?

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Hmmm, let me play with the idea of overwriting the env variable....

rfcs/2020-05-25-2685-dev-workflow-simplification.md
MOZGIII: Inside `minikube`? Well, `minikube` has various drivers, interesting ones are `docker` (default), `none` and `podman`. `podman` is experimental though, and we won't be able to use it in the integration yet. `none` might be a good fit when running in the in-docker environment. Actually, no, it only seems like it's a good fit, but it won't work properly in docker container with docker socket from host passed inside...

rfcs/2020-05-25-2685-dev-workflow-simplification.md
bruceg: This is a pretty common idiom for setting (overriding) settings in makefiles on the command line. Mind you, the convention for settings is all-caps as indicated before, but I have seen both ways. For example, a common pattern for installing into an alternate path is `make install PREFIX=/tmp/fake/root`

rfcs/2020-05-25-2685-dev-workflow-simplification.md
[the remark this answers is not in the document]
bruceg: But it isn't "shell-command like", the `export` explicitly says it's a variable here, and it has been vanishingly rare for me to see `export lowercase`.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
bruceg: > Then `make environment blah` will first switch `RUN`, then invoke `$RUN blah` aka `scripts/environment.sh blah`.

Will it do that even in the presence of parallel make options (like having `MAKEFLAGS=-j8`)? That looks like a race condition. Better would be `make MODE=environment blah`.

Hoverbear: So I played with this a bit: 
image

This works, but we'd need to add some complexity around `build` etc to make sure artifacts are correctly exported.

MOZGIII: > we'd need to add some complexity around `build` etc to make sure artifacts are correctly exported.

:frowning:

What if, instead of adding magic to `build` we just add a separate target, like `environment-copy-artifacts` or sth?

IMO, having extra complexity in tasks might be the root cause of the problems that we have currently - a few iterations and we get the same exploding amounts combinations.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: I'm fine with all caps :)

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Sure but then we're running `make environment generate environment-copy-artifacts` instead of `make environment-generate` and that's not awesome.

Hoverbear: This is why I was falling back to having `make environment-generate` etc. It's simple, explicit, opt in.

Hoverbear: Maybe `make generate ENVIRONMENT=true` is best.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Breaks windows/mac compat so no.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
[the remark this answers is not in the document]
Hoverbear: Ok :)

rfcs/2020-05-25-2685-dev-workflow-simplification.md
lukesteensen: > Maybe `make generate ENVIRONMENT=true` is best.

I like this idea.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Could we spawn a container in the kube and run in there? So it'd be a knob like `CONTAINER_TOOL=podman`? You'd set `CONTAINER_TOOL=minikube` for kubernetes and get a shell into a kubernetes pod running the env with what you need?

Hoverbear:
Noting to myself for later I can solve some potential OpenSSL issues with:

```nix
        export OPENSSL_DIR="${openssl.dev}"
        export OPENSSL_LIB_DIR="${openssl.out}/lib"
```

rfcs/2020-05-25-2685-dev-workflow-simplification.md
MOZGIII: Keeping them makes sense from the  "single entry" point of view. I wonder how it would work if we remove them though. I won't miss them. Let's try removing them and see what it does?

rfcs/2020-05-25-2685-dev-workflow-simplification.md
MOZGIII: `minikube` itself with driver `docker` runs containers. It'll run them in `docker`. We'll then have a docker-in-docker. So, if minikube itself runs in docker, we might get a
- docker-in-docker-in-docker situation - this is hell for doing any kind of work against
- minikube in docker, but with using host docker socket - this might work, but we'll likely have issues with overlay networks being isolated - i.e. `environment` container (where `minikube` and our other stuff is running) won't have access to the network of the k8s cluster node container that `minkube` spawned.

I initially estimated the complexity of keeping `minikube` in docker-powered environment at the k8s RFC, and I think my estimate still holds. It's simply not worth it. The amount of complexity that docker adds is way too high. It's not going to be a helpful tool, but rather a constant cause of issues.
Managing the binaries with nix for use on host is already more than enough, and super cool! :smile:

rfcs/2020-05-25-2685-dev-workflow-simplification.md
MOZGIII: > Sure but then we're running `make environment generate environment-copy-artifacts` instead of `make environment-generate` and that's not awesome.

The good thing about it is that disabling unnecessary steps becomes a very natural thing - just don't specify them. But then, there's no way to specify those commands to run by default... Or is there? Yeah, I'm pretty sure we could make a thing where we just do `make environment qwe`, and it runs `qwe-environment-post` step automatically if we're in the `environment` mode, and doesn't if we're at "host" mode.
Not insisting on this though, so if you like `make environment-qwe` better - let's discuss that.

If we go with `make environment-*`, I'd think we ideally want a wildcard definition for them, so that they all have the same predictable behavior, and, most importantly - that it's defined at one place, and not for every command. Not sure about how to achieve it with `make`, but at least from the perspective of the goals - I think this is what we aim for. Otherwise, every command will have two definitions again, that can potentially diverge.
Do you share this point of view?

MOZGIII: `make generate ENVIRONMENT=true` also works fine, I guess.

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: Okay, so no action right now. :)

rfcs/2020-05-25-2685-dev-workflow-simplification.md
Hoverbear: I'd miss them. :)

Hoverbear:
@Jeffail Good idea! I'll add that in docs-level proposal!

PR #2692, opened 2020-05-26
feat(operations): More usable event RFC

Hoverbear:
Adds a new RFC proposing some minor changes to Event.

Ref #2414.


rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+- **Programmatically, through our internal code.** When used in this way, the `LogEvent` is being interacted with through static, safe interfaces and we can utilize the full type system available to us.
+- **Through configuration or FFI via knobs or scripts.** When used this way, the configuration format (at this time, TOML) may impact how aesthetic or expressive our configuration is. For example, TOML values like `a.b.c` as defined in a configuration. This is TOML table syntax, and we can't just magically treat it like a `Path`, a conversion step is necessary.
+
+## Understanding `LogEvent` in Vector 0.9.0
+
+Before we explore how to improve `LogEvent`, we must make sure we're familiar with the current one! Let's review the structure and some examples of lifetimes of this structure.
+
+### `LogEvent` Internal Structure
+
+It's current implementation meets our needs currently from several ergonomics angles (notably the flattening code), but it we must twist it in several painful ways to work with it (notably buffers to flatmaps to `BTree` style APIs).
```
bruceg, line 73: "it we must"? Also the leading "It's" should be "Its" and is somewhat dangling.

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+    Number(Number),
+    String(String),
+    Array(Vec<Value>),
+    Object(Map<String, Value>),
+}
+```
+
+Differences are:
+
+- `Event` uses a `BTreeMap` , where `serde_json`'s `Map` type is only `BTreeMap` when the `preserve_order` feature flag is turned off. We do not enable this, feature, so they are equivalent.
```
bruceg, line 111: This isn't really a difference, is it?

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+- A `Bytes` type in our `Value` type exists because `serde_json::value::Value` and most JSON implementations support this type through the `String` type, which isn't ideal for applying `Bytes` related functions on. We hold this special `Bytes` variant to support better in-pipeline processing for Vector.
```
bruceg, line 112: "applying … on" ⇒ "applying … to" or even just drop the "on". Might want to mention the lossy nature of Bytes-to-String conversions here or below.

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+- Our `Value` type lacks a `String` variant, using the `Bytes` variant instead, preventing us from having optimized `String` operations.
+- A `Timestamp` type in our `Value` type exists for largely the same reason the `Bytes` variant exists: Optimize in-pipeline processing.
+- A `Float` and `Integer` type exist where `serde_json::value::Value::Number` exists also exist for optimizing in-pipeline processing.
```
bruceg, line 115: Suggestion: Separate `Float` and `Integer` types exist whereas `serde_json` only has `Number`, also for optimizing …

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+    pub fn remove_prune(&mut self, key: &Atom, prune: bool) -> Option<Value>;
+    pub fn keys<'a>(&'a self) -> impl Iterator<Item = String> + 'a;
+    pub fn all_fields<'a>(&'a self) -> impl Iterator<
+				Item = (String, &'a Value)
+		> + Serialize;
+    pub fn is_empty(&self) -> bool;
+}
+```
+
+Many of these APIs offer functionality ***coarsely similar*** to a `FlatMap`. These APIs were added ad-hoc, as needed over time, and were not designed intentionally to all fit together. In addition, some changes were added unobtrusively to aid in the migration path.
```
bruceg, line 150: Which migration path are you referencing here?

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+
+Modify the [Coercer transform](https://vector.dev/docs/reference/transforms/coercer/#field-name) Types section.
+
+```bash
+Enum, must be one of: "bool" "float" "int" "string" "timestamp", "bytes"
+```
+
+# Prior Art
+
+In the [WASM transform](https://github.com/timberio/vector/pull/2006/files) our demo POC was using protobufs, which by their nature are not always UTF-8. We had to patch Vector to support Bytes during this time. It currently does so via a fallback when non-UTF-8 bytes are found.
```
bruceg, line 178: I would think protobufs are *rarely* valid UTF-8, not merely *not always*.

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+
+# Sales Pitch
+
+This RFC ultimately proposes the following steps:
+
+1. Rename `Value::Bytes` to `Value::String` .
+2. Add `Value::Bytes`.
+3. Significant UX improvements on `LogEvent`, particularly turning JSON ↔ `LogEvent`.
```
bruceg, line 186: Wording issue: All the other steps but 3 and 5 start with a verb. Also, we should perhaps prefer words over symbols (ref "↔")

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+4. Refactor the `PathIter` to make `vector::Event::Path` type.
+5. Significant UX improvements on `Path` , particularly an internal `String` ↔ `Path` with an `Into` that does not do path parsing, as well as a `Path::new(s: String)` that does.
+6. Refactor all `LogEvent` to accept `Into<Path>` values.
+    1. Remove obsolete functionality like `insert_path` and `remove_prune` since the new `Path` type covers this.
```
bruceg, line 190: I'm not clear how a new `Path` type can include the distinction between `remove` and `remove_prune`.

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+
+# Sales Pitch
+
+This RFC ultimately proposes the following steps:
+
+1. Rename `Value::Bytes` to `Value::String` .
+2. Add `Value::Bytes`.
+3. Significant UX improvements on `LogEvent`, particularly turning JSON ↔ `LogEvent`.
+4. Refactor the `PathIter` to make `vector::Event::Path` type.
+5. Significant UX improvements on `Path` , particularly an internal `String` ↔ `Path` with an `Into` that does not do path parsing, as well as a `Path::new(s: String)` that does.
```
bruceg, line 188: If there are two ways to produce a path, one with parsing and one without, I think that should be clear from the name, or use an interface *like* the `FromStr` trait that is failable (but even `FromStr` might not make it clear that it is parsing).

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+6. Refactor all `LogEvent` to accept `Into<Path>` values.
+    1. Remove obsolete functionality like `insert_path` and `remove_prune` since the new `Path` type covers this.
+    2. Refactor the `keys` function to return an `Iterator<Path>`
+7. Add an `Entry` style API to `LogEvent`.
+    1. Remove functionality rendered obsolete by the Entry API like `try_insert`, moving them to use the new Entry API
+8. Provide `iter` and `iter_mut` functions that return `(Path, Value)`.
```
bruceg, line 194: Nit pick: these functions won't return that type, they will yield it (or yield elements of type …).

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,223 @@
...
+
+We believe these steps will provide a more ergonomic and consistent API.
+
+# Drawbacks
+
+The changes to `Path` and `PathIter` may lead to confusion or footgunning. We **must** focus on ensuring safety from ourselves when designing this API.
+
+Supporting `Value::Bytes` means we may take on additional maintenance burden and have to make new transforms that deal with bytes.
+
+Some naming/API changes, producing a mental burden on our developers.
```
bruceg, line 205: Might be good to make this a complete sentence.

rfcs/2020-05-25-more-usable-logevents.md
Hoverbear: Gosh you're right that's a terrible sentence.

rfcs/2020-05-25-more-usable-logevents.md
Hoverbear: It's not. I originally suspected it was a difference and left it here. I'll remove it.

rfcs/2020-05-25-more-usable-logevents.md
Hoverbear: `insert_path` etc. I'll add an example. :)

rfcs/2020-05-25-more-usable-logevents.md
Hoverbear: True!

rfcs/2020-05-25-more-usable-logevents.md
Hoverbear: Okay, agreed. We should make it obvious.

Hoverbear:
@lukesteensen 

> Bytes vs Strings
> 
> Renaming Value::Bytes to Value::String definitely feels like a move in the right direction, but I think there will be some interesting questions that pop up if we try to enforce UTF8 and especially when we reintroduce a new separate Bytes.
> 
> For example, what about a Kafka source that feeds into a regex transform? Kafka is definitely byte-oriented instead of string, but we currently have the benefit of that pipeline just working as long as the bytes are valid UTF8. Should we encourage transforms to operate on bytes instead of strings wherever possible?
> 
> What about sources that logically return strings but don't strictly guarantee UTF8? We could push up the to_string_lossy to that point instead of the current strategy where we deal with it downstream, but there is the downside that certain use cases will become lossy that previously were not.
> 
> We could opt not to enforce UTF8, but that opens its own set of questions around how best to use the two different types.

Generally agree. We already enforce UTF-8 on our byte strings though. I'd be **very, very, very much preferring it** if we didn't actually have this "fallback" and we made Bytes a type you'd get from some sources via a `raw` encoding. That `raw` encoding would be a different RFC though. Would you be ok just introducing the bytes type but not using it yet?

Hoverbear:
@lukesteensen 

> Path type UX
> 
> I'd be curious to see a few more details on this topic. It seems like you're referring to two separate code paths, one for internal use and one for user-provided input. That sounds like a great idea. I might even go as far as a UserProvidedPath type that goes through an explicitly fallible conversion before turning into a useable Path.
> 
> I may have interpreted this wrong, but I am a bit hesitant to have Path::new(s: String) behave differently from Path::from(s: String).

@bruceg had similar [feedback](https://github.com/timberio/vector/pull/2692#discussion_r431481971). I'm pondering options. :)

Hoverbear:
@bruceg 

> Has any consideration been given to optimizations like LRU caching destroyed events to speed up their re-creation (like the Linux kernel's slab cache)? Is there any way of using templates to speed up event creation or modification when the "shape" of the result can be known in advance (ie simple sources, some transforms, etc)? This is probably out of scope for this RFC.

Absolutely! We've been chatting about several ideas like this. I'd love to chat with you more about this template idea. :) This RFC is kind of intended to lead into a future one on perf. This is kind of trying to cut at some of the UX to enable future work as well.

rfcs/2020-05-25-more-usable-logevents.md
Hoverbear: We've discussing removing those conversions now

Hoverbear:
Pushed changes:

* Removes implicit converstion for strings/bytes, leaving bytes as unused but for future use.
* Path type's `::from()` and `::new()` take strings as is, and `Path::build()` parses it, appropriate for user input or configs.

lukesteensen:
> We already enforce UTF-8 on our byte strings though

We coerce in a way that doesn't fail if they're not UTF-8, which is different from enforcing it. I just want to be careful about changing existing behavior which is rather lenient. 

> Would you be ok just introducing the bytes type but not using it yet?

We can probably hold off if it's not used _at all_, but I'd be perfectly fine to add it exclusively for some small wasm use cases at first and expand usage over time.

Hoverbear:
@lukesteensen I'd propose we add it to something like the kafka source just so we have it as live code and tested, then. :)

rfcs/2020-05-25-more-usable-logevents.md
Hoverbear: Hmmm,yeah, you're right I think. We might not be able to remove it.

After looking at the `remove_prune` test(!) I feel like I need to explore the functionality a bit more.

https://github.com/timberio/vector/blob/b6e53d52249370e8b5bd4006fca674c3c99b8302/src/event/util/log/remove.rs#L156-L182

lukesteensen:
We could do that, we'd just need to figure out if we'd be changing the behavior of the kafka source in the process and if that's desirable.

rfcs/2020-05-25-more-usable-logevents.md
Hoverbear: I'm striking this from  the RFC as I don't really have a clear conclusion yet. Let's leave it for later or a spec'd issue instead of this RFC.

Hoverbear:
I've removed the `remove_prune` removal, did a bit more explicit `Path`<->`String` barrier, and and fixed the nit. I think this is good to merge.

rfcs/2020-05-25-more-usable-logevents.md
```diff
@@ -0,0 +1,222 @@
...
+# Sales Pitch
+
+This RFC ultimately proposes the following steps:
+
+1. Rename `Value::Bytes` to `Value::String`.
+2. Add `Value::Bytes`. (Unused for now -- to be used later in raw codecs and WASM)
+3. Add UX improvements on `LogEvent`, particularly turning JSON into or from `LogEvent`.
+4. Refactor the `PathIter` to make `vector::Event::Path` type.
+5. Add UX improvements on `Path` , particularly an internal `String` ↔ `Path` with an `Into`/`From` that does not do path parsing, as well as a `Path::parse(s: String)` that does.
+6. Refactor all `LogEvent` to accept `Into<Path>` values.
```
lukesteensen, line 188: I'm still a little uneasy about this. My concern is just that allowing indexing into an event with a string opens the door for an unknowing contributor to add a `key: String` config field and have things "just work" with actually doing any parsing.

On the other hand, I get that it's much more convenient for internal use. So as long as we're careful I'm fine to give it a try.

Hoverbear: I would say my intention is that when we make a component, that any field which takes a path would accept a `Path` instead of a `String`. If they add a string and it "just works" that's fine, our review process should be rigorous enough that we notice during review or notice from the lack of testing.

Hoverbear:
See https://github.com/timberio/vector/milestone/42

PR #2693, opened 2020-05-26
chore: Merge duplicate RetryAction and RetryLogic

bruceg:

### DIFF
--- a/src/sinks/util/retries2.rs
+++ b/src/sinks/util/retries2.rs
@@ -1,3 +1,4 @@
+pub use super::retries::{RetryAction, RetryLogic};
 use crate::Error;
 use futures::FutureExt;
 use std::{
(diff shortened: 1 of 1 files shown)


fanatid:
While code really duplicated I think better left as it now. After upgrading sinks which uses code from `retries` to `retries2`, `retries2` can be renamed to `retries` much easier than with these changes.

Sinks for upgrade:
- `aws_cloudwatch_logs`
- `aws_cloudwatch_metrics`
- `aws_kinesis_firehose`
- `aws_kinesis_streams`
- `aws_s3`
- `clickhouse`
- `cloud_storage`
- `elasticsearch`

bruceg:
For background, this came about when trying to deal with some `RetryLogic` rework and running into some issues that potentially fell out from having two different copies of the code. That work looks like it will be done a different way, or at least not at this time, so we can drop this.

PR #2698, opened 2020-05-27
chore(kafka sink): Fix integration test

fanatid:
Should fail, because https://github.com/timberio/vector/pull/2694 not here yet.

Not sure that this is right solution, maybe better put `rdkafka-plain` to `sinks-kafka`? (but vector also have `rdkafka-cmake`)

Edit:
details: https://github.com/timberio/vector/pull/2694#issuecomment-634809830


fanatid:
Thanks. What about adding `rdkafka-plain` / `rdkafka-cmake` feature to `sinks-kafka` feature? If I understand correctly vector will not work with kafka for: `--no-default-features --features sinks-kafka`?

MOZGIII:
As I understand it, the purpose of those features to be separate is to let users pick the compilation mechanism they like better. Thus, if we include `rdkafka-plain` / `rdkafka-cmake` into `sinks-kafka` this will prevent users from actually choosing the alternative.

I'm not really aware though why do we have it implemented like that. However, if we need to locally decide on whether to include `rdkafka-plain` / `rdkafka-cmake` into `sinks-kafka` - I'd suggest we don't, since in the current design that would be problematic.

PR #2653, opened 2020-05-20
feat(new source): Initial `kubernetes_logs` implementation

MOZGIII:
This PR implements Kubernetes source according to #2222.
There are a few things are different from the RFC:
- the name for the source is `kubernetes_logs` - it's wasn't properly addressed at the RFC, but the logs and other source kinds for the kubernetes would very likely have very different internal design, so we should probably keep them as separate units; to disambiguate from the very beginning I modified the name to be `kubernetes_logs`;
- had to bump MSKV to 1.15 - watch bookmarks are introduced in that version, and this is what we want to have to avoid extra desync issues.

Done:

- closes #2626;
- closes #2627;
- closes #2628;
- closes #2629;
- closes #2630;
- closes #2631.

To do (before the merge):

- [x] shutdown logic fix;
- [x] address the fingerprinting issue (#2890);
- [x] move the annotation process to the create_event to optimize out access to the `file` field;
- [x] ensure that `MultiResponseDecoder` doesn't leak memory;
- [x] add new dependencies to the environment (`minikube`, `skaffold`, etc);
- [x] bookmark parsing error (https://github.com/Arnavion/k8s-openapi/issues/70).

To do (after the merge):

- more configuration and knobs;
- #2632 (integration tests);
- #2633 (documentation).

---

The overall approach was to build highly composable components, so that we can reuse them for further work - like for adding a sidecar-oriented transform for pod annotation, or exclusions based on namespace labels.

PR #2702, opened 2020-05-28
enhancement(tests): Kubernetes E2E test framework

### LINKED ISSUE #2632: K8s: integration tests for the k8s source
Implement integration tests for the k8s source.

To address together with #2631.

Part of #2222.

Since every individual building block for the k8s implementation is to be thoroughly tested, only a simple smoke test is required in the integration test. Mainly to ensure the glue code works accordingly, and that we didn't go out-of-spec somewhere in our implementation.

Things to test:
- log files selection (aka filtering);
- format parsing, including automatic partial merges;
- pod metadata annotations;
- race conditions around pod deletion (https://github.com/timberio/vector/pull/2653#discussion_r431690299).

MOZGIII:
~Will be rebased after #2653 is merged.~ :tada: 

This PR adds E2E (end-to-end) test for our Kubernetes integration.

It also renames `kubernetes-integration-tests` into `kubernetes-e2e-tests`, and introduces a concept of an E2E test itself - a kind of test where we build a full Vector binary, deploy it and assert it's observable behavior. Very similar conceptually to the test harness, but it's part of the main Vector tree.

Check out CI jobs to see it in action, and take a look at updated [`CONRTIBUTING.md`](https://github.com/timberio/vector/blob/k8s-test-framework/CONTRIBUTING.md#kubernetes-e2e-tests) for how to run tests locally.

Closes #2632.


lib/kubernetes-test-framework/src/resource_file.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub fn path(&self) -> &Path {
+        self.path.as_path()
+    }
+}
+
+impl Drop for ResourceFile {
+    fn drop(&mut self) {
+        std::fs::remove_file(&self.path).expect("unable to clean up custom resource file");
+    }
+}
```
lukesteensen, line 27: I'm pretty sure this is redundant with the use of a `Tempdir`. It should also delete itself on drop.

lib/kubernetes-test-framework/src/wait_for_resource.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Wait for resource deletion.
+    Delete,
+    /// Wait for the specified condition.
+    Condition(C),
+}
+
+/// Wait for a set of `resources` within a `namespace` to reach a `wait_for`
+/// condition.
+/// Use `extra` to pass additional arguments to `kubectl`.
+pub async fn namespace<CMD, NS, R, COND, EX>(
```
lukesteensen, line 23: I _think_ the convention for type parameters is `PascalCase` not all uppercase.

MOZGIII: I don't like those either. Pascal variant looks better here, I'll change it. Thx!

lib/kubernetes-test-framework/src/resource_file.rs
MOZGIII: Good point. I think there was some kind of trouble I had to work around with this, and I didn't really focus on it, and just did it to advance. I'll revisit this part again.

MOZGIII: Just removing it worked

lib/kubernetes-test-framework/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "kubernetes-test-framework"
+version = "0.1.0"
+authors = ["MOZGIII <mike-n@narod.ru>"]
```
ktff, line 4: Should this be `Vector Contributors <vector@timber.io>` ?

lib/kubernetes-test-framework/src/interface.rs
```diff
@@ -0,0 +1,27 @@
...
+
+/// An interface between the test framework and external CLI commands and test
+/// utilities.
+#[derive(Debug)]
+pub struct Interface {
+    /// A command used to deploy `vector` into the kubernetes cluster and
+    /// delete if from there.
+    pub deploy_vector_command: String,
+
+    /// A `kubectl` command used to
```
ktff, line 13: Used for what?

tests/kubernetes-e2e.rs
```diff
@@ -0,0 +1,725 @@
...
+        .wait(
+            "test-vector-test-pod",
+            vec!["pods/test-pod"],
+            WaitFor::Condition("initialized"),
+            vec!["--timeout=30s"],
+        )
+        .await?;
+
+    // Wait for some extra time to ensure pod completes.
+    tokio::time::delay_for(std::time::Duration::from_secs(10)).await;
```
ktff, line 320: It would be nice to wait for `Succeeded` pod phase instead of a delay, or maybe in conjuction if there is a race with other parts of the system, like filesystem and container runtime.

tests/kubernetes-e2e.rs
```diff
@@ -0,0 +1,725 @@
...
+
+        // Remember that we've seen a control marker.
+        got_control_marker = true;
+
+        // Request termination in a while.
+        let mut stop_tx = stop_tx.clone();
+        tokio::spawn(async move {
+            // Wait for two minutes - a reasonable time for vector internals to
+            // pick up new `Pod` and collect events from them in idle load.
+            let duration = std::time::Duration::from_secs(120);
```
ktff, line 623: The same with before mentioned `Succeeded` pod phase. We could decrease some of that delay.

MOZGIII:
> it would be great if we could reduce the runtime of tests, ~1h is a lot and will slow down dev process.

The majority of time is spent at the build stage, so it's a known issue, and the one shared with Test Harness invocation. The test logic itself is relatively quick, only takes a couple of minutes.

For local development and iteration, I recommend using `QUICK_BUILD` mode - this uses the default debug build that your system is likely to build very quickly. I could make multiple iterations on the test code within just 10 minutes on my system.

tests/kubernetes-e2e.rs
MOZGIII: I didn't find a way to wait for `Succeesed` state of a raw `Pod` (the one not in a `Job`). If you can show me `kubectl` command example - that would be great!
For now, tests rely on waiting for `initialized` state, and that test, in particular, assumes the `echo` in a `Pod` completes within 10 seconds. It's a race, and it would definitely be better to have it work based on explicit synchronization. I'll take a look to see if there's any state change we rely on to improve it.

tests/kubernetes-e2e.rs
MOZGIII: This one is racy by nature - the longer we wait, the more certainty we have that the excluded `Pod` logs weren't picked up because it was excluded, rather than them being just slow to load.
We can think about other ways to assert that, but this one seemed very straightforward and fair.

tests/kubernetes-e2e.rs
ktff: True, I was still thinking in the context of `http` api. Shame `kubectl` doesn't have this.

tests/kubernetes-e2e.rs
ktff: > This one is racy by nature 

That's true. The wait for `Succeeded` was to be used to achieve same/greater level of probability. Unfortunately it's not possible as you commented before.

tests/kubernetes-e2e.rs
MOZGIII: I thougt about it. Technically it’s doable, and it’s not that much much work. One provlem is, though, the solution would itself be racy. I’m thinking about calling `kubectl -ogo -f '{{...}}'` in a loop until we see the status we’re looking for. To safeguard from the races, we’d have to restrict the API such that we can only wait for certtain  statuses, instead of the arbitrary ones.

MOZGIII: I created #3150 so we don't lose track of it.

MOZGIII:
Is anyone in the process of reviewing this still? Ping @binarylogic, @Hoverbear. :smile:

.github/workflows/tests.yml
```diff
@@ -265,33 +265,52 @@ jobs:
...
       matrix:
-        kubernetes:
-          - v1.18.2
-          - v1.17.5
-          - v1.16.9
-          - v1.15.11
-          - v1.14.10
+        minikube_version:
+          - 'v1.11.0' # https://github.com/kubernetes/minikube/issues/8799
+        kubernetes_version:
```
Hoverbear, line 275: So this is... 5 * 3 test runners **just** for k8s?

.github/workflows/tests.yml
```diff
@@ -265,33 +265,52 @@ jobs:
...
+          - 'v1.18.6'
+          - 'v1.17.9'
+          - 'v1.16.13'
+          - 'v1.15.12'
```
Hoverbear, line 279: Why do we test these versions?

CONTRIBUTING.md
```diff
@@ -627,6 +631,74 @@ required.
 Essentially, the steps you have to take to deploy manually are the same that
 `skaffold` will perform, and they're outlined at the previous section.
 
+#### Kubernetes E2E tests
+
+Kubernetes integration is quite complicated, a lot of things can go wrong.
```
Hoverbear, line 636:
```suggestion
Kubernetes integration has a lot of parts that can go wrong.
```

Slight reword so we don't imply that kubernetes is too complicated for the user.

CONTRIBUTING.md
```diff
@@ -627,6 +631,74 @@ required.
...
+> E2E tests normally run at CI, so there's typically no need to run them
+> manually.
+
+##### Requirements
+
+* `kubernetes` cluster (`minikube` has special support, but any cluster should
+  work)
+* `docker`
+* `kubectl`
+* `bash`
```
Hoverbear, line 650: So these tests can't work from a Windows machine?

CONTRIBUTING.md
```diff
@@ -627,6 +631,74 @@ required.
...
+
+Vector release artifacts are prepared for E2E tests, so the ability to do that
+is required too, see Vector [docs](https://vector.dev) for more details.
+
+##### Running the E2E tests
+
+To run the E2E tests, use the following command:
+
+```shell
+CONTAINER_IMAGE_REPO=<your name>/vector-test make test-e2e-kubernetes
```
Hoverbear, line 660: Why not use `timberio/vector:dev` or something for a default so users don't need this?

CONTRIBUTING.md
```diff
@@ -627,6 +631,74 @@ required.
...
+```
+
+Where `CONTAINER_IMAGE_REPO` is the docker image repo name to use, without part
+after the `:`. Replace `<your name>` with your Docker Hub username.
+
+This will run Kubernetes E2E tests suite. If the command exit with 0 status
```
Hoverbear, line 666: I don't think this line is necessary. I think most users are able to tell that the output of a command has run and that it did not fail.

CONTRIBUTING.md
```diff
@@ -627,6 +631,74 @@ required.
...
+code - tests passed.
+
+You can also pass additional parameters to adjust the behavior of the test:
+
+* `QUICK_BUILD=true` - use development build and a skaffold image from the dev
```
Hoverbear, line 671: How about `DEVELOPMENT=true`? I'd like to use that same env in other tasks so it'd be nice to keep the number of vars down.

CONTRIBUTING.md
```diff
@@ -627,6 +631,74 @@ required.
...
+  build. Useful for development of the tests or Vector code to speed up the
+  iteration cycles.
+
+* `USE_MINIKUBE_CACHE=true` - instead of pushing the built docker image to the
+  registry under the specified name, directly load the image into
+  a `minikube`-controlled cluster node.
+  Requires you to test against a `minikube` cluster. Eliminates the need to have
+  a registry to run tests.
+
+* `CONTAINER_IMAGE_REPO=<your name>/vector-test:tag` - completely skip the step
```
Hoverbear, line 683: I guess this exists because there is no `make build-docker-....` images and the `release` tasks are kind of awkward?

CONTRIBUTING.md
```diff
@@ -627,6 +631,74 @@ required.
...
+CONTAINER_IMAGE_REPO=<your name>/vector-test make test-e2e-kubernetes
+```
+
+Where `CONTAINER_IMAGE_REPO` is the docker image repo name to use, without part
+after the `:`. Replace `<your name>` with your Docker Hub username.
+
+This will run Kubernetes E2E tests suite. If the command exit with 0 status
+code - tests passed.
+
+You can also pass additional parameters to adjust the behavior of the test:
```
Hoverbear, line 669: I suggest you namespace the k8s specific variables to like `KUBERNETES_USE_CACHE` and `KUBERNETES_CONTAINER_IMAGE_REPO` so we don't pollute the global environment namespace with variables whose names are relatively unclear.

scripts/test-e2e-kubernetes.sh
```diff
@@ -0,0 +1,108 @@
...
+  local CHARS="abcdefghijklmnopqrstuvwxyz0123456789"
+  # shellcheck disable=SC2034
+  for i in {1..8}; do
+    echo -n "${CHARS:RANDOM%${#CHARS}:1}"
+  done
+  echo
+}
+
+# Build a docker image if it wasn't provided.
+if [[ -z "${CONTAINER_IMAGE:-}" ]]; then
```
Hoverbear, line 20: Do we not have an existing way to build docker images?

scripts/test-e2e-kubernetes.sh
```diff
@@ -0,0 +1,108 @@
...
+  # any value for `CONTAINER_IMAGE_REPO` (for instance, `vector-test` will do).
+  #
+  CONTAINER_IMAGE_REPO="${CONTAINER_IMAGE_REPO:?"You have to specify CONTAINER_IMAGE_REPO to upload the test image to."}"
+
+  # Assign a default test run ID if none is provided by the user.
+  TEST_RUN_ID="${TEST_RUN_ID:-"$(date +%s)-$(random-string)"}"
+
+  if [[ "${QUICK_BUILD:-"false"}" == "true" ]]; then
+    # Build in debug mode.
+    cargo build
```
Hoverbear, line 38: You should build vector correctly with the makefile here.

distribution/kubernetes/vector-namespaced.yaml
```diff
@@ -49,6 +49,11 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            # Set a reasonable log level to avoid issues with internal logs
+            # overwriting console output at E2E tests. Fell free to change at
```
Hoverbear, line 53:
```suggestion
            # overwriting console output at E2E tests. Feel free to change at
```

.github/workflows/tests.yml
MOZGIII: Well, it was worth it during development - helped to find a couple of interesting bugs along the way in unexpected places, that were only present in non-trivial combinations.
I don't mind reducing it, but it's not clear what subset of those is reasonable.

On a side note, there are a lot of ways we could optimize the E2E tests in general, to make them run quicker so that they don't create as much load on the system. Technically, the "payload" of these tests is very lightweight, and the most time, yet again, is spent during env prep and build. If we could share build artifacts - it would be great.

I feel like if we optimize the process in general we won't have to worry about having 15 tests...

We could move them into a separate workflow though, such that they'll live under their own section.

Hoverbear: Yeah this would consume actually 15% of our total possible github actions runners so we should consider the value here very, very carefully.

.github/workflows/tests.yml
MOZGIII: According to the RFC, we test all versions from MSKV to the latest released version. MSKV is 1.15.

lib/k8s-test-framework/src/reader.rs
```diff
@@ -0,0 +1,123 @@
...
+        // Ensure wait doesn't fail, and that we exit status is success.
+        let exit_status = reader.wait().await.expect("wait failed");
+        assert!(exit_status.success());
+    }
+
+    #[tokio::test]
+    async fn test_reader_inifinite() {
+        let mut command = Command::new("bash");
+        command.arg("-c");
+        command.arg(r#"NUM=0; while true; do echo "Line $NUM"; NUM=$((NUM+=1)); sleep 0.01; done"#);
```
Hoverbear, line 88: Put a limit on this `while true` please. Maybe 1k?

.github/workflows/tests.yml
MOZGIII: We don't really have to test 1.14, but we do it cause people asked if we'll support it, and our code works for 1.14 too. We might want to reduce MSKV to 1.14.

CONTRIBUTING.md
MOZGIII: Technically, you should be able to run tests on Windows via git bash. I didn't test it though...
Also, I want to note that running tests *on* Windows doesn't correlate to testing *Windows clusters*, and that we don't address that yet.

tests/kubernetes-e2e.rs
```diff
@@ -0,0 +1,732 @@
...
+};
+use std::collections::HashSet;
+
+const VECTOR_CONFIG: &str = r#"
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: vector-config
+data:
+  vector.toml: |
```
Hoverbear, line 21: I really think we should teach Vector to handle YAML so this can look slick!

tests/kubernetes-e2e.rs
```diff
@@ -0,0 +1,732 @@
...
+    drop(test_namespace);
+    drop(vector);
+    Ok(())
+}
+
+/// This test validates that vector properly annotates log events with pod
+/// metadata obtained from the k8s API.
+#[tokio::test]
+async fn pod_metadata_annotation() -> Result<(), Box<dyn std::error::Error>> {
+    let _guard = lock();
```
Hoverbear, line 438: Since this is isolated to a cargo run you could use `RUST_TEST_THREADS` to force the runner to only run 1 test at a time here. That'd avoid the complexity of the mutex maybe?

CONTRIBUTING.md
MOZGIII: Users don't have push access to `timberio/vector:dev`, they'll have to specify something either way.
I was thinking about providing a default value if minikube cache is used, sane value in that case would be `localhost/<something>` due to some CRI internals.

scripts/test-e2e-kubernetes.sh
```diff
@@ -0,0 +1,108 @@
...
+export KUBE_TEST_DEPLOY_COMMAND="scripts/deploy-kubernetes-test.sh"
+
+# Prepare args.
+CARGO_TEST_ARGS=()
+if [[ -n "${SCOPE:-}" && "$SCOPE" != '""' ]]; then
+  CARGO_TEST_ARGS+=("$SCOPE")
+fi
+
+# Run the tests.
+cargo test \
```
Hoverbear, line 101: Really really happy to see I can run this on my localhost.

CONTRIBUTING.md
MOZGIII: I'm pretty sure `DEVELOPMENT` didn't exist when I implemented this :smile: I'll see it if fits - we might want to deliberately have those different. I'll check and respond here.

scripts/test-e2e-kubernetes.sh
MOZGIII: I this case, the idea is to use the same build flow that devs usually use to reuse as much of the flow as possible. I'm using `cargo check` / `cargo test` usually, and not `make`. Thus `cargo build` if preferable to `make` here. Does it make sense?

lib/k8s-test-framework/src/reader.rs
MOZGIII: This is supposed to be infinite. What's the problem with having no limit?

tests/kubernetes-e2e.rs
MOZGIII: Good idea! This could make Helm config much more flexible too.

tests/kubernetes-e2e.rs
MOZGIII: This one is sort of a primary safety measure, while `RUST_TEST_THREADS` is secondary. This is to avoid a situation where running without `RUST_TEST_THREADS` causes test failures. It makes more sense to keep such restriction as an explicit part of the test.

At the other hand, `RUST_TEST_THREADS` makes test improves output readability - when tests run in parallel and lock on the mutex, they work alright, but the start and the relevant output of the test can go separately - as multiple tests print their starting lines at the same time, and then they run in sequence - very confusing.

CONTRIBUTING.md
MOZGIII: Oops, there's a mistake. This is supposed to be `CONTAINER_IMAGE`, not `CONTAINER_IMAGE_REPO`. Hope it makes more sense.
This can be set if you want to bypass image building altogether, and use a completely custom image. This might be useful, for instance, for testing release images. I used it for experimenting with vector building, and to avoid rebuilding vector each time I change tests. Overall, it's very useful to have such option. Sorry for an error - this is definitely confusing as is.

MOZGIII: Also, if the `CONTAINER_IMAGE` is not set - a new image is built using the `CONTAINER_IMAGE_REPO` and a unique random tag. This process uses `make` (unless `QUICK_BUILD` is requested).

.github/workflows/tests.yml
Hoverbear: So over time as k8s releases grow, so will this list?  Meaning more and more of our CI jobs will be just k8s tests?

CONTRIBUTING.md
Hoverbear: Yeah I have bash through busybox so I think this might work. I think this is fine for now. We'll need to support windows fully at some point though. We do support users developing on windows.

CONTRIBUTING.md
Hoverbear: That sounds reasonable. :)

CONTRIBUTING.md
MOZGIII: Prefixing all of them with `KUBERNETES_` is way too much characters to type, but `K8S_` would be fine I guess.
That said, I don't think global env pollution is an issue here. Unless the variables set globally at the system level (i.e. like in `~/.profile`) - but then it'd require more than just `K8S_` for them to make sense universally. I woundn't worry about that case.

Given that those are used in a very specific command, I don't think it's big issue. It's easy to type and remember though - and this is a big plus IMO.
If using them in `.envrc` or similar - it's often possible to use comments.

CONTRIBUTING.md
Hoverbear: It doesn't yet! :( You've probably noticed we use separate make jobs for this distinction. Maybe that's appropriate here too?

lib/k8s-test-framework/src/reader.rs
Hoverbear: So you're sure that this test will never, ever, result in a process running indefinitely on a users machine?

tests/kubernetes-e2e.rs
Hoverbear: Yeah it's kind of a toss up. I think having both is a sane solution that measures the costs of both. :)

CONTRIBUTING.md
MOZGIII: You're thinking about `make test-e2e-kubernetes` and a `make test-e2e-kubernetes-quick`, sth like that? I thought about doing it like that, but I discarded it because we'd still need to pass env vars is some cases, and adding entry points complicates things worse than just having one configurable one imo. It's a good idea though, thx for bringing this up.

I understand what you mean with `DEVELOPMENT` now. Well, I wouldn't worry about it now. Let's streamline it when it becomes a problem. For now, `QUICK_BUILD` captures the distinction with the regular flow better (for the "how do I make it quick!!1" situations :smile: ), and with lack of widely used `DEVELOPMENT` var - I think it'd be easier to find.

scripts/test-e2e-kubernetes.sh
MOZGIII: We do! This test just invokes, see below.

scripts/test-e2e-kubernetes.sh
MOZGIII: It runs from localhost, AND you can use regular `kubectl` to debug it if something goes wrong, because the test itself uses that same `kubectl` to drive it's operations! No more headaches with "test fails, but how do I repro it and start tinkering with it".

scripts/test-e2e-kubernetes.sh
MOZGIII: Oh, and there's a non-quick-build that uses makefile below. And that one is the primary one (used at CI), while this is an optional for quicker iteration.

lib/k8s-test-framework/src/reader.rs
MOZGIII: Yes, it's fine. Tests clean up all test pods at termination, and will kill this one too. And the tests themselves have a limit of the maximum amount of lines to churn through awaiting for the needle, until the test is force-failed.

CONTRIBUTING.md
MOZGIII: > We do support users developing on windows.
Yeah, that's a bit tricky to do for k8s, but we'll get there.

The dev flow (the one that uses skaffold) *is* currently limited to linux (on the contrary to e2e tests, what can be launched from Linux/macOS/Windows in theory). We can add support for using Windows/macOS there if we cross-compile `vector` for linux - but that would significantly increas the build time and won't be fit for quick iterations anymore.
I want to address it at some point. But not now - and most likely after we ship the "MVP for the integration".

.github/workflows/tests.yml
MOZGIII: I'm afraid so, at least that's the plan. We'll probably bump MSKV along. K8s unfortunately have plenty of edge cases, and we'd better test as many variants as we can.
The good thing is, though, it's quite easy to maintain. I'll improve this a little bit later so we get alerted to add new k8s versions.

.github/workflows/tests.yml
MOZGIII: Right. Well, we can exclude some of the matrix incarnations if we start facing issues... Just need to figure out which ones are the safest. Alternatively, this can be transformed into a single sequential invocation - but I'd hold on with that until we hit an actual problem.

CONTRIBUTING.md
MOZGIII: True :smile: It's mostly here for completeness, but it is indeed redundant for the majority of the users.
I'll remove it, good find!

CONTRIBUTING.md
MOZGIII: Added the default when `USE_MINIKUBE_CACHE=true` is set.

CONTRIBUTING.md
MOZGIII: Tried it on Windows - got stuck at the building vector into release artifacts. This should be easy to fix.
We just need to give Windows a little bit more attention to solve all the necessary dependencies before thinking about k8s-specific stuff.

CONTRIBUTING.md
Hoverbear: Maybe you missed it in the commit? 
image

CONTRIBUTING.md
Hoverbear: I do not think `QUICK_BUILD` is a good name for this value.

CONTRIBUTING.md
```diff
@@ -627,6 +631,79 @@ required.
...
+
+* `QUICK_BUILD=true` - use development build and a skaffold image from the dev
+  flow instead of a production docker image. Significantly speeds up the
+  preparation process, but doesn't guarantee the correctness in the release
+  build. Useful for development of the tests or Vector code to speed up the
+  iteration cycles.
+
+* `USE_MINIKUBE_CACHE=true` - instead of pushing the built docker image to the
```
Hoverbear, line 674: So by default when someone runs `make test-e2e-kubernetes CONTAINER_IMAGE_REPO=foo/bar` that pushes to the Docker hub? And this is the default?

CONTRIBUTING.md
MOZGIII: You still have to pass `USE_MINIKUBE_CACHE=true`.
We can actually autodetect whether `minikube` is used or now based on the `kubectl` context.
However, this would complicate the test setup a bit too much, and I chose not to do it. Besides technical reasons, the user experience won't change much, as k8s e2e tests still require a lot of involvement from the user. There was an attempt in eliminating this involvement before, with our previous k8s impl, but it wasn't very successful due to the amounts of edge cases. So, this time we took a different route and made things more robust, but less automatic.

`CONTRIBUTING.md` clearly documents the proper way of invoking the command, so this shouldn't be a problem for our users.

MOZGIII:
You're hitting a `minikube` bug here: https://github.com/kubernetes/minikube/issues/8799
Please use `minikube` v0.11.0 until it's resolved.

CONTRIBUTING.md
MOZGIII: Yep. This shouldn't be surprising. Delivering images via registry is a standard way to make an image available to the Kubernetes. There's simply no other generic way of doing it. `minikube` has special support for delivering images into a cluster it manages without the registry, but we don't want to reimplement all that logic for a generic k8s cluster.
Users can spin up local registry if they don't want to use a cloud one.

Hoverbear:
Ok. :) Consider this unblocked then.

MOZGIII:
Another issue: https://github.com/kubernetes/minikube/issues/8840
Rolling back v1.16.12 to v1.16.13 for now.

binarylogic:
@MOZGIII just checking in. What's remaining here? @Hoverbear given your review, do you approve or still need changes made?

Hoverbear:
I'll see if minikube has been updated and try to rerun.

Hoverbear:
So I'd like `make test-e2e-kubernetes` to work like our other test commands and not require additional env vars to run. Could we make `USE_MINIKUBE_CACHE=true` default?

MOZGIII:
`minikube` still at v1.12.1, so none of the issues were addressed.

> Could we make `USE_MINIKUBE_CACHE=true` default?

We can autodetect it. See https://github.com/timberio/vector/pull/2702#discussion_r460244100
I gave it some thought, and I think we autodetecting it isn't that bad. Defaulting it to jut `true` is very surprising, imo.

However, the command will still not be really "usable" as the rest of the test commands: those are optimized for local run, while E2E is targeted towards optimizing the CI run use case - the reason for that is it's main goal is to test release builds, rather than debug. The overall theme of this test requiring manual involvement from the user remains, and it's sort of by design.

Hoverbear:
Okay so we need to wait for minikube to update before we merge?

It still seems like `USE_MINIKUBE_CACHE` is basically asking you to:

* If true, use the local minihub docker registry
* If false, push images to the docker hub.

Can you explain why you think it would be more surprising to a user that a test run is using their docker credentials to push images? As opposed to use a local cache on the minikube they'll be testing against?

Hoverbear:
I think https://github.com/timberio/vector/pull/2702/commits/ec3f9c5308b1ccffb2867ca1d983f0e3358f01cd solves the `USE_MINIKUBE_CACHE` issue :)

MOZGIII:
I think everything's sorted out. Are we ready to merge?

> * If true, use the local minihub docker registry

It's slightly more complicated. `minikube cache` doesn't use local registry - it copies an image directly to the node it manages, and then it imports it and sets up a "lock" so that it's not GC'd by the `kubelet`. They ship their own custom system components into the `kube-system` to support it. So, even technically, no registry is involved.

MOZGIII:
> Okay so we need to wait for minikube to update before we merge?

I don't think it worth it. We can just keep using a working version. There are relevant TODOs in place, and issues at upstream are opened.

Hoverbear:
Ok!

Hoverbear:
So I noted as a ZFS user minikube can't work. So I had to use microk8s which does work very well if you do not have `docker-ce` from docker installed.

Hoverbear:
@MOZGIII I wonder if we can make `QUICK_BUILD` default for the `test-e2e-kubernetes` runs?

MOZGIII:
> @MOZGIII I wonder if we can make `QUICK_BUILD` default for the `test-e2e-kubernetes` runs?

It doesn't work on mac, and has some other limitations that make it poorly suitable to be the default.
We can work towards better E2E tests support in the future - once we have some issues sorted out we can make it the default.

Hoverbear:
Ok!

PR #2706, opened 2020-05-29
fix(Humio sink): Removing trailing slash from host parameter before constructing HEC URL

jmountifield:
This PR addresses https://github.com/timberio/vector/issues/2705

### DIFF
--- a/src/sinks/humio_logs.rs
+++ b/src/sinks/humio_logs.rs
@@ -65,7 +65,12 @@ impl SinkConfig for HumioLogsConfig {
 
 impl HumioLogsConfig {
     fn build_hec_config(&self) -> HecSinkConfig {
-        let host = self.host.clone().unwrap_or_else(|| HOST.to_string());
+        let mut host = self.host.clone().unwrap_or_else(|| HOST.to_string());
+
+        if host.ends_with('/') {
+            host.pop();
+        }
+
 
         HecSinkConfig {
             token: self.token.clone(),


fanatid:
Thanks! But this only solve humio sink, actual problem in splunk_hec:
https://github.com/timberio/vector/blob/f6e0eb2eb87fe874d56261d61522fdc2928d8966/src/sinks/splunk_hec.rs#L181

src/sinks/humio_logs.rs
```diff
@@ -65,7 +65,12 @@ impl SinkConfig for HumioLogsConfig {
 
 impl HumioLogsConfig {
     fn build_hec_config(&self) -> HecSinkConfig {
-        let host = self.host.clone().unwrap_or_else(|| HOST.to_string());
+        let mut host = self.host.clone().unwrap_or_else(|| HOST.to_string());
+
+        if host.ends_with('/') {
+            host.pop();
+        }
+
```
bruceg, line 73: Could this be done just by adding a `.trim_end_matches('/').to_string()` to the end of the first line?

fanatid:
@bruceg do we still need this PR? #2711 will fix host passed from humio

bruceg:
I think this is a separate issue as you noted earlier. At least, I don't see how #2711 resolves this issue.

fanatid:
`host` from `HumioLogsConfig` passed to `HecSinkConfig`, `HecSinkConfig` trim `/` at the end of the `host` (#2711), so we can not care will `host` in `HumioLogsConfig` have `/` or not

bruceg:
Ah, I missed that. Sounds good.

PR #2711, opened 2020-05-29
fix(splunk_hec sink): Remove extra slash on Uri join

### LINKED ISSUE #2705: Humio host with trailing slash fails to build usable HEC URL
When configuring Humio sink, if the `host` value has a training slash it is passed to HecSinkConfig and that creates an invalid URL:

`host = "https://cloud-es.us.humio.com/"`

Creates the URL:

`http://cloud-es.us.humio.com//services/collector/event`

Ideally Vector will remove any trailing forward slash as this will be common with users copying URLs from browsers etc.

fanatid:
Closes #2705 (improved version of #2706)

Fixing problem in `splunk_hec` immediately fix problem in `humio_logs`.

### DIFF
--- a/src/sinks/splunk_hec.rs
+++ b/src/sinks/splunk_hec.rs
@@ -230,12 +230,8 @@ pub fn validate_host(host: &str) -> crate::Result<()> {
     }
 }
 
-fn build_uri(mut host: &str, path: &str) -> Result<Uri, http02::uri::InvalidUri> {
-    if host.ends_with('/') {
-        host = &host[0..host.len() - 1];
-    }
-
-    format!("{}{}", host, path).parse::<Uri>()
+fn build_uri(host: &str, path: &str) -> Result<Uri, http02::uri::InvalidUri> {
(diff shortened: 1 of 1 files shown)


src/sinks/splunk_hec.rs
```diff
@@ -235,6 +230,14 @@ pub fn validate_host(host: &str) -> crate::Result<()> {
     }
 }
 
+fn build_uri(mut host: &str, path: &str) -> Result<Uri, http02::uri::InvalidUri> {
+    if host.ends_with('/') {
+        host = &host[0..host.len() - 1];
+    }
```
bruceg, line 236: This is simply `let host = host.trim_end('/');` (which avoids the `mut`) and could possibly be merged into the format line below.

bruceg (changes requested):

bruceg (approved):
Looks good to me.

lukesteensen (approved):

PR #2715, opened 2020-05-31
chore(tests): Use `test_util::runtime()` in tests

### LINKED ISSUE #716: Replace usage of `tokio::runtime::Runtime` with `test_util::runtime()` in tests
We want to avoid setting up a complete runtime as it will spawn many threads that are not needed and will exhaust resources.

ktff:
Closes #716.

Only replaces uses of `Runtime::new` in tests that can run with one thread.

### DIFF
--- a/src/shutdown.rs
+++ b/src/shutdown.rs
@@ -286,14 +286,14 @@ impl SourceShutdownCoordinator {
 
 #[cfg(test)]
 mod test {
-    use crate::runtime;
     use crate::shutdown::SourceShutdownCoordinator;
+    use crate::test_util::runtime;
     use futures01::future::Future;
     use std::time::{Duration, Instant};
 
     #[test]
     fn shutdown_coordinator_shutdown_source_clean() {
-        let mut rt = runtime::Runtime::new().unwrap();
+        let mut rt = runtime();
         let mut shutdown = SourceShutdownCoordinator::new();
         let name = "test";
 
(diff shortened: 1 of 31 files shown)


fanatid:
There are two more places in code with `Runtime::single_threaded()`:
```
$ grep 'Runtime::single_threaded' -rn src/
src/topology/unit_test.rs:318:    let rt = Runtime::single_threaded().unwrap();
src/test_util.rs:237:    Runtime::single_threaded().unwrap()
```

Out of interest, why `runtime` function exists in `test_util.rs`? `Runtime::single_threaded().unwrap()` / `Runtime::with_thread_count(1).unwrap()` in tests looks more clear, it's just directly says that runtime will have 1 thread without any thoughts in mind compare to `runtime` function where dev need check it or remember.

ktff:
> src/test_util.rs:237:    Runtime::single_threaded().unwrap()

That one is the `runtime` function.

> src/topology/unit_test.rs:318:    let rt = Runtime::single_threaded().unwrap();

That one isn't used for `vector` tests but for users `unit tests`.

> Out of interest, why runtime function exists in test_util.rs? 

Good question, from what I see it's main purpose is to abstract what runtime we are using. It was added some 11 months ago, and it proved it's usefulness 4 months later when we changed out Runtime interface.  

> runtime function where dev need check it or remember.

In line with the above, it's fine if it's sometimes forgotten, it just means that any future `Runtime` change will be just a bit larger. 

> Runtime::single_threaded().unwrap() / Runtime::with_thread_count(1).unwrap() in tests looks more clear

It is clearer, but I would argue, it's just noise in the tests.

PR #2723, opened 2020-06-02
feat(operations): Introduce Dev Workflow improvements

Hoverbear:
Conforms to and satisfies much of #2685 .

Notes:

* We don't yet remove `docker-compose`, it's still used in integration testing.
* We will remove `docker-compose` in a future PR.
* Some `docker-compose` jobs remain for backwards compat with releases/etc that would otherwise be deleted.

TODOS:

* [x] Write the docs in from the RFC
* [x] Final integration kinks


Hoverbear:
Impacted by https://github.com/timberio/vector/issues/2749

Hoverbear:
Tentatively marking this ready for review. I'd appreciate if it @Jeffail could take a look at try doing `nix-shell --pure --run "make test"` on his Mac?

Hoverbear:
@bruceg I've noticed that starting specific services with `podman-compose` isn't supported yet. :( I think I might have to fall back to starting them all if you use `podman` until we fully move off the `docker-compose` file :(

binarylogic:
I think it makes sense for everyone on the team to review this before merging. I want to make sure everyone is comfortable with these changes.

bruceg:
> @bruceg I've noticed that starting specific services with `podman-compose` isn't supported yet.

Well, that's disappointing. The use of the docker-compose setup had effectively blocked me from using podman previously. I had hoped this project would at least reduce those issues.

Ref: [podman-compose issue](https://github.com/containers/podman-compose/issues/125)

.github/workflows/environment.yml
```diff
@@ -0,0 +1,21 @@
...
+  VERBOSE: true
+  ENVIRONMENT_UPSTREAM: docker.pkg.github.com/timberio/vector/environment:${{ github.sha }}
+
+jobs:
+  publish-new-environment:
+    runs-on: ubuntu-latest
+    env:
+      ENVIRONMENT_AUTOBUILD: true
+    steps:
+      - run: docker login https://docker.pkg.github.com -u hoverbear -p ${{ secrets.GITHUB_TOKEN }}
```
MOZGIII, line 18: How about something like this?

```suggestion
      - run: docker login https://docker.pkg.github.com -u "$GITHUB_ACTOR" -p "$GITHUB_TOKEN"
```

Plus `docker.pkg.github.com/$GITHUB_REPOSITORY/environment` for image name.

.github/workflows/benches.yml
```diff
@@ -0,0 +1,35 @@
...
+
+
+  bench:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: cachix/install-nix-action@v10
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+      - run: nix-shell --pure --run "echo 'Env installing...'"
```
MOZGIII, line 33: What if we set [`shell`](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_iddefaultsrun) to `nix-shell --pure --run` for the whole job? If it works, it'll save us a few characters on every command. :smile: Maybe we can even set is as [default](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun).

scripts/check-scripts.sh
```diff
@@ -10,4 +10,4 @@ set -euo pipefail
 FILES=()
 while IFS='' read -r LINE; do FILES+=("$LINE"); done < <(git ls-files | grep '\.sh')
 
-shellcheck "${FILES[@]}"
+shellcheck --shell bash "${FILES[@]}"
```
MOZGIII, line 13: :+1:

scripts/environment/entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#! /usr/bin/env nix-shell
+#! nix-shell --pure /vector/shell.nix --keep INSIDE_ENVIRONMENT -i bash
```
MOZGIII, line 2: Magic

website/package.json
```diff
@@ -36,7 +36,8 @@
     "strftime": "^0.10.0",
     "svg.js": "^2.6.2",
     "timeago-react": "^3.0.0",
-    "titleize": "^2.1.0"
+    "titleize": "^2.1.0",
+    "markdownlint-cli": "0.22"
```
MOZGIII, line 40: :+1:

scripts/environment/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs
+
+# Setup the env
+RUN mkdir -p vector/{scripts,website,scripts/environment}
+ADD default.nix shell.nix rust-toolchain .envrc /vector/
+ADD scripts/environment/definition.nix /vector/scripts/environment/
+ADD scripts/Gemfile scripts/Gemfile.lock /vector/scripts/
+ADD website/package.json website/yarn.lock /vector/website/
+WORKDIR /vector
+SHELL [ "/usr/bin/env", "nix-shell", "/vector/shell.nix", "--run" ]
```
MOZGIII, line 12: :+1:

scripts/check-component-features.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
MOZGIII, line 1: I remember changing all of those to `#!/usr/bin/env bash` at #2458, seems like I missed some. Maybe we need to add a linter for this...

Makefile
```diff
@@ -22,17 +37,98 @@ help:
...
 	@echo "---------------------------------------------------------------------------------------"
+	@echo -e "Nix user? You can use \033[0;33m\`direnv allow .\`\033[0m or \033[0;33m\`nix-shell --pure\`\033[0m"
+	@echo -e "Want to use \033[0;33m\`docker\`\033[0m or \033[0;33m\`podman\`\033[0m? See \033[0;33m\`ENVIRONMENT=true\`\033[0m commands. (Default \033[0;33m\`CONTAINER_TOOL=docker\`\033[0m)"
 	@echo ""
 	@awk 'BEGIN {FS = ":.*##"; printf "Usage: make \033[36m<target>\033[0m\n"} /^[a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36m%-46s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-##@ Default
+##@ Environment
+# We use a volume here as non-Linux hosts are extremely slow to share disks, and Linux hosts tend to get permissions clobbered.
+define ENVIRONMENT_EXEC
```
MOZGIII, line 47: I feel like these (`ENVIRONMENT_EXEC`, `ENVIRONMENT_COPY_ARTIFACTS` and `ENVIRONMENT_PREPARE`) are good candidates to be manually invokable commands at `scripts/`. But then, it's very like I personally wouldn't be invoking them. They'll be useful for automation that you can't drive from the environment shell but that you want to run in the environment - for instance, editor integration and etc...

MOZGIII:
I need some more time to test-run everything, but I can already say that I like the approach and think this is going to be a big improvement.

Makefile
```diff
@@ -22,17 +37,98 @@ help:
 	@echo "                                      V E C T O R"
 	@echo ""
 	@echo "---------------------------------------------------------------------------------------"
+	@echo -e "Nix user? You can use \033[0;33m\`direnv allow .\`\033[0m or \033[0;33m\`nix-shell --pure\`\033[0m"
+	@echo -e "Want to use \033[0;33m\`docker\`\033[0m or \033[0;33m\`podman\`\033[0m? See \033[0;33m\`ENVIRONMENT=true\`\033[0m commands. (Default \033[0;33m\`CONTAINER_TOOL=docker\`\033[0m)"
```
MOZGIII, line 41: Looks odd on my end:
image

src/sinks/clickhouse.rs
```diff
@@ -244,7 +244,6 @@ mod integration_tests {
     use serde_json::Value;
     use std::time::Duration;
     use tokio01::util::FutureExt;
-
```
MOZGIII, line 247: This might not be intended

Makefile
```diff
@@ -22,17 +37,98 @@ help:
...
+	@$(CONTAINER_TOOL) cp vector-build-outputs:/target/criterion ./target/criterion || true
+	@$(CONTAINER_TOOL) rm -f vector-build-outputs
+endef
+
+
+ifeq ($(ENVIRONMENT_AUTOBUILD), true)
+define ENVIRONMENT_PREPARE
+	@echo "Building the environment. (ENVIRONMENT_AUTOBUILD=true) This may take a few minutes..."
+	$(CONTAINER_TOOL) build \
+		$(if $(findstring true,$(VERBOSE)),,--quiet) \
```
MOZGIII, line 87: :+1: for `VERBOSE`, really useful

Makefile
```diff
@@ -22,17 +37,98 @@ help:
...
+		--name vector-build-outputs \
+		busybox true
+	@$(CONTAINER_TOOL) cp vector-build-outputs:/target/release/vector ./target/release/ || true
+	@$(CONTAINER_TOOL) cp vector-build-outputs:/target/debug/vector ./target/debug/ || true
+	@$(CONTAINER_TOOL) cp vector-build-outputs:/target/criterion ./target/criterion || true
+	@$(CONTAINER_TOOL) rm -f vector-build-outputs
+endef
+
+
+ifeq ($(ENVIRONMENT_AUTOBUILD), true)
```
MOZGIII, line 83: Can we add an option to skip `ENVIRONMENT_PREPARE` altogether? It adds a tiny bit of time to each invocation, and it would be great to have an option to skip it sometimes.

.github/workflows/environment.yml
Hoverbear: Oh nice thank you!

Hoverbear:
@bruceg I'd be very happy to move all the services to raw docker calls. :) I was going to leave that for a follow up but if it's blocking you we can do it now.

Makefile
Hoverbear: I suggest you use Nix in this case. :)

src/sinks/clickhouse.rs
Hoverbear: Whoops!

Makefile
Hoverbear: Oh interesting! Is this a mac?

Makefile
Hoverbear: I generally suggest using Nix, all `ENVIRONMENT_*` things are only really existing to support the choice to using Docker for a dev env, which is not a good choice.

scripts/check-component-features.sh
Hoverbear: Good news, Nix won't run them at all without using this, so we have a built in lint since the CI literally won't run those scripts. :)

scripts/environment/entrypoint.sh
Hoverbear: It really, really is. :)

.github/workflows/benches.yml
Hoverbear: I was looking for that option! Thanks!

Hoverbear: So one note, we can't use `nix-shell` until we've gotten cachix. :(

Hoverbear: image

Makefile
Hoverbear: Let's make `environment_exec` a script!

.github/workflows/benches.yml
MOZGIII: Oh, unexpected. Here's the relevant code: https://github.com/actions/runner/blob/2bd0b1af0e4daa0ed3d8115c79ba6aa6633dd732/src/Runner.Worker/Handlers/ScriptHandler.cs#L220
It might be helpful if you want to dig into that - I've spent enough time with Github Actions to know checking the code directly is quicker than trying to find something like this in the docs.

MOZGIII: > So one note, we can't use `nix-shell` until we've gotten cachix. :(

Don't we use `cachix` by default?

Makefile
MOZGIII: I agree :smiley: Yeah, we probably don't want to add more flexibility into docker since we have nix. :+1:

.github/workflows/benches.yml
Hoverbear: The first step of most CI jobs now is installing nix, which can't be done in nix-shell since we don't have one. :)

Hoverbear: Ahhh ok I have an idea then

Makefile
MOZGIII: Nope, Ubuntu 18.04. Quite odd.
```
$ make --version
GNU Make 4.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

$ bash --version
GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

.github/workflows/benches.yml
MOZGIII: We can just set `shell: bash` specifically for those first steps

MOZGIII: I.e., `shell: nix` at the `job` level (on the contrary to `default` - to not spill the default to newly added jobs that don't expect shell to be nix), and `shell: bash` for the setup step.

MOZGIII: Also, I think quotes around `{0}` aren't needed

bruceg:
> @bruceg I'd be very happy to move all the services to raw docker calls. :) I was going to leave that for a follow up but if it's blocking you we can do it now.

No, I think that's relatively unrelated, so should wait for a follow up.

Hoverbear:
@bruceg would something like https://github.com/lilyball/nix-env.fish work? Found it in https://github.com/NixOS/nix/issues/440

Makefile
Hoverbear: What the heck!

MOZGIII: Idk what's the issue is there. Must be `make` specifics.

Also:

```
$ /bin/echo --version
echo (GNU coreutils) 8.28
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Brian Fox and Chet Ramey.
```

bruceg:
> @bruceg would something like https://github.com/lilyball/nix-env.fish work? Found it in [NixOS/nix#440](https://github.com/NixOS/nix/issues/440)

Looks like it should, yes.

Hoverbear:
I'm trying to fix these two failing tests `tcp_stream_detects_disconnect` and `tcp_stream` as they are failing fairly reliably. I suspect getting 'real' network sockets instead of the 'fake' docker ones might have exposed this. Can't get it to happen locally though. :(

MOZGIII:
Judging from the name (disconnect), docker is highly likely the issue here. I've had issues with disconnection management at docker proxies before at another project, must be a similar problem.

Hoverbear:
In this case, the CI is using Nix so docker is no longer involved

MOZGIII:
Oh, it's `sinks::socket::test::tcp_stream_detects_disconnect`, I though it was at a source. Interesting. It might just be a race: https://github.com/timberio/vector/blob/2937c46e0f8f061c120a8c21840cc3101434266c/src/sinks/socket.rs#L194
The test coroutines are not synchronized, and thus the code has full right to crash as we observe at CI.

Hoverbear:
@MOZGIII Oh! That might be it too! I was looking at https://github.com/timberio/vector/pull/2106/files and saw we still use the tokio01 shutdown in some test utils, so I've been poking with that.

Makefile
```diff
@@ -2,37 +2,148 @@
...
 
 ##@ Building
+build: ## Build the project in release mode (Supports `ENVIRONMENT=true`)
+ifeq ($(ENVIRONMENT), true)
+	${ENVIRONMENT_PREPARE}
+	${ENVIRONMENT_EXEC} make build
+	${ENVIRONMENT_COPY_ARTIFACTS}
+else
+	cargo build --release --no-default-features --features ${DEFAULT_FEATURES}
+endif
```
bruceg, line 137: You can probably use a canned recipe to DRY these:
```make
ifeq ($(ENVIRONMENT), true)
define magic =
${ENVIRONMENT_PREPARE}
${ENVIRONMENT_EXEC} make $@
endef
else
define magic =
$1
endef
endif

build:
        $(call magic,"cargo build --release --no-default-features --features ${DEFAULT_FEATURES}")
```

Of course, you would need separate macros where one did the `${ENVIRONMENT_COPY_ARTIFACTS}`, and it would probably get ugly for multi-line recipes. Another possibility would be defining a sub-rule, which would allow for multi-line recipes:
```make
ifeq ($(ENVIRONMENT), true)
define magic = 
${ENVIRONMENT_PREPARE}
${ENVIRONMENT_EXEC} make REAL_$@
endef
else
define magic =
@$(MAKE) REAL_$@
endef
endif

build:
        $(magic)
REAL_build:
        cargo build --release --no-default-features --features ${DEFAULT_FEATURES}
```

Makefile
```diff
@@ -2,37 +2,148 @@
...
+define ENVIRONMENT_PREPARE
+	$(CONTAINER_TOOL) pull $(ENVIRONMENT_UPSTREAM)
+endef
+endif
+
 
-##@ Default
+environment: ## Enter a full Vector dev shell in $CONTAINER_TOOL, binding this folder to the container.
+	${ENVIRONMENT_PREPARE}
+	@export ENVIRONMENT_TTY=true
```
bruceg, line 116: Note that this may cause surprises in a parallel make environment. Does it need to be set for all recipe lines that run after it?

Makefile
```diff
@@ -2,37 +2,148 @@
...
+	@printf -- "\n"
+	@printf -- "---------------------------------------------------------------------------------------\n"
+	@printf -- "Nix user? You can use ${FORMATTING_BEGIN_YELLOW}\`direnv allow .\`${FORMATTING_END} or ${FORMATTING_BEGIN_YELLOW}\`nix-shell --pure\`${FORMATTING_END}\n"
+	@printf -- "Want to use ${FORMATTING_BEGIN_YELLOW}\`docker\`${FORMATTING_END} or ${FORMATTING_BEGIN_YELLOW}\`podman\`${FORMATTING_END}? See ${FORMATTING_BEGIN_YELLOW}\`ENVIRONMENT=true\`${FORMATTING_END} commands. (Default ${FORMATTING_BEGIN_YELLOW}\`CONTAINER_TOOL=docker\`${FORMATTING_END})\n"
+	@printf -- "\n"
+	@awk 'BEGIN {FS = ":.*##"; printf "Usage: make ${FORMATTING_BEGIN_BLUE}<target>${FORMATTING_END}\n"} /^[a-zA-Z0-9_-]+:.*?##/ { printf "  ${FORMATTING_BEGIN_BLUE}%-46s${FORMATTING_END} %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+##@ Environment
+# We use a volume here as non-Linux hosts are extremely slow to share disks, and Linux hosts tend to get permissions clobbered.
+define ENVIRONMENT_EXEC
```
bruceg, line 62: If I'm seeing right, every use of `ENVIRONMENT_EXEC` runs `ENVIRONMENT_PREPARE` first. Would it make sense to roll that right in here?

src/sinks/socket.rs
```diff
@@ -152,6 +154,9 @@ mod test {
         let pump = sink.send_all(events);
         let _ = rt.block_on(pump).unwrap();
 
+        // Some CI machines are very slow, be generous.
+        std::thread::sleep(std::time::Duration::from_secs(5));
+
         let output = receiver.wait();
```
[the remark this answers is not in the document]
Hoverbear: I think so, let me try yanking this. I took a very liberal approach.

tests/buffering.rs
```diff
@@ -62,6 +62,7 @@ fn test_buffering() {
     let mut rt = test_util::runtime();
 
     let (topology, _crash) = topology::start(config, &mut rt, false).unwrap();
+    std::thread::sleep(std::time::Duration::from_secs(1)); // Give topo a moment to start up.
```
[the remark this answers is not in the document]
Hoverbear: Yeah -- I really hated writing this, but it seemed to ease things. I don't really know a better short-term solution though

Makefile
Hoverbear: You're totally right, and I considered making it work like our `RUN` val right now... However I also noted as you did that some jobs would require special concerns like moving files etc, so I opted not to. I tried to balance reusability and understanding, maybe I lost. :)

Let me try to incorporate this feedback. :) Will report back.

Makefile
Hoverbear: You're definitely right, if someone depended on this target as a dependency it would be an issue. I'll localize it better!

Makefile
Hoverbear: Agree. :)

Makefile
Hoverbear: image
image


Much better

src/sinks/socket.rs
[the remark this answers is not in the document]
Hoverbear: Trying a smaller timeout since that seemed to not work...

PR #2735, opened 2020-06-04
chore(aws_kinesis_firehose sink): Upgrade rusoto

fanatid:
[rusoto](https://github.com/rusoto/rusoto) use tokio 0.2 from 0.43 and 0.44 include compression support (required for #1411).

This PR only update `aws_kinesis_firehose`.


src/sinks/util/rusoto2.rs
```diff
@@ -261,12 +246,16 @@ impl Body for RusotoBody {
                 }
             }
             Some(SignedRequestPayload::Stream(stream)) => {
-                match with_notify(cx, || stream.poll())? {
-                    Async::Ready(Some(buffer)) => {
-                        Poll::Ready(Some(Ok(io::Cursor::new(buffer.into_iter().collect()))))
-                    }
-                    Async::Ready(None) => Poll::Ready(None),
-                    Async::NotReady => Poll::Pending,
+                let stream = unsafe { Pin::new_unchecked(stream) };
```
fanatid, line 249: Not sure is it possible create pinned stream without `unsafe` here or not.... :thinking:

Hoverbear: Please include some reasoning why `Pin::new` is not appropriate and why this is safe. :)

fanatid: haha, I thought about this so much, though about `pin_project` or something, but `Pin::new` just works.
I hope that I'll get clear understanding of pin things at one day. Thanks Ana!

src/sinks/aws_kinesis_firehose.rs
```diff
@@ -170,36 +178,33 @@ impl RetryLogic for KinesisFirehoseRetryLogic {
...
     #[snafu(display("Stream name does not match, got {}, expected {}", name, stream_name))]
     StreamNamesMismatch { name: String, stream_name: String },
 }
 
-fn healthcheck(
-    config: KinesisFirehoseSinkConfig,
-    resolver: Resolver,
-) -> crate::Result<super::Healthcheck> {
+async fn healthcheck(config: KinesisFirehoseSinkConfig, resolver: Resolver) -> crate::Result<()> {
     let client = create_client(config.region.try_into()?, config.assume_role, resolver)?;
```
bruceg, line 188: I think we want the `region.try_into()?` to happen outside of the `async fn` so as to catch config errors up front rather than having them happen after the system starts running.

src/sinks/util/rusoto2.rs
```diff
@@ -91,34 +89,15 @@ impl AwsCredentialsProvider {
...
-        match self {
-            Self::Default(f) => f.poll(),
-            Self::Role(f) => f.poll(),
-            Self::Static(f) => f.poll(),
-        }
+    async fn credentials(&self) -> Result<AwsCredentials, CredentialsError> {
+        let fut = match self {
+            AwsCredentialsProvider::Default(p) => p.credentials(),
+            AwsCredentialsProvider::Role(p) => p.credentials(),
+            AwsCredentialsProvider::Static(p) => p.credentials(),
```
bruceg, line 98: Can these use `Self` instead of spelling out `AwsCredentialsProvider` each time or does that cause problems with `async_trait`?

fanatid: It was problem in `async_trait`, I think that it was fixed in 1.29 so I update to latest version.
https://github.com/dtolnay/async-trait/search?q=e0401&type=Issues

src/sinks/aws_kinesis_firehose.rs
fanatid: `region.try_into()?` also happen on creating `KinesisFirehoseService` which is not async, so config error should be catched up front.
https://github.com/timberio/vector/pull/2735/files#diff-6b010670b04b03f3921950a0b304868dR76

PR #2737, opened 2020-06-04
fix: Handle templates with invalid strftime elements

### LINKED ISSUE #2736: Template from invalid strftime
Currently if string for `Template` have [Item::Error](https://docs.rs/chrono/0.4.6/chrono/format/enum.Item.html#variant.Error) (added in https://github.com/timberio/vector/pull/1003) nothing will happened. As result issues like https://github.com/timberio/vector/issues/2729 arise.
Proper way will be return `crate::Result<Template>` from `Template::from`, but for start maybe add `panic!` at least?

bruceg:
Please make sure I did all the error handling correctly. There were a couple of cases where I guessed at the right error handling result.

Closes #2736 

### DIFF
--- a/src/sinks/gcp/cloud_storage.rs
+++ b/src/sinks/gcp/cloud_storage.rs
@@ -12,7 +12,7 @@ use crate::{
         },
         Healthcheck, RouterSink,
     },
-    template::Template,
+    template::{Template, TemplateError},
     tls::{TlsOptions, TlsSettings},
     topology::config::{DataType, SinkConfig, SinkContext, SinkDescription},
 };
@@ -26,8 +26,9 @@ use hyper::{
 };
 use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
-use snafu::Snafu;
+use snafu::{ResultExt, Snafu};
 use std::collections::HashMap;
+use std::convert::TryFrom;
 use tower::{Service, ServiceBuilder};
 use tracing::field;
 use uuid::Uuid;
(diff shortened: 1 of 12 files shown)


src/template.rs
```diff
@@ -29,20 +31,53 @@ pub struct Template {
...
+            .filter(|item| matches!(item, Item::Error))
+            .count()
+        {
+            0 => Ok(Template {
+                src: src.into(),
+                src_bytes: src.into(),
+                has_ts: StrftimeItems::new(src).filter(is_dynamic).count() > 0,
+                has_fields: RE.is_match(src),
+            }),
+            _ => Err(TemplateError::StrftimeError),
```
fanatid, line 63: `StrftimeItems::new(src)` called twice, i.e. `src` also will be parsed twice.
Instead `.filter(|item| matches!(item, Item::Error)).count()` [.any](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) can be used.

fanatid:
Thank you Bruce, rest things looks good!

src/template.rs
bruceg: Yeah, that was not my finest piece of code, just wanted to get it written. I'll see how I can rework it.

src/sinks/gcp/cloud_storage.rs
```diff
@@ -197,11 +200,12 @@ impl GcsSink {
 
         let batch = config.batch.unwrap_or(bytesize::mib(10u64), 300);
 
-        let key_prefix = if let Some(kp) = &config.key_prefix {
-            Template::from(kp.as_str())
-        } else {
-            Template::from("date=%F/")
-        };
+        let key_prefix = config
```
Hoverbear, line 203: It seems like lines 200-204 already do similar? Is one redundant?

fanatid (approved):

Hoverbear (approved):

lukesteensen (approved):

PR #2763, opened 2020-06-07
chore(aws_ec2_metadata transform): Upgrade `hyper`

ktff:
Ref. #2117

### DIFF
--- a/src/transforms/aws_ec2_metadata.rs
+++ b/src/transforms/aws_ec2_metadata.rs
@@ -1,14 +1,13 @@
 use super::Transform;
 use crate::{
     event::Event,
-    sinks::util::http::HttpClient,
+    sinks::util::http2::HttpClient,
     topology::config::{DataType, TransformConfig, TransformContext, TransformDescription},
 };
-use bytes::Bytes;
-use futures::compat::Future01CompatExt;
-use futures01::Stream;
-use http::{uri::PathAndQuery, Request, StatusCode, Uri};
-use hyper::Body;
+use bytes::{Bytes, BytesMut};
+use futures::{compat::Future01CompatExt, TryStreamExt};
+use http02::{uri::PathAndQuery, Request, StatusCode, Uri};
+use hyper13::Body;
 use serde::{Deserialize, Serialize};
 use std::collections::{hash_map::RandomState, HashSet};
 use std::time::{Duration, Instant};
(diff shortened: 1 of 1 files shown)


src/transforms/aws_ec2_metadata.rs
```diff
@@ -132,7 +131,7 @@ impl TransformConfig for Ec2Metadata {
         let host = self
             .host
             .clone()
-            .map(|s| Uri::from_shared(s.into()).unwrap())
+            .map(|s| Uri::from_maybe_shared(s).unwrap())
```
fanatid, line 134: Can be done without clone (but `HOST.clone()` better move to closure):
```rust
        let host = self
            .host
            .as_ref()
            .map(|s| s.parse::<Uri>().unwrap())
            .unwrap_or_else(|| HOST.clone());
```
Another thing, that it would be better handle error instead unwrap (not sure that creating snafu error from sinks mod is correct):
```rust
            .map(|s| {
                s.parse::<Uri>()
                    .map_err(|source| crate::sinks::BuildError::UriParseError2 { source })
            })
            .unwrap_or_else(|| Ok(HOST.clone()))?;
```

src/transforms/aws_ec2_metadata.rs
```diff
@@ -274,8 +273,14 @@ impl MetadataClient {
...
-        let body = res.into_body().concat2().compat().await?;
-        let token = body.into_bytes();
+        let body = res
+            .into_body()
+            .try_fold(BytesMut::new(), |mut store, bytes| async move {
+                store.extend_from_slice(&bytes);
+                Ok(store)
+            })
+            .await?;
+        let token: Bytes = body.into();
```
fanatid, line 283: We can not use `hyper13::body::to_body` because another parties use bytes@0.4?
Maybe use with extra copy?
```
        let body = hyper13::body::to_bytes(res.into_body()).await?;
        let token = Bytes::from(&body[..]);
```

Or add function `to_body` with `try_fold` inside at top-level which later can be easily changed to `hyper13::body::to_body`?

src/transforms/aws_ec2_metadata.rs
bruceg: I think using a sink error in a transform is suspect at best, although this is exactly the kind of error we want to use. It might make sense to have a crate-level set of errors for some of these general errors, but that's outside the scope of this PR. Please open an issue to discuss it.

src/transforms/aws_ec2_metadata.rs
ktff: Yes, there is a difference in version for `bytes` crate. 

Better some extra code than an extra copy of every incoming byte. 

`to_body` function sounds good. I'll add it to this file, and when we get a need for it in some other location, then we should move it, probably to a new file, or as an extension in `src/stream.rs`, will see.

PR #2773, opened 2020-06-10
enhancement(aws_s3 sink): Correct AWS Console behaviour via S3 object Metadata

tyrken:
Following on from #2769 I played around with setting content-type and content-encoding options of the `aws s3` CLI command, and found `text/plain` has an unfortunate side-effect of forcing the file extension from the uploaded `.log` to `.txt` when Downloading in Chrome & Firefox, so is a no-go.

Also & more seriously, content-encoding=gzip, i.e. compression=gzip in the sink, leads to corrupt files when downloading via the browser (CLI/API access is fine).  I recommend dropping that completely, instead using just content-type=application/x-gzip.

Playing around with some other content-types, these all avoid the renaming problems but have their own issues:

* application/x-ndjson, application/octet-stream: OK, but "Open" merely acts the same as "Download".
* application/ld+json: Open works in Chrome & Mozilla, but Mozilla thinks it's simple JSON and so shows a parsing error (but you can switch to view the raw text on a tab so it's not that bad).
* text/event-stream: Open works in Chrome, Mozilla just acts like Download.

As a mostly Chrome user I might prefer either of the latter two options for non-compressed files, but really they are being mis-understood by the browsers & only working by accident.  Hence this PR as I'd expect you'd prefer it simple & least-bad instead.

PR against your v0.9 branch as that's what I'm using currently, and master looks a little different.

### DIFF
--- a/src/sinks/aws_s3.rs
+++ b/src/sinks/aws_s3.rs
@@ -287,7 +287,7 @@ impl Service<Request> for S3Sink {
             body: Some(request.body.into()),
             bucket: request.bucket,
             key: request.key,
-            content_encoding: request.content_encoding,
+            content_type: request.content_type,
             acl: options.acl.map(to_string),
             grant_full_control: options.grant_full_control,
             grant_read: options.grant_read,
@@ -342,7 +342,11 @@ fn build_request(
         body: inner,
         bucket,
         key,
-        content_encoding: compression.content_encoding().map(|ce| ce.to_string()),
+        content_type: if compression == Compression::Gzip {
+            Some("application/x-gzip".to_string())
+        } else {
+            Some("application/octet-stream".to_string())
+        },
         options,
     }
 }
@@ -352,7 +356,7 @@ struct Request {
     body: Vec<u8>,
     bucket: String,
     key: String,
-    content_encoding: Option<String>,
+    content_type: Option<String>,
     options: S3Options,
 }
 
@@ -572,6 +576,10 @@ mod integration_tests {
 
             let obj = get_object(key).await;
             assert_eq!(obj.content_encoding, None);
+            assert_eq!(
+                obj.content_type,
+                Some("application/octet-stream".to_string())
+            );
 
             let response_lines = get_lines(obj).await;
             assert_eq!(lines, response_lines);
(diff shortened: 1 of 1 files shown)


tyrken:
It looks like the two aws-integration-tests that fail are flakes or out-of-step with the current integration test hardware spec, and the cargo-deny fails look like they are out-of-date CI instructions compare to using cargo-deny-action@v1 in master.

Hence ready for review please!  You may want to make use of the below commands to check my discussion above:

```bash
echo '{"message":"something1"}' >dummy.log
echo '{"message":"something2"}' >>dummy.log
gzip -k dummy.log
aws s3 cp dummy.log s3://$MYBUCKET/$MYPATH/dummy.log --content-type application/octet-stream
aws s3 cp dummy.log.gz s3://$MYBUCKET/$MYPATH/dummy.log.gz --content-type application/x-gzip
aws s3 cp dummy.log.gz s3://$MYBUCKET/$MYPATH/bad-encode.log.gz --content-type application/octet-stream --content-encoding gzip
```

src/sinks/splunk_hec.rs
```diff
@@ -125,7 +125,8 @@ impl HttpSink for HecSinkConfig {
             chrono::Utc::now().timestamp_nanos()
         };
 
-        let sourcetype = event.get(&event::log_schema().source_type_key()).cloned();
+        // Do not merge this change into master
+        // let sourcetype = event.get(&event::log_schema().source_type_key()).cloned();
```
lukesteensen, line 129: Could we get rid of this diff?

tyrken:
Updated as you requested, but...

* If you check the comments on the related issue #2769 then @fanatid is also thinking about this but not responded to my last post yet.
* Looking at the nice global sink::utils::Compression this PR sort of breaks the design idea, sorry.
* Given the confusion, I've bumped this up to AWS for a suggestion from the source.  They are redesigning the S3 GUI at the moment, we might get lucky & they fix the "Download" button to not decompress according to Content-Encoding...

fanatid:
@tyrken I had seen that comment. Wanted tests it again (today I hope). I think there is no clear winner between `text/plain` / `text/x-log` / etc, what can be good for one user can be not good for another. But currently I tend to left `content-encoding` as gzip if compression applied and add `content-type` = `text/x-log` because this set by AWS when `.log` file uploaded through console.

Sorry for delay.

fanatid:
Resolved with #2848

PR #2777, opened 2020-06-10
chore: update rusoto to 0.44.0

fanatid:
Part of #2118 

With `async move` there are many places where we need clone some struct (for example rusoto Client). Does it make sense use `Arc` for this structs for cheaper clone? (from other side this will require locks for counting refs)

I found that aws integration tests was not compiled, so changed dependencies in futures.


fanatid:
I thought that rusoto update PR for all aws sinks will be simple and easy for review, I was _wrong_. There a lot of changes which can be not obvious and I'm sure hard to review. I also have few questions, what makes PR hard to merge or will take some extra time.
Should we consider split this to few PR (one per each sink) or just force this big update? I'm ok with extra efforts for small PR's if this will be helpful.

Cargo.toml
```diff
@@ -399,7 +397,15 @@ docker = [
...
-aws-integration-tests = ["sinks-aws_cloudwatch_logs", "sinks-aws_cloudwatch_metrics", "sinks-elasticsearch", "sinks-aws_kinesis_firehose", "sinks-aws_kinesis_streams", "sinks-aws_s3", "transforms-aws_ec2_metadata"]
+aws-integration-tests = [
+  "aws-cloudwatch-logs-integration-tests",
+  "aws-cloudwatch-metrics-integration-tests",
+  "aws-ec2-metadata-integration-tests",
+  "aws-kinesis-firehose-integration-tests",
+  "aws-kinesis-streams-integration-tests",
+  "aws-s3-integration-tests",
+  "es-integration-tests",
+]
```
fanatid, line 408: Otherwise integration tests will not be included in build and will not be executed.

src/sinks/elasticsearch.rs
```diff
@@ -657,6 +659,10 @@ mod integration_tests {
     }
 
     fn run_insert_tests(mut config: ElasticSearchConfig, break_events: bool) {
+        // Need async version of ElasticSearchCommon::parse_config otherwise
+        // function called in blocked runtime will panic with error:
+        // default Tokio reactor already set for execution context
+
```
fanatid, line 665: I'm not sure, but probably elasticsearch on aws sink currently broken. If sink build called outside runtime block (`rt.block_on`) `ElasticSearchCommon::parse_config` works, but if we parse config happened in block vector will panic (see message in code).
It's not only about rusoto:0.44, I tried with 0.41 too. Test is ok because parse outside block_on.

src/sinks/aws_cloudwatch_logs/mod.rs
```diff
@@ -1026,41 +1011,40 @@ mod integration_tests {
...
+        // Now and future in +2h limit
         lines.push(add_event(Duration::zero()));
-        // ------------------ future
         lines.push(add_event(Duration::hours(1)));
         lines.push(add_event(Duration::minutes(110)));
-        // ------------------ 2h limit
-        add_event(Duration::minutes(125));
+        // 14 days limit
+        lines.push(add_event(Duration::days(-1)));
+        lines.push(add_event(Duration::days(-13)));
```
fanatid, line 1024: I changed lines order here because:
- In function `process_events` events sorted by timestamp -- `events.sort_by_key(|e| e.timestamp);` to buckets 24h each
- In `CloudwatchFuture` (`request.rs`) sorted events in buckets pushed to aws from end

i.e. recent events pushed first, is this correct? Or oldest should be pushed first?

src/sinks/aws_cloudwatch_logs/mod.rs
```diff
@@ -146,29 +150,31 @@ pub enum CloudwatchError {
...
-        let request = self.request.unwrap_with(&REQUEST_DEFAULTS);
+        // let request = self.request.unwrap_with(&REQUEST_DEFAULTS);
 
         let log_group = self.group_name.clone();
         let log_stream = self.stream_name.clone();
 
         let svc = ServiceBuilder::new()
-            .concurrency_limit(request.in_flight_limit)
+            // .concurrency_limit(request.in_flight_limit)
+            .concurrency_limit(1)
```
fanatid, line 160: This is what I do not understand. If I return commented line back, then test `cloudwatch_insert_log_event_batched` will hang up tests.
I think this related with `CloudwatchLogsSvc::poll_ready` because context in oneshot receiver rewritten by another `poll_ready` call.

scripts/test-integration-aws.sh
```diff
@@ -8,4 +8,5 @@ set -euo pipefail
 #   Run integration tests for AWS components only.
 
 docker-compose up -d dependencies-aws
+docker-compose up -d dependencies-elasticsearch
```
bruceg, line 11: You can join this with the previous line and bring both up at once. Also, why is this needed for the AWS tests? Is it because the elasticsearch sink is dual-mode?

src/sinks/elasticsearch.rs
```diff
@@ -424,7 +426,7 @@ fn finish_signer(
     credentials: &AwsCredentials,
     mut builder: http02::request::Builder,
 ) -> http02::request::Builder {
-    signer.sign_with_plus(&credentials, true);
+    signer.sign(&credentials);
```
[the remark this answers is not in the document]
fanatid: Yes, `sign_with_plus` was removed.
https://github.com/rusoto/rusoto/pull/1661

scripts/test-integration-aws.sh
fanatid: Oh, indeed, I can specify both services.
I added this service because `aws-integration-tests` have dependency `es-integration-tests`, but without service tests will fail. I think right thing will remove service and dependency, which was probably added to test aws credentials.

src/sinks/aws_cloudwatch_logs/request.rs
```diff
@@ -66,121 +71,107 @@ impl CloudwatchFuture {
...
-    type Item = ();
-    type Error = CloudwatchError;
+    type Output = Result<(), CloudwatchError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        macro_rules! ready2 {
+            ($fut:ident) => {
+                ready!(Pin::new($fut).poll(cx))
+            };
+        }
```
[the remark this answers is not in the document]
fanatid: This is just shortcut for 4 places, 5 lines of it not costs it.
Local in method because use local variable.
I'll remove it. Got it, no macroses.

fanatid:
I'll split this pull request.

PR #2793, opened 2020-06-12
chore(topology): Allow disconnected components, and split `start` step

### LINKED ISSUE #2775: Panic on changing topology that has a closed source
In a scenario where a source has closed it self, changing to a different topology that reuses that source, or subsequently closed transforms, will panic on one of our checks.   

One solution is to remove those checks, more precisely instead of the `unwrap`, when we are sending `ControlMessage` in `.unbounded_send(...).unwrap()`, we can use a more fine grained check that doesn't panic when it's disconnected. The reason being that we now have a scenario where it's valid for a components output to be closed.

Theoretically the panic can occur, without a change in topology, during it's construction, so long as the source manages to close itself and the change propagates fast enough while we are building it.

First observed by @Hoverbear in https://timberio.slack.com/archives/CCG0F0NJE/p1591630215447700, and I've reproduced it.

cc. @lukesteensen

ktff:
Closes #2775 

Removes the checks and separates `start` step into `connect` and `spawn`.

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -541,7 +541,7 @@ impl RunningTopology {
                 if let Some(input) = self.inputs.get(sink_name) {
                     output
                         .unbounded_send(fanout::ControlMessage::Add(sink_name.clone(), input.get()))
-                        .expect("Components shuldn't be spawned before connecting them together.");
+                        .expect("Components shouldn't be spawned before connecting them together.");
                 }
             }
         }
@@ -554,7 +554,7 @@ impl RunningTopology {
                             transform_name.clone(),
                             input.get(),
                         ))
-                        .expect("Components shuldn't be spawned before connecting them together.");
+                        .expect("Components shouldn't be spawned before connecting them together.");
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


ktff:
It seams that

```
    topology::transient_state_tests::remove_sink
    topology::transient_state_tests::remove_transform
```

are now passing.

Hoverbear:
Yess!

After `multi_file_update` is fixed we can support the same suite on windows/mac as linux!

src/topology/mod.rs
```diff
@@ -529,7 +541,7 @@ impl RunningTopology {
                 if let Some(input) = self.inputs.get(sink_name) {
                     output
                         .unbounded_send(fanout::ControlMessage::Add(sink_name.clone(), input.get()))
-                        .unwrap();
+                        .expect("Components shuldn't be spawned before connecting them together.");
```
lukesteensen, line 544:
```suggestion
                        .expect("Components shouldn't be spawned before connecting them together.");
```

src/topology/mod.rs
```diff
@@ -542,7 +554,7 @@ impl RunningTopology {
                             transform_name.clone(),
                             input.get(),
                         ))
-                        .unwrap();
+                        .expect("Components shuldn't be spawned before connecting them together.");
```
lukesteensen, line 557:
```suggestion
                        .expect("Components shouldn't be spawned before connecting them together.");
```

PR #2796, opened 2020-06-12
chore(loki sink): Make encoding optional

ktff:
Ref. #2778

Makes `encoding` optional and enhances `Unhandled error` message.

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -702,7 +702,7 @@ fn handle_errors(
         .map_err(|_| ())
         .flatten()
         .or_else(move |()| {
-            error!("Unhandled error");
+            error!("An error occured that vector can't fix");
             let _ = abort_tx.unbounded_send(());
             Err(())
         })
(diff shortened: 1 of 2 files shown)


src/topology/mod.rs
```diff
@@ -687,7 +687,7 @@ fn handle_errors(
         .map_err(|_| ())
         .flatten()
         .or_else(move |()| {
-            error!("Unhandled error");
+            error!("An error occured that vector can't fix");
```
bruceg, line 690: I don't see what this has to do with making encoding optional.

ktff: Directly nothing, they were both a minor issues reported in #2778, this one is a clarification of what `unhandled` means in this context. Do you think this deserves a separate PR?

bruceg: No, I guess not, however this is still vague. I suggest passing the task typetag and name to `handle_errors` and doing something like the following:
```rust
error!(message = "An error occurred that vector couldn't handle", r#type = typetag, %name);
```

ktff: True, that would be better.

ktff: I've increased the scope of instrumentation span instead of passing name and typetag directly.

bruceg: Even better

PR #2812, opened 2020-06-14
chore(networking)!: Remove custom DNS resolution

### LINKED ISSUE #2635: Consider removal of custom DNS resolution
Opening this issue to solicit feedback. **To be clear, no decision has yet been made here.** If you feel strongly one way or the other (or know someone who might) please let us know!

# Background

Custom DNS resolution was suggested in #827 and implemented in #1118. Notably, these were both in anticipation of user demand, not in response to any actual demand (as far as I know). 

Since then there have been numerous issues opened to "finish" implementing this everywhere (e.g. #2359, #1954). Support is still technically incomplete.

# The Problem

The immediate problem that sparked this discussion was #2511, which has led @LucioFranco down a rabbit hole of very strange failures that shouldn't have anything to do with DNS. It has been a huge expenditure of time and there is still no solution in sight.

The larger problem that has been on my mind for a while is how invasive this feature is. As shown by #2359, it adds an extra wrinkle to anything that could make a network request, forcing us to customize HTTP clients that are built into many higher-level libraries. This is a significant barrier to contributors, often for virtually no benefit. And once it's done, all the extra wiring has a meaningful ongoing maintenance cost.

# Proposal

Unless it turns out that a meaningful number of users are relying on this feature, we should rip out the whole thing and add documentation on how to achieve the same result with a [system-level resolver](https://wiki.archlinux.org/index.php/Domain_name_resolution).

As far as I can tell, there is no meaningful use of this feature and it has proven to be a significant burden to maintain. Combined with the fact that there are very reasonable alternatives that do not involve Vector at all, this seems like a clear case for cutting the feature.

**I propose that we circulate this issue to give users a chance to weigh in, and then make a decision based on that feedback in two weeks time.**

ktff:
Closes #2635.

### Open questions

- [x] If we don't have intention to have `dns` configurable and it's fine to not call `spawn_blocking` through a runtime handle, then we can further remove passing `Resolver` through contexts and functions, and instead construct `Resolver` directly at call site. EDIT: We'll keep it as it is.

cc. @lukesteensen

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -30,7 +30,10 @@ impl Resolver {
         // Any port will do, but `9` is a well defined port for discarding
         // packets.
         let dummy_port = 9;
+        // https://tools.ietf.org/html/rfc6761#section-6.3
         if name == "localhost" {
+            // Not all operating systems support `localhost` as IPv6 `::1`, so
+            // we resolving it to it's IPv4 value.
             Ok(LookupIp(
                 vec![SocketAddr::new(Ipv4Addr::LOCALHOST.into(), dummy_port)].into_iter(),
             ))
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -4,7 +4,7 @@ use hyper::client::connect::dns::{Name, Resolve};
 use hyper13::client::connect::dns::Name as Name13;
 use snafu::ResultExt;
 use std::{
-    net::{IpAddr, SocketAddr, ToSocketAddrs},
+    net::{IpAddr, Ipv4Addr, SocketAddr, ToSocketAddrs},
     task::{Context, Poll},
 };
 use tokio::task::spawn_blocking;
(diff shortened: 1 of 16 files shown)


ktff:
@Jeffail I'll remove you form review since your file is only changed in tests. But you are welcome to review nevertheless.

ktff:
> spawn_blocking doesn't take a runtime handle, so it's fine

But it does require the runtime, that's why it must be called from inside of one. The `tokio v0.2` that we will transition to eventually does [expose](https://docs.rs/tokio/0.2.21/tokio/runtime/struct.Handle.html#method.spawn_blocking)  a way to make this dependency explicit via a handle.

src/dns.rs
```diff
@@ -24,18 +24,23 @@ impl Resolver {
...
-        .context(JoinError)?
-        .map(LookupIp)
-        .context(UnableLookup)
+        // We need to add port with the name so that `to_socket_addrs`
+        // resolves it properly. We will be discarding the port afterwards.
+        //
+        // Any port will do, but `9` is a well defined port for discarding
+        // packets.
+        let dummy_port = 9;
+        if name == "localhost" {
```
MOZGIII, line 33: This is great. Would be good to a ref to https://tools.ietf.org/html/rfc6761#section-6.3 for clarity.

ktff: Done.

ktff:
Noting, do not merge this, we are getting
`(signal: 11, SIGSEGV: invalid memory reference)` 
on MAC

ktff:
The error started popping elsewhere too so it's not related to this PR.

PR #2819, opened 2020-06-15
chore: update goauth to new futures

fanatid:
Part of #2118 

Thread blocking on `goauth::get_token` can be removed on chaning sink build to async.

### DIFF
--- a/src/sinks/gcp/mod.rs
+++ b/src/sinks/gcp/mod.rs
@@ -4,7 +4,7 @@ use goauth::scopes::Scope;
 use goauth::{
     auth::{JwtClaims, Token, TokenErr},
     credentials::Credentials,
-    error::GOErr,
+    GoErr,
 };
 use hyper13::{
     header::{HeaderValue, AUTHORIZATION},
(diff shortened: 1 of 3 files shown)


src/sinks/gcp/mod.rs
```diff
@@ -90,11 +90,19 @@ fn get_token_implicit() -> Result<Token, GcpError> {
     }
 }
 
+fn get_token_with_creds_blocking(
+    jwt: &Jwt<JwtClaims>,
+    credentials: &Credentials,
+) -> Result<Token, GoErr> {
+    let mut rt = tokio_compat::runtime::current_thread::Runtime::new()?;
```
bruceg, line 97: I'm certainly not excited about this, but I understand it's necessary and I see the elasticsearch sink has to do the same thing..

fanatid: We should remove this once sink build function will be async. Here also `Credentials::from_file` and `get_token_implicit` which can be async.

bruceg: Exactly right.

PR #2821, opened 2020-06-15
chore(dependencies): Configure dependabot to use correct issue names

Hoverbear:
This PR also reduces the number of times dependabot checks down to weekly.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,6 +1,7 @@
 version: 1
 update_configs:
   - package_manager: "rust:cargo"
+    directory: "/"
     update_schedule: "weekly"
     default_reviewers:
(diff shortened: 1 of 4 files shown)


Hoverbear:
@binarylogic Bit of an issue here:

image

Checked at https://dependabot.com/docs/config-file/validator/

We'd need to use a shorter label prefix. :(

binarylogic:
`chore(deps)` is fine. Can you update it in the `semantic.yml` file as well?

Hoverbear:
Yup!

Hoverbear:
@binarylogic I updated it, before I merge just wanted to make sure there isn't something else to be done -- I know they have hooks into GA etc.

PR #2831, opened 2020-06-16
feat(filter transform): Adding a condition to filter using IP CIDR

bill-bateman:
This allows you to use filter/swimlane using IP CIDRs.

Second part of #2738


bill-bateman:
Uh-oh -- the pull-request / checklist job is failing.

It says: `##[error]Resource not accessible by integration`

Is it because of the change to `Cargo.lock`? New dependency is `cidr_utils` crate (what I use to parse CIDR and test if an IP is in a given CIDR).

bruceg:
@Hoverbear I added you to this one to see if you have any insight into the checklist failure.

src/conditions/check_fields.rs
```diff
@@ -326,6 +329,57 @@ impl CheckFieldsPredicate for ExistsPredicate {
...
+
+impl CheckFieldsPredicate for IpCidrPredicate {
+    fn check(&self, event: &Event) -> bool {
+        match event {
+            Event::Log(l) => l.get(&self.target).map_or(false, |v| {
+                let v = v.to_string_lossy();
+                let ip_addr = match IpAddr::from_str(&v) {
+                    Ok(addr) => addr,
+                    Err(e) => {
+                        error!("Invalid IP \"{}\": {}", v, e);
```
bruceg, line 370: I don't think an error is appropriate here. The other predicate that does conversion (`EqualsPredicate`) simply returns `false` with no notice when conversion fails.

bruceg (changes requested):
Looks pretty good except for the error message. I appreciate you added some good looking tests.

src/conditions/check_fields.rs
bill-bateman: Good call - removing that error

bruceg (approved):
Looks great!

Hoverbear:
Failures are https://github.com/timberio/vector/pull/2873

bill-bateman:
@Hoverbear I see the splunk_hec unit tests are failing like you said - should I uh fix that? By rebasing so I get the splunk_hec fix? Not super good with Git, not sure best way to do that. Or is it ok to leave it since it's been fixed in master?

But the other test fail was the checklist - using this GitHub action: https://github.com/wyozi/contextual-qa-checklist-action
If I understand the tool correctly, it should see that I edited `Cargo.toml` and then create a checklist like:

- [ ] Has at least one other team member approved the dependency changes?

But instead it fails:
```
Run wyozi/contextual-qa-checklist-action@master
  with:
    gh-token: ***
    input-file: .github/checklist.yml
##[error]Resource not accessible by integration
```

Oh actually just saw an issue opened in the repo for the action (https://github.com/wyozi/contextual-qa-checklist-action/issues/11) - made by @fanatid about this same error happening in Vector. No activity on that issue though.

Hoverbear:
@bill-bateman We're good to go. :) Thank you so much for this awesome job!

PR #2832, opened 2020-06-16
fix(kafka sink): Fix typo integration tests

bruceg:
Signed-off-by: Bruce Guenter <bruce@timber.io>

PR #2834, opened 2020-06-16
chore(kafka sink): Fix command for integration tests

fanatid:
Currently Kafka tests not compiled because `rdkafka` feature required.

Also see #2832 

Edit: should fail without #2832 (will rebase after merge).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -246,7 +246,7 @@ ifeq ($(AUTOSPAWN), true)
 	${MAYBE_ENVIRONMENT_EXEC} $(CONTAINER_TOOL)-compose up -d dependencies-kafka
 	sleep 5 # Many services are very lazy... Give them a sec...
 endif
-	cargo test --no-default-features --features kafka-integration-tests ::kafka:: -- --nocapture
+	cargo test --no-default-features --features "kafka-integration-tests rdkafka-plain" ::kafka:: -- --nocapture
 ifeq ($(AUTODESPAWN), true)
 	${MAYBE_ENVIRONMENT_EXEC} $(CONTAINER_TOOL)-compose stop
 endif
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -232,7 +232,7 @@ jobs:
       - run: nix-shell --run "make test-integration-influxdb"
 
   test-integration-kafka:
-    name: Integration - Linux, Kafka)
+    name: Integration - Linux, Kafka
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/scripts/environment/definition.nix
+++ b/scripts/environment/definition.nix
@@ -58,8 +58,6 @@ scope@{ pkgs ? import <nixpkgs> {} }:
     pkg-config
     openssl
     protobuf
-    rdkafka
-    openssl
     ruby_2_7
     nodejs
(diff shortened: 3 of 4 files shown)


fanatid:
hmm tests fails with:

2020-06-17T05:12:05.1378170Z thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: KafkaCreateFailed { source: KafkaError (Client config error: No such configuration property: "ssl.ca.location" ssl.ca.location tests/data/Vector_CA.crt) }', src/sinks/kafka.rs:432:20

Hoverbear:
Hmmmm, very interesting.

Makefile
```diff
@@ -246,7 +246,7 @@ ifeq ($(AUTOSPAWN), true)
 	${MAYBE_ENVIRONMENT_EXEC} $(CONTAINER_TOOL)-compose up -d dependencies-kafka
 	sleep 5 # Many services are very lazy... Give them a sec...
 endif
-	cargo test --no-default-features --features kafka-integration-tests ::kafka:: -- --nocapture
+	cargo test --no-default-features --features kafka-integration-tests --features rdkafka-plain ::kafka:: -- --nocapture
```
Hoverbear, line 249:
```suggestion
	cargo test --no-default-features --features kafka-integration-tests rdkafka-plain ::kafka:: -- --nocapture
```

fanatid: No, this will not work:
```bash
$ cargo test --no-default-features --features kafka-integration-tests rdkafka-plain ::kafka:: -- --nocapture
error: Found argument '::kafka::' which wasn't expected, or isn't valid in this context

USAGE:
    cargo test --features <FEATURES>... --no-default-features

For more information try --help
```

Hoverbear: Sorry, you're right. :) You need to do it your way or wrap it all in a string 

```
--features "kafka-integration-tests  rdkafka-plain"
```

Hoverbear:
I'm not sure why we're seeing this SSL error :thinking:

fanatid:
Locally all tests passed. @bruceg maybe you have any ideas?

bruceg:
That could possibly be an issue with a different version of the kafka library, although that property has been present in librdkafka since SSL/TLS support was added in Sept 2015. Is it possible that the library is built without TLS support?

Hoverbear:
@fanatid You can force your shell to only have our nix pkgs in the path with `nix-shell --pure`. I left you some messages on slack that might help too!

Hoverbear:
@fanatid Can you try adding 

openssl.dev to https://github.com/timberio/vector/blob/6963b5807a79dcef54423e7436d794e979cadb5a/scripts/environment/definition.nix#L62?

fanatid:
Not sure that this is correct @Hoverbear but tests :heavy_check_mark: 

Kafka tests previously fail because `pkg-config` had empty list (`--list-all`), as result event with flag `--enable-ssl` `librdkafka` used in `rdkafka-sys` crate compiled without SSL (because `WITH_SSL=1` was not in `config.h`).

`pkgs.stdenv.mkDerivation` instead `pkgs.mkShell` add libs to `pkg-config` what fixes kafka case.

Hoverbear:
Wowww!!!! Amazing detective work! Thanks so much @fanatid !

PR #2838, opened 2020-06-17
fix(sinks): address potential deadlock

### LINKED ISSUE #2577: Investigate poll_complete possible hang
This poll_complete that we have in our `BatchedHttpSink` should attempt to only return `NotReady` if `poll_complete` applies back pressure.

Reference https://github.com/timberio/vector/pull/2573#discussion_r423233151

lukesteensen:
Closes #2577

While likely not an issue in practice, our `start_send` implementation here was theoretically vulnerable to deadlock.

The issue would occur if the sink's slot was full and the subsequent call to `poll_complete` returned `Async::Ready`. We would then incorrectly return `NotReady` when we actually had the ability to accept the item. This is a problem because no underlying component had returned `NotReady` and our task would therefore not be registered to wake on any future notification.

### DIFF
--- a/src/sinks/util/http.rs
+++ b/src/sinks/util/http.rs
@@ -136,15 +136,15 @@ where
 
     fn start_send(&mut self, item: Self::SinkItem) -> StartSend<Self::SinkItem, Self::SinkError> {
         if self.slot.is_some() {
-            self.poll_complete()?;
-            return Ok(AsyncSink::NotReady(item));
+            if self.poll_complete()?.is_not_ready() {
+                return Ok(AsyncSink::NotReady(item));
+            }
         }
+        assert!(self.slot.is_none(), "poll_complete did not clear slot");
 
         if let Some(item) = self.sink.encode_event(item) {
-            if let AsyncSink::NotReady(item) = self.inner.start_send(item)? {
-                self.poll_complete()?;
-                self.slot = Some(item);
-            }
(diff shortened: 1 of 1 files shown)


bruceg:
I looked at the logic, and I don't think it's actually possible to loop infinitely. Can you describe a case where it could? `poll_complete` takes the slot, which leaves it `None` which breaks the loop. The only way for the slot to be refilled there returns not ready, which also breaks the loop with a return. That said, I don't see a way for it to loop more than once, making the `while` pointless, so maybe I'm missing something too.

Hoverbear:
That's kind of where I'm getting hung up! :thinking:

lukesteensen:
> That said, I don't see a way for it to loop more than once, making the `while` pointless, so maybe I'm missing something too.

Yeah, it ends up just being a terse way to check the conditional twice. I agree that's confusing. How would we feel about something like this instead?

```rust
if self.slot.is_some() {
    if self.poll_complete()?.is_not_ready() {
    }
}
assert!(self.slot.is_none(), "poll did not clear slot");
```

lukesteensen:
The difference is really just how it handles bugs in `poll_complete`. The `while` loop will continue looping in the hope that progress is eventually made, while the assert will just panic due to the contract being violated.

lukesteensen:
`http2.rs` doesn't exist anymore :tada:

PR #2858, opened 2020-06-18
chore(tests): Debug `topology_reuse_old_port` test

ktff:
This PR doesn't fix anything at moment, it just adds more logging for `topology_reuse_old_port` test to determine why it's failing in CI on Windows.


ktff:
I wasn't able to reproduce it here, so I'll leave it be until we observe it somewhere again. 

Noting: 
Current speculation is that Windows hasn't yet finished the process of making the port, that we have released a moment before, available again before we ask for it again.

PR #2862, opened 2020-06-18
chore(tests): Adjust buffering tests

ktff:
This PR does three things:
* Resolves races exclusive to `buffering` tests.
* Moves around `thread::sleep` in `buffering` tests, reducing waiting in total.
* Splits `crate::test_util::CountReceiver::wait` method into `wait` and `cancel`. All of the usages of this method were having a race as they were expecting that the method would wait for the `tcp` channel to close, which wasn't the case. Instead the method was canceling the channel which caused a race.

### DIFF
--- a/tests/buffering.rs
+++ b/tests/buffering.rs
@@ -8,7 +8,7 @@ use std::sync::atomic::Ordering;
 use tempfile::tempdir;
 use tracing::trace;
 use vector::event;
-use vector::test_util::{self, next_addr, wait_for};
+use vector::test_util::{self, next_addr, wait_for, wait_for_atomic_usize};
 use vector::topology::{self, config};
 use vector::{buffers::BufferConfig, runtime, sinks};
 
@@ -71,8 +71,17 @@ fn test_buffering() {
     let _ = rt.block_on(send).unwrap();
 
     // We need to wait for at least the source to process events.
-    wait_for(|| source_event_counter.load(Ordering::Acquire) == num_events);
-    // Now we give it some time for the events to propagate to Disk.
+    // This is to avoid a race after we send all events, at that point two things
+    // can happen in any order, we reaching `terminate_abruptly` and source processing
+    // all of the events. We need for the source to process events before `terminate_abruptly`
+    // so we wait for that here.
+    wait_for_atomic_usize(source_event_counter, |x| x == num_events);
+    // Now we give it some time for the events to propagate to File.
+    // This is to avoid a race after the source processes all events, at that point two things
+    // can happen in any order, we reaching `terminate_abruptly` and events being written
+    // to file. We need for the events to be written to the file before `terminate_abruptly`.
+    // We can't know when exactly all of the events have been written, so we have to guess.
+    // But it should be shortly after source processing all of the events.
     std::thread::sleep(std::time::Duration::from_secs(1));
     // Simulate a crash.
     terminate_abruptly(rt, topology);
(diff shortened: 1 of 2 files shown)


tests/buffering.rs
```diff
@@ -71,17 +70,16 @@ fn test_buffering() {
...
     let _ = rt.block_on(send).unwrap();
 
-    // There was a race caused by a channel in the mock source, and this
-    // check is here to ensure it's really gone.
-    assert_eq!(source_event_counter.load(Ordering::Acquire), num_events);
-
-    // Give the topology some time to process the received data and simulate
-    // a crash.
+    // We need to wait for at least the source to process events.
+    wait_for(|| source_event_counter.load(Ordering::Acquire) == num_events);
```
MOZGIII, line 74: So, this means that the changes from  https://github.com/timberio/vector/commit/e0761a9681da6b7e8cb69c750c41a7be89d24fed  don't fix the issue :( 
Can we restore the comment that was there before the said comment? I think it's more on-point, and gives hints on where the root cause of the issue is. Unless it's not relevant anymore.

tests/buffering.rs
```diff
@@ -352,10 +336,6 @@ fn test_reclaim_disk_space() {
 
     let output_events = test_util::receive_events(out_rx);
 
-    // There was a race caused by a channel in the mock source, and this
-    // check is here to ensure it's really gone.
-    assert_eq!(source_event_counter.load(Ordering::Acquire), num_events);
```
MOZGIII, line 357: I'm pretty sure the potential problem remains here, and this is a meaningful assertion.
We can remove it, of course, but that'd be effectively muting the failure.

tests/buffering.rs
ktff: It fixed races relevant to the PR, but there is another race in MockSourceConfig that we can't avoid without changing the channel impl to some that doesn't use buffering, but even then just source processing events isn't enough, they still need to reach the buffers of source and yet still they need to reach file to which we are buffering those events. Only some intrusive synchronizing could fix this.

Because of the first race, the assert isn't appropriate.

tests/buffering.rs
ktff: https://github.com/timberio/vector/pull/2862/files#r444125759 also applies to this.

tests/buffering.rs
MOZGIII: Yeah! The code before https://github.com/timberio/vector/commit/e0761a9681da6b7e8cb69c750c41a7be89d24fed was:

```rust
    // A race caused by `rt.block_on(send).unwrap()` is handled here. For some
    // reason, at times less events than were sent actually arrive to the
    // `source`.
    // We mitigate that by waiting on the event counter provided by our source
    // mock.
    test_util::wait_for_atomic_usize(source_event_counter, |x| x == num_events);
```

It's pretty much what you said, so I suggest just restoring it as it was.

MOZGIII: So, to improve the overall situation, we have to reimplement the `MockSourceConfig`, right?

ktff: Ok, I'll document the race for because of which we need to wait. And use that `test_util::wait_for_atomic_usize`.


> So, to improve the overall situation, we have to reimplement the MockSourceConfig, right?

I don't think its possible to get better than waiting for `source_event_counter`.

ktff:
New error started occurring, 
```
panicked at 'attempt to add with overflow', src\buffers\disk\leveldb_buffer.rs:86:12
```
I think it's a bug that existed before this PR so I'll be merging this in spite of it.

PR #2865, opened 2020-06-18
chore(dependencies): file-source bytes update

fanatid:
Do not like that for creating `BytesMut` from `Bytes` data need to be copied, but looks like there is no way do this in bytes:0.5 without it.

### DIFF
--- a/src/sources/file/line_agg.rs
+++ b/src/sources/file/line_agg.rs
@@ -1,4 +1,4 @@
-use bytes::{Bytes, BytesMut};
+use bytes05::{Bytes, BytesMut};
 use futures01::{Async, Poll, Stream};
 use regex::bytes::Regex;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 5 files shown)


MOZGIII:
> Do not like that for creating `BytesMut` from `Bytes` data need to be copied, but looks like there is no way do this in bytes:0.5 without it.

I've stublem upon this before, and I think it's possible to pass the payload around via `Vec<u8>`. Both `bytes` versions should have a cheap conversion to/from `Vec`, so might it work.

src/sources/file/line_agg.rs
```diff
@@ -248,6 +248,12 @@ where
     }
 }
 
+fn make_mut(bytes: Bytes) -> BytesMut {
+    let mut bytes_mut = BytesMut::new();
+    bytes_mut.extend_from_slice(&bytes);
+    bytes_mut
```
MOZGIII, line 254: For the record, this is fine, the copying was there before: https://docs.rs/bytes/0.4.12/src/bytes/bytes.rs.html#1653-1658.

fanatid: Yes. it was. But there also was possibility that clone will not actually happen..

PR #2873, opened 2020-06-19
fix(splunk_hec source): fix wrap path

### LINKED ISSUE #2864: Tests in splunk_hec splunk_hec fails
Probably result of warp update? #2837

fanatid:
Resolve #2864 

From docs https://docs.rs/warp/0.2.3/warp/macro.path.html

> The path! macro automatically assumes the path should include an end() filter. To build up a path filter prefix, such that the end() isn't included, use the / .. syntax.

Not sure how this passed tests in #2837

### DIFF
--- a/src/sources/splunk_hec.rs
+++ b/src/sources/splunk_hec.rs
@@ -85,7 +85,7 @@ impl SourceConfig for SplunkConfig {
         let health_service = source.health_service(out.clone());
         let options = SplunkSource::options();
 
-        let services = path!("services" / "collector")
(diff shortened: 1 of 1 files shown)


lukesteensen (commented):
Could we write a test that would have caught this? It'd be nice to see one that failed without this fix.

fanatid:
Not sure how this can be caught. :thinking:  Current tests already fail with http status 404 without this changes.

Hoverbear (approved):

PR #2874, opened 2020-06-19
chore: Bump rust-toolchain to 1.44.1

fanatid:
https://blog.rust-lang.org/2020/06/18/Rust.1.44.1.html
https://github.com/rust-lang/rust/blob/stable/RELEASES.md#version-1441-2020-06-18


MOZGIII:
https://github.com/rust-lang/rust/issues/71397 - segfaults on macos that we've seen in tests could be this.

PR #2896, opened 2020-06-24
feat(windows platform): add a way to register and run vector as a service

oktal:
On Windows, Vector should be able to run as a service.

This must be done in two distinct steps:

First, the user must register vector as a service in the Service Control Manager. A new `--service` command-line parameter has been introduced (windows only), with multiple possible values:
* `--service install` install vector as a service 
* `--service uninstall` uninstall the service
* `--service start` start the service
* `--service stop` stop the service

Then, vector must provide a service entry point and register itself with the Service Control Manager. A dependency to the [windows-service](https://crates.io/crates/windows-service) has been introduced to handle service install and registration.

The `main` function now has been split into two distinct main (old main `#[cfg(not(windows))]` and a new main for `#[cfg(windows)]`.

The new `main` for Windows first attempts to start as a service, and then falls back to console mode.
The launch code has been extracted to a new `service.rs` file so that it could be called from the part that handles the startup of the service on Windows.

I think this PR still needs some work. I'm pretty new to Rust (coming from a C++ background), so my coding style might not fit your standards. I also need some suggestion on how to properly handle and wait for the sources to shutdown properly before stopping the service on Windows (`vector_windows.rs:60`)

This feature has been inspired from telegraf.


Hoverbear:
Hi @oktal ! This looks super cool! Thanks so much for contributing this!

src/main.rs
```diff
@@ -1,359 +1,117 @@
...
-    if opts.watch_config {
-        // Start listening for config changes immediately.
-        vector::topology::config::watcher::config_watcher(
-            config_paths.clone(),
-            vector::topology::config::watcher::CONFIG_WATCH_DELAY,
-        )
-        .unwrap_or_else(|error| {
-            error!(message = "Unable to start config watcher.", %error);
-            std::process::exit(exitcode::CONFIG);
+#[cfg(windows)]
```
Hoverbear, line 92: I feel like starting as a service should be opt-in. It seems like running `vector` via Powershell after this would mean we started vector as a service, not directly? I'd prefer that do be done via the dedicated service start command if we could?

What kind of user experience difference does this bring?

src/service.rs
```diff
@@ -0,0 +1,388 @@
+#![allow(clippy::cognitive_complexity)]
```
Hoverbear, line 1: So right now this file seems to have most of the old `main.rs` content. I'd prefer that it go back there and we have this file include most of the `vector_windows.rs` code. It'd be particularly awesome if we can have it structured so that we could add other OS's later in a very clean way...

src/main.rs
oktal: It is actually opt-in. If you start `vector` via Powershell or windows command line, it won't start as a service. However, since console mode and windows service mode have the same entry point (the `main` function), I need to check whether the binary has been started as a service since I need to register the service with the Windows Service Control Manager when it is started by Windows as a service.

Do detect whether or not `vector` has been started as a service, I then attempt to call the Win32 `StartServiceCtrlDispatcher` function (through `windows-service-rs` library) and check its result code.

This is why I first try to start as a serevice and fall-back to console mode if I fail. This is somehow counter-intuitive, I had some trouble understanding the whole principle, see for  example the [MSDN Service Sample](https://docs.microsoft.com/en-us/windows/win32/services/svc-cpp)

src/service.rs
oktal: I extracted parts of this file because I needed to parse command-line in `vector_windows.rs` when starting as a service.
I wanted to keep windows specific code and boilerplate stuff separated in its own source file.

 So what you say is that content of this file should be directory put in `main.rs`, inside a `#[cfg(windows)]` directive ?

We could work on some interface that would make it easy to add support for other OS's later. That's what I started to do initially before giving up because Linux is already supported through `systemd`. 

`telegraf` uses this [library](https://github.com/kardianos/service) that supports Windows XP+, Linux, OSX

oktal:
Thanks for the feedback and comments !

Yes, we could instead isolate service related commands in a specific `service` subcommand and keep the rest of the command line untouched. I like the idea of isolating the service behavior.

What about `vector service --action start` instead ?

I'm also planning to add `--name` and `--display-name` options if you want to install multiple instances of the same service under different names.

src/main.rs
Hoverbear: OKay! This makes sense. Thanks. :)

src/service.rs
Hoverbear: Hmmm yeah I think we had the same feeling here. Ok!

Hoverbear:
@oktal I like that UX too! Maybe we could make `start` and `stop` etc optional subcommands?

```
vector service start --name foo
vector service stop --name foo
vector service --name foo # Status
vector service restart --name foo
```

Hoverbear:
Noting for myself I should write some docs for this. :)

src/config_paths.rs
```diff
@@ -49,3 +49,16 @@ pub fn expand(config_paths: Vec<PathBuf>) -> Option<Vec<PathBuf>> {
     }
     Some(paths)
 }
+fn default_config_paths() -> Vec<PathBuf> {
+    let config_paths = if cfg!(unix) {
+        DEFAULT_UNIX_CONFIG_PATHS.clone()
+    } else if cfg!(windows) {
+        let program_files = env!("ProgramFiles");
```
Hoverbear, line 56: Maybe this would be more understandable if it was built in https://github.com/timberio/vector/blob/64b0754b34e08b4449270e510262040341a4ea12/src/config_paths.rs#L7 ?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2966056be5134e91cb37089dbfe301b5

oktal:
> Maybe we could make start and stop etc optional subcommands?

I guess that works too. I just need to check if you can have recursive subcommands with `structopt` ?

I let you decide before moving on with this :)

src/vector_windows.rs
```diff
@@ -0,0 +1,247 @@
...
+            // Handle stop
+            ServiceControl::Stop => {
+                shutdown_tx.send(()).unwrap();
+                ServiceControlHandlerResult::NoError
+            }
+
+            _ => ServiceControlHandlerResult::NotImplemented,
+        }
+    };
+    let status_handle =
```
Hoverbear, line 212: This is wild! Hahah, I had no idea Windows did this.

Hoverbear:
@oktal Yeah let's not worry too much about the UX... I think this is going to get followed up with a RFC to add a more generic impl that can hammer it out. (You are welcome to do this, but I won't ask it of you to merge this!)

src/vector_windows.rs
```diff
@@ -0,0 +1,247 @@
...
+    use std::ffi::OsString;
+    use std::time::Duration;
+    use std::{fmt, thread};
+
+    #[derive(Debug, Clone, PartialEq)]
+    pub enum ControlAction {
+        Install,
+        Uninstall,
+        Start,
+        Stop,
```
Hoverbear, line 32: Restart/reload?

src/vector_windows.rs
```diff
@@ -0,0 +1,247 @@
...
+        let arguments = config_paths
+            .iter()
+            .flat_map(|p| vec![OsString::from("--config"), p.as_os_str().into()])
+            .collect::<Vec<OsString>>();
+
+        let service = ServiceDefinition {
+            name: OsString::from(SERVICE_NAME),
+            display_name: OsString::from("Vector Service"),
+
+            description: OsString::from("Vector is a lightweight, ultra-fast, open-source tool for building observability pipelines."),
```
Hoverbear, line 81: We use https://docs.rs/built/0.4.2/built/ which can provide these. `PKG_DESCRIPTION` in this case

src/vector_windows.rs
```diff
@@ -0,0 +1,247 @@
...
+            account_password: None,
+        };
+
+        debug!(
+            message = "Registering service.",
+            service_info = format!("{:?}", service_info).as_str(),
+        );
+
+        service_manager.create_service(service_info, ServiceAccess::empty())?;
+        // It is currently not possible to change the description of the service.
```
Hoverbear, line 149:
```suggestion
        // TODO: It is currently not possible to change the description of the service.
```

lukesteensen:
> Maybe we could make `start` and `stop` etc optional subcommands?

I like this idea! It lines up a little better with how `systemctl` manages services on Linux. That said, I don't have a strong opinion here.

src/config_paths.rs
oktal: Yes, this is better !

I also need to target the right recommended default configuration folder. I guess I'll follow what you proposed here ? https://github.com/timberio/vector/issues/1738#issuecomment-583493898

Maybe `%APPDATA%` ?

oktal:
@lukesteensen I like the idea of following `systemctl` conventions, will use optional subcommands then.

src/service.rs
oktal: What do you think we should do about `main.rs` ? Moving code back or keeping things as they are right now ?

Hoverbear:
You are so, so awesome!

src/service.rs
```diff
@@ -0,0 +1,273 @@
...
+    };
+
+    code
+}
+
+#[cfg(windows)]
+fn control_service(service: &ServiceInfo, action: ControlAction) -> exitcode::ExitCode {
+    use crate::vector_windows;
+
+    if action == ControlAction::Status {
```
oktal, line 217: I don't think  this is idiomatic Rust. I'm open to suggestions on how to improve this code

src/vector_windows.rs
```diff
@@ -0,0 +1,222 @@
...
+                controls_accepted: ServiceControlAccept::STOP,
+                exit_code: ServiceExitCode::Win32(0),
+                checkpoint: 0,
+                wait_hint: Duration::default(),
+            })?;
+
+            let topology = vector.topology;
+
+            shutdown_rx.recv();
+            let shutdown = topology.stop();
```
oktal, line 204: I guess I need to wait for the topology to stop here. Should I use the same code pattern than `main.rs` and `await` on the `Future` inside a  `Runtime::block_on` ? Ideally, we also need some kind of timeout and force the server to shut down.

src/service.rs
```diff
@@ -0,0 +1,273 @@
...
+        }
+        None => exitcode::USAGE,
+    };
+
+    exit_code
+}
+
+#[cfg(unix)]
+fn control_service(service: &ServiceInfo, action: ControlAction) -> exitcode::ExitCode {
+    error!("Service commands are currently not supported on this platform.");
```
oktal, line 255: I made the choice of making the `service` subcommand available on all platforms, but the only implementation is for Windows right now.

This gives the opportunity  to add other platforms implementation later on.

oktal:
Hey guys, a new version of the Pull Request has landed !

As discussed, I moved service code inside a `service` subcommand in `service.rs`. The options are now:

```
vector.exe-service 0.10.0
Control the state of the vector service

USAGE:
    vector.exe service [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    install      Install vector as a service
    uninstall    Uninstall the service
    start        Start the service
    stop         Stop the service
    restart      Restart the service
    status       Print the status of the service
    help         Prints this message or the help of the given subcommand(s)
```

All subcommands except `install` have an optional `--name` option in case the service was installed with a custom name.

Speaking on that, we also need a way to pass  the name of the service that was specified during installation since this is needed when vector is started as a service by the windows Service Control Manager (`vector_windows::run_service`).

Currently, the exe is started with a `--config` argument. We would need something like `--service-name` on Windows to register with the right service name. Suggestions ?

@Hoverbear

src/service.rs
```diff
@@ -0,0 +1,297 @@
...
+    name: Option<String>,
+
+    /// The display name to be used by interface programs to identify the service like Windows Services App
+    #[structopt(long)]
+    display_name: Option<String>,
+
+    /// The configuration files that will be used by the service.
+    /// If no configuration file is specified, will target default configuration file.
+    #[structopt(name = "config", short, long)]
+    config_paths: Vec<PathBuf>,
```
oktal, line 35: I initially wanted to reuse the same `--config` parameter from the `RootOpts` but failed to do so.

```
release\vector.exe --config C:\Users\oktal\tmp\vector.toml -- service install
error: The subcommand 'service' wasn't recognized
        Did you mean 'service'?

If you believe you received this message in error, try re-running with 'vector.exe -- service'

USAGE:
    vector.exe --config <config>...

For more information try --help
```

src/service.rs
Hoverbear: I think for now it's good to make it only appear on windows. We can make it appear later for other OS. :)

src/service.rs
Hoverbear: I wonder if it's worth us making it global (https://kbknapp.github.io/clap-rs/clap/struct.Arg.html#method.global) in the future...

Either way, I think this is fine for now.

src/vector_windows.rs
Hoverbear: Will Windows not issue something like a SIGKILL at a certain timeout? Whoa.

oktal: Windows does not have unix signals. An application can use the WinAPI [SetConsoleCtrlHandler](https://docs.microsoft.com/en-us/windows/console/setconsolectrlhandler) for events like Ctrl-C but the system won't emit any signal.

I'm not sure what the SCM does when the timeout is reached though.

src/service.rs
oktal: Ok so I'll make the sub command only available on Windows, then.

Hoverbear: Thank you! :)

src/vector_windows.rs
Hoverbear: Hmmm okay. I think it might be worthwhile for us to use a timeout yes. So I guess the same strategy main.rs uses then is best.

Hoverbear:
Noting this could use some documentation and I'll add that. :)

Hoverbear:
While trying this I noted that this PR seems to break `vector generate stdin//console` on Windows:

```
PS C:\Users\Operator\git\timberio\vector> cargo run --no-default-features --features default-msvc -- generate stdin//console
warning: unused variable: `shutdown`
   --> src\vector_windows.rs:204:17
    |
204 |             let shutdown = topology.stop();
    |                 ^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_shutdown`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused `std::result::Result` that must be used
   --> src\vector_windows.rs:203:13
    |
203 |             shutdown_rx.recv();
    |             ^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
    = note: this `Result` may be an `Err` variant, which should be handled

warning: 2 warnings emitted

warning: unused imports: `handle_config_errors`, `read_configs`
 --> src\main.rs:9:19
  |
9 | use vector::cli::{handle_config_errors, read_configs, start_service};
  |                   ^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^
  |

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.77s
     Running `target\debug\vector.exe generate stdin//console`
Jul 30 20:57:52.412 ERROR vector::config_paths: Config file not found in path. path="C:\\Program Files\\Vector\\config\\vector.toml"
error: process didn't exit successfully: `target\debug\vector.exe generate stdin//console` (exit code: 78)
```

I suspect this is because `vector::vector_windows::run()` tries to read the config which does not exist. :(

Perhaps this should detect if the user has passed in a subcommand?

Hoverbear:
I don't think this is expected, the user sees an error when running `vector service` from a non-elevated prompt:

```
PS C:\Users\Operator\git\timberio\vector> cargo run --no-default-features --features default-msvc -- service
warning: unused variable: `shutdown`
   --> src\vector_windows.rs:204:17
    |
204 |             let shutdown = topology.stop();
    |                 ^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_shutdown`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused `std::result::Result` that must be used
   --> src\vector_windows.rs:203:13
    |
203 |             shutdown_rx.recv();
    |             ^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
    = note: this `Result` may be an `Err` variant, which should be handled

warning: 2 warnings emitted

warning: unused imports: `handle_config_errors`, `read_configs`
 --> src\main.rs:9:19
  |
9 | use vector::cli::{handle_config_errors, read_configs, start_service};
  |                   ^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^
  |

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.80s
     Running `target\debug\vector.exe service`
Jul 30 21:47:09.583  INFO vector::cli: Log level "info" is enabled.
Jul 30 21:47:09.586 ERROR vector::service: Error opening service. error=IO error in winapi call
error: process didn't exit successfully: `target\debug\vector.exe service` (exit code: 70)
```

Hoverbear:
It would be very nice to see a better error here when the user  doesn't have a service installed:

```
PS C:\Users\Operator\git\timberio\vector> sudo cargo run --no-default-features --features default-msvc -- service status
warning: unused variable: `shutdown`
   --> src\vector_windows.rs:204:17
    |
204 |             let shutdown = topology.stop();
    |                 ^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_shutdown`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused `std::result::Result` that must be used
   --> src\vector_windows.rs:203:13
    |
203 |             shutdown_rx.recv();
    |             ^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
    = note: this `Result` may be an `Err` variant, which should be handled

warning: 2 warnings emitted

warning: unused imports: `handle_config_errors`, `read_configs`
 --> src\main.rs:9:19
  |
9 | use vector::cli::{handle_config_errors, read_configs, start_service};
  |                   ^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^
  |

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.67s
     Running `target\debug\vector.exe service status`
Jul 30 21:55:52.374  INFO vector::cli: Log level "info" is enabled.
Jul 30 21:55:52.377 ERROR vector::service: Error opening service. error=IO error in winapi call
error: process didn't exit successfully: `target\debug\vector.exe service status` (exit code: 70)
```

Hoverbear:
Checking the status on an installed service seems to yield some issues too:

```
PS C:\Users\Operator\git\timberio\vector> sudo cargo run --no-default-features --features default-msvc -- service install
    Finished dev [unoptimized + debuginfo] target(s) in 0.67s
     Running `target\debug\vector.exe service install`
Jul 30 21:56:54.455  INFO vector::cli: Log level "info" is enabled.
PS C:\Users\Operator\git\timberio\vector> sudo cargo run --no-default-features --features default-msvc -- service status
    Finished dev [unoptimized + debuginfo] target(s) in 0.68s
     Running `target\debug\vector.exe service status`
Jul 30 21:57:04.960  INFO vector::cli: Log level "info" is enabled.
error: process didn't exit successfully: `target\debug\vector.exe service status` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

Hoverbear:
Gosh this is so cool: 
image

Install and uninstall work great :)

Hoverbear:
I'm noting `start` provides similar access errors: 

```
PS C:\Users\Operator\git\timberio\vector> sudo cargo run --no-default-features --features default-msvc -- service start
warning: unused variable: `shutdown`
   --> src\vector_windows.rs:204:17
    |
204 |             let shutdown = topology.stop();
    |                 ^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_shutdown`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused `std::result::Result` that must be used
   --> src\vector_windows.rs:203:13
    |
203 |             shutdown_rx.recv();
    |             ^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
    = note: this `Result` may be an `Err` variant, which should be handled

warning: 2 warnings emitted

warning: unused imports: `handle_config_errors`, `read_configs`
 --> src\main.rs:9:19
  |
9 | use vector::cli::{handle_config_errors, read_configs, start_service};
  |                   ^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^
  |

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.77s
     Running `target\debug\vector.exe service start`
Jul 30 22:00:28.560  INFO vector::cli: Log level "info" is enabled.
Jul 30 22:00:28.572 ERROR vector::service: Error controlling service. error=IO error in winapi call
error: process didn't exit successfully: `target\debug\vector.exe service start` (exit code: 70)

```

src/service.rs
Hoverbear: I think this is totally ok. :) You might want to use a `match` statement though and only catch the commands you need here.

oktal:
Thanks a lot for the review and feedback @Hoverbear !

My goal is to have all the issues you raised sorted by the end of the week :)

oktal:
> Checking the status on an installed service seems to yield some issues too:
> 
> ```
> PS C:\Users\Operator\git\timberio\vector> sudo cargo run --no-default-features --features default-msvc -- service install
>     Finished dev [unoptimized + debuginfo] target(s) in 0.67s
>      Running `target\debug\vector.exe service install`
> Jul 30 21:56:54.455  INFO vector::cli: Log level "info" is enabled.
> PS C:\Users\Operator\git\timberio\vector> sudo cargo run --no-default-features --features default-msvc -- service status
>     Finished dev [unoptimized + debuginfo] target(s) in 0.68s
>      Running `target\debug\vector.exe service status`
> Jul 30 21:57:04.960  INFO vector::cli: Log level "info" is enabled.
> error: process didn't exit successfully: `target\debug\vector.exe service status` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
> ```

I have the same issue. It's weird because it was initially working when I first submitted the PR and tried on my side.

The issue seems to be in `windows-service`, the crate I'm depending on to handle all the service-related calls to the Windows API. I submitted a PR that should fix the issue https://github.com/mullvad/windows-service-rs/pull/52

oktal:
> It would be very nice to see a better error here when the user doesn't have a service installed:
> 
> ```
> PS C:\Users\Operator\git\timberio\vector> sudo cargo run --no-default-features --features default-msvc -- service status
>     Finished dev [unoptimized + debuginfo] target(s) in 0.67s
>      Running `target\debug\vector.exe service status`
> Jul 30 21:55:52.374  INFO vector::cli: Log level "info" is enabled.
> Jul 30 21:55:52.377 ERROR vector::service: Error opening service. error=IO error in winapi call
> error: process didn't exit successfully: `target\debug\vector.exe service status` (exit code: 70)
> ```

Error messages have been improved. For example, here is now what you get when attempting to retrieve status while  the service has been installed yet:

```
Aug 06 15:06:07.005 ERROR vector::service: Error controlling service. error=The specified service does not exist as an installed service. (os error 1060)
```

oktal:
> While trying this I noted that this PR seems to break vector generate stdin//console or any other subcommand on Windows:

Fixed in https://github.com/timberio/vector/pull/2896/commits/79b00d20a610022561fe7940026bbb13ea688901

Reading of configuration files is now deferred after subcommand evaluation.

src/vector_windows.rs
oktal: Fixed in https://github.com/timberio/vector/pull/2896/commits/79b00d20a610022561fe7940026bbb13ea688901

The future returned by `RunningTopology::stop()` is now correctly await-ed. It already has an internal deadline timer, so I did not add one.

oktal:
Hi @Hoverbear, updated PR that should address all the issues you raised 🎉

The only remaining issue is the `service status` one, which I'm waiting on the `windows-service-rs` author to merge my PR.

Hoverbear:
Hey @oktal !

Sorry about the review delay, I had a vacation! ⛺ 

I'm going to give this some QA testing tomorrow and see if there is anything left to fiddle with while we wait on the new release upstream.

src/service.rs
```diff
@@ -0,0 +1,278 @@
...
+            &service_definition,
+            vector_windows::service_control::ControlAction::Stop,
+        ),
+        _ => unreachable!(),
+    };
+
+    let exit_code = match res {
+        Ok(()) => exitcode::OK,
+        Err(error) => {
+            win_error!("Error controlling service.", error);
```
Hoverbear, line 248: It seems you only use this macro once? Is it worth keeping?

src/cli.rs
```diff
@@ -153,3 +165,159 @@ impl std::str::FromStr for LogFormat {
...
+
+pub struct Vector {
+    pub config_paths: Vec<PathBuf>,
+    pub opts: RootOpts,
+    pub topology: RunningTopology,
+    pub graceful_crash: mpsc::UnboundedReceiver<()>,
+    pub runtime: Runtime,
+}
+
+pub fn prepare_service() -> Result<Vector, exitcode::ExitCode> {
```
Hoverbear, line 177: Do you think this might be better structured as part of a `impl Vector` block?

src/main.rs
```diff
@@ -1,133 +1,27 @@
...
-                format!("vector={}", level),
-                format!("codec={}", level),
-                format!("file_source={}", level),
-                "tower_limit=trace".to_owned(),
-                format!("rdkafka={}", level),
-            ]
-            .join(","),
-        },
-    };
+fn run_vector(vector: Vector) {
```
Hoverbear, line 15: Perhaps this too might find a good home in the `Vector` struct? 🤔

src/service.rs
```diff
@@ -0,0 +1,278 @@
...
+    let config = service.query_config()?;
+
+    let name_string = format!("{:?}", service_name);
+    let state_string = format!("{:?}", status.current_state);
+
+    let state_width = state_string.len() + 2;
+    let name_width = name_string.len() + 2;
+
+    println!(
+        "{0: <state_width$} {1: <name_width$} Display Name",
```
Hoverbear, line 187: Based on this, I would expect more output from this command? 
image

src/service.rs
```diff
@@ -0,0 +1,278 @@
...
+    let status = service.query_status()?;
+    let config = service.query_config()?;
+
+    let name_string = format!("{:?}", service_name);
+    let state_string = format!("{:?}", status.current_state);
+
+    let state_width = state_string.len() + 2;
+    let name_width = name_string.len() + 2;
+
+    println!(
```
Hoverbear, line 186: Could you make these `info!`? This way it will hook into our `tracing` implementation.

src/service.rs
```diff
@@ -0,0 +1,278 @@
...
+        "{0: <state_width$} {1: <name_width$} Display Name",
+        "State",
+        "Name",
+        state_width = state_width,
+        name_width = name_width
+    );
+    println!(
+        "{0:-<state_width$} {1:-<name_width$} -------------",
```
Hoverbear, line 194: Is this an idiom or convention?

Hoverbear:
```powershell
PS C:\Users\ana\CLionProjects\vector> .\target\debug\vector.exe service install --config foo.toml
Aug 20 08:42:37.136  INFO vector::cli: Log level "info" is enabled.
Aug 20 08:42:37.152 ERROR vector::service: Error controlling service. error=Access is denied. (os error 5)
PS C:\Users\ana\CLionProjects\vector> sudo .\target\debug\vector.exe service install --config foo.toml
Aug 20 08:43:23.465  INFO vector::cli: Log level "info" is enabled.
```

Great, nice. 👍

src/service.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// If no configuration file is specified, will target default configuration file.
+    #[structopt(name = "config", short, long)]
+    config_paths: Vec<PathBuf>,
+}
+
+impl InstallOpts {
+    fn service_info(&self) -> ServiceInfo {
+        let service_name = self.name.as_deref().unwrap_or(*DEFAULT_SERVICE_NAME);
+        let display_name = self.display_name.as_deref().unwrap_or("Vector Service");
+        let description = built_info::PKG_DESCRIPTION;
```
Hoverbear, line 51: I unfortunately did not see this get set?

image

src/vector_windows.rs
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn install_service(service: &ServiceDefinition) -> Result<()> {
+        let manager_access = ServiceManagerAccess::CONNECT | ServiceManagerAccess::CREATE_SERVICE;
+        let service_manager = ServiceManager::local_computer(None::<&str>, manager_access)?;
+
+        let service_info = ServiceInfo {
```
Hoverbear, line 106: Could I ask you to make this a call to a (new) `ServiceInfo::new()`?

src/vector_windows.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        debug!(
+            message = "Registering service.",
+            service_info = format!("{:?}", service_info).as_str(),
+        );
+
+        service_manager.create_service(service_info, ServiceAccess::empty())?;
+        // TODO: It is currently not possible to change the description of the service.
+        // Waiting for the following PR to get merged in
+        // https://github.com/mullvad/windows-service-rs/pull/32
```
Hoverbear, line 127: I see why https://github.com/timberio/vector/pull/2896/files#r474094595 is happening! Ok. :) Thank you. This is fine.

src/service.rs
```diff
@@ -0,0 +1,278 @@
...
+macro_rules! win_error {
+    ($msg:expr, $e:ident) => {
+       match $e {
+           windows_service::Error::Winapi(error) => error!(message=$msg, %error),
+           _ => error!(message=$msg, %$e)
+       }
+    };
+}
+
+lazy_static! {
```
Hoverbear, line 19: I'm not sure this needs to be a `lazy_static`. Can you make this a `const`?

src/vector_windows.rs
```diff
@@ -0,0 +1,229 @@
...
+        //
+        // service.set_description(&self.description);
+        Ok(())
+    }
+
+    fn uninstall_service(service: &ServiceDefinition) -> Result<()> {
+        let service_access =
```
Hoverbear, line 134: It seems like services only disappear on halt (and can only be reinstalled after reboot), is this correct? If so, can you `warn!` that?

Hoverbear:
@oktal Could you help me understand what the problem here is?

```powershell
PS C:\Users\ana\CLionProjects\vector> sudo .\target\debug\vector.exe service install --config foo.toml
Aug 20 09:14:12.228  INFO vector::cli: Log level "info" is enabled.
PS C:\Users\ana\CLionProjects\vector> sudo .\target\debug\vector.exe service status
Aug 20 09:14:22.252  INFO vector::cli: Log level "info" is enabled.
PS C:\Users\ana\CLionProjects\vector> sudo .\target\debug\vector.exe service start
Aug 20 09:14:30.020  INFO vector::cli: Log level "info" is enabled.
Aug 20 09:14:30.031 ERROR vector::service: Error controlling service. error=The service did not respond to the start or control request in a timely fash
ion. (os error 1053)
PS C:\Users\ana\CLionProjects\vector> sudo .\target\debug\vector.exe service stop
Aug 20 09:14:35.732  INFO vector::cli: Log level "info" is enabled.
Aug 20 09:14:35.736 ERROR vector::service: Error controlling service. error=The service has not been started. (os error 1062)
PS C:\Users\ana\CLionProjects\vector> sudo .\target\debug\vector.exe service uninstall
Aug 20 09:14:40.784  INFO vector::cli: Log level "info" is enabled.
PS C:\Users\ana\CLionProjects\vector> cat .\foo.toml
data_dir = "/var/lib/vector/"

[sources.source0]
type = "generator"
lines = ["Line 1", "Line 2"]


[sinks.sink0]
healthcheck = true
inputs = ["source0"]
type = "console"
encoding = "json"

[sinks.sink0.buffer]
type = "memory"
max_events = 500
when_full = "block"
```

src/service.rs
oktal: I tried to reproduce the same output from the PowerShell `Get-Service` command:

```console
PS C:\WINDOWS\system32> Get-Service -Name "vector"

Status   Name               DisplayName
------   ----               -----------
Stopped  vector             Vector Service
```

src/service.rs
oktal: Is there any particular reason to hook into the `tracing` implementation ?

Here is what the output looks like with `info!`

```console
Aug 31 16:13:33.182  INFO vector::cli: Log level "info" is enabled.
Aug 31 16:13:33.186  INFO vector::service: State       Name       Display Name
Aug 31 16:13:33.186  INFO vector::service: ----------- ---------- -------------
Aug 31 16:13:33.186  INFO vector::service: "Stopped"   "vector"   "Vector Service"
```

With `println!`

```console
Aug 31 16:14:43.363  INFO vector::cli: Log level "info" is enabled.
State       Name       Display Name
----------- ---------- -------------
"Stopped"   "vector"   "Vector Service"
```

I don't think there is any particular reason to dump full log info in this case

oktal:
> @oktal Could you help me understand what the problem here is?

My guess is that the binary stops right after being started by the Service Control Manager because your `data_dir` is not a valid Windows path.

```ps
c:\Dev\vector (windows-service){hg}
λ target\debug\vector.exe service install --config "C:\Dev\vector.toml"
Aug 31 16:27:15.739  INFO vector::cli: Log level "info" is enabled.

c:\Dev\vector (windows-service){hg}
λ target\debug\vector.exe service start
Aug 31 16:27:44.043  INFO vector::cli: Log level "info" is enabled.

PS C:\WINDOWS\system32> Get-Service -Name "vector"

Status   Name               DisplayName
------   ----               -----------
Running  vector             Vector Service
```

src/vector_windows.rs
oktal: Can you please provide more details on what you tried ? Did you try to uninstall the service while it was running ?

Hoverbear:
I see the discussion on the windows-service PR is continuing. :)

oktal:
Hi @Hoverbear 

Currently, this PR is blocked by the windows-service PR. However, the windows-service PR is only needed for the `service status` sub-command.

I propose removing this extra sub-command to avoid adding extra delay to this Pull Request (I have to resolve conflicts with `main.rs` every time I update the PR).

Getting the status of the service can easily be done with the `Get-Service` PowerShell cmdlet anyway.

What do you think ?

Hoverbear:
I think that sounds reasonable, Oktal! Could you stub it out with a TODO tracking the issue? We can re-enable it after.

oktal:
Hi @Hoverbear 

I updated the PR and completely removed the `service status` sub-command. After some thinking, I did not want to keep commented-out code everywhere in the code.

I think we do not need this command, after all. `Get-Service` from PowerShell can do the exact same thing.

I merged master back and had to resolve conflicts. If this PR hangs for some time, I get conflicts in `main.rs` (resolving is tricky sometimes).

Hoverbear:
@oktal Okay! Great!

So based on my earlier reviews I think this is basically done. I want to take some time to QA it and add docs. Do you want me to take this to the finish line for you? This way you don't need to worry about merge conflicts.

oktal:
@Hoverbear yes, that would be great !

Let me know how it goes :)

Hoverbear:
@oktal Thanks so much! :) I am so excited to support windows better.

If you want to ever chat with us about development we'd love to invite you to our community space: https://discord.gg/yEdpAUb

Hoverbear:
@oktal Can I ask you to resolve the DCO before you leave this to me? :)

Hoverbear:
Impacted by #4025

Hoverbear:
I added some events  in https://github.com/oktal/vector/pull/1

src/app.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub fn run(self) {
+        let mut rt = self.runtime;
+
+        let graceful_crash = self.config.graceful_crash;
+        let mut topology = self.config.topology;
+
+        let config_paths = self.config.config_paths;
+
+        #[cfg(feature = "api")]
+        // assigned to prevent the API terminating when falling out of scope
```
Hoverbear, line 187: (Not a change request) A rust note for you, if you called `_api` as `_` instead, this wouldn't work since Rust would automatically drop it. "Load bearing `_`"

src/app.rs
```diff
@@ -0,0 +1,266 @@
...
+                SignalTo::Shutdown => {
+                    emit!(VectorStopped);
+                    tokio::select! {
+                    _ = topology.stop().compat() => (), // Graceful shutdown finished
+                    _ = signals.next() => {
+                        // It is highly unlikely that this event will exit from topology.
+                        emit!(VectorQuit);
+                        // Dropping the shutdown future will immediately shut the server down
+                    }
+                }
```
Hoverbear, line 255: This passes Rustfmt??? Wild.

src/service.rs
```diff
@@ -0,0 +1,212 @@
...
+        ),
+        Err(errs) => {
+            handle_config_errors(errs);
+            None
+        }
+    }
+}
+
+#[allow(unused)]
+mod built_info {
```
Hoverbear, line 210: This seems unnecessary.

Hoverbear: https://github.com/oktal/vector/pull/2

Hoverbear:
https://github.com/oktal/vector/pull/2 should fix shutdown tests

Hoverbear:
It looks like the CI failures on the shutdowns are real so I'm going to need to explore them more. :)

Hoverbear (approved):

Hoverbear:
Hm, I think this is hanging:

image
image

src/app.rs
```diff
@@ -211,6 +211,8 @@ impl Application {
                 tokio::select! {
                 Some(signal) = signals.next() => {
                     if signal == SignalTo::Reload {
+                        // Reload paths
+                        config_paths = config::process_paths(&opts.config_paths).unwrap_or(config_paths);
```
ktff, line 215: All of these changes are in the master, it was probably lost in a bad merge.

ktff (changes requested):
I'll advise to recheck the mergers with master, it seems some code was lost in the main. Example of config path reload code.

src/app.rs
Hoverbear: @oktal and I were just ironing those out. :)

ktff (approved):

Hoverbear:
Yayyy! Way to go, @oktal !

Epicism:
Hello,

I had a question - is there any way to set the Startup Type to "Automatic" or "Automatic (Delayed Start)"?

PR #2897, opened 2020-06-24
enhancement: Add SASL to kafka

fanatid:
Previous attempt -- #1276 

Should we add integration tests for this?

One important thing here is that `librdkafka` silently ignore build flags (like `--enable-ssl` / `--enable-gssapi`), so clients can get not working binaries (see #2834 as example). Possible solution that `librdkafka` will throw error on `./configure` step, but we too far from this. Developers and contributors also can have not working sasl right now if not install `cyrus-sasl-devel` (or similar package).

Edit: `librdkafka` master branch fail on `./configure` if thing is not present, but this not included in any release yet (https://github.com/edenhill/librdkafka/commit/d38de95b210022736f9d6df2eaf050386ea83ec0)

### DIFF
--- a/.meta/_partials/fields/_kafka_options.toml.erb
+++ b/.meta/_partials/fields/_kafka_options.toml.erb
@@ -26,3 +26,39 @@ examples = [
 description = """\
 The options and their values. Accepts `string` values.
 """
+
+[<%= namespace %>.sasl]
+type = "table"
+category = "SASL"
+common = false
+description = "Options for SASL/SCRAM authentication support."
+
+[<%= namespace %>.sasl.children.enabled]
+type = "bool"
+common = true
+description = "Enable SASL/SCRAM authentication to the remote."
+
+[<%= namespace %>.sasl.children.username]
+type = "string"
+common = true
+examples = ["username"]
+description = "The Kafka SASL/SCRAM authentication username."
+
+[<%= namespace %>.sasl.children.password]
+type = "string"
+common = true
+examples = ["password"]
+description = "The Kafka SASL/SCRAM authentication password."
+
+[<%= namespace %>.sasl.children.mechanism]
+type = "string"
+common = true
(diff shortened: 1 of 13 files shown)


scripts/environment/definition.nix
```diff
@@ -38,53 +38,54 @@ scope@{ pkgs ? import <nixpkgs> {} }:
 
   packages = with pkgs; [
     # Core CLI tools
-    dnsutils
-    curl
     bash
-    nix
-    direnv
+    bashInteractive
```
bruceg, line 42: Is this just sorting the entries in this file?

bruceg:
> Should we add integration tests for this?

If the kafka docker image we test against supports SASL authentication, then we should test against that, yes.

scripts/environment/definition.nix
fanatid: Yes, just sorting dependencies and adding `cyrus_sasl`.

Hoverbear: Looks good to me!

.meta/_partials/fields/_kafka_options.toml.erb
```diff
@@ -26,3 +26,39 @@ examples = [
...
 The options and their values. Accepts `string` values.
 """
+
+[<%= namespace %>.sasl]
+type = "table"
+category = "SASL"
+common = false
+description = "Options for SASL/SCRAM authentication support."
+
+[<%= namespace %>.sasl.children.enabled]
```
Hoverbear, line 36: Do you think we can more intelligently detect if it needs to enabled? Like if username/password is present?

fanatid: Yes, I think better detect sasl by `username`/`password`/`mechanism`. But I added `enabled` here for consistency with `tls` option.

fanatid:
Integration test added, ready for review.

bruceg (approved):

PR #2899, opened 2020-06-24
feat(platforms): Enable Nix based builds

Hoverbear:


Cargo.toml
```diff
@@ -441,3 +437,6 @@ required-features = ["transforms-wasm", "transforms-lua"]
 
 [patch.'https://github.com/tower-rs/tower']
 tower-layer = "0.3"
+
+[patch.crates-io]
+leveldb = { git = "https://github.com/timberio/leveldb/", branch = "optional-snappy-build" }
```
fanatid, line 442: Just curious, why `rev` added to `tower` and `rlua`? As I understand `leveldb` updated and `snappy` moved to optional feature, but why not just remove it completely? Is it used somewhere? https://github.com/timberio/leveldb/commit/827aee5f23e258b66b53347e113564697773a8cc

PR #2901, opened 2020-06-24
docs: Adding spaces to condition docs

bill-bateman:
Noticed I messed up the spaces in the docs I added for #2745 

In the website there wasn't any space between sentences, e.g.

> Check whether a fields contents exactly matches the value specified.This may be a single string or a list of strings, in which case this evaluates to true if any of the list matches.


bill-bateman:
hmm looks like `make check-generate` failed... thought it was all good but let me double-check and fix that...

PR #2905, opened 2020-06-25
chore: reduce http:0.1 usage

fanatid:
This pull request:

- [x] rename `http02` => `http` and `http` => `http01`
- [x] reduce `http01` usage (`http::Uri`)
- [x] remove duplicated code

`http01` still used in few places, and can be removed if:

- [ ] `reqwest` update (can do this in next PR) (dependabot PR: https://github.com/timberio/vector/pull/2782)
- [ ] compare `StatusCode` as `u16` instead object for `shiplift` dependency (`docker` source)

### DIFF
--- a/src/sinks/gcp/cloud_storage.rs
+++ b/src/sinks/gcp/cloud_storage.rs
@@ -21,7 +21,7 @@ use bytes::Bytes;
 use chrono::Utc;
 use futures::TryFutureExt;
 use futures01::{stream::iter_ok, Future, Sink};
-use http::{Method, StatusCode, Uri};
+use http::{StatusCode, Uri};
 use hyper::{
     header::{HeaderName, HeaderValue},
     Body, Request, Response,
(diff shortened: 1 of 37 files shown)


src/sinks/gcp/cloud_storage.rs
```diff
@@ -226,30 +224,19 @@ impl GcsSink {
         Ok(Box::new(sink))
     }
 
-    fn healthcheck(&mut self) -> crate::Result<Healthcheck> {
-        let builder = Request::builder()
-            .method(Method::HEAD)
-            .uri(self.base_url.parse::<Uri>()?);
+    async fn healthcheck(mut self) -> crate::Result<()> {
+        let uri = self.base_url.parse::<Uri>()?;
+        let mut request = http::Request::get(uri).body(Body::empty())?;
```
bruceg, line 229: This appears to have turned a HEAD request into a GET.

fanatid: My bad. Thanks!

lukesteensen:
This is exciting! :tada:

PR #2930, opened 2020-07-01
chore(kafka sink): disable sasl feature

fanatid:
Revert part of https://github.com/timberio/vector/pull/2915.

`sasl` is problematic right now. Problems with building statically linked, problems with using new not published vendored feature in rdkafka-sys (win32 not supported yet). `sasl` still will be available, but users will need build binaries themselves.

This PR close #2928 and #2929. After landing we need create issue for tracking.

### DIFF
--- a/.meta/_partials/fields/_kafka_options.toml.erb
+++ b/.meta/_partials/fields/_kafka_options.toml.erb
@@ -30,29 +30,24 @@ The options and their values. Accepts `string` values.
 [<%= namespace %>.sasl]
 type = "table"
 category = "SASL"
-common = false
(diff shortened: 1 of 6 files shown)


MOZGIII:
How about merging this? It'll unblock the k8s E2E tests at CI.
We still don't approval from @Hoverbear. I think we can merge this, and open another follow-up PR is this ends up problematic.

Hoverbear:
I thought I approved this! Sorry.

Hoverbear:
Do you think we could re-enable this anytime soon? I was hoping to announce SASL in 0.10.0 :(

PR #2955, opened 2020-07-05
enhancement(syslog source): Accept `octet-counting` encoded messages

### LINKED ISSUE #2690: feature request: syslog does not allow for RFC6587 when set to "TCP"
We've got devices that will follow the [RFC6587](https://tools.ietf.org/html/rfc6587) specification. They send logs using Octet Counting as specified in 3.4.1. of that same RFC. It improves on syslog to allow for all characters to be sent, so you can send logs with `\n` characters if you need to, like sending lists of things or stack traces.

Logs sent by those devices look like this in raw text mode:
```
136 <14>1 2020-05-26T18:36:57.000000+01:00 zc4430kno-4234A71FAD01 kernel - - - [  128.604348] bcmgenet f0480000.ethernet eth0: Link is Down
259 <14>1 2020-05-26T18:36:57.000000+01:00 zc4430kno-4234A71FAD01 syslog 1952 - - 200526-17:36:57.976506 [mod=evo-platform, lvl=INFO] [tid=1992] SPlatformEthernetD  INFO SPlatformEtherne : platform_ethernet_driver_netlink_callback:network eth0 link is 'DOWN'
256 <14>1 2020-05-26T18:36:57.000000+01:00 zc4430kno-4234A71FAD01 syslog 1952 - - 200526-17:36:57.977317 [mod=evo-platform, lvl=INFO] [tid=1992] SPlatformEthernetD  INFO SProxy :  S_proxy_network_interface_event_handler : Update MAC to 44:34:a7:1f:ad:01
282 <14>1 2020-05-26T18:36:57.000000+01:00 zc4430kno-4234A71FAD01 syslog 1952 - - 200526-17:36:57.977466 [mod=evo-devconfig, lvl=INFO] [tid=1992] SPlatformEthernetD  INFO RemoteDeviceConfig : onChanged_callback: Something happened to config: clientDevice.DeviceId value: 4234a71fad01
127 <30>1 2020-05-26T18:36:57.000000+01:00 zc4430kno-4234A71FAD01 WPEFramework 1669 - - Interface: eth0 changed state new state: 2
365 <14>1 2020-05-26T18:36:57.000000+01:00 zc4430kno-4234A71FAD01 syslog 1952 - - 200526-17:36:57.977623 [mod=evo-platform, lvl=INFO] [tid=1992] SPlatformEthernetD  INFO SPlatformEtherne : platform_ethernet_driver_debug_print_interfaces: 
interface name='lo' 
link='UP' 
ipaddr='127.0.0.1' 
netmask='255.0.0.0' 
mac='00:00:00:00:00:00' 
gateway='' 
DNS='127.0.0.1'
```
can you see on the last log message as a multiline log might be useful?

right now these logs fail to be parsed by the sink, and are just processed as 
```
{"host":"89.152.250.225","message":"138 <30>1 2020-05-26T17:10:04.000000+01:00 zc4430kno-4234A71FAD01 WPEFramework 1616 - - 4.46.0.m.lab_development NCL ocdm_cb_thread();Looping","source_type":"socket","timestamp":"2020-05-26T16:10:04.778532001Z"}
```
as if they were from the `socket` sink. 

It'd be nice to also support this RFC as well as the other two already supported by the sink. It involves:
1. not using the `\n` delimiter as signal of a message ending, and
2. using the byte count given at the top to know how many bytes to load into a message before applying parsing.

As of the low level details, the byte count is sent on a separate TCP packet, 
image

and then after an ACK the actual message is sent.
image

and that can be useful too for pre-allocating some buffer needed for the entire message.

ktff:
Closes #2690

### DIFF
--- a/src/sources/syslog.rs
+++ b/src/sources/syslog.rs
@@ -9,18 +9,18 @@ use crate::{
     tls::{MaybeTlsSettings, TlsConfig},
     topology::config::{DataType, GlobalOptions, SourceConfig, SourceDescription},
 };
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use chrono::{Datelike, Utc};
 use derive_is_enum_variant::is_enum_variant;
 use futures01::{future, sync::mpsc, Future, Sink, Stream};
 use serde::{Deserialize, Serialize};
 use std::net::SocketAddr;
 #[cfg(unix)]
-use std::path::PathBuf;
+use std::{io, path::PathBuf};
 use syslog_loose::{self, IncompleteDate, Message, ProcId, Protocol};
 use tokio01::{
     self,
-    codec::{BytesCodec, LinesCodec},
+    codec::{BytesCodec, Decoder, LinesCodec},
     net::{UdpFramed, UdpSocket},
 };
 use tracing::field;
@@ -99,7 +99,7 @@ impl SourceConfig for SyslogConfig {
             #[cfg(unix)]
             Mode::Unix { path } => Ok(build_unix_source(
                 path,
-                self.max_length,
+                SyslogDecoder::new(self.max_length),
                 host_key,
                 shutdown,
                 out,
@@ -124,10 +124,10 @@ struct SyslogTcpSource {
 }
 
 impl TcpSource for SyslogTcpSource {
-    type Decoder = LinesCodec;
+    type Decoder = SyslogDecoder;
 
     fn decoder(&self) -> Self::Decoder {
-        LinesCodec::new_with_max_length(self.max_length)
+        SyslogDecoder::new(self.max_length)
     }
 
     fn build_event(&self, frame: String, host: Bytes) -> Option<Event> {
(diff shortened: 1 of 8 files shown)


src/sources/syslog.rs
```diff
@@ -142,6 +143,106 @@ impl TcpSource for SyslogTcpSource {
...
+        // Encoding scheme:
+        //
+        // len ' ' data
+        // |    |  | len number of bytes that contain syslog message
+        // |    |
+        // |    | Separating whitespace
+        // |
+        // | ASCII decimal number of unknown length
+
+        if let Some((i, _)) = src.iter().enumerate().find(|&(_, &b)| b == b' ') {
```
lukesteensen, line 169: You should be able to use `position` instead of `enumerate` + `find`

src/sources/syslog.rs
```diff
@@ -142,6 +143,106 @@ impl TcpSource for SyslogTcpSource {
...
+    }
+
+    /// None if this is not octet counting encoded
+    fn checked_decode(&self, src: &mut BytesMut) -> Option<Result<Option<String>, io::Error>> {
+        if let Some(&first_byte) = src.get(0) {
+            if 49 <= first_byte && first_byte <= 57 {
+                // First character is non zero number so we can assume that
+                // octet count framing is used.
+                trace!("Octet counting encoded event detected.");
+                return Some(self.octet_decode(src));
```
lukesteensen, line 215: This makes me a little bit nervous. We try not to make too many assumptions about the syslog "format", and I'm not sure if there's any variant we accept that start with a number.

I also don't love the fact that this could theoretically change modes on a per-read basis. It's very clever and the UX of auto-detecting the format is great, I just want to be sure that it won't lead to confusing or incorrect behavior. I don't think the alternative of adding a configuration option is that onerous.

tests/syslog.rs
```diff
@@ -214,6 +215,75 @@ fn test_unix_stream_syslog() {
...
+            tls: None,
+        }),
+    );
+    config.add_sink("out", &["in"], tcp_json_sink(out_addr.to_string()));
+
+    let mut rt = runtime();
+
+    let output_lines = receive(&out_addr);
+
+    let (topology, _crash) = topology::start(config, &mut rt, false).unwrap();
```
lukesteensen, line 240: This will need an update after #2964

src/sources/syslog.rs
ktff: The [`rfc6587`](https://tools.ietf.org/html/rfc6587) isn't quite a standard, but more of an observation, so yes some caution is needed. 

In code we always first match on `<` which is used as a start in both `rfc5424` and `rfc3164` so at the moment we don't have other formats that start with a number. 

> change modes on a per-read basis

We can detect format on first message and assume it on the rest of the stream. It has been noted in `rfc6587` that changing format on read per read basis is not a good practice, but has been observed in legacy systems.

It is not a strong guarantee, but again the `rfc6587` noted that:
> It can be assumed that octet-counting framing is used if a syslog frame starts with a digit.

Having configuration option would mean that user:
1. Has to know which formatting they will have.
2. If they have sources with different formatting they will need to have two `syslog` sources, one for each formatting.

Going by `rfc` only, I would say it's fine to auto detect.

lukesteensen: That makes sense to me. If the RFC says we can do it, that's good enough :smile:

PR #2902, opened 2020-06-25
chore: move topology building into tokio context

lukesteensen:

PR #2964, opened 2020-07-06
chore: move topology building into tokio context, take 2

lukesteensen:
Outside of `src/topology`, these changes are almost entirely mechanical. Reading each commit individually should make that clearer.

We can't go all the way to `#[tokio::test]` until we no longer depend on tokio-compat, which is why this isn't quite as red as #2902. Otherwise, this still helps unblock making `build` methods async and tackles a little bit of the code organization tech debt in `src/main.rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4703,11 +4703,11 @@ dependencies = [
 
 [[package]]
 name = "tokio-executor"
-version = "0.1.9"
+version = "0.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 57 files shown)


fanatid:
For fixing CI `scripts/check-component-features.sh` should be changed:
https://github.com/timberio/vector/blob/35330c77049c928adfaad755743aca4c2bcd47f3/scripts/check-component-features.sh#L28
```
if (echo "$COMPONENTS" | grep -E -v "(Log level|^(Sources:|Transforms:|Sinks:|)$)" >/dev/null); then
```

Issue #2970, opened 2020-07-07
Remove Nix based environments


Hoverbear:
After some talks with the team we're looking to remove the `nix` environment and opt for something a bit more boring and practical.

I'm not really sure we need to exhaustive details about why. This PR entails two things:

* Amend https://github.com/timberio/vector/blob/master/rfcs/2020-05-25-2685-dev-workflow-simplification.md with a different environment, remove the nix stuff.
* Implement that change.

My recommended action is we make the base environment image `ubuntu:20.04` or `alpine` since they are modern and familiar with a good package set. We can then use centos7 etc to build our releases for older glibcs.

Hoverbear:
@lukesteensen Can you give approval to do this? Do you have an opinion on which baseimage?

lukesteensen:
Sure, approved and I'd probably go with Ubuntu so we have more flexibility with glibc vs musl.

PR #2971, opened 2020-07-07
chore(operations): Make environment use ubuntu 20.04

Hoverbear:
Addresses #2970


Hoverbear:
Just wrapping up some final knobs on CI.

PR #2975, opened 2020-07-08
chore(operations): Build debian docker image from the archive instead of package

MOZGIII:
The motivation for this change is that currently we, either way, unpack `.deb` at a builder image and only use the binary and configs in the final image.
We can eliminate the extra step of building a `.deb` file and just use the `.tar.gz` archive, just like we do with `alpine` build. This saves some time when iterating on things that require docker image rebuilds (E2E tests for instance).

Some extra work might be required to make nightly and release builds work, CC @Hoverbear.

### DIFF
--- a/distribution/docker/debian/Dockerfile
+++ b/distribution/docker/debian/Dockerfile
@@ -1,14 +1,19 @@
 FROM debian:buster-slim AS builder
 
-COPY vector-*.deb ./
-RUN dpkg -i vector-$(dpkg --print-architecture).deb
+WORKDIR /vector
+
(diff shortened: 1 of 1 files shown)


MOZGIII:
If we want correctness - I'm fine with that. But then we should install `.deb` in the final image, instead of doing it at the builder and copying the binary afterward. Otherwise, it's effectively the same as using a binary from the archive, but with extra steps.

Hoverbear:
Yeah -- Agree!

MOZGIII:
Sounds good, I'm closing this one then.

Hoverbear:
Thanks Mike. :)

PR #2980, opened 2020-07-08
chore(deps): Fixup dependabot config

Hoverbear:
Fix #2979 , we talked about this. I just forgot to actually push the code...

PR #2990, opened 2020-07-08
chore(deps): Fix dependency label to be longer again.

Hoverbear:
Continuing on #2781 #2980 , I didn't like the shortened label names, this reverts us to the longer label names and the short PR prefixes.


Hoverbear:
Oh, heck, I get it... Ok. We have to correct the dependabot label then.

Hoverbear:
Merging as this is trivial.

PR #2996, opened 2020-07-09
fix(operations): Fix the cmark-gfm installation at builder image

MOZGIII:


MOZGIII:
> I'm kind of surprised the `rm` is necessary.

Yeah, I'm not sure why it's required. I'm not very into centos and fedora, so I don't have enough context on how it works there. It's what stackoverflow suggests, and I'm pretty sure I tried without `rm` and it didn't work. I'll double-check though.

MOZGIII:
Works without `rm`! :thinking:

PR #3001, opened 2020-07-09
chore(operations): Remove CircleCI

Hoverbear:
Removes all the circleCI stuff. Based on #2869 I think we can just delete it all now? @binarylogic can you confirm?

Soft blocks #2961 since the two pipelines might interact during the publish phase.


Hoverbear:
@binarylogic I noticed `sync-install` is not part of our GA  workflow, should that be migrated?


https://github.com/timberio/vector/blob/863923b359afeb61742f88284e844f395f52afa2/Makefile#L456-L458
https://github.com/timberio/vector/blob/f758963ad28a698a35d772ea8d0cea3d9c2ef46c/.circleci/config.yml#L656-L664

PR #3009, opened 2020-07-10
chore(security): remove outdated security advisories

fanatid:
Only remove ignored security advisories which not relevant anymore with current vector dependencies.
Related with https://github.com/timberio/vector/issues/2464


binarylogic (approved):
Thanks

PR #3010, opened 2020-07-10
chore(sinks): Build `HttpClient` once

ktff:
Ref #2237 

Moves building of `HttpClient` from two locations, one for sink and one for healthcheck, to `build` method.

There are two sinks that had different behavior for sink and healthcheck:
- `honeycomb`
- `logdna`

In both cases healthchecks were using `TLS`, but sinks weren't. This PR changes that to not using `TLS`.

### DIFF
--- a/src/sinks/elasticsearch.rs
+++ b/src/sinks/elasticsearch.rs
@@ -1,5 +1,4 @@
 use crate::{
-    dns::Resolver,
     emit,
     event::Event,
     internal_events::{ElasticSearchEventReceived, ElasticSearchMissingKeys},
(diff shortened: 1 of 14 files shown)


src/sinks/elasticsearch.rs
```diff
@@ -571,6 +570,7 @@ mod tests {
 mod integration_tests {
     use super::*;
     use crate::{
+        dns::Resolver,
```
lukesteensen, line 573: Is this needed?

ktff: Yes, we are using it here
https://github.com/timberio/vector/blob/5b610a312eaf67f05fc546d3f66ede679709ad69/src/sinks/elasticsearch.rs#L810
Before it was imported through the above `use super::*;` from which we removed it.

PR #3032, opened 2020-07-11
enhancement(sinks): add compression level

fanatid:
This pull request adds compression level to `Compression` enum.

I implemented custom deserializer, so the `compression` option is backward compatible (like `encoding` option). Level valid only for interval `[0; 9]`, keys like `none`, `fast`, `default`, `best` also supported. The default value is `6`.


fanatid:
Another idea set compression level in codec option, like `gzip-8` / `gzip-fast` / `gzip-best` / `gzip` :thinking:

src/sinks/util/buffer/compression.rs
```diff
@@ -0,0 +1,144 @@
...
+pub enum Compression {
+    #[derivative(Default)]
+    None,
+    Gzip(usize),
+}
+
+impl Compression {
+    pub fn default_gzip() -> Compression {
+        Compression::Gzip(6)
+    }
```
bruceg, line 15: Consider marking functions like this as `const` to allow them to be used as initializers for static data.

src/sinks/util/buffer/compression.rs
```diff
@@ -0,0 +1,144 @@
...
+            Compression::Gzip(level) => {
+                rusoto_core::encoding::ContentEncoding::Gzip(None, level as u32)
+            }
+        }
+    }
+}
+
+impl<'de> de::Deserialize<'de> for Compression {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
```
bruceg, line 46: I wonder if this couldn't be handled automatically by serde with a flattened enum, something like the following (untested):
```
#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
#[serde(rename_all = "lowercase")]
enum CompressionName {
    None,
    Gzip,
}

#[derive(Clone, Debug, Deserialize, Serialize)]
#[serde(untagged)]
enum CompressionConfig {
    Name(String),
    NameLevel { codec: String, level: String },
}

impl TryInto<Compression> for CompressionConfig { … }
```

.meta/_partials/fields/_compression_options.toml.erb
```diff
@@ -3,17 +3,32 @@
...
+type = "table"
+description = """\
+Configures the compression specific sink behavior.\
+"""
+
+[<%= namespace %>.compression.children.codec]
 <%= options.to_toml %>
 
-[<%= namespace %>.compression.enum]
+[<%= namespace %>.compression.children.codec.enum]
```
bruceg, line 25: Given that we already use the term _codec_ to refer to things like JSON encoding, I would recommend something other than that here. I suggest either _algorithm_ or _method_ instead (probably the latter since the former is harder to type).

bruceg (changes requested):
We should decide on a preference between hybrid names like `gzip-1` or `zstd-3` or a separate option like `compression.level = X`. While the former makes for an abbreviated configuration file, the latter can accommodate things other than a level value if desired, and is less of an overloading of the compression algorithm name. My preference would be for what is in this PR now, though this may require either manual deserialization or a breaking config change.

.meta/_partials/fields/_compression_options.toml.erb
binarylogic: Agree with this. I prefer `algorithm`.

.meta/_partials/fields/_compression_options.toml.erb
```diff
@@ -3,17 +3,32 @@
...
 <%= enum.to_toml %>
+
+[<%= namespace %>.compression.children.level]
+type = "uint"
+default = 6
+examples = [0, 1, 6, 9]
```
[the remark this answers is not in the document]
binarylogic: Agree. We can just remove the default and note each algorithm's default in the description. Also, the `type` should be changed to `any` since it allows for integers and strings.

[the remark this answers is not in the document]
jszwedko: It'd be nice if we could present unions in the docs so that the type could be something like `string | uint8`.

src/sinks/util/buffer/compression.rs
fanatid: I think it can, but I still will need to verify the decoded config in `TryInto`... so I think it's fine to go with custom Deserializer

fanatid:
@binarylogic how I can generate `config/vector.spec.toml`?

binarylogic:
No need, just update the `.toml` files and it'll update automatically when we release.

fanatid:
bors t

binarylogic:
@kirillt what's the status on this? Are you blocked with anything?

kirillt:
@fanatid

fanatid:
@binarylogic pull request is ready for review (except that this also requires new docs files update for affected sinks).

binarylogic:
@bruceg mind giving this another review?

src/sinks/util/buffer/compression.rs
```diff
@@ -0,0 +1,154 @@
...
+        let mut map = serializer.serialize_map(None)?;
+        match self {
+            Compression::None => map.serialize_entry("algorithm", "none")?,
+            Compression::Gzip(level) => {
+                map.serialize_entry("algorithm", "gzip")?;
+                match level {
+                    0 => map.serialize_entry("level", "none")?,
+                    1 => map.serialize_entry("level", "fast")?,
+                    6 => map.serialize_entry("level", "default")?,
+                    9 => map.serialize_entry("level", "fast")?,
```
bruceg, line 147:
```suggestion
                    9 => map.serialize_entry("level", "best")?,
```

src/sinks/util/buffer/compression.rs
```diff
@@ -0,0 +1,154 @@
...
+pub enum Compression {
+    #[derivative(Default)]
+    None,
+    Gzip(usize),
+}
+
+impl Compression {
+    pub const fn default_gzip() -> Compression {
+        Compression::Gzip(6)
+    }
```
bruceg, line 15: I note that our default when converting to a rusoto parameter is the gzip default of `6`, but for compressed buffers we had been defaulting to "fast". This makes this default awkward here.

bruceg: Also, since the constant `6` is used in a number of places, it might be helpful to move it into a constant to make it more explicit.

bruceg (changes requested):
One obvious typo, and one question about the default. Otherwise looks ok.

src/sinks/util/buffer/compression.rs
fanatid: I'm not sure, but can not use values with serde? We need function for this for default value on deserialization. Once compressed buffers use another value, maybe we add `gzip_fast`/`gzip_default` instead `default_gzip`?

bruceg: Regarding the constant, I just meant to add `const GZIP_DEFAULT: u32 = 6;` or similar, and use that in place of `6`. You could use `flate2::Compression::default().level()` as the constant, except that none of those functions are marked as `const` in `flate2`.

Regarding the split default functions, I would have to see how that would work. My thinking was specifically about defaulting to 6 in the serde deserializer when just `"gzip"` is specified. Were you meaning to use `gzip_fast` there?

bruceg: Heh, they just did the `const` marking, it just isn't released yet: https://github.com/rust-lang/flate2-rs/commit/0a631bbd19873c50ab87a5d120dcec28725631a9

fanatid: @bruceg I added constants and returned back fast compression for `Buffer`.

StephenWakely (approved):

binarylogic:
@fanatid do we need to update docs for this?

fanatid:
:scream: I completely forgot about this! We need. Will fix it.

bruceg (approved):

jamtur01:
@binarylogic @fanatid All good to merge?

binarylogic (approved):

Issue #3033, opened 2020-07-11
Flaky `topology::config::watcher::tests::multi_file_update`


ktff:
Ref #2978 

Tests if we detect multiple updates in a row to a config file one after another.
Common scenario for update - save - observe loop.

Flaky on mac.
https://github.com/timberio/vector/pull/3010/checks?check_run_id=857600333

Probably a bug.

jszwedko:
Closing since this test was deleted and evidently doesn't appear to have been needed.

PR #3034, opened 2020-07-11
chore(tests): Delete `topology::config::watcher::tests::multi_file_update`

ktff:
Ref.  #3033


ktff:
@lukesteensen I'll move forward with merging these test deleting PRs.

PR #3037, opened 2020-07-11
chore(tests): Split `Receiver::cancel` out of `Receiver::wait` 

ktff:
Splits `crate::test_util::Receiver::wait` method into `wait` and `cancel`. Somel of the usages of this method were having a race as they were expecting that the method would `wait` for the `tcp` channel to close, which wasn't the case. Instead the method was canceling the channel which caused a race. Only test that needs the `cancel` was `tcp_stream`. 

Similar how in #2862 `crate::test_util::CountReceiver::wait` was split.

Ref. #2978 
Fixes [`test_max_size_resume`](https://github.com/timberio/vector/issues/2978#issuecomment-657133377)

### DIFF
--- a/src/sinks/socket.rs
+++ b/src/sinks/socket.rs
@@ -144,7 +144,7 @@ mod test {
         let context = SinkContext::new_test();
         let (sink, _healthcheck) = config.build(context).unwrap();
 
-        let receiver = receive(&addr);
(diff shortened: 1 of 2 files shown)


ktff:
Nope, it's not that.

PR #3045, opened 2020-07-12
chore(sinks): Build `aws` sinks clients only once

ktff:
Ref. #2237 

Moves building of various `aws` sink clients from two locations, one for sink and one for healthcheck, to build method.


ktff:
Yep

PR #3059, opened 2020-07-13
chore(kafka source): remove owning_ref

fanatid:
Close #2972 

- update `rdkafka` crate
- remove `owning_ref` crate
- remove extra clones
- add more async code
- fix `kafka-integration-tests` feature (before part of tests was not compiled for integration tests)

### DIFF
--- a/src/sources/kafka.rs
+++ b/src/sources/kafka.rs
@@ -162,7 +162,8 @@ fn kafka_source(
                     }
                 }
             })
-            // Not work =\
+            // Try `forward` after removing old futures.
+            // Error: implementation of `futures_core::stream::Stream` is not general enough
             // .forward(
             //     out.sink_compat()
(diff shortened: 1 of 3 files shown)


src/sources/kafka.rs
```diff
@@ -97,78 +96,95 @@ impl SourceConfig for KafkaSourceConfig {
...
+                            Ok(event)
+                        }
+                    }
+                }
+            })
+            // Not work =\
+            // .forward(
+            //     out.sink_compat()
+            //         .sink_map_err(|e| error!(message = "Error sending to sink", error = ?e)),
+            // )
```
bruceg, line 169: This looks like a useful note to future developers, but it's missing a "why". For example, "We can't use `.forward` here because …"

fanatid: Yeah, totally agree. Added such a comment.

PR #3065, opened 2020-07-14
chore(transforms): update runtime_transform

fanatid:
Update weak select method in internal StreamExt. Will remove `take_until` from this trait in future updates.


src/stream.rs
```diff
@@ -160,36 +122,91 @@ where
...
-                &mut self.stream1 as &mut dyn Stream<Item = _, Error = _>,
-            )
-        } else {
-            (
-                &mut self.stream1 as &mut dyn Stream<Item = _, Error = _>,
-                &mut self.stream2 as &mut dyn Stream<Item = _, Error = _>,
-            )
-        };
-        self.flag = !self.flag;
+mod new_futures {
```
bruceg, line 127: Why is this in an internal module? It's not wrong, just wondering why the additional cognitive overhead.

src/stream.rs
```diff
@@ -160,36 +122,91 @@ where
...
+    impl<St1, St2> Stream for SelectWeak<St1, St2>
+    where
+        St1: Stream,
+        St2: Stream<Item = St1::Item>,
+    {
+        type Item = St1::Item;
+
+        fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<St1::Item>> {
+            let this = self.project();
+            if !*this.flag {
```
bruceg, line 184: Does this need the negation?

src/stream.rs
fanatid: Wanted to leave global import from `futures01` and use `futures03` imports, will be removed in future on removing `take_until`.

src/stream.rs
fanatid: Good catch, this not required.

PR #3066, opened 2020-07-14
chore(sinks): add SinkConfig::build_async

### LINKED ISSUE #2687: [RFC] Make SinkConfig::build async

### LINKED ISSUE #3019: Remove async code in sync Sink::build
In #2977 and #2993 I updated `gcp` and `pulsar` sinks. These sinks for building sink (`Sink::build`) need to run async code (spawn a thread, move sink creation to Sink impl). Once #2687 will be resolved (through #2934) this code will be outdated and should be removed.

lukesteensen:
Closes #2687 
Closes #3019 

This adds a new `build_async` fn to `SinkConfig` that can be used in place of `build`. This was a smaller change than making `build` itself async and has the advantage of not requiring `async-trait` anywhere it's not needed, which improves error reporting.

We can follow up this PR and either add a default implementation of `build` or split the sync version into a subtrait. I don't have a strong opinion there so decided to punt for now.

### DIFF
--- a/src/sinks/gcp/mod.rs
+++ b/src/sinks/gcp/mod.rs
@@ -49,24 +49,14 @@ pub struct GcpAuthConfig {
 }
 
 impl GcpAuthConfig {
-    pub fn make_credentials(&self, scope: Scope) -> crate::Result<Option<GcpCredentials>> {
-        let this = self.clone();
-        // We can not run new Runtime in thread which already used by other Runtime.
-        // Without new thread we get error: `default Tokio timer already set for execution context`
-        std::thread::spawn(|| {
-            let mut rt = crate::runtime::Runtime::single_threaded().unwrap();
-            rt.block_on_std(async move {
-                let gap = std::env::var("GOOGLE_APPLICATION_CREDENTIALS").ok();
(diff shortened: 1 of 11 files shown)


src/sinks/gcp/mod.rs
```diff
@@ -49,24 +49,14 @@ pub struct GcpAuthConfig {
 }
 
 impl GcpAuthConfig {
-    pub fn make_credentials(&self, scope: Scope) -> crate::Result<Option<GcpCredentials>> {
-        let this = self.clone();
-        // We can not run new Runtime in thread which already used by other Runtime.
-        // Without new thread we get error: `default Tokio timer already set for execution context`
-        std::thread::spawn(|| {
-            let mut rt = crate::runtime::Runtime::single_threaded().unwrap();
-            rt.block_on_std(async move {
```
bruceg, line 58: That was mercifully short lived. :)

PR #3089, opened 2020-07-17
chore: Bump rust-toolchain to 1.45.0

fanatid:
https://blog.rust-lang.org/2020/07/16/Rust-1.45.0.html


tests/support/mod.rs
```diff
@@ -110,13 +110,11 @@ impl SourceConfig for MockSourceConfig {
         let event_counter = self.event_counter.clone();
         let mut recv = wrapped.lock().unwrap().take().unwrap();
         let mut shutdown = Some(shutdown);
-        let mut token = None;
         let source = future::lazy(move || {
             stream::poll_fn(move || {
                 if let Some(until) = shutdown.as_mut() {
                     match until.poll() {
-                        Ok(Async::Ready(res)) => {
-                            token = Some(res);
```
MOZGIII, line 119: Wouldn't this lead to a bug with dropping the token too soon? I suggest adding an explicit `drop` at the end - this might help with letting clippy (and people) know that it must be held on to until the end.

src/sources/util/tcp.rs
```diff
@@ -162,13 +162,12 @@ fn handle_stream(
...
 ) {
     let mut shutdown = Some(shutdown);
-    let mut token = None;
     let mut reader = FramedRead::new(socket, source.decoder());
     let handler = stream::poll_fn(move || {
         // Gracefull shutdown procedure
         if let Some(future) = shutdown.as_mut() {
             match future.poll() {
-                Ok(Async::Ready(tk)) => {
+                Ok(Async::Ready(_)) => {
```
ktff, line 170: We need to hold onto `ShutdownSignalToken` to have proper shutdown procedure. If we drop it here, we are at the mercy of the other side. Until they close the channel, the whole process will hang on it.

ktff (changes requested):
I'm guessing clippy or compiller started to complain that those token variables aren't being used, but we need to hold on to them until we finish with the whole processing so as @MOZGIII suggested a explicit `drop` is the way to go.

tests/support/mod.rs
fanatid: Returned back. Wanted to check that CI actually fail if this will be removed. (probably was need create draft PR for this)

src/sources/util/tcp.rs
```diff
@@ -190,7 +191,9 @@ fn handle_stream(
         }
 
         // Actual work
-        reader.poll()
+        let result = reader.poll();
+        let _ = token.take();
```
bruceg, line 195: Is this functionally any different than `drop(token)`?

bruceg (commented):
This is failing tests now, could the token changes be responsible?

src/sources/util/tcp.rs
fanatid: ~`drop(token)` still produces error message~ tested again and no messages :thinking: 
Will change to drop.

MOZGIII (approved):
Looks good now!

ktff (approved):
This is fine, although we can avoid the unused warning/error/cell with a `_` naming, so 
`let mut _token=None;` should work.

We can use explicit `drop` but now I realize it will require more code, to be specific, something like this
```
let poll=reader.poll();
if let Ok(Async::Ready(None))=poll{
   std::mem::drop(token.take());
}
poll
``` 

But this is not the PR for it, so I'm not so for it.

fanatid:
Nice find @ktff! I checked `_token` and it's work. Changes pushed.

Hoverbear (approved):

MOZGIII (approved):

bruceg (approved):

lukesteensen (approved):

PR #3090, opened 2020-07-17
fix(prometheus source): Allow '=' in tag values

adamncasey:
I started using vector to pull metrics from an endpoint which exposed a tag containing `=`:
```
telemetry_scrape_size_bytes_count{registry="default",content_type="text/plain; version=0.0.4"}
```

This metric wasn't supported by vector, which logged a 'Malformed' error when parsing this:
```
Jul 16 11:52:13.222 ERROR source{name=rabbitin type=prometheus}: vector::internal_events::prometheus: parsing error error=Malformed
```

The metric tag parsing code works by first splitting the string between the brackets by `,`, then each substring by `=`. Any of these splits which more than two components returned a parse error. E.g.

```
1 -> registry="default",content_type="text/plain; version=0.0.4"
2 -> ['registry="default"'], ['content_type="text/plain; version=0.0.4']"
3 -> [
    ['registry', '"default"']
],
[
    ['content_type', '"text/plain; version', '=0.0.4"']
]
```

This patch fixes this by only splitting the tag value input string on the first `=` to correctly split tag key/values for this case.

The change from `split` -> `splitn` could also in theory speed up processing of all tags,
as it won't need to check every character in the string after the first `=`.
I haven't performed any benchmarks.

Signed-off-by: Adam Casey <acasey23@bloomberg.net>

### DIFF
--- a/src/sources/prometheus/parser.rs
+++ b/src/sources/prometheus/parser.rs
@@ -105,25 +105,30 @@ fn parse_tags(input: &str) -> Result<BTreeMap<String, String>, ParserError> {
             continue;
         }
         let pair = pair.trim();
-        let parts = pair.splitn(2, '=').collect::<Vec<_>>();
-        if parts.len() != 2 {
-            return Err(ParserError::Malformed {
-                s: "expected 2 values separated by '='",
-            });
+
+        match pair.find('=') {
+            None => {
+                return Err(ParserError::Malformed {
+                    s: "expected 2 values separated by '='",
+                });
+            }
+            Some(equals) => {
+                let key = pair[..equals].trim().to_string();
+                let mut value = pair[equals + 1..].trim();
+
+                if value.starts_with('"') {
+                    value = &value[1..];
+                };
+                if value.ends_with('"') {
+                    value = &value[..value.len() - 1];
+                };
+                let value = value.trim();
+                let value = value.replace(r#"\\"#, "\\");
+                let value = value.replace(r#"\n"#, "\n");
+                let value = value.replace(r#"\""#, "\"");
+                result.insert(key, value);
+            }
         }
-        let key = parts[0].trim().to_string();
-        let mut value = parts[1].trim();
-        if value.starts_with('"') {
-            value = &value[1..];
-        };
-        if value.ends_with('"') {
-            value = &value[..value.len() - 1];
-        };
-        let value = value.trim();
-        let value = value.replace(r#"\\"#, "\\");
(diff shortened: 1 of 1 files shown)


adamncasey:
one CI failure for OSX:

```
...
yarn install
yarn install v1.22.4
[1/4] Resolving packages...
[2/4] Fetching packages...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
error An unexpected error occurred: "https://registry.yarnpkg.com/core-js/-/core-js-2.6.11.tgz: ESOCKETTIMEDOUT".
info If you think this is a bug, please open a bug report with the information provided in "/Users/runner/work/vector/vector/website/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
##[error]Process completed with exit code 1.
```

It seems unlikely this relates to my change

src/sources/prometheus/parser.rs
```diff
@@ -105,7 +105,7 @@ fn parse_tags(input: &str) -> Result<BTreeMap<String, String>, ParserError> {
             continue;
         }
         let pair = pair.trim();
-        let parts = pair.split('=').collect::<Vec<_>>();
+        let parts = pair.splitn(2, '=').collect::<Vec<_>>();
         if parts.len() != 2 {
```
bruceg, line 109: While this is the simplest change, this might be better expressed as the following to avoid creating a new Vec for what should always be exactly two elements:
```rust
match pair.find('=') {
    None => Err(…),
    Some(equals) => {
        let key = pair[..equals].trim().to_string();
        let mut value = pair[equals + 1..].trim();
        …
    }
```

bruceg (approved):

adamncasey:
I adopted your suggestion & added anoth test as I wasn't totally clear how this would work for some edge cases.

Thanks for the fast initial feedback!

bruceg:
Thanks for this fix @adamncasey. It's great to see good tests for it too.

PR #3095, opened 2020-07-17
chore: remove tokio01::codec

fanatid:
Ref. https://github.com/timberio/vector/issues/2945#issuecomment-658868665


src/sources/socket/udp.rs
```diff
@@ -39,41 +42,44 @@ pub fn udp(
...
+                .expect("failed to bind to udp listener socket");
             info!(message = "listening.", %address);
 
-            Ok(socket)
-        })
-        .and_then(move |socket| {
-            let host_key = host_key.clone();
-            // UDP processes messages per packet, where messages are separated by newline.
-            // And stretch to end of packet.
-            UdpFramed::with_decode(socket, BytesDelimitedCodec::new(b'\n'), true)
```
fanatid, line 53: [tokio_util::udp::UdpFramed](https://docs.rs/tokio-util/0.3.1/tokio_util/udp/struct.UdpFramed.html) do not have `with_decode` method, is it ok drop it?
In `tokio:0.1.22` [UdpFramed](https://docs.rs/tokio/0.1.22/tokio/net/struct.UdpFramed.html) also do not have this method in docs....

Cargo.toml
```diff
@@ -192,7 +190,7 @@ tower-test01 = { package = "tower-test", version = "0.1" }
 serde_yaml = "0.8.9"
 dirs = "2.0.2"
 tokio-test = "0.2"
-tokio = { version = "0.2", features = ["test-util"] }
+tokio = { version = "0.2", features = ["test-util", "uds"] }
```
MOZGIII, line 193: This might be redundant, just noting.

src/async_read.rs
```diff
@@ -1,80 +1,55 @@
...
-                    // to provide a mechanism for the creator to let the stream run forever,
-                    // we interpret this as "run forever".
-                    self.free = true;
-                }
-                Ok(Async::NotReady) => {}
+    fn poll_read(self: Pin<&mut Self>, cx: &mut Context, buf: &mut [u8]) -> Poll<IoResult<usize>> {
+        let this = self.project();
+        match this.until.poll(cx) {
+            Poll::Ready(_) => {
+                // TODO: Need proper shutdown
```
MOZGIII, line 45: What's the idea here?

src/internal_events/unix.rs
```diff
@@ -20,7 +20,7 @@ impl InternalEvent for UnixSocketConnectionEstablished<'_> {
 
 #[derive(Debug)]
 pub struct UnixSocketConnectionFailure<'a> {
-    pub error: std::io::Error,
+    pub error: tokio::io::Error,
```
MOZGIII, line 23: It looks like this can be generalized around error type too.

src/sources/socket/udp.rs
MOZGIII: Here's the relevant PR to tokio: https://github.com/tokio-rs/tokio/pull/1517/files#diff-9ceb10657fd3e9c67559c899f9bf59e7R182

It seems like the `UdpFramed` implementation was bugged, and `with_decode` is just a fix that should've been in the `new` fn.

MOZGIII: New impl at https://github.com/tokio-rs/tokio/blob/7e4edb89638065a041c1093b292f709d1f2190a0/tokio-util/src/udp/frame.rs#L44 looks suspicious.

Does `clear` at https://github.com/tokio-rs/tokio/blob/7e4edb89638065a041c1093b292f709d1f2190a0/tokio-util/src/udp/frame.rs#L65 means that when multiple frames arrive in a single read, all frames after the first one are dropped? This looks like a problem.

@LucioFranco, could you take a look into it if you have a minute?

src/internal_events/unix.rs
fanatid: Like `UnixSocketError`?

src/async_read.rs
fanatid: I thought about shutdown stream on vector shutdown. Removed comment and added relevant code in https://github.com/timberio/vector/pull/3095/commits/d9dc85360505e7e9182df82c712cebd837124e65

src/internal_events/unix.rs
MOZGIII: Yep.

fanatid: I'm not sure that I understand, but isn't specified type better than generalized? `UnixSocketError` was generalized, because type there can be `std::io::Error` or [tokio_util::codec::LinesCodecError](https://docs.rs/tokio-util/0.3.1/tokio_util/codec/enum.LinesCodecError.html).

MOZGIII: Oh, that's fine then, nvm! I was worried about consistency, but it doesn't matter that much really.

src/sources/socket/udp.rs
ktff: With this new `UdpFramed` `Decoder` can expect the whole payload so it can/should decode all frames. So a `Decoder` like this is required, I already wrote it in the other PR so here it is:
```

struct UdpDecoder;

impl Decoder for UdpDecoder {
    type Item = Vec<Bytes>;
    type Error = io::Error;
    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
        // UDP processes messages per payload, where messages are separated by newline
        // and stretch to end of payload.
        let mut lines = Vec::new();
        let mut decoder = BytesDelimitedCodec::new(b'\n');
        while let Some(line) = decoder.decode_eof(src)? {
            lines.push(line);
        }
        Ok(Some(lines))
    }
}
```

ktff: An alternative is to manage the read buffer ourselves.

ktff:
Can you add your name, for all of the done files in here, to #2945.

fanatid:
Added. It's can be hard to make changes while this not merged, but I think changing shutdown to future:0.3 should not be blocked and require a lot of rebases in future. I'll address all comments here tomorrow, so hope that we will able merge this quickly.

src/sources/socket/udp.rs
fanatid: Looks like pull request for this already exists https://github.com/tokio-rs/tokio/pull/1445

fanatid: Fixed in https://github.com/timberio/vector/pull/3095/commits/3cf9fb5fe6f05158b71cd5a6bc8a9128b48e9e4e
`max_length` will be fixed in another PR https://github.com/timberio/vector/issues/3117

LucioFranco: Yeah @MOZGIII seems like we didn't port the fix from 0.1 to 0.2 and the PR that @fanatid mentioned is the fix for always calling `decode_eof` for datagram based protocols. I just pinged the author so we could try to get that into a release.

src/sinks/util/tcp.rs
```diff
@@ -79,10 +86,11 @@ enum TcpSinkState {
     ResolvingDns(crate::dns::ResolverFuture),
     Connecting(MaybeTlsConnector),
     Connected(TcpOrTlsStream),
-    Backoff(Delay),
+    Backoff(Box<dyn Future<Item = (), Error = ()> + Send>),
```
MOZGIII, line 89: I wan wondering how difficult would it be to convert this part to `futures` `0.3` and user actual `Delay` here? Not that it's required for this PR, but if it's easy it would be nice to have as well.

fanatid: I think should be pretty simple when we move `trait Sink` to futures:0.3 :thinking:

src/sinks/statsd.rs
```diff
@@ -329,75 +328,65 @@ mod test {
...
+                            .sink_map_err(|e| error!("error sending event: {:?}", e)),
+                    )
+                    .await
+                    .unwrap()
+            });
+
+            // Add a delay to the write side to let the read side
+            // poll for read interest. Otherwise, this could cause
+            // a race condition in noisy environments.
+            delay_for(Duration::from_millis(100)).await;
```
MOZGIII, line 381: This delay is new, wasn't there before. I wonder if we could get rid of it if we do socket bind on the main test fn, and move it into the `tokio::spawn` instead of doing it in there? As I understand, the only possible new race here is with the socket not being bound when we start collecting messages from it.

src/sinks/util/tcp.rs
```diff
@@ -198,7 +213,8 @@ impl Sink for TcpSink {
                 //
                 // If this returns `WouldBlock` we know the connection is still
                 // valid and the write will most likely succeed.
-                match connection.get_mut().read(&mut [0u8; 1]) {
+                let stream = connection.get_mut().get_mut().get_mut().get_mut();
```
MOZGIII, line 216: I feel like this line needs a type annotation to kind of pin the exact type we're expecting here, after the chain of `get_mut`s.
It seems like we want a raw TCP socket here, because we're doing wouldblock assertions below. It looks like an important detail that can be lost.

src/sinks/statsd.rs
fanatid: This delay is not new: https://github.com/timberio/vector/pull/3095/files#diff-42fa092ff1c47ae33a665f6f0df08c4fL371-L376
I think we can move `bind` out of `spawn` and remove delay, hope that we remember what we should change in future if this test start fail :grin:

src/sinks/util/tcp.rs
```diff
@@ -225,13 +241,20 @@ impl Sink for TcpSink {
                         emit!(TcpEventSent {
                             byte_size: line.len()
                         });
+                        let line = Bytes::copy_from_slice(&line);
```
MOZGIII, line 244: This a conversion between two `bytes` crates in unfortunate. The extra copies are probably not that big of a hit to the performance, so I guess it's ok. Just pointing it out.

fanatid: Yeah. The bad part of such slow update that in the future we will need to find and rid all such conversions. Also new `bytes` have extra copying on converting between Bytes <=> BytesMut :disappointed:

src/sinks/statsd.rs
MOZGIII: The link to the source doesn't open for me :frowning_face: Which file is it?

fanatid: It's the same file (`src/sinks/statsd.rs`), was removed from lines 371-376

src/sinks/util/tcp.rs
MOZGIII: True, yet it seems kind of odd that `Bytes` were convertible to `BytesMut` before in the first place. The API seemes to have been incorrect, thus we're probably using it incorrectly... We probably should do an audit of the code as part of old `bytes` removal - I bet there are a lot of places where we can avoid copying where we could just switch to `Bytes` to `BytesMut`.

fanatid: An interesting thing, will `clippy` catch extra operations with Bytes or not. I already started improving `bytes` from BytesMut => Bytes through freeze, but this pretty hard for my current Rust level. https://github.com/tokio-rs/bytes/pull/418

src/sinks/util/tcp.rs
fanatid: Added types annotations for this and another 2 places.

src/sinks/util/unix.rs
```diff
@@ -41,27 +40,23 @@ impl UnixSinkConfig {
...
 #[derive(Debug, Snafu)]
 enum HealthcheckError {
     #[snafu(display("Connect error: {}", source))]
-    ConnectError { source: std::io::Error },
+    ConnectError { source: tokio::io::Error },
 }
 
-fn unix_healthcheck(path: PathBuf) -> Healthcheck {
-    // Lazy to avoid immediately connecting
-    let check = future::lazy(move || {
```
lukesteensen, line 58: I think the reasoning for `lazy` is still valid here unless you've verified that `connect` doesn't bind to the port until polled.

src/sinks/util/unix.rs
```diff
@@ -115,27 +112,28 @@ impl UnixSink {
...
                         let out = FramedWrite::new(stream, BytesCodec::new());
-                        UnixSinkState::Open(out)
+                        UnixSinkState::Open(CompatSink::new(out))
                     }
                 },
                 UnixSinkState::Backoff(ref mut delay) => match delay.poll() {
                     Ok(Async::NotReady) => return Ok(Async::NotReady),
-                    // Err can only occur if the tokio runtime has been shutdown or if more than 2^63 timers have been created
-                    Err(err) => unreachable!(err),
+                    Err(_) => unreachable!(),
```
lukesteensen, line 128: Did the type here change? If it's unit, I prefer to show that with `Err(()) => `.

src/sinks/util/tcp.rs
```diff
@@ -137,24 +150,24 @@ impl TcpSink {
...
                         error!(message = "unable to resolve dns.", %error);
-                        TcpSinkState::Backoff(self.next_delay())
+                        TcpSinkState::Backoff(self.next_delay01())
                     }
                 },
                 TcpSinkState::Backoff(ref mut delay) => match delay.poll() {
                     Ok(Async::NotReady) => return Ok(Async::NotReady),
                     // Err can only occur if the tokio runtime has been shutdown or if more than 2^63 timers have been created
-                    Err(err) => unreachable!(err),
+                    Err(_) => unreachable!(),
```
lukesteensen, line 170: Same here

src/sources/util/tcp.rs
```diff
@@ -172,8 +174,8 @@ fn handle_stream(
                     debug!("Start gracefull shutdown");
                     // Close our write part of TCP socket to signal the other side
                     // that it should stop writing and close the channel.
-                    if let Some(socket) = reader.get_ref().get_ref() {
-                        if let Err(error)=socket.shutdown(Shutdown::Write){
+                    if let Some(socket) = reader.get_ref().get_ref().get_ref().get_ref().get_ref() {
```
lukesteensen: A type annotation would be good here.

src/sources/syslog.rs
```diff
@@ -256,34 +259,43 @@ pub fn udp(
     let out = out.sink_map_err(|e| error!("error sending line: {:?}", e));
 
     Box::new(
-        future::lazy(move || {
-            let socket = UdpSocket::bind(&addr).expect("failed to bind to udp listener socket");
-
+        async move {
+            let socket = UdpSocket::bind(&addr)
```
lukesteensen, line 263: Same as above for `lazy` and `bind`.

src/sources/socket/udp.rs
```diff
@@ -39,41 +42,46 @@ pub fn udp(
     let out = out.sink_map_err(|e| error!("error sending event: {:?}", e));
 
     Box::new(
-        future::lazy(move || {
-            let socket = UdpSocket::bind(&address).expect("failed to bind to udp listener socket");
```
lukesteensen, line 43: Here too.

src/sources/statsd/mod.rs
```diff
@@ -44,35 +45,46 @@ fn statsd(addr: SocketAddr, shutdown: ShutdownSignal, out: mpsc::Sender<Event>)
     let out = out.sink_map_err(|e| error!("error sending metric: {:?}", e));
 
     Box::new(
-        future::lazy(move || {
-            let socket = UdpSocket::bind(&addr).expect("failed to bind to udp listener socket");
-
+        async move {
+            let socket = UdpSocket::bind(&addr)
```
lukesteensen, line 47: Here too :smile:

src/sources/util/tcp.rs
fanatid: This added in https://github.com/timberio/vector/pull/3095/commits/4e5c05c8487e540c37761522c7f7f4e763d1ea58#diff-36ca91cf07d40e3dc9a8f98f2c4a2762

src/sinks/util/unix.rs
fanatid: Good note. I actually do not understand why new futures need `lazy` function, because new futures not executed while not pulled, for example: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=1903bb7388f8d2965ee127bf8db2675c

src/sinks/util/unix.rs
fanatid: :+1: done in https://github.com/timberio/vector/pull/3095/commits/9c5708450da833faa66ce260b7e786c4b2e648dd

src/sinks/util/unix.rs
MOZGIII: I think it's because Rust can't yet pass bare async closures around, thus this `future::lazy`.

Here: https://docs.rs/futures-util/0.3.5/src/futures_util/future/lazy.rs.html#34-38, it's very similar to `poll!` macro: https://docs.rs/futures-util/0.3.5/src/futures_util/async_await/poll.rs.html#13-17

src/sinks/util/tcp.rs
MOZGIII: Ah, we probably don't want to hold this off until that arrives... :+1:

PR #3099, opened 2020-07-19
fix(cli): Cleanup `list` command

### LINKED ISSUE #3007: list command no longer clean
When starting Vector to use the `generate` command, users may be surprised to see their stdout is polluted and not valid TOML.

```
hoverbear@OBSIDIAN:/git/timberio/vector$ cargo run -- generate stdin//console
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/vector generate stdin//console`
Jul 09 16:06:21.268  INFO vector: Log level "info" is enabled.
data_dir = "/var/lib/vector/"

[sources.source0]
max_length = 102400
type = "stdin"

[sinks.sink0]
healthcheck = true
inputs = ["source0"]
type = "console"

[sinks.sink0.buffer]
type = "memory"
max_events = 500
when_full = "block"


hoverbear@OBSIDIAN:/git/timberio/vector$ cargo run -- generate stdin//console > test.toml
    Finished dev [unoptimized + debuginfo] target(s) in 0.26s
     Running `target/debug/vector generate stdin//console`


hoverbear@OBSIDIAN:/git/timberio/vector$ cat test.toml 
Jul 09 16:07:31.098  INFO vector: Log level "info" is enabled.
data_dir = "/var/lib/vector/"

[sources.source0]
max_length = 102400
type = "stdin"

[sinks.sink0]
healthcheck = true
inputs = ["source0"]
type = "console"

[sinks.sink0.buffer]
type = "memory"
max_events = 500
when_full = "block"
```

This issue requests that `vector generate stdin//console > test.toml` works again as expected.

---

Todo:

- [ ] Add test

ktff:
Closes #3007 

Changes default logging level for `list` command from `info` to `warn`.

Adds tests for `generate` and `list` commands that check absence of log lines in their output.


.github/workflows/tests.yml
```diff
@@ -104,6 +104,21 @@ jobs:
...
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - run: sudo bash scripts/environment/bootstrap-ubuntu-20.04.sh
+      - run: bash scripts/environment/prepare.sh
+      - run: make slim-builds
+      - run: make test-cli
```
Hoverbear, line 121: Do you think we could run this right after the `test` job above? It would save us a lot of CI time. :)

scripts/test-cli.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# test-cli.sh
+#
+# SUMMARY
+#
+#   Run cli tests only.
+
+cargo test --test cli --features cli-tests -- --test-threads 4
```
Hoverbear, line 10: I don't think this file is required? The makefile does this already.

Cargo.toml
```diff
@@ -427,6 +427,7 @@ pulsar-integration-tests = ["sinks-pulsar"]
 splunk-integration-tests = ["sinks-splunk_hec", "warp"]
 
 shutdown-tests = ["sources","sinks-console","sinks-prometheus","sinks-blackhole","unix","rdkafka","transforms-log_to_metric","transforms-lua"]
+cli-tests = []
```
Hoverbear, line 430: I think this is fine as a default test? I'm not sure we need to add another feature for this quite yet.

scripts/test-cli.sh
ktff: True, force of habit.

.github/workflows/tests.yml
ktff: Yea, we could bundle it with shutdown tests.

Cargo.toml
ktff: Fair enough.

PR #3100, opened 2020-07-19
chore(socket source): Update `udp`

ktff:
Ref. #2945

In `src/sources/socket/udp.rs `
 - Removes direct dependency on `tokio 0.1`
 - New `UdpFramed` wasn't usable, mostly because of crate `bytes` version mismatch, so we are now directly managing buffer for reading the data from socket. 
 - Bonus of not using `UdpFramed` is that now we accept `max_length` configuration for `udp` mod which is documented as available, but in reality wasn't implemented.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ tokio-openssl = "0.3.0"
 tokio-retry = "0.2.0"
 tokio-signal = "0.2.7"
 tokio-compat = { version = "0.1", features = ["rt-full"] }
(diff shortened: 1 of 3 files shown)


fanatid:
https://github.com/timberio/vector/pull/3095 :disappointed:

ktff:
@fanatid so there was an overlap. Ok, I'll close this pr, at least I figured out why `tcp_gracefull_shutdown` and `udp_shutdown_infinite_stream` are flaky.

PR #3106, opened 2020-07-19
chore: Remove tokio 0.1 from shutdown

ktff:
Ref.  #2945.

Removes direct dependency on `tokio01` in `src/shutdown.rs`.

### DIFF
--- a/src/shutdown.rs
+++ b/src/shutdown.rs
@@ -1,9 +1,10 @@
+use futures::{FutureExt, TryFutureExt};
 use futures01::{future, Async, Future};
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Instant;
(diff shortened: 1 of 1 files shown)


src/shutdown.rs
```diff
@@ -281,29 +282,24 @@ impl SourceShutdownCoordinator {
...
                     shutdown_force_trigger.cancel();
+                    false
                 }
-                success
-            })
-            .map_err(|_| ())
+            }
+        };
+
+        Box::pin(done.map(Ok)).compat()
```
fanatid, line 302: For method chaining `fut.boxed()` can be used `Box::pin(fut)` :slightly_smiling_face: 
https://docs.rs/futures/0.3.5/futures/future/trait.FutureExt.html#method.boxed

Issue #2917, opened 2020-06-29
Rework all the sinks that use `VecBuffer` to allow `batch.max_bytes`


bruceg:
The older `struct VecBuffer` is unable to count the byte size of events that are added to it. This prevents them from being able to enforce the `batch.max_bytes` limit. The newer `struct VecBuffer2` introduced in PR #2916 is able to track this size, but requires that the input type implements `trait Length`. For composite types, this requires a bit of thought and may not be possible without actually encoding the event.

Once all sinks are migrated over to the new `struct VecBuffer2`, it can be moved over the old name.

* [X] aws_cloudwatch_logs (PR #2916 )
* [ ] aws_kinesis_firehose
* [ ] aws_kinesis_streams
* [ ] loki
* [ ] BatchSink tests

PR #3147, opened 2020-07-21
chore(sinks): Batch buffer size rework

### LINKED ISSUE #3024: Batch buffer sizing potential bugs
The current batch buffer framework has us handling batch sizes in two steps: first the batch configuration is "unwrapped" with a set of defaults applied; then a batch buffer is created with those size parameters. This 2-step process results in the potential for several bugs due to the disconnection between parsing the configuration and actually using it.

1. Some of the buffers only use one of the sizes—specifically, `MetricBuffer` and `VecBuffer` only look at events since their size in bytes can only be calculated after the batch is encoded. Attempts to restrict such buffer's byte size should trigger an error. Currently this error is ignored. Further, with the current arrangement, it must be done separately for each use of `MetricBuffer` or `VecBuffer` instead of being handled in the buffer.
2. To simplify calculations when a default for one size is not specified, the internal default for each size is `usize::max_value()`. However, some of the buffers use the given size limit to pre-allocate the batch buffer. This leads to situations like #3006, where the maximum number of events was unset, resulting in `VecBuffer::new` trying (and obviously failing) to allocate a vector with an impossibly large number of slots.

These two steps should be combined in such a way that the batch buffers can ensure the correct size(s) are set or not to prevent future bugs and configuration confusion. This should preferably be done in such a way that the compiler can reject missing or extraneous defaults, to make such internal bugs impossible to deliver.

bruceg:
Closes #3024 

This doesn't directly merge batch settings parsing with later handling, but ties the two together using type inference, such that the correct handler is used automatically.

### DIFF
--- a/src/sinks/aws_cloudwatch_logs/mod.rs
+++ b/src/sinks/aws_cloudwatch_logs/mod.rs
@@ -7,7 +7,7 @@ use crate::{
     sinks::util::{
         encoding::{EncodingConfig, EncodingConfiguration},
         retries::{FixedRetryPolicy, RetryLogic},
-        rusoto, BatchConfig, BatchSettings, Compression, Length, PartitionBatchSink,
+        rusoto, Batch, BatchConfig, BatchSettings, Compression, Length, PartitionBatchSink,
         PartitionBuffer, PartitionInnerBuffer, TowerRequestConfig, TowerRequestSettings,
         VecBuffer2,
     },
@@ -160,12 +160,13 @@ impl CloudwatchLogsSinkConfig {
 #[typetag::serde(name = "aws_cloudwatch_logs")]
 impl SinkConfig for CloudwatchLogsSinkConfig {
     fn build(&self, cx: SinkContext) -> crate::Result<(super::RouterSink, super::Healthcheck)> {
-        let batch = self.batch.use_size_as_events()?.get_settings_or_default(
+        let batch = VecBuffer2::<InputLogEvent>::get_settings_defaults(
+            self.batch,
             BatchSettings::default()
                 .bytes(1_048_576)
                 .events(10_000)
                 .timeout(1),
-        );
+        )?;
         let request = self.request.unwrap_with(&REQUEST_DEFAULTS);
 
         let log_group = self.group_name.clone();
--- a/src/sinks/aws_cloudwatch_logs/mod.rs
+++ b/src/sinks/aws_cloudwatch_logs/mod.rs
@@ -164,7 +164,7 @@ impl SinkConfig for CloudwatchLogsSinkConfig {
             .bytes(1_048_576)
             .events(10_000)
             .timeout(1)
-            .parse_config::<VecBuffer2<InputLogEvent>>(self.batch)?;
+            .parse_config(self.batch)?;
         let request = self.request.unwrap_with(&REQUEST_DEFAULTS);
 
         let log_group = self.group_name.clone();
--- a/src/sinks/aws_cloudwatch_logs/mod.rs
+++ b/src/sinks/aws_cloudwatch_logs/mod.rs
@@ -7,7 +7,7 @@ use crate::{
     sinks::util::{
         encoding::{EncodingConfig, EncodingConfiguration},
         retries::{FixedRetryPolicy, RetryLogic},
-        rusoto, Batch, BatchConfig, BatchSettings, Compression, Length, PartitionBatchSink,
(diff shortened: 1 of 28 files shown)


src/sinks/aws_cloudwatch_metrics.rs
```diff
@@ -117,11 +117,10 @@ impl CloudWatchMetricsSvc {
         client: CloudWatchClient,
         cx: SinkContext,
     ) -> crate::Result<super::RouterSink> {
-        let batch = config
-            .batch
-            .disallow_max_bytes()?
-            .use_size_as_events()?
-            .get_settings_or_default(BatchSettings::default().events(20).timeout(1));
+        let batch = BatchSettings::default()
+            .events(20)
```
Hoverbear, line 121: Do you think it would be beneficial to conventionalize having these defaults as some `const` values inside sinks that use them?

bruceg: Oh yeah, since all the methods are `const`, these could totally get moved out of the function.

bruceg: Um, actually, no they can't, or at least not without repeating the buffer type in the const declaration. The actual type of these settings are `BatchSettings<B> where B: Batch`, and the `B` is inferred by passing the settings to the buffer creation function. This allows the batch parsing to be controlled by the buffer type, which was the whole point of this PR.

src/sinks/util/batch.rs
```diff
@@ -116,6 +137,19 @@ impl BatchSettings {
             ..self
         }
     }
+
+    // Would like to use `trait From` here, but that results in
```
Hoverbear, line 141: So if you can limit `B2` such that Rust can know `BatchSettings` is not a valid input you might be able to do this?

src/sinks/util/buffer/vec.rs
```diff
@@ -21,6 +21,16 @@ impl<T> Batch for VecBuffer<T> {
     type Input = T;
     type Output = Vec<T>;
 
+    fn get_settings_defaults(
```
Hoverbear, line 24: I see this fn interface enough times I wonder if it should be its own trait?

src/sinks/util/batch.rs
bruceg: AFAICT the issue is excluding `B1` from being the same as `B2`, because where they are the same this collapses to just `impl From<T> for T` (where T is `BatchSettings<B>`) which is pre-defined as an identity transform.

src/sinks/util/buffer/vec.rs
bruceg: It's an interesting idea, though I think out of scope for this PR, unless you are seeing it in different places in this set of interfaces.

src/sinks/aws_cloudwatch_metrics.rs
Hoverbear: Totally makes sense!

src/sinks/util/batch.rs
Hoverbear: Aw shucks, ok. :(

src/sinks/util/buffer/vec.rs
Hoverbear: Agreed!

bruceg:
FYI I have a PR prepared for #2917 based on this one.

Hoverbear:
@bruceg I'd be happy to review that if you have a cleanish commit history with it so I can get a relevant diff!

src/sinks/aws_cloudwatch_logs/mod.rs
```diff
@@ -160,12 +160,11 @@ impl CloudwatchLogsSinkConfig {
...
 impl SinkConfig for CloudwatchLogsSinkConfig {
     fn build(&self, cx: SinkContext) -> crate::Result<(super::RouterSink, super::Healthcheck)> {
-        let batch = self.batch.use_size_as_events()?.get_settings_or_default(
-            BatchSettings::default()
-                .bytes(1_048_576)
-                .events(10_000)
-                .timeout(1),
-        );
+        let batch = BatchSettings::default()
+            .bytes(1_048_576)
```
lukesteensen, line 164: If we wanted to get fancy, we could add `megabytes`/`kilobytes` methods :smile:

bruceg: But would that be `megabytes` or `mebibytes`? :grimacing:

PR #3153, opened 2020-07-22
chore(operations): Fixup version invocation to not have info

### LINKED ISSUE #3152: Fix release-docker job for 0.10.0
See https://github.com/timberio/vector/runs/896263455?check_suite_focus=true

Hoverbear:
Fixes the `make version` command from outputting content that would break the release process. Closes #3152


Hoverbear:
This might be causing `release-github` to fail too: 
image

PR #3185, opened 2020-07-24
chore: Moves platform-specific signal code out of `main`

### LINKED ISSUE #1383: Move platform-specific signal handling code out of `main.rs`
After Windows support was added, the `main.rs` started to contain a lot of duplicated code for handling interruption signals on `cfg(unix)` and `cfg(windows)`. We need to factor it out by putting it to a separate source file or even a new crate in `lib`. See https://github.com/timberio/vector/pull/1219#discussion_r354382434 for details.

ktff:
Closes #1383 

With moving signal code "slight" updates to usage of `futures01` in handling code was applied. We can split this update out to a separate PR if needed, it was too tempting not to collapse all of the `select`,`compat`,`match`, etc., into a simpler form.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,18 +4,18 @@ extern crate tracing;
 mod cli;
 
 use cli::{Color, LogFormat, Opts, SubCommand};
-use futures::compat::Future01CompatExt;
+use futures::{compat::Future01CompatExt, StreamExt, TryStreamExt};
 use futures01::{future, Future, Stream};
 use std::{
     cmp::max,
     fs::File,
     path::{Path, PathBuf},
 };
-#[cfg(unix)]
-use tokio_signal::unix::{Signal, SIGHUP, SIGINT, SIGQUIT, SIGTERM};
 use topology::Config;
 use vector::{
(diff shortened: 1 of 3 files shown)


lukesteensen:
Does need a `cargo fmt` before merge though.

ktff:
@lukesteensen I'll be merging this, `cargo fmt` changed nothing, what did you see?

src/main.rs
```diff
@@ -128,120 +132,67 @@ fn main() {
...
+                            {
+                                Ok(true) => (),
+                                Ok(false) => error!("Reload was not successful."),
+                                // Trigger graceful shutdown for what remains of the topology
+                                Err(()) => break SignalTo::Shutdown,
+                            }
+                        } else {
+                            error!("Reload aborted.");
+                        }
+                    }else{
```
lukesteensen, line 169: I think this is all I saw. I wonder if the maco confuses rustfmt?

ktff: It does, outside of the macro it formats it fine, I'll change it in other PR.

PR #3188, opened 2020-07-24
chore: update tokio in tls

fanatid:
Ref. #2945

Tokio update in `src/tls`. Still need figure out lifetime problem for warp, see https://github.com/seanmonstar/warp/pull/675 (currently works with patched `warp`: https://github.com/timberio/warp/tree/serve-incoming-lifetime)


fanatid:
I'm not sure that we need these cyclical dependencies in features:
```toml
sources-http = ["warp", "sources-tls"]
sources-logplex = ["warp", "sources-tls"]
sources-socket = ["bytesize", "listenfd", "tokio/uds", "tokio-util/udp", "sources-tls"]
sources-splunk_hec = ["bytesize", "warp", "sources-tls"]
sources-tls = ["sources-http", "sources-logplex", "sources-socket", "sources-splunk_hec"]
```
I think that `sources-tls` should be equal `[]`, tls do not depend on listed features.

src/sources/util/tcp.rs
```diff
@@ -98,117 +94,133 @@ pub trait TcpSource: Clone + Send + 'static {
...
-    .forward(out)
+    .filter_map(move |frame| future::ready(match frame {
+        Ok(frame) => {
+            let host = host.clone();
+            source.build_event(frame, host).map(Ok)
+        }
+        Err(_) => {
+            // Drop error, or?
+            Some(Err(()))
+        }
```
bruceg, line 218: I don't think we should ever just drop a specific error, but it's not obvious to me what should happen here.

src/tls/mod.rs
```diff
@@ -19,26 +16,27 @@ mod settings;
...
 pub(crate) use maybe_tls::MaybeTls;
-pub(crate) use outgoing::MaybeTlsConnector;
 pub use settings::{MaybeTlsSettings, TlsConfig, TlsOptions, TlsSettings};
 
+// TODO: Rename to TlsResult?
 pub type Result<T> = std::result::Result<T, TlsError>;
 
 pub type MaybeTlsStream<S> = MaybeTls<S, SslStream<S>>;
 
+// TODO: check what not used and remove
```
bruceg, line 26: Are these still outstanding?

src/tls/maybe_tls.rs
```diff
@@ -54,47 +60,76 @@ impl<R: Clone, T: Clone> Clone for MaybeTls<R, T> {
 }
 
 // Conditionally implement Debug for Debugable types
-impl<R: Debug, T: Debug> Debug for MaybeTls<R, T> {
-    fn fmt(&self, fmt: &mut Formatter) -> fmt::Result {
+impl<R: fmt::Debug, T: fmt::Debug> fmt::Debug for MaybeTls<R, T> {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
```
bruceg, line 64: I'm not sure why to add the `fmt::` prefix to these.

src/sources/util/tcp.rs
fanatid: Added `warn!` and return `None` instead `Some(Err(()))`

src/topology/config/mod.rs
```diff
@@ -127,11 +127,13 @@ pub enum DataType {
...
-        shutdown: ShutdownSignal,
-        out: mpsc::Sender<Event>,
-    ) -> crate::Result<sources::Source>;
+        _name: &str,
+        _globals: &GlobalOptions,
+        _shutdown: ShutdownSignal,
+        _out: mpsc::Sender<Event>,
+    ) -> crate::Result<sources::Source> {
+        unimplemented!()
+    }
```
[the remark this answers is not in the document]
fanatid: Yeah, I added this here first and then opened #3182 because such change required separate PR. Later forgot to remove this change. Removed.

src/tls/maybe_tls.rs
fanatid: Just removed extra imports (Debug/Formatter) and changed to familiar Debug trait impl. After this PR all Debug impls in codebase will looks the same. But not a big deal for me can change back.

src/tls/mod.rs
fanatid: Yes :( need check everything, make slight changes and remove draft pr mark.

src/tls/mod.rs
```diff
@@ -19,26 +16,27 @@ mod settings;
...
 #[derive(Debug, Snafu)]
 pub enum TlsError {
```
fanatid: Can I rename this enum to `Error`? Then import can be used as `crate::tls::Result` / `crate::tls::Error` (same as in root: `crate::Result` / `crate::Error`).

src/sources/util/tcp.rs
```diff
@@ -60,7 +55,8 @@ fn make_listener(
     }
 }
 
-pub trait TcpSource: Clone + Send + 'static {
+// TODO: add async_trait for making `run` async.
+pub trait TcpSource: Clone + Send + Sync + 'static {
```
fanatid, line 58: Should we make `run` method in this trait async? cc @lukesteensen

src/tls/mod.rs
bruceg: I don't see that as much of a benefit. I would tend to go the other way, to rename `Result` to `TlsResult` so we don't have to use the module path on each use. Looking at our sources, we appear to have far more instances of `pub (enum|type|struct) SomethingError` and `SomethingResult` than the other. In fact, the above is the only `type Result` other than the crate level type alias.

src/sources/util/tcp.rs
bruceg: It will need to be made async eventually, though that would probably be worth a separate PR unless it makes this one significantly easier.

fanatid: Sure, separate PR. Just wanted know the opinion in general.

fanatid:
TcpSink do not work properly now on shutdown :( need learn how compat works for fixing code =\

src/sources/socket/mod.rs
```diff
@@ -382,7 +385,7 @@ mod test {
             MaybeTlsSettings::Raw(()),
         );
         rt.spawn(
-            futures01::stream::iter_ok::<_, ()>(std::iter::repeat(()))
+            futures01::stream::iter_ok::<_, ()>(std::iter::repeat(()).take(1_000))
```
fanatid, line 388: I added `.take(1_000)` because if left unlimited stream test will fail.
Will fail because once `TcpSource` receive shutdown signal and close socket with `Shutdown::Write`, `poll_complete` on `TcpSink` side will not be ready because `mpsc` in `TcpSource` will full, as result `TcpSink` will not shutdown and `TcpSource` will live while `TcpSink` live. Looks like some deadlock. Maybe this problem go away with moving more code to new futures? Not sure.
https://github.com/timberio/vector/blob/fb97d37dc30a700bd74eda9bd37f42faa2cd654e/src/sinks/util/tcp.rs#L224-L233

src/sources/http.rs
```diff
@@ -61,8 +62,19 @@ impl HttpSource for SimpleHttpSource {
...
+        &self,
+        _name: &str,
+        _globals: &GlobalOptions,
+        _shutdown: ShutdownSignal,
+        _out: mpsc::Sender<Event>,
+    ) -> crate::Result<super::Source> {
+        unimplemented!()
+    }
+
+    async fn build_async(
```
bruceg, line 77: This is not for this PR

src/sources/logplex.rs
```diff
@@ -32,16 +33,29 @@ impl HttpSource for LogplexSource {
...
+        &self,
+        _name: &str,
+        _globals: &GlobalOptions,
+        _shutdown: ShutdownSignal,
+        _out: mpsc::Sender<Event>,
+    ) -> crate::Result<super::Source> {
+        unimplemented!()
+    }
+
+    async fn build_async(
```
bruceg, line 48: Ditto

src/sources/splunk_hec.rs
```diff
@@ -70,8 +67,19 @@ fn default_socket_address() -> SocketAddr {
...
+        &self,
+        _name: &str,
+        _globals: &GlobalOptions,
+        _shutdown: ShutdownSignal,
+        _out: mpsc::Sender<Event>,
+    ) -> crate::Result<super::Source> {
+        unimplemented!()
+    }
+
+    async fn build_async(
```
bruceg, line 82: Ditto

src/sources/http.rs
fanatid: Why? `HttpSource::run` is async now:
https://github.com/timberio/vector/blob/fb97d37dc30a700bd74eda9bd37f42faa2cd654e/src/sources/util/http.rs#L57

bruceg: Oh, my bad. I misread this and the other two as a remnant of the other PR that you had maybe only partially reverted.

fanatid: One note about this change that now `bind` happened in `build_async`, I remember that `future::lazy` was used to remove such effect before -- is this change ok?

src/sources/socket/mod.rs
ktff: That's exactly what's happening, is what I wanted to say, but after some digging, even in that case it should have been forcibly shutdown. Something else is going on, try increasing the thread count.

ktff: More context, `TcpSource` should live so long all it's received events haven't been processed, this is graceful shutdown, but for this case, and others, on top of graceful we have forceful which should have shutdown it.

ktff: Maybe it's an issue of using dedicated threads for blocking operations, I don't think they can be forcibly stopped.

src/sinks/socket.rs
```diff
@@ -221,37 +226,45 @@ mod test {
...
+                        let mut cx = Context::from_waker(&noop_waker_ref());
+                        loop {
+                            if close_rx.lock().await.try_recv().is_ok() {
+                                stream.get_ref().unwrap().shutdown(Shutdown::Write).unwrap();
                                 break;
-                            } else {
-                                msg_counter1.fetch_add(1, Ordering::SeqCst);
+                            }
+
+                            match Pin::new(&mut stream).poll_read(&mut cx, &mut [0u8; 11]) {
```
lukesteensen, line 252: This seems to have gotten a lot more complicated because you switched from `read` to `poll_read`. Is that something we need to do? I belive the initial logic was that this socket will already be in non-blocking mode and a normal read works just fine.

src/sinks/util/tcp.rs
```diff
@@ -211,17 +206,18 @@ impl Sink for TcpSink {
...
+                // If this returns `Poll::Pending` we know the connection is still
                 // valid and the write will most likely succeed.
-                let stream: &mut MaybeTlsStream<TcpStream> =
-                    connection.get_mut().get_mut().get_mut().get_mut();
-                match stream.read(&mut [0u8; 1]) {
-                    Err(error) if error.kind() != ErrorKind::WouldBlock => {
+
+                let stream: &mut MaybeTlsStream<TcpStream> = connection.get_mut().get_mut();
+                let mut cx = Context::from_waker(&noop_waker_ref());
+                match Pin::new(stream).poll_read(&mut cx, &mut [0u8; 1]) {
```
lukesteensen, line 214: Again, why the change here? This doesn't seem directly related to upgrading tokio.

src/sources/http.rs
```diff
@@ -73,7 +85,7 @@ impl SourceConfig for SimpleHttpConfig {
             encoding: self.encoding,
             headers: self.headers.clone(),
         };
-        source.run(self.address, "", &self.tls, out, shutdown)
+        source.run(self.address, "", &self.tls, out, shutdown).await
```
lukesteensen, line 88: This doesn't quite seem right. We're going to `await` the return value of `build_async` when we're building the topology, and this seems to imply that it will immediately start running and not return until it's done processing the input.

src/sources/logplex.rs
```diff
@@ -32,16 +33,29 @@ impl HttpSource for LogplexSource {
...
         _: &str,
         _: &GlobalOptions,
         shutdown: ShutdownSignal,
         out: mpsc::Sender<Event>,
     ) -> crate::Result<super::Source> {
         let source = LogplexSource::default();
-        source.run(self.address, "events", &self.tls, out, shutdown)
+        source
+            .run(self.address, "events", &self.tls, out, shutdown)
+            .await
```
lukesteensen, line 58: Same here.

src/sources/socket/mod.rs
lukesteensen: I would much rather we fix the root cause here than add the `take`. We need to make sure shutdown still works properly.

src/sources/http.rs
fanatid: This partially incorrect, I mentioned this in https://github.com/timberio/vector/pull/3188#discussion_r461653474.
`run` is async because bind in tokio:0.2 is async, previously bind happened on topology building as sync. `run` result is still future: https://github.com/timberio/vector/blob/9e13da7960a2ecca0878d38fd15ddf65ad23dae7/src/sources/util/http.rs#L120-L130
I remember that `future01::lazy` was used in other sinks/sources for moving bind out of topology building, but here was no lazy usage, so everything saved. Should I move bind out of topology building?

src/sinks/socket.rs
fanatid: [TcpStream](https://docs.rs/tokio/0.2.22/tokio/net/struct.TcpStream.html) from new tokio not implement [std::io::Read](https://doc.rust-lang.org/std/io/trait.Read.html), I think this particular example can be simplified, I'll try.

src/sources/http.rs
lukesteensen: Mentioned this in slack, but for the record we should not bind within `build`, but only when the future returned by `build` is polled.

src/sinks/socket.rs
```diff
@@ -221,37 +227,45 @@ mod test {
...
-                            } else {
-                                msg_counter1.fetch_add(1, Ordering::SeqCst);
+        let jh = rt.spawn_handle_std(async move {
+            let tls = MaybeTlsSettings::from_config(&config, true).unwrap();
+            let mut listener = tls.bind(&addr).await.unwrap();
+            listener
+                .incoming()
+                .take(2)
+                .for_each(|connection| {
+                    let mut close_rx = close_rx.take();
```
lukesteensen, line 239: Why did this need to become an option?

src/sinks/util/tcp.rs
```diff
@@ -77,20 +78,28 @@ pub struct TcpSink {
...
-    CompatSink<FramedWrite<Compat<Compat01As03<MaybeTlsStream<TcpStream>>>, BytesCodec>, Bytes>;
+type TcpOrTlsStream = CompatSink<FramedWrite<MaybeTlsStream<TcpStream>, BytesCodec>, Bytes>;
+
+// enum TcpSinkState2 {
+//     Disconnected,
+//     ResolvingDns(BoxFuture<'static, Result<crate::dns::LookupIp, crate::dns::DnsError>>),
+//     Connecting(BoxFuture<'static, Result<MaybeTlsStream<TcpStream>, TlsError>>),
+//     Connected(FramedWrite<MaybeTlsStream<TcpStream>, BytesCodec>),
+//     Backoff(BoxFuture<'static, ()>),
+// }
```
lukesteensen, line 102: We should clean up all the commented out code before merging.

fanatid: Yeah, I understand that this code not related anyhow with current TcpSink, but can be very useful in future for saving time of anybody who will work on migrating to new Sink trait. If you insist to delete this, can I at least add a link to relevant code in this PR with a comment about new Sink trait impl?

src/sinks/socket.rs
fanatid: [tokio::sync::oneshot::Receiver](https://docs.rs/tokio/0.2.22/tokio/sync/oneshot/struct.Receiver.html) not `Copy` and not `Clone` and can be used only once, but we have two connections, so I need say for a second that I do not have `Receiver` to it.

Issue #1940, opened 2020-02-26
Avoid cross-dependencies between components


a-rodin:
Currently some components of Vector import and use code from other components. The list of such dependencies is the following:

* `sources-kubernetes` depends on `sources-file`, `transforms-json_parser`, `transforms-regex_parser`;
* `sources-syslog` depends on `sources-socket`;
* `sources-vector` depends on `sources-socket`;
* `sinks-humio_logs` depends on `sinks-splunk_hec`;
* `sinks-new_relic_logs` dependes on `sinks-http`;
* `sinks-sematext_logs` depends on `sinks-elasticsearch`.

I think in long term this might make it harder to refactor individual components, as change is one component might break some other (seemingly) unrelated component. So ideally we need to identify the common code and factor it out to `lib` crates or, if the code is too specific, at least as util modules.

The above relates only to implementation of components, but not to their tests. Using other components in tests I think is less an issue, although some tests for components might be changed to not depend on other components. In particular, I think most of the tests depending on `json_parser` can be modified to avoid such a dependency.

See also https://github.com/timberio/vector/pull/1924.

ktff:
These have already been resolved:
* `syslog` source
* `vector` source
* `honeycomb` sink

- [ ] although some of their feature dependency can be updated to reflect this. 

---

`kubernetes_logs` source uses transforms:
* `merge` whole so there is little to gain in separating common code.
* `regex` partially so there is common code that could be extracted, although it would be quite specific.
     - [ ] To extract common `regex` or not?
* `json_parser` partially so there is common code to be extracted, and it's something that can be shared, even specialized since tests use it in same/similar way as `kuberentes_logs` source.   
     - [ ] Extract 

---

Others: 
* `humio_logs` sink uses whole `splunk_hec` sink
* `new_relic_logs` sink uses whole `http` sink
* `sematext_logs` sink uses whole `elasticsearch` sink

They are using whole of the other sink, as they are basically wrappers with some extra/default configuration. Their common code would be the whole sink so we wouldn't gain much.

cc. @binarylogic

PR #3210, opened 2020-07-26
chore: Remove unneeded cross feature dependency

ktff:
Ref. #1940 

Some feature dependencies were a leftover, and some were gating shared code so those were removed.


Cargo.toml
```diff
@@ -261,13 +261,13 @@ sources-journald = []
...
-sources-socket = ["bytesize", "listenfd", "tokio/uds", "tokio-util/udp", "sources-tls"]
+sources-socket = ["bytesize", "listenfd", "tokio-util/udp", "sources-tls"]
 sources-splunk_hec = ["bytesize", "warp", "sources-tls"]
 sources-statsd = ["tokio-util/udp"]
 sources-stdin = ["bytesize"]
-sources-syslog = ["sources-socket", "syslog_loose"]
-sources-tls = ["sources-http", "sources-logplex", "sources-socket", "sources-splunk_hec"]
-sources-vector = ["sources-socket"]
+sources-syslog = ["bytesize", "listenfd", "tokio-util/udp", "sources-tls", "syslog_loose"]
+sources-tls = []
```
fanatid, line 269: Looks like this solve my question https://github.com/timberio/vector/pull/3188#issuecomment-663712811

Cargo.toml
```diff
@@ -43,7 +43,7 @@ tracing-limit = { path = "lib/tracing-limit" }
 futures01 = { package = "futures", version = "0.1.25" }
 futures = { version = "0.3", default-features = false, features = ["compat", "io-compat"] }
 tokio01 = { package = "tokio", version = "0.1.22", features = ["io", "tcp", "rt-full", "experimental-tracing", "sync"], default-features = false }
-tokio = { version = "0.2.13", features = ["blocking", "fs", "sync", "macros", "test-util", "rt-core", "io-std"] }
+tokio = { version = "0.2.13", features = ["blocking", "fs", "sync", "macros", "test-util", "rt-core", "io-std", "uds"] }
```
fanatid, line 46: Why move `uds` feature from components to dependency?

Cargo.toml
ktff: Yep.

Cargo.toml
ktff: Our `unix` util requires this feature, and without moving it here the `unix` util  would need to be feature guarded by  sources that use it, and in turn they would need to depend on features that `unix` util needs which kinda backwards, so this is a compromise. 

An alternative would be to have separate feature for `unix` util.

PR #3214, opened 2020-07-26
chore: Remove dependency on `transforms-json_parser`

ktff:
Ref. #1940 

PR doesn't extract shared code, but specializes it for it's call site. It's a bit faster, shorter, clearer, and has a dependency less.

### DIFF
--- a/src/sources/kubernetes_logs/parser/docker.rs
+++ b/src/sources/kubernetes_logs/parser/docker.rs
@@ -1,6 +1,6 @@
 use crate::{
     event::{self, Event, LogEvent, Value},
-    internal_events::KubernetesLogsMessageParseFailed,
+    internal_events::KubernetesLogsDockerFormatParseFailed,
     transforms::Transform,
 };
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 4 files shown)


src/sources/kubernetes_logs/parser/docker.rs
```diff
@@ -22,32 +20,29 @@ lazy_static! {
...
+    let to_parse = log.remove(&event::log_schema().message_key())?.as_bytes();
+
+    match serde_json::from_slice(to_parse.as_ref()) {
+        Ok(JsonValue::Object(object)) => {
+            for (key, value) in object {
+                log.insert_flat(key, value);
+            }
+            Some(())
+        }
+        Ok(_) | Err(_) => None,
```
MOZGIII, line 45: We probably want to emit an event here with a counter.

ktff: Like this?

src/sources/kubernetes_logs/parser/docker.rs
```diff
@@ -42,7 +43,10 @@ fn parse_json(log: &mut LogEvent) -> Option<()> {
             }
             Some(())
         }
-        Ok(_) | Err(_) => None,
+        Ok(_) | Err(_) => {
+            emit!(KubernetesLogsMessageParseFailed { message: &to_parse });
```
MOZGIII, line 47: Looks good, but let's put the `Docker` somewhere in the name. This should be a docker-specific parser error - pretty much mirroring the similar emit at JSON.

How about `KubernetesLogsDockerFormatParseFailed`?

Similarly, the `counter!` can emit `k8s_docker_format_parse_failures`.

ktff: Sounds good.

PR #3275, opened 2020-07-30
chore(sources): introduce Pipeline type

### LINKED ISSUE #2935: Introduce Push trait for sources in place of channel
Blocked on #2933 

See [Architecture RFC](https://github.com/timberio/vector/blob/master/rfcs/2020-06-18-2625-architecture-revisit.md).

lukesteensen:
Closes #2935

This is just a mechanical change to introduce pipelines first as a wrapper around the old `Sender`s. Now that we have the wrapper in place, we can get going on some more interesting changes:

1. Embed transforms in pipelines
2. Add APIs for partitioning
3. Replace old channels with new ones
4. etc, etc

### DIFF
--- a/src/sources/socket/mod.rs
+++ b/src/sources/socket/mod.rs
@@ -5,12 +5,12 @@ mod unix;
 
 use super::util::TcpSource;
 use crate::{
-    event::{self, Event},
+    event,
     shutdown::ShutdownSignal,
     tls::MaybeTlsSettings,
     topology::config::{DataType, GlobalOptions, SourceConfig, SourceDescription},
+    Pipeline,
 };
-use futures01::sync::mpsc;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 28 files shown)


src/sources/socket/mod.rs
```diff
@@ -227,7 +225,7 @@ mod test {
 
     #[test]
     fn tcp_continue_after_long_line() {
-        let (tx, rx) = mpsc::channel(10);
+        let (tx, rx) = Pipeline::new_test();
```
bruceg, line 228: I notice some of these had differing sizes in the channel creation, but `new_test` uses just one. Will that be an issue?

lukesteensen: It doesn't seem to be an issue. A lot of these were just guesses at a reasonable default, but if the test depends on a specific size we can switch back to specifying.

PR #3298, opened 2020-07-31
fix: conditional compilation for unix source

fanatid:
Fixes messages like this when compiling with `--no-default-features --features ...`:
```bash
warning: function is never used: `build_unix_source`
  --> src/sources/util/unix.rs:22:8
   |
22 | pub fn build_unix_source<D, E>(
   |        ^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```


fanatid:
Will update/rebased after https://github.com/timberio/vector/pull/3188 because looks like need update tls related code.

fanatid:
Also fixes `sources-journald` cfg from https://github.com/timberio/vector/pull/2558#discussion_r464930315

PR #3300, opened 2020-07-31
feat(humio_logs sink): Allow configuration of type field

jszwedko:
Targeting #3297  

This adds a configuration option to tell Humio what the type is of the
events so that it can parse them accordingly.

I think this usually won't be needed, but could be useful if the `text`
encoding is used.

Closes https://github.com/timberio/vector/issues/3082

### DIFF
--- a/src/sinks/humio_logs.rs
+++ b/src/sinks/humio_logs.rs
@@ -18,6 +18,10 @@ pub struct HumioLogsConfig {
         default
     )]
     encoding: EncodingConfigWithDefault<Encoding>,
+
+    #[serde(rename = "type")]
+    sourcetype: Option<String>,
+
     #[serde(default)]
     pub compression: Compression,
 
@@ -72,6 +76,7 @@ impl HumioLogsConfig {
         HecSinkConfig {
             token: self.token.clone(),
             host,
+            sourcetype: self.sourcetype.clone(),
             encoding: self.encoding.clone().transmute(),
             compression: self.compression,
             batch: self.batch,
@@ -129,6 +134,7 @@ mod integration_tests {
         topology::config::{SinkConfig, SinkContext},
         Event,
     };
+    use chrono::Utc;
     use futures::compat::Future01CompatExt;
     use futures01::Sink;
     use serde_json::json;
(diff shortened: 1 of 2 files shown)


src/sinks/humio_logs.rs
```diff
@@ -18,6 +18,10 @@ pub struct HumioLogsConfig {
         default
     )]
     encoding: EncodingConfigWithDefault<Encoding>,
+
+    #[serde(rename = "type")]
```
jszwedko, line 22: Using `type` here rather than `sourcetype` as this matches Humio's terminology. I couldn't call it `type` in Rust so I just decided to use `sourcetype` internally since it is used with the HEC sink.

https://docs.humio.com/foundational-concepts/events/#type

src/sinks/humio_logs.rs
```diff
@@ -287,7 +345,7 @@ mutation {{
         humio_type: String,
 
         #[serde(rename = "@error")]
-        error: Option<bool>,
+        error: Option<String>,
```
jszwedko, line 348: Realized it is actually sending back `"true"` rather than `true` when errors occur.

ktff: Good catch.

src/sinks/humio_logs.rs
```diff
@@ -174,6 +180,58 @@ mod integration_tests {
         });
     }
 
+    #[test]
+    fn humio_type() {
+        let mut rt = runtime();
+
+        rt.block_on_std(async move {
```
fanatid, line 187: I think [#[tokio::test]](https://docs.rs/tokio/0.2.22/tokio/attr.test.html) can be used instead creating runtime and call `block_on_std`.

jszwedko: I actually tried that on Friday after Bruce suggested it in #3175 but it looks like the Humio sink still has a dependency on tokio01 so I ended up getting an error that the sink was shutdown when I tried to send to it. I think the last remaining dependency is through the tls package so it can probably be swapped soon when https://github.com/timberio/vector/pull/3188 is merged.

jszwedko:
> LGTM, this is for #3250 and not #3298 ?

Doh, yeah, I meant to say that this PR was targeting the branch in https://github.com/timberio/vector/pull/3297 . It fixes #2717 .

PR #3304, opened 2020-08-01
chore: Refactor line_agg

MOZGIII:
This is a minor refactoring of the `sources::file::line_agg`, in preparation to extract it for reuse among other sources.

- Extracted logic into a reusable piece (experiments have shown that this will be very useful for some implementation variants).
- Switched to futures 0.3; an attempt to rewrite the implementation to `async_stream` was made, but it turned out to be more complicated than the raw poll-based implementation. The poll-based approach is easier to reason about. This isn't definite though, and possibly another attempt will be more successful.

### DIFF
--- a/src/sources/file/line_agg.rs
+++ b/src/sources/file/line_agg.rs
@@ -1,12 +1,20 @@
+#![deny(missing_docs)]
+
 use bytes05::{Bytes, BytesMut};
-use futures01::{Async, Poll, Stream};
+use futures::{Stream, StreamExt};
+use pin_project::pin_project;
 use regex::bytes::Regex;
 use serde::{Deserialize, Serialize};
 use std::collections::{hash_map::Entry, HashMap, VecDeque};
 use std::hash::Hash;
 use std::time::Duration;
-use tokio01::timer::DelayQueue;
+use std::{
+    pin::Pin,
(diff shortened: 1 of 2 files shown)


src/sources/file/line_agg.rs
```diff
@@ -38,8 +46,9 @@ pub enum Mode {
     HaltWith,
 }
 
+/// Configuration parameters of the line aggregator.
 #[derive(Debug, Clone)]
-pub(super) struct Config {
+pub struct Config {
```
bruceg, line 51: Are these structs used beyond `super`? I'm wondering why the visibility change is necessary. Does this have to do with the pin projection?

MOZGIII: Cisibility changes are nit nesessary technically - it’s just a preparation to move this out of `file` mod.
I probably should do it right in this PR really.

MOZGIII: I'll do the move in another PR. There are pending changes to the `file` source - I'd like to merge those first to ease the conflicts. The plan is to move `src/sources/file/line_agg.rs` to `src/line_agg.rs` and `src/sources/file/mod.rs` back to `src/sources/file.rs`.

Issue #3043, opened 2020-07-12
Flaky `test_udp_syslog`


ktff:
Ref. #2978 

Tests `syslog` source in `udp` mode.

https://github.com/timberio/vector/pull/2955/checks?check_run_id=861560287
https://github.com/timberio/vector/pull/3040/checks?check_run_id=862920206

ktff:
#3308 doesn't seem as the main cause.

jszwedko:
Closing since this test was deleted and evidently doesn't appear to have been needed.

Issue #3039, opened 2020-07-12
Flaky `test_max_size_resume`


ktff:
Ref. #2978 

Tests that we don't loose any events if internal buffers get full.

[test_max_size_resume.log](https://github.com/timberio/vector/files/4905004/test_max_size_resume.log)
https://github.com/timberio/vector/pull/3044/checks?check_run_id=862915559

Probably a race, but it may be a bug.

ktff:
#3308 doesn't seem as the main cause.

jszwedko:
Closing since this test was deleted and evidently doesn't appear to have been needed.

PR #3308, opened 2020-08-01
chore(tests): Fix race in `CountReceiver`

ktff:
### Problem

There was a race in some of our usages of `CountReceiver`, and there still is (explained later). The race was happening between `CountReceiver` establishing first connection and:
* waiting on it
* shutting down `topology`

If connection is established before both of them, everything is fine, if not, the events weren't being collected so the tests failed.

### Solution

 `CountReceiver` now exposes a way to wait for the first connection which tests can use to avoid the race. This theoretically doesn't remove the race entirely since we don't sync for other connections, but our usage of it is for one connection only so this fixes the race in practice. This behavior is commented in code.

### Affected

First found while examining flaky `merge_and_fork` test, but was affecting dozens of tests. They were determined to be affected by ,well, first using the `CountReceiver` and second they started to fail once an artificial delay was introduced [here](https://github.com/timberio/vector/blob/6244e68d206eb614311ab950634850d0d36ab69a/src/test_util.rs#L441), to amplify the race. So with `merge_and_fork`, all other affected tests were also preemptively modified:
* `tcp_stream`
* `basic_unix_sink`, which is also now enabled
* `crash` tests
* `tcp` tests

Other tests and benches weren't failing so they weren't changed.

Some of before removed flaky tests were also using this code so they probably had this race:
* #3043 
* #3041 
* #3039 

- [ ] It will be nice to check if those tests were really affected and if this fixes them completely.

Ref. #2978

### DIFF
--- a/src/sinks/util/unix.rs
+++ b/src/sinks/util/unix.rs
@@ -231,7 +231,7 @@ mod tests {
             let out_path = temp_uds_path("unix_test");
 
             // Set up server to receive events from the Sink.
-            let receiver = CountReceiver::receive_lines_unix(out_path.clone());
+            let mut receiver = CountReceiver::receive_lines_unix(out_path.clone());
 
             // Set up Sink
             let config = UnixSinkConfig::new(out_path, Encoding::Text.into());
@@ -241,6 +241,9 @@ mod tests {
             // Send the test data
             let (input_lines, events) = random_lines_with_stream(100, num_lines);
             let _ = sink.send_all(events).compat().await.unwrap();
+
+            // Wait for output to connect
+            receiver.connected().await;
 
             // Receive the data sent by the Sink to the receiver
             assert_eq!(input_lines, receiver.wait().await);
--- a/src/sinks/util/unix.rs
+++ b/src/sinks/util/unix.rs
@@ -201,7 +201,7 @@ impl Sink for UnixSink {
     }
 }
 
-#[cfg(all(feature = "tokio/uds", test))]
+#[cfg(test)]
 mod tests {
     use super::*;
     use crate::test_util::{random_lines_with_stream, runtime, shutdown_on_idle, CountReceiver};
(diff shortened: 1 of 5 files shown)


src/sinks/util/unix.rs
```diff
@@ -201,7 +201,7 @@ impl Sink for UnixSink {
     }
 }
 
-#[cfg(all(feature = "tokio/uds", test))]
```
ktff, line 204: I was also confused by this, but this style of `crate/feature` doesn't work in `cfg` annotations.

src/test_util.rs
```diff
@@ -419,7 +442,10 @@ impl CountReceiver<String> {
         stream
             .take_until(tripwire)
             .map_ok(|socket| FramedRead::new(socket, LinesCodec::new()))
-            .map(|x| x.unwrap())
+            .map(|x| {
+                connected.take().map(|trigger| trigger.send(()));
+                x.unwrap()
+            })
```
fanatid, line 448: No `unwrap` for `trigger.send()` because channel can be closed to this moment? Probably `inspect` is more appropriate here.

src/test_util.rs
```diff
@@ -367,50 +368,72 @@ impl<T: Send + 'static> CountReceiver<T> {
         self.count.load(Ordering::Relaxed)
     }
 
+    /// Succeds once first connection has been maid.
```
bruceg, line 371: Typo: "made"

src/test_util.rs
ktff: > No unwrap for trigger.send() because channel can be closed to this moment?

Yes. If main code doesn't care about it, then this code doesn't either.

> Probably inspect is more appropriate here.

`send` requires `self`.

src/test_util.rs
ktff: Done.

PR #3309, opened 2020-08-01
chore(tests): Fix `splunk_hec` source tests

ktff:
#3275 introduced `Pipeline` which has less of a capacity than the previously used channel which caused a deadlock in tests that were depending on that capacity being larger than a certain size.

This PR removes this dependency.

### DIFF
--- a/src/sources/splunk_hec.rs
+++ b/src/sources/splunk_hec.rs
@@ -746,13 +746,11 @@ mod tests {
     };
     use chrono::{TimeZone, Utc};
     use futures::compat::Future01CompatExt;
-    use futures01::{stream, sync::mpsc, Sink};
+    use futures01::{stream, sync::mpsc, Future, Sink};
     use std::net::SocketAddr;
 
     /// Splunk token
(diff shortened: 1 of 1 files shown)


src/sources/splunk_hec.rs
```diff
@@ -813,12 +811,8 @@ mod tests {
         rt: &mut Runtime,
     ) -> Vec<Event> {
         let n = messages.len();
-        assert!(
-            n <= CHANNEL_CAPACITY,
-            "To much messages for the sink channel"
-        );
         let pump = sink.send_all(stream::iter_ok(messages.into_iter().map(Into::into)));
-        let _ = rt.block_on(pump).unwrap();
+        rt.spawn(pump.map(|_| ()));
```
fanatid, line 815: `unwrap` not required in `spawn`?

ktff: It would be nice. 
`.map_err(|()| panic!())` is not quite as effective, but it's closest that we can get.

PR #3310, opened 2020-08-01
chore: Add `rdkafka` dependency

ktff:
Adds missing feature dependency. 

Fixes `Component Features - Linux` check.

I'll be merging this as soon as that check passes.


fanatid:
I had seen this before, but it's not clear what should be used `rdkafka-plain` or `rdkafka-cmake`. I think just `rdkafka` is good, because included in both `plain` and `cmake`.
https://github.com/timberio/vector/blob/fbfea04a74c7adc92953200b7354052b53428aaa/Cargo.toml#L219-L224

PR #3342, opened 2020-08-04
chore: Remap v2 RFC

### LINKED ISSUE #3325: Remap v2 RFC

Jeffail:
Closes #3325 

[Rendered](https://github.com/timberio/vector/blob/remap-v2-rfc/rfcs/2020-08-04.3325.remap-v2.md)

### DIFF
--- a/rfcs/2020-08-04.3325.remap-v2.md
+++ b/rfcs/2020-08-04.3325.remap-v2.md
@@ -1,0 +1,55 @@
+# RFC #3325 - Remap V2
+
+## Scope
+
+This RFC covers incremental changes to the `remap` mapping syntax. Previous RFC for context: https://github.com/timberio/vector/blob/master/rfcs/2020-07-21-2744-remapping-syntax.md
+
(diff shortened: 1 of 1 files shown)


rfcs/2020-08-04.3325.remap-v2.md
```diff
@@ -0,0 +1,55 @@
...
+.baz = "something static"
+```
+
+Sometimes it might be more convenient to express `if` statements as an expression for assigning to a value:
+
+```coffee
+.content = if .type == "foo" { .foo.content } else { .bar.body }
+```
+
+In which case each block must contain a query. It would be possible to expand this to allow assignments within the block but we need to think through and experiment with the issues of mutating context and how we'd document the behavior we choose.
```
bruceg, line 49: Suggest "In this case" instead of "In which case".

By this, do you mean that there must be an `else` block, such that the expression has a value in either case, or something else? If the former, I suggest making that more explicit.

PR #3345, opened 2020-08-04
enhancement(docker source): Add instrumentation

### LINKED ISSUE #3196: Define events for the `docker` source
As part of finishing https://github.com/timberio/vector-product/issues/14, we need to define internal events for the `docker` source.

## Defining events

## Requirements

- [x] Add a `DockerEventReceived` event.
- [x] Audit and add events for all component-specific behavior. ([example](https://github.com/timberio/vector/blob/fb71f5ad76266d9a44164e6804723376353e4ff0/src/internal_events/tcp.rs#L9))

ktff:
Closes #3196

Besides adding internal events, this PR also fixes one bug that was discovered while writing this. It occurred when we failed in fetching container metadata from docker, causing us not to create and return `ContainerLogInfo`, which prevents the source from ever collecting the logs of that container. Was introduced in #1149.

### DIFF
--- a/src/internal_events/docker.rs
+++ b/src/internal_events/docker.rs
@@ -1,6 +1,5 @@
 use super::InternalEvent;
 use metrics::counter;
-use serde_json::Error;
 
 #[derive(Debug)]
 pub struct DockerEventReceived {
@@ -32,7 +31,7 @@ pub struct DockerContainerEventReceived<'a> {
 
 impl<'a> InternalEvent for DockerContainerEventReceived<'a> {
     fn emit_logs(&self) {
-        trace!(
+        debug!(
             message = "received one container event.",
             container_id = %self.container_id,
             action = %self.action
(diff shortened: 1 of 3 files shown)


src/internal_events/docker.rs
```diff
@@ -0,0 +1,194 @@
...
+
+#[derive(Debug)]
+pub struct DockerContainerEventReceived<'a> {
+    pub container_id: &'a str,
+    pub action: &'a str,
+}
+
+impl<'a> InternalEvent for DockerContainerEventReceived<'a> {
+    fn emit_logs(&self) {
+        debug!(
```
fanatid, line 41:
```suggestion
        trace!(
```

Previously level for this level was `trace`.

fanatid (approved):
Looks great.

src/internal_events/docker.rs
ktff: It was, but it's more important than `DockerEventReceived` so it would be nice to have it more visible than those, but it is less important than `DockerContainer*watch`, so `debug` seems like a nice middle.

PR #3346, opened 2020-08-04
chore: update tokio in sources

fanatid:
Ref #2945. Finish tokio update in sources. `line_agg.rs` updated by @MOZGIII in https://github.com/timberio/vector/pull/3304


MOZGIII:
Note someone will have to do another pass on these after both this and #3304 is merged.

PR #3347, opened 2020-08-04
chore: update tokio for tests and transforms

fanatid:
Ref. #2945 

This PR update tokio in `test_util` and transforms (only one tranform use old tokio: `aws_ec2_metadata`).


fanatid:
Blocked by #3355

benches/http.rs
```diff
@@ -51,7 +51,7 @@ fn benchmark_http_no_compression(c: &mut Criterion) {
                 let mut rt = runtime::Runtime::new().unwrap();
 
                 let (topology, _crash) = rt.block_on_std(topology::start(config, false)).unwrap();
-                wait_for_tcp(in_addr);
+                rt.block_on_std(async move { wait_for_tcp(in_addr).await });
```
bruceg, line 54: I take it there's no async in this benchmark.

src/transforms/aws_ec2_metadata.rs
```diff
@@ -143,7 +141,7 @@ impl TransformConfig for Ec2Metadata {
         let refresh_interval = self
             .refresh_interval_secs
             .map(Duration::from_secs)
-            .unwrap_or(Duration::from_secs(10));
+            .unwrap_or_else(|| Duration::from_secs(10));
```
bruceg, line 144: I'm curious why this is needed. I presume it's a clippy lint. `Duration::from_secs` is `const` and `Duration` implements `Copy` according to the tokio docs, so a closure should be just extra work.

fanatid: Indeed, `Duration::from_secs` is `const` fn. Thanks @bruceg.

MOZGIII: https://godbolt.org/z/csrc88
https://godbolt.org/z/vocav6

I'd like to point out that the generated code is effectively the same with optimization, and without closures are still better:

https://godbolt.org/z/43dhh9
https://godbolt.org/z/x4b61n

With closures, the default `Duration` gets created lazily.

Thus, I don't think it's worth bothering to change `unwrap_or_else` to `unwrap_or` for const defaults in general or prefer it for that matter.

Issue #3338, opened 2020-08-04
K8s: reduce the load of E2E tests on the CI


MOZGIII:
## The problem

As was noted during the review, we might see CI contention with Kubernetes E2E tests. This happened.

The reason for this is:
- there are 15 jobs total, checking various cluster-version/runtime combinations
- the jobs take very long to run, cause each job builds a release docker image individually (actual tests are a minor fraction of the whole process)
- all this runs for each push at PRs

## Motivation

We want to make our CI work (i.e. not stall on these jobs), but at the same time, we also want to provide a very high quality of integration with Kubernetes. We need a solution that won't compromise *too much* of the quality for CI performance.

## Context & constraints

The obvious solution is to move the E2E jobs to nightly flow, but that will be a big issue because we need to let people know that Kubernetes integration is broken while they're in the process of working on the code - not after it's been merged to `master`. By that point, it's already too late.

Another trivial solution is to simply test fewer combinations. This is also discouraged, because, in general, we want all of the supported combinations covered by tests. The reason for this is there are just way too many edge cases, and the initial development process showed there can be very subtle bugs tied to a particular combination - so it's absolutely necessary to have the test coverage across all the supported environments. Our E2E tests currently provide a *very basic* coverage, and we *intend to grow* it in the future.

However, there's an intuitive observation that not every code change (PR) is likely to affect Kubernetes integration specifically. The majority of PRs have potential to either break vector for all platforms (rather than breaking just the k8s integration), or just generically work (again, working for k8s too). We can probably omit *some* E2E test cases for *most* of the PRs.  Thus, if we have the ability for PR owners to *very easily, and at any point* opt-in to running the full E2E suite, we can, by default, run only a small subset of the tests (aka smoke tests). A serious consideration on determining the details here is required though.

If we look on the test from the perspective of a developer waiting for his push to get a green CI status, the longest part of that wait, for the E2E test cases in general (we currently have K8s E2E tests and Test Harness) is the release build. It takes an enormous amount of time. It is a prerequisite for any kind of E2E test due to the nature of the tests.

Related to the point above, the artifacts resulting from the release build can be shared among the E2E tests.

## Proposed solution

1. Move E2E tests into a dedicated workflow file. We'll resue the steps and will run them via various triggers, so this makes a lot of sense.
2. Take a small subset of E2E jobs, and "mark" them as "running unconditionally".
[...]

PR #3348, opened 2020-08-04
chore(operations): Reduce number of CI jobs

Hoverbear:
Compress a lot of our CI jobs into fewer jobs.

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -24,39 +24,7 @@ jobs:
         with:
           access_token: ${{ secrets.GITHUB_TOKEN }}
 
-  check-clippy:
-    name: Clippy - Linux
-    runs-on: ubuntu-20.04
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/cache@v2
-        with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - run: sudo bash scripts/environment/bootstrap-ubuntu-20.04.sh
-      - run: bash scripts/environment/prepare.sh
-      - run: make slim-builds
-      - run: make check-clippy
-
-  check-fmt:
-    name: Format
-    runs-on: ubuntu-20.04
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/cache@v2
-        with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - run: sudo bash scripts/environment/bootstrap-ubuntu-20.04.sh
-      - run: bash scripts/environment/prepare.sh
-      - run: make slim-builds
-      - run: make check-fmt
-
-  check-markdown:
+  checks:
     name: Markdown
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/lints.yml
```diff
@@ -104,86 +64,3 @@ jobs:
...
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - run: bash scripts/environment/bootstrap-macos-10.sh
-      - run: bash scripts/environment/prepare.sh
-      - run: make slim-builds
-      - run: make check
-
-  check-windows:
```
jszwedko, line 172: I take it we don't need to do these checks on Windows and OSX?

.github/workflows/lints.yml
```diff
@@ -71,22 +39,14 @@ jobs:
...
-      - uses: actions/checkout@v2
-      - uses: actions/cache@v2
-        with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - run: sudo bash scripts/environment/bootstrap-ubuntu-20.04.sh
-      - run: bash scripts/environment/prepare.sh
-      - run: make slim-builds
```
jszwedko, line 88: Would it make sense to still use `slim-builds` for this check?

.github/workflows/tests.yml
```diff
@@ -288,9 +288,9 @@ jobs:
           - "v1.11.0" # https://github.com/kubernetes/minikube/issues/8799
         kubernetes_version:
           - "v1.18.6"
-          - "v1.17.9"
-          - "v1.16.12" # v1.16.13 is broken, see https://github.com/kubernetes/minikube/issues/8840
-          - "v1.15.12"
+          #- "v1.17.9"
```
jszwedko, line 291: I assume this'll be augmented by https://github.com/timberio/vector/issues/3338 ?

.github/workflows/tests.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       - run: bash scripts/environment/prepare.sh
       - run: make slim-builds
       - run: make test-shutdown
-      # Currently has too few tests to justify a separate task in CI.
+      - run: make test-behavior
```
jszwedko, line 119: It seems like this is also run by `test-behavior`. Do we need it here too?

.github/workflows/lints.yml
Hoverbear: `check` is a subset of `test` which is already run, so yes.

.github/workflows/tests.yml
Hoverbear: Whoops I grabbed the wrong one! Nice catch! omg!

.github/workflows/tests.yml
Hoverbear: Yeah! I just disabled the mid-versions for now. I'm not sure that's a safe bet, but it'll ease off a lot of load. I think #3338 is doing to augment/resolve this.

distribution/nix/default.nix.erb
```diff
@@ -1,4 +1,4 @@
-# Taken from https://github.com/nixos/nixpkgs
+I # Taken from https://github.com/nixos/nixpkgs
```
jszwedko, line 1: :eyes: Should this be here?

Hoverbear: Definitely not...

.github/workflows/lints.yml
jszwedko: Doh, missed that :smile:

PR #3351, opened 2020-08-04
feat(splunk_hec sink): Make sourcetype templatable

jszwedko:
Allows for setting the sourcetype dynamically based on an event field.

Related: #3251 

### DIFF
--- a/src/sinks/splunk_hec.rs
+++ b/src/sinks/splunk_hec.rs
@@ -1,12 +1,13 @@
 use crate::{
     event::{self, Event, LogEvent, Value},
-    internal_events::{SplunkEventEncodeError, SplunkEventSent},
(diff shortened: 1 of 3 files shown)


src/sinks/splunk_hec.rs
```diff
@@ -122,6 +123,19 @@ impl HttpSink for HecSinkConfig {
...
+            .as_ref()
+            .map(|sourcetype| {
+                sourcetype
+                    .render_string(&event)
+                    .map_err(|missing_keys| {
+                        emit!(SplunkSourceTypeMissingKeys { keys: missing_keys });
+                    })
+                    .ok()
+            })
+            .unwrap_or(None);
```
lukesteensen, line 137: To avoid nesting `Option`s here you should be able to replace `map` with `and_then` and get rid of the unwrap.

jszwedko: And so I can :smile: That is better. 3271c029

PR #3367, opened 2020-08-06
fix(operations): Add build.rs to CI trigger paths

MOZGIII:


fanatid:
Should be aligned with #3364

PR #3415, opened 2020-08-10
chore(administration): Roll back to Rust 1.44 to fix nightlies

Hoverbear:
The nighlies are broken, this should fix them until we upgrade our builders to LLVM10.


jszwedko:
This fixes the binary builds but it looks like the release steps of the nightly workflow fail due to armhf artifacts being referred to, but no longer build. I'll open a followup PR for that.

https://github.com/timberio/vector/actions/runs/205841251

PR #3416, opened 2020-08-10
chore: break out top-level config module

lukesteensen:
There's some light cleanup and renaming here, but the vast majority is extracting the existing config-related code into its own module outside of `topology`. This should make the upcoming diffs cleaner and easier to review as we pull more logic out of topology and into config compilation.


ktff:
It seems `topology_required_healthcheck_fails_start` is failing for a reason, but `remove_file` and `test_reclaim_disk_space` are just flaky. It seems #3306 didn't help, should they be yanked out?

lukesteensen:
Going to go ahead and merge since the only remaining failure are known issues.

@ktff Good catch on `topology_required_healthcheck_fails_start`, fixed an incorrect edit there. Only remaining is `test_reclaim_disk_space` which should probably be pulled out.

PR #3418, opened 2020-08-11
enhancement(split transform): Add instrumentation

### LINKED ISSUE #3407: Define events for the `split` transform
As part of finishing https://github.com/timberio/vector-product/issues/14, we need to define internal events for the `split` transform.

## Defining events

## Requirements

- [ ] Add a `SplitEventProcessed` event.

ktff:
Closes #3407 

### Open questions

- [x] There is one more error to be added, but I'm not sure that it actually isn't a feature. Transform accepts defined number of fields but also less. Didn't find any mention of this in the documentation, so it may be a parse error.  (Edit: let's consider it a robustness feature)


lukesteensen (approved):

ktff:
Noting that `sinks::aws_kinesis_firehose::integration_tests::firehose_put_records` has [failed](https://github.com/timberio/vector/pull/3418/checks?check_run_id=980470948), but it doesn't seem to be because of this PR.

PR #3432, opened 2020-08-12
chore(administration): Remove some more armv7 bits

jszwedko:
The armhf assets were still being referred to in the nightly build, but
are no longer being built.


jszwedko:
Kicked off nightly workflow to test: https://github.com/timberio/vector/actions/runs/205978120

jszwedko:
Hmm, it skipped the steps I was interested in in the nightly workflow. I'll merge and see what happens overnight when it runs as normal.

PR #3436, opened 2020-08-13
 enhancement(file source): Even more instrumentation

### LINKED ISSUE #3197: Define events for the `file` source

ktff:
Closes #3197 
Follow up of #3417

Does:
* Renames `FileFingerprintFailed` to `FileChecksumFailed`, which @binarylogic suggested, as it's more accurate.
* Adds `FileCheckpointed`

### DIFF
--- a/lib/file-source/src/file_server.rs
+++ b/lib/file-source/src/file_server.rs
@@ -133,6 +133,7 @@ where
                 checkpointer
                     .write_checkpoints()
                     .map_err(|error| self.emitter.emit_file_checkpoint_write_failed(error))
+                    .map(|count| self.emitter.emit_file_checkpointed(count))
                     .ok();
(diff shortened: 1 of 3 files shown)


MOZGIII:
`FileChecksumFailed` reveals the implementation details in the interface. We can allow it, since it's under our control and we can always change it to be more in-line with the implementation - but generally speaking, I think the previous name was a better fit for the code point that emits the message.

MOZGIII (approved):
Great, the more events the better!

PR #3451, opened 2020-08-13
chore: Extract line_agg from file source

MOZGIII:
This PR extracts the `line_agg` module from the `file` source, and renames the standalone `file/mod.rs` into `file.rs`.

Pretty trivial change.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -26,6 +26,7 @@
 /src/tls.rs @bruceg
 /src/trace.rs @LucioFranco
 /src/types.rs @bruceg
+/src/line_agg.rs @MOZGIII
 
 /src/sinks/*_metrics.rs @lukesteensen
 /src/sinks/aws_kinesis_streams.rs @LucioFranco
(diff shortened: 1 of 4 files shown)


.github/CODEOWNERS
```diff
@@ -45,8 +46,7 @@
 /src/sinks/util/auto_concurrency @bruceg
 
 /src/sources/docker.rs @LucioFranco
-/src/sources/file/mod.rs @LucioFranco
-/src/sources/file/line_agg.rs @MOZGIII
+/src/sources/file.rs @LucioFranco
```
bruceg, line 49: This can probably be dropped, as he is no longer part of the team.

MOZGIII: True. I don't want this PR to introduce any changes like that though - to keed it simple and get in quickly.
I'd suggest we create a dedicated PR (or a series of dedicated PRs) to drop/reassign ownership.

MOZGIII:
We can think about that. There are pros and cons to that. Since it's not a blocker - let's create an issue to discuss it.

MOZGIII:
I've overlooked that we also need to share the multiline config among the sources - decided to add it here rather than to another PR.
@bruceg sorry to bother, but could you re-review this?

bruceg:
Is this PR blocking something? If not, the nature of this PR seems to be ideal for discussing where it should be moved.

MOZGIII:
> Is this PR blocking something? If not, the nature of this PR seems to be ideal for discussing where it should be moved.

I agree, let's discuss. I need this for my work to add the line aggregation at the docker source, but I can work off this PR's tree for now.

MOZGIII:
Rolled back the multiline extract change - I'll create a follow-up PR with it to address it separately.

MOZGIII:
I misinterpreted the concern regarding discussing the location of the `line_agg`. I'll create a dedicated issue.

PR #3460, opened 2020-08-14
chore: remove bytes:0.4

fanatid:
Not sure that issue for such PR exists :thinking: 

There can be places with extra bytes copies, not sure that I removed all of them.. =\


MOZGIII:
> There can be places with extra bytes copies, not sure that I removed all of them.. =\

I think we chould create an issue to make an audit for performance issues after these upgrades. Looking for potential extra copies mingt be one of the items there.

Also, I think it makes sense to use `BytesMut` at `Event`. It worth a separate issue to discuss properly.

juchiast:
> There can be places with extra bytes copies, not sure that I removed all of them.. =\

@fanatid I noticed some copies when working on #1484,

https://github.com/timberio/vector/blob/bytes-update/src/sinks/util/unix.rs#L156

https://github.com/timberio/vector/blob/bytes-update/src/sinks/util/tcp.rs#L245

fanatid:
@juchiast thank you for note, removed.
@MOZGIII sounds good, I'll create issue about `BytesMut` at `Event` after merge.

PR #3461, opened 2020-08-14
fix: doc test in event/metric

fanatid:
Introduced in #3434

### DIFF
--- a/src/event/metric.rs
+++ b/src/event/metric.rs
@@ -233,7 +233,7 @@ impl Display for Metric {
     /// and `Y` is the value or bucket.
     ///
     /// example:
-    /// ```
+    /// ```ignore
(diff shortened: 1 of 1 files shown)


src/event/metric.rs
```diff
@@ -233,7 +233,7 @@ impl Display for Metric {
     /// and `Y` is the value or bucket.
     ///
     /// example:
-    /// ```
+    /// ```ignore
```
jszwedko, line 236: This seems fine, but it also looks like we can use ```text to tell Rust that it is just text.

https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes

jszwedko (approved):

bruceg (approved):

src/event/metric.rs
fanatid: Good note, I'll change!

PR #3474, opened 2020-08-16
chore: replace tokio::test by own macros

### LINKED ISSUE #3360: Add own attribute macros for tests like tokio::test
Slowly we started to use [tokio::test](https://docs.rs/tokio/0.2.22/tokio/attr.test.html) macros for tests but in the same time, we need initialize logs for tracing in each test with [trace_init](https://github.com/timberio/vector/blob/c12ab4a103cf02e904d8be69fa07d006b5c1f4fa/src/test_util.rs#L48-L59) function. With own attributes macros (maybe with re-usage `tokio::test`) we will able activate trace logs outside the test function body. Also, sometimes we need a function name for the test, with own macros we can pass it as function argument.

fanatid:
Closes #3360.

New macros add `crate::test_util::trace_init()` in tests, we can not do this with `tokio::test`.

### DIFF
--- a/lib/test-util-macros/src/lib.rs
+++ b/lib/test-util-macros/src/lib.rs
@@ -1,0 +1,143 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2019 Tokio Contributors
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+use proc_macro::TokenStream;
+use quote::quote;
+use std::num::NonZeroUsize;
+
+#[derive(Clone, Copy, PartialEq)]
+enum Scheduler {
+    Basic,
+    Threaded,
+}
+
+#[proc_macro_attribute]
+pub fn test(args: TokenStream, item: TokenStream) -> TokenStream {
+    let args = syn::parse_macro_input!(args as syn::AttributeArgs);
+    let input = syn::parse_macro_input!(item as syn::ItemFn);
+    parse(args, input).unwrap_or_else(|e| e.to_compile_error().into())
+}
+
+fn parse(args: syn::AttributeArgs, mut input: syn::ItemFn) -> Result<TokenStream, syn::Error> {
+    let sig = &mut input.sig;
+    let body = &input.block;
+    let attrs = &input.attrs;
+    let vis = input.vis;
+
+    for attr in attrs {
+        if attr.path.is_ident("test") {
+            let msg = "second test attribute is supplied";
+            return Err(syn::Error::new_spanned(&attr, msg));
+        }
+    }
+
+    if !sig.inputs.is_empty() {
+        let msg = "the test function cannot accept arguments";
+        return Err(syn::Error::new_spanned(&sig.inputs, msg));
+    }
+
+    if sig.asyncness.is_none() {
+        let msg = "the async keyword is missing from the function declaration";
+        return Err(syn::Error::new_spanned(sig.fn_token, msg));
+    }
+
+    sig.asyncness = None;
+
+    let mut scheduler = None;
+    let mut core_threads = None;
+
+    for arg in args {
+        match arg {
+            syn::NestedMeta::Meta(syn::Meta::NameValue(namevalue)) => {
+                let ident = namevalue.path.get_ident();
+                if ident.is_none() {
+                    let msg = "Must have specified ident";
+                    return Err(syn::Error::new_spanned(namevalue, msg));
+                }
+                match ident.unwrap().to_string().to_lowercase().as_str() {
+                    "core_threads" => match &namevalue.lit {
+                        syn::Lit::Int(expr) => {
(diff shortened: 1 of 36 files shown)


src/config/watcher.rs
```diff
@@ -139,9 +141,8 @@ mod tests {
         tokio::time::timeout(timeout, fut).await.is_ok()
     }
 
-    #[tokio::test]
+    #[test_util::test]
```
jszwedko, line 144: I feel like `test_util::test` doesn't give a great idea of what this macro does. Without looking at it, I'd expect that I'd need to add this to _every_ test.

I feel like I lack the knowledge to suggest a better name though :smile: Maybe `test_util::async_test`? Or, following Tokio's naming, `util::async_test`? Or simply `async_test` if we move it to the main crate.

lib/test-util-macros/src/lib.rs
```diff
@@ -0,0 +1,143 @@
+// The MIT License (MIT)
```
jszwedko, line 1: Is there a way we could just include their macro? I'm concerned we'll drift from the upstream if we copy it like this.

src/sinks/util/buffer/metrics.rs
```diff
@@ -334,7 +334,7 @@ mod test {
         (buffered, sent_requests)
     }
 
-    #[tokio::test(core_threads = 2)]
+    #[test_util::test]
```
MOZGIII, line 337: I'm pretty sure this `core_threads` specification is important

src/config/watcher.rs
MOZGIII: What if we call the crate `vector_test`?

lib/test-util-macros/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    let result = quote! {
+        #[::core::prelude::v1::test]
+        #(#attrs)*
+        #vis #sig {
+            #rt
+                .enable_all()
+                .build()
+                .unwrap()
+                .block_on(async {
+                    crate::test_util::trace_init();
```
MOZGIII, line 136: This won't work with non-vector crates. So, does it make sense to have it in a separate crate, rather than in our `vector` crate?

src/sinks/util/buffer/metrics.rs
fanatid: I did not checked `tokio::test` deeply when initially modified this test, `tokio::test(core_threads = 2)` worked because this change basic_scheduler to threaded. Yesterday when explored code I realized that `tokio::test(threaded_scheduler)` will also work.
In our macros threaded_scheduler by default, so we ok with `test_util::test`.

lib/test-util-macros/src/lib.rs
fanatid: Yes, this will not work with non-vector crate, but [procedural macros](https://doc.rust-lang.org/reference/procedural-macros.html) can be defined only as external crates, so we can not define macros in `vector` crate itself.

src/config/watcher.rs
fanatid: We can use `test_util::test` for <i>every</i> test because this macros initialize trace logs. I can change macros in a way that `Runtime` will not be created for not `async` tests. This sounds like magic from another side.. :thinking: maybe two macroses: `sync_test` and `async_test`?

lib/test-util-macros/src/lib.rs
fanatid: Do you mean copy their code and only add `crate::test_util::trace_init()`?

bruceg, line 1: Does this have implications for the license on the main Vector package?

src/config/watcher.rs
fanatid: I'll change current name to `async_test` and add `sync_test`. I placed macros under `test_util::` because everything test related in this scope. Is there better place? `crate::async_test`?

jszwedko: > We can use test_util::test for every test because this macros initialize trace logs

If this is the case, would it be possible to just have that automatically done rather than annotate every test?

I like `sync_test` and `async_test`.

lib/test-util-macros/src/lib.rs
fanatid: Good question. Unfortunately, I do not have an answer. I placed license to file because `src/stream.rs` had the same solution: https://github.com/timberio/vector/blob/8e7a3674ffca11b13a3a87dcf69d7254f2083e98/src/stream.rs#L1-L23

jszwedko: I was more just wondering if we could embed their macro somehow, but I don't know if that is possible. I just didn't want to miss updates to the upstream macro.

MOZGIII:
Overall, I'm kind of not sure it's worth doing. I understand the appeal of putting all the boilerplate test initialization into a macro, but I'm not sure it's worth it, for the following reasons:

1. The core idea of having a macro that does all the initialization is not very appealing. Maybe it's a good tradeoff, but in general, it seems to be bound to either do extra work, or to be very complex and needing a lot of configurability.
2. Having our own macro to implement `async` wrapping is "extra weight". It's not trivial, and we might want to keep using the `tokio` one, or tap into their implementation instead of just copying. Generally speaking, copying code from other projects seems worse than just using it as a dependency - and in this case, I think, we must consider carefully if we really want this.

What if we have a macro that just enables logs in the test instead? We should be able to use it in combination with `tokio::test`:

```rust
#[vector_test::init_logs]
#[tokio::test]
async fn qwe() { ... }
```

and with non-async tests

```rust
#[vector_test::init_logs]
fn qwe() { ... }
```

Re-using the `tokio::test` explicitly is also better for the community - a lot of people will be familiar with `tokio::test` and how to use it.

One of the problems with having `vector_test::init_logs` and `tokio::test` separate is people might forget to use `vector_test::init_logs`. In some sense, the key benefit of having a single macro that would combine the standard `#[test]` or `#[tokio::test]` and custom setup required for `vector` is that it eliminates the need to remember to invoke our custom setup.

If the goal here is help with ensuring users don't forget to initialize logs - I suggest we also consider other options. Frankly, having a macro still doesn't ensure every test will have a properly initialized log - people might still use the usual `#[test]` or `#[tokio::test]` invocations.

I've heard we can use a custom test harness for cargo tests - maybe that would be a better solution? The key benefits are that we could make it so every test runs with a custom wrapper of our choosing - and we can add the proper log initialization there.
Then we won't even have to manually alter our tests. It also possible to implement logs initialization in a more sane way - to have the test executor initialize the logger once - it's effectively a shared resource among all the tests either way, even currently, so it makes sense to me.

Relevant refs:
- https://internals.rust-lang.org/t/test-and-external-test-harnesses/3145
- https://rust-lang.github.io/rfcs/2318-custom-test-frameworks.html

fanatid:
@jszwedko @MOZGIII @bruceg @lukesteensen thank you for comments. Initially, when I opened #3360 I thought that it will be cool have macro which initializes logs for us, because I often was needed to add `trace_init()` for checking logs when investigating issues with crates update. I checked a few times how `tokio::test` works in past and maybe because this macro does not look complex for me now. But, I completely agree that this additional macro can be strange for contributors, they will need to check how our macro work instead use what they used everywhere else (`#[test]` / `#[tokio::test]`). I also agree that we will need to support it, what is not good.

I didn't knew about test framework (I had seen `harness` keyword, but did not explore for what it required), thanks to @MOZGIII that mentioned it. I'll check can we use it or not.

@lukesteensen I'll add `trace_init` to tests on moving them to `tokio:0.2` runtime.

PR #3481, opened 2020-08-17
chore: remove tower:0.1

### LINKED ISSUE #2640: Upgrade service based sinks

### LINKED ISSUE #3482: Upgrade tower

fanatid:
Closes #3482
Closes #2640 

Currently blocked by #3422.~, so created as draft.~

Better left review aside while #3422 not merged, because the same changes in both pull requests.


fanatid:
Mac fail again with `process didn't exit successfully` :disappointed:

PR #3493, opened 2020-08-18
fix(performance): Only update auto-concurrency RTT from OK responses

bruceg:
Error responses from a service may happen much faster (or rarely much slower) than a successful response. This can dramatically skew the measurements, leading to incorrect changes to the concurrency limit.

Note that this does not directly affect the handling of HTTP 429 responses, as they are returned as an error within a `Response` rather than the `Error` types handled here. I will do that part next.

### DIFF
--- a/src/sinks/util/auto_concurrency/controller.rs
+++ b/src/sinks/util/auto_concurrency/controller.rs
@@ -255,7 +255,7 @@ impl EWMA {
         self.average
     }
 
-    /// Update and return the current average
+    /// Update the current average
     fn update(&mut self, point: f64) {
         self.average = match self.average {
             avg if avg == 0.0 => point,
--- a/src/sinks/util/auto_concurrency/controller.rs
+++ b/src/sinks/util/auto_concurrency/controller.rs
@@ -142,77 +142,81 @@ impl<L> Controller<L> {
             inner.current_rtt.update(rtt);
         }
         let current_rtt = inner.current_rtt.average();
-        let past_rtt = inner.past_rtt.average();
 
-        // No past measurements, set up initial values.
-        if past_rtt == 0.0 {
-            if let Some(current_rtt) = current_rtt {
-                inner.past_rtt.update(current_rtt);
-                inner.next_update = now + Duration::from_secs_f64(current_rtt);
-            }
-        } else if past_rtt > 0.0 && now >= inner.next_update {
-            #[cfg(test)]
-            {
+        match inner.past_rtt.average() {
+            None => {
+                // No past measurements, set up initial values.
                 if let Some(current_rtt) = current_rtt {
-                    stats.averaged_rtt.add(current_rtt, now);
+                    inner.past_rtt.update(current_rtt);
+                    inner.next_update = now + Duration::from_secs_f64(current_rtt);
                 }
-                stats.concurrency_limit.add(inner.current_limit, now);
-                drop(stats); // Drop the stats lock a little earlier on this path
             }
+            Some(mut past_rtt) => {
+                if now >= inner.next_update {
+                    #[cfg(test)]
+                    {
+                        if let Some(current_rtt) = current_rtt {
+                            stats.averaged_rtt.add(current_rtt, now);
+                        }
+                        stats.concurrency_limit.add(inner.current_limit, now);
+                        drop(stats); // Drop the stats lock a little earlier on this path
+                    }
 
-            if let Some(current_rtt) = current_rtt {
-                emit!(AutoConcurrencyAveragedRtt {
-                    rtt: Duration::from_secs_f64(current_rtt)
-                });
-            }
+                    if let Some(current_rtt) = current_rtt {
+                        emit!(AutoConcurrencyAveragedRtt {
+                            rtt: Duration::from_secs_f64(current_rtt)
+                        });
+                    }
 
-            // Only manage the concurrency if in_flight_limit was set to "auto"
-            if self.in_flight_limit.is_none() {
-                let threshold = past_rtt * THRESHOLD_RATIO;
+                    // Only manage the concurrency if in_flight_limit was set to "auto"
+                    if self.in_flight_limit.is_none() {
+                        let threshold = past_rtt * THRESHOLD_RATIO;
 
-                // Normal quick responses trigger an increase in the
-                // concurrency limit. Note that we only check this if we had
-                // requests to go beyond the current limit to prevent
-                // increasing the limit beyond what we have evidence for.
-                if inner.current_limit < super::MAX_CONCURRENCY
-                    && inner.reached_limit
-                    && !inner.had_back_pressure
-                    && current_rtt.is_some()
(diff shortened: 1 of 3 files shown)


jszwedko (commented):
I'm seeing some interesting behavior with this change. I added a `CLOSE` mode to the test server (https://github.com/timberio/http_test_server/pull/1/commits/eaf85e5706fdb9a83455418496b9721540f266f9) that simply closes the connection if the limit is hit. I tried it before and after these changes with a rate limit of 10 req / s.

On master:

concurrent_requests

On this branch:
concurrent_requests

It looks like, with these changes, it is correctly not counting them in the mean calculation, but it also doesn't appear to be backing off like I expected. What do you think? It's possible that my server implementation isn't manifesting in the correct signal here (it simply closes the connection without writing any HTTP response).

src/sinks/util/auto_concurrency/controller.rs
```diff
@@ -234,12 +251,11 @@ impl EWMA {
     }
 
     /// Update and return the current average
```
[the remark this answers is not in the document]
bruceg: Oops, thanks.

bruceg:
What it looks like is happening illustrates a discrepancy between how many requests Vector thinks is outstanding vs how many the test server knows is outstanding. When the test server closes the connection, it immediately counts the request as "done". However, to Vector, the request is still being processed until the timeout hits. So, it pauses with no request being issued until the timeouts completes, and then issues a bunch more. My confusion is why the concurrency didn't back off. It appears that when it starts sending requests again, it goes right back to the peak concurrency of 10, but a timeout should show as backpressure and cause it to clamp down. The trace logs and metrics from Vector would be useful here. I'll try to run the test as well.

jszwedko:
> So, it pauses with no request being issued until the timeouts completes, and then issues a bunch more

Hmm, I might be missing something, but in this scenario the test server just closes the connections; would it still wait for the timeout in that case? The graphs seem to show it issuing a cumulative 10 requests per second which is what I'd expect with a rate limit of 10 req/s. It's the lack of back-off that I'm uncertain about.

Some new plots with the added request rate graph:

master (b2fea480):

plot

this branch:

plot

The test command I'm using is:

`HTTP_TEST_RATE_LIMIT_BEHAVIOR=CLOSE HTTP_TEST_RATE_LIMIT_BUCKET_FILL_INTERVAL=1s HTTP_TEST_RATE_LIMIT_BUCKET_CAPACITY=10 HTTP_TEST_RATE_LIMIT_BUCKET_QUANTUM=10 HTTP_TEST_LATENCY_NORMAL_MEAN=250ms TEST_TIME=15 ./bin/run-concurrency-test`

If it is helpful, here is the vector output:

[test_cmd.log](https://github.com/timberio/vector/files/5105635/test_cmd.log)

bruceg:
That's interesting, but if I'm seeing it right, it's actually close to optimal behavior based on the request rate graph. It has "found" the 10 requests per second limit and stays there.

jszwedko:
Hmm. I guess that's true. I was thinking it should back off because of the error condition, but in this case, if a server really did just drop anything over 10 requests / sec with no other server impacts (like causing other requests to degrade), this would be the optimal vector behaviour. I'll have to think about this more.

src/sinks/util/auto_concurrency/controller.rs
```diff
@@ -131,37 +138,47 @@ impl<L> Controller<L> {
...
 
-        let current_rtt = inner.current_rtt.update(rtt);
+        if use_rtt {
+            inner.current_rtt.update(rtt);
+        }
+        let current_rtt = inner.current_rtt.average();
         let past_rtt = inner.past_rtt.average();
 
+        // No past measurements, set up initial values.
         if past_rtt == 0.0 {
```
MOZGIII, line 148: Have you considered making `EWMA::average` return `Option` like `Mean`? It looks like it would be better for consistency, and it'll also make for a better-structured use of the value - is would ensure there'd be no way to interpret the first measurement incorrectly, and address other problems of that sort.

src/sinks/util/auto_concurrency/controller.rs
```diff
@@ -131,37 +138,47 @@ impl<L> Controller<L> {
...
+                inner.next_update = now + Duration::from_secs_f64(current_rtt);
+            }
         } else if past_rtt > 0.0 && now >= inner.next_update {
             #[cfg(test)]
             {
-                stats.averaged_rtt.add(current_rtt, now);
+                if let Some(current_rtt) = current_rtt {
+                    stats.averaged_rtt.add(current_rtt, now);
+                }
                 stats.concurrency_limit.add(inner.current_limit, now);
```
MOZGIII, line 159: Is this correct behavior? If we had a non-rtt response - should we include it into the stats even?

If not - it seems like we can turn off this whole codepath - starting from the `if` starting the chain and up until `if self.in_flight_limit.is_none() {`... Or maybe even more than that.

src/sinks/util/auto_concurrency/controller.rs
```diff
@@ -182,14 +200,16 @@ impl<L> Controller<L> {
...
                 });
             }
 
             // Reset values for next interval
-            let new_avg = inner.past_rtt.update(current_rtt);
-            inner.next_update = now + Duration::from_secs_f64(new_avg);
+            if let Some(current_rtt) = current_rtt {
+                inner.past_rtt.update(current_rtt);
+            }
+            inner.next_update = now + Duration::from_secs_f64(inner.past_rtt.average());
```
MOZGIII, line 212: This seems like it can end up being `inner.next_update = now + Duration::from_secs_f64(0.0);` - in case the `EWMA` didn't update.

It's a bit hard to follow along that are the consequences of `current_rtt.average()` being `None`.

src/sinks/util/auto_concurrency/controller.rs
bruceg: I hadn't, but that makes sense.

src/sinks/util/auto_concurrency/controller.rs
bruceg: It is possible for there to be valid data in the `current_rtt` average despite the most recent response should not be used for RTT calculation. Consider if we receive 4 acknowledgements and one error before `inner.next_update` is reached. Then `current_rtt` will contain the average of those 4, even though the most recent was an error, and we should use that data for this interval.

src/sinks/util/auto_concurrency/controller.rs
bruceg: No, if the `past_rtt` wasn't updated, it still has a value since this part of the condition cannot be entered unless `past_rtt` had a value. I agree this is unclear so I'll rework the logic to make it clearer.

jszwedko:
> So are you good with merging this as a step forward, or is there something remaining to adjust in the logic?

:+1: it seems like this is a step forward. I'm fine with seeing it merged.

jszwedko (approved):

Issue #3478, opened 2020-08-17
Remove tokio-compat


fanatid:
#3449 remove `tokio:0.1`, but we still have `tokio-compat` because use `futures:0.1` in a lot of places.

~Before remove `tokio-compat` I propose update all tests to `crate::test_util::test` macros (instead manually created `Runtime` in each test). Macros added in #3474.~

Plan:
- [x] Replace `runtime().block_on*` to `#[tokio::test]`. #3505 #3626 
- [x] Investigate why some test require `threaded_scheduler` and/or `core_threads = 2`. #3578 #3596 
- [x] Update benches #3626
- [x] Remove tokio-compat #3627

MOZGIII:
Is the plan to also remove our `Runtime` abstraction along with the `tokio-compat`? The reason I'm asking is I think it may be useful even after tokio upgrade - however it's not totally *necessary* so we might want to cut it for simplicity...

lukesteensen:
> Is the plan to also remove our `Runtime` abstraction along with the `tokio-compat`?

My assumption is that we will be using `Runtime` in extremely few places as we move off `tokio-compat` and onto `#[tokio::test]`. That makes it much less important and I'd be in favor of just removing it unless there is a good reason to keep it.

PR #3505, opened 2020-08-19
chore: add more tokio::test

fanatid:
Part of #3478.

Add `tokio::test` from `tokio:0.2` instead `runtime()` whish uses `tokio-compat` almost to each tests. I tried do not change code itself, except removing not required functions anymore.

Some tests require `threaded` scheduler and `core_threads = 2` for working, will investigate and fix this in future PR's.

### DIFF
--- a/src/config/watcher.rs
+++ b/src/config/watcher.rs
@@ -121,7 +121,7 @@ fn add_paths(watcher: &mut RecommendedWatcher, config_paths: &[PathBuf]) -> Resu
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_util::temp_file;
+    use crate::test_util::{temp_file, trace_init};
     use futures::compat::Future01CompatExt;
     use futures01::{Future, Stream};
     use std::time::Duration;
@@ -141,7 +141,8 @@ mod tests {
 
     #[tokio::test]
     async fn file_update() {
-        crate::test_util::trace_init();
+        trace_init();
+
         let delay = Duration::from_secs(3);
         let file_path = temp_file();
         let mut file = File::create(&file_path).unwrap();
@@ -155,7 +156,8 @@ mod tests {
 
     #[tokio::test]
     async fn sym_file_update() {
-        crate::test_util::trace_init();
+        trace_init();
+
         let delay = Duration::from_secs(3);
         let file_path = temp_file();
         let sym_file = temp_file();
--- a/src/config/watcher.rs
+++ b/src/config/watcher.rs
@@ -121,7 +121,7 @@ fn add_paths(watcher: &mut RecommendedWatcher, config_paths: &[PathBuf]) -> Resu
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_util::temp_file;
+    use crate::test_util::{trace_init, temp_file};
     use futures::compat::Future01CompatExt;
     use futures01::{Future, Stream};
     use std::time::Duration;
@@ -141,6 +141,8 @@ mod tests {
 
     #[tokio::test]
     async fn file_update() {
+        trace_init();
+
         let delay = Duration::from_secs(3);
         let file_path = temp_file();
         let mut file = File::create(&file_path).unwrap();
@@ -154,6 +156,8 @@ mod tests {
 
     #[tokio::test]
     async fn sym_file_update() {
+        trace_init();
+
         let delay = Duration::from_secs(3);
         let file_path = temp_file();
         let sym_file = temp_file();
--- a/src/config/watcher.rs
+++ b/src/config/watcher.rs
@@ -121,7 +121,7 @@ fn add_paths(watcher: &mut RecommendedWatcher, config_paths: &[PathBuf]) -> Resu
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_util::{trace_init, temp_file};
+    use crate::test_util::{temp_file, trace_init};
     use futures::compat::Future01CompatExt;
     use futures01::{Future, Stream};
     use std::time::Duration;
--- a/src/config/watcher.rs
+++ b/src/config/watcher.rs
@@ -121,7 +121,7 @@ fn add_paths(watcher: &mut RecommendedWatcher, config_paths: &[PathBuf]) -> Resu
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_util::{temp_file, trace_init};
+    use crate::test_util::temp_file;
     use futures::compat::Future01CompatExt;
     use futures01::{Future, Stream};
     use std::time::Duration;
@@ -141,8 +141,6 @@ mod tests {
 
     #[tokio::test]
     async fn file_update() {
-        trace_init();
-
         let delay = Duration::from_secs(3);
         let file_path = temp_file();
         let mut file = File::create(&file_path).unwrap();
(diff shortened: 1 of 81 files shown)


src/config/watcher.rs
```diff
@@ -141,7 +141,6 @@ mod tests {
 
     #[tokio::test]
     async fn file_update() {
-        crate::test_util::trace_init();
```
MOZGIII, line 144: We probably shouldn't remove these from here.

MOZGIII:
I'd also suggest that we keep the tests intact for now. It's quite hard to validate on a scope this large.

src/config/watcher.rs
MOZGIII: It looks like there are a lot of removals of the `crate::test_util::trace_init`, while this PR looks like it should add them. I'm confused, was the intent to remove or add them?

fanatid: Oops, looks like I removed this by accident. Will return back.

> It looks like there are a lot of removals of the `crate::test_util::trace_init`

In some places yes, I removed `trace_init` from setup function and add to functions itself.

> while this PR looks like it should add them.

Initially, I wanted to add `trace_init` to every function, but after @lukesteensen comment decide left everything. Changed title to more appropriate.

src/kubernetes/reflector.rs
```diff
@@ -447,581 +446,574 @@ mod tests {
...
-                );
-
-                // Send pod addition to a stream.
-                watch_stream_tx
-                    .send(mock_watcher::ScenarioActionStream::Ok(WatchResponse::Ok(
-                        WatchEvent::Added(make_pod("uid0", "10")),
-                    )))
-                    .await
-                    .unwrap();
+    #[tokio::test(threaded_scheduler, core_threads = 2)]
```
MOZGIII, line 538: Why `core_threads = 2` here but not in other tests?

The amount of the threads (or even the runtime kind) should not be relevant for these tests - if they fail in any of the possible configurations - it's a bug.

fanatid: `threaded_scheduler` / `core_threads = 2` added because otherwise test hung up. I completely understand that we should able to use `tokio::test` in the default configuration and want to address this issue in future PRs.

MOZGIII: You mind creating an issue for this, and other things like this you've encountered? It's hard to tell if the change was problematic from the review. This one is easy to spot, but I assume there are other tiny details without which tests fail but and that we should look into more closely.

src/kubernetes/state/delayed_delete.rs
```diff
@@ -166,131 +166,134 @@ mod tests {
...
-        })
+        // Unfreeze time.
+        tokio::time::resume();
     }
 
-    #[test]
-    fn resync() {
+    #[tokio::test(threaded_scheduler)]
+    async fn resync() {
         no_maintenance_test_flow(
```
MOZGIII, line 293:
```suggestion
        trace_init();
        no_maintenance_test_flow(
```

src/kubernetes/state/evmap.rs
```diff
@@ -169,7 +169,6 @@ mod tests {
     async fn test_with_debounce() {
         // Due to https://github.com/tokio-rs/tokio/issues/2090 we're not
         // pausing the time.
-
```
MOZGIII, line 172: This comment is not related to the lines directly below it, but rather to the beginning of the test body. This newline is effectively a separator that indicates that. Let's preserve it.

MOZGIII:
Regarding the use of `#[tokio::test(threaded_scheduler)]` - https://docs.rs/tokio/0.2.22/tokio/attr.test.html doesn't mention the `threaded_scheduler` option, and the last version it's mentioned is `tokio` `0.2.11` - https://docs.rs/tokio/0.2.11/tokio/attr.test.html. I wonder what's up with that. Are they deprecating it? If so, we probably can just use the standard `#[tokio::test]`.

src/kubernetes/state/delayed_delete.rs
fanatid: Changes. Thanks!

fanatid:
> #[tokio::test(threaded_scheduler)]

I accidentally found this when checked `tokio::test` source code, I do not see any comments / lines about deprecating scheduler selection.

https://docs.rs/tokio-macros/0.2.5/src/tokio_macros/entry.rs.html#105-121

src/kubernetes/reflector.rs
fanatid: I added an update plan in https://github.com/timberio/vector/issues/3478#issue-680213492. Sorry that did not mention all ths things before :disappointed:

MOZGIII:
Yeah, they're re-adding the docs: https://github.com/tokio-rs/tokio/pull/2697 (for `tokio::main`, but it shows they just somehow lost the docs in the process).

~On a separate note - isn't `threaded_scheduler` the default?~ It's not for tests, we have to set it.

tests/tcp.rs
```diff
@@ -203,55 +192,52 @@ fn merge_and_fork() {
         sinks::socket::SocketSinkConfig::make_basic_tcp_config(out_addr2.to_string()),
     );
 
-    let mut rt = Runtime::with_thread_count(2).unwrap();
```
MOZGIII, line 206: This one looks like it requires `core_threads = 2`

fanatid: Currently, CI is running but if I remember correctly this test is fine with 1 thread.

MOZGIII: The change to use two threads originated from here: https://github.com/timberio/vector/pull/3340
It seems you have enough context on what's going on, so this should be fine if you say so :smile:

src/kubernetes/reflector.rs
MOZGIII: I'll create an issue for the k8s reflector tests in particular for myself. It looks like this change introduces the behaviour I didn't account for, and I'd like to ensure tests remain reliable in the long run.

tests/topology.rs
```diff
@@ -38,7 +42,7 @@ fn into_message(event: Event) -> String {
         .to_string_lossy()
 }
 
-#[tokio::test(core_threads = 2)]
+#[tokio::test(threaded_scheduler)]
```
MOZGIII, line 45: Why change this one? It looked fine...

fanatid: I do not know why this was not worked with basic scheduler, I guess that because we have spawns in the test. Anyway, `threaded_scheduler` / `core_threads` is fine for now, but will be removed later. This PR only addresses `tokio::test` macro addition, otherwise will be hard review all changes.

MOZGIII: Ah, so it was broken, and `threaded_scheduler` instead of `core_threads = 2` fixes things? Just looks odd to me as this test already uses `#[tokio::test]` macro, and there are no changes to the internals of the test itself. I assumed it was fine (i.e. not broken), but if it is broken - makes sense.

I'd be surprised if our use of `spawn`s required multiple threads. That'd mean we have some major concurrency management flaws in the code.

fanatid: Yeah, a lot of tests had `core_threads = 2` before, I added this attribute when I moved them to `tokio::test` (because otherwise tests were hanged up). `threaded_scheduler` works because of `core_threads` with value more than 1 produce threaded scheduler in macro.

MOZGIII (approved):
I didn't look through all of the code, but I've completed the review of the parts I've worked on recently.
There's an issue but it can be resolved outside of this PR.

PR #3513, opened 2020-08-20
chore(lua transform): Add events to lua v2 transform.

Hoverbear:
Add metrics to lua v2 transform, resolving #3400 .

(Please give a bit of extra care with this review, I am not that familiar with the lua stuff.)


src/transforms/lua/v2/mod.rs
```diff
@@ -329,7 +334,7 @@ impl RuntimeTransform for Lua {
 }
 
 #[cfg(test)]
-fn format_error(error: &rlua::Error) -> String {
+pub(crate) fn format_error(error: &rlua::Error) -> String {
```
lukesteensen, line 337: Why this change? Doesn't seem like we're using it anywhere new.

Hoverbear: I... have no idea.

PR #3520, opened 2020-08-21
chore(wasm transform): Make WASM transform conform to RFC

Hoverbear:
Add metrics to wasm transform, resolving #3411 .

Note this deviates a bit from the RFC, but this is with @lukesteensen 's blessing.


src/internal_events/wasm/compilation.rs
```diff
@@ -5,76 +5,80 @@ use metrics::counter;
...
+        emit!(me.clone());
         me
     }
     pub fn complete(self) {
         emit!(Self {
             state: State::Completed,
             role: self.role,
-            #[cfg(feature = "wasm-timings")]
+            error: self.error,
+            epoch: self.epoch,
```
lukesteensen, line 35: Could these be reduced down to `..self`?

src/internal_events/wasm/compilation.rs
```diff
@@ -5,76 +5,80 @@ use metrics::counter;
...
+            elapsed: self.epoch.elapsed()
+        })
+    }
+    pub fn error(self, error: String) {
```
lukesteensen, line 39: I'm a little surprised `rustfmt` doesn't enforce an extra newline between the previous method's closing brace and the start of this one.

src/wasm/hostcall/mod.rs
```diff
@@ -95,50 +95,78 @@ mod wrapped_for_ffi {
...
     #[lucet_hostcall]
     #[no_mangle]
     pub extern "C" fn emit(vmctx: &Vmctx, data: u32, length: u32) -> u32 {
-        let internal_event = internal_events::Hostcall::begin(Role::Transform, "register");
-        // TODO: Handle error.
-        let ret = super::emit(vmctx, data, length).unwrap();
-        internal_event.complete();
-        ret
+        let internal_event =
+            internal_events::WasmHostcallProgress::begin(Role::Transform, "register");
```
lukesteensen, line 111: Looks like we have `"register"` being passed in `emit` and `"emit"` being passed in `register`.

src/internal_events/wasm/hostcall.rs
```diff
@@ -44,30 +60,31 @@ impl Hostcall {
...
                     role = self.role.as_const_str(),
-                );
-            } else {
-                trace!(
+                    elapsed_micros = self.elapsed.as_micros() as u64,
+                },
+                "WASM Hostcall invocation.",
+            ),
+            State::Errored => event!(
+                tracing::Level::ERROR,
```
lukesteensen, line 77: Why `event!` with a hardcoded level instead of the normal level-specific macros?

src/internal_events/wasm/compilation.rs
Hoverbear: Way nicer. :)

src/internal_events/wasm/compilation.rs
Hoverbear: Added it!

src/internal_events/wasm/hostcall.rs
Hoverbear: I should use those instead. :) I had no reason.

Hoverbear:
#3606

PR #3526, opened 2020-08-22
chore: less work on shutdown test

fanatid:
Noticed when investigating issue introduced in #2149


ktff:
Good catch but they are still failing. 

`env = "VECTOR_QUIET"` is definitely the cause, no clue why.

fanatid:
It's not a bugfix. I wanted to fix it but wasn't able to do this quickly, so submit only `Makefile` changes.

ktff (approved):

jamtur01:
Thanks for catching this. 👍

MOZGIII (approved):

PR #3560, opened 2020-08-25
fix: Fixed language on code fence in CONTRIBUTING.md

jamtur01:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -172,7 +172,7 @@ Currently Vector uses Github Actions to run tests. The workflows are defined in
 
 Tests are run for all changes except those that have the label:
 
-```
+```sh
 ci-condition: skip
 ```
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -172,7 +172,7 @@ Currently Vector uses Github Actions to run tests. The workflows are defined in
 
 Tests are run for all changes except those that have the label:
 
-```
+```sh
```
jszwedko, line 175: It seems like we want `text` here? This isn't shell code.

jamtur01: Is `text` a language GH's markdown recognizes? I couldn't find it in the list.

jamtur01: Hmmm it is in Linguist. I'll update.

jszwedko: Hmm, yeah, they refer to https://github.com/github/linguist/blob/master/lib/linguist/languages.yml for the list in https://docs.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks#syntax-highlighting but it isn't clear to me how those map to the the language identifier allowed for codeblock fences.

Text is listed in there though: https://github.com/github/linguist/blob/082fd09369b607b38112bfaf68640b4d44b5f328/lib/linguist/languages.yml#L5541-L5580

PR #3561, opened 2020-08-25
enhancement: Rewrite LineAgg to allow passing context data and be more memory efficient

MOZGIII:
This PR alters `LineAgg` such that:

- it can pass around an extra piece of data together with each `line` - the `context`; this is required for some use cases - more on this below;
- it uses `Vec<Bytes>` instead of `ByteMut` to aggregate lines, and only does `BytesMut` once when the final event is emitted; this should reduce memory pressure and allocs count;
- it has different arguments order - the way it was before was natual for `file` source, but not for the current generic implementation.

## Motivation for adding context

`LineAgg` has to keep all the data nesessary to produce the `Event` from the aggragted `Bytes`. This is because we sometimes emit more than one line from the aggregation results - and the consumers of those merged line have to somehow transfer that data through the line aggregation mechnism.

Now, for `file` source it was very convenient that the `LineAgg` implementation naturally passed around all the required data - the `line` - for the line being processed - and the `src` - which has the semantics of the "file name" for the `file` source, and the aggregation key for the `LineAgg`. Given just the `line` and the `src`, `file` source could work.

The same approach doesn't work for sources that want to pass around additional metadata together with events. For instance, `docker` source parses the raw message into

- a timestamp, 
- labels and 
- message line. 

To properly implement `LineAgg`, we need to memorize those thought the merging process. Why can't we just keep that data outside of the merging logic? Well, that's because of the case when we want to emit multiple messages from the aggregation. If not that - it would be fine, but emitting multiple messages required us to somehow persist the content for the first message - otherwise what would we use to build the event from it?

In docker case, what's interesting is that we don't need the `src` at all - the processing there is already scoped by the container, and it's enough to just consume the stream and run the `LineAgg` logic.

### DIFF
--- a/src/line_agg.rs
+++ b/src/line_agg.rs
@@ -2,6 +2,7 @@
 
 #![deny(missing_docs)]
 
+use bytes::{Bytes, BytesMut};
 use futures::{Stream, StreamExt};
 use pin_project::pin_project;
 use regex::bytes::Regex;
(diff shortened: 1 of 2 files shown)


src/line_agg.rs
```diff
@@ -340,25 +349,54 @@ where
...
+            bytes_mut.extend_from_slice(&line);
+        }
+        (bytes_mut.freeze(), self.context)
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use bytes05::Bytes;
```
fanatid, line 399: We have only one version of `bytes` now :slightly_smiling_face:

PR #3565, opened 2020-08-25
chore: Move CI tests to container image and eliminate docker-compose for speed

jamtur01:
This provides some selective speed-ups to CI runs. It does it by:

1. Removing the bootstrap and prepare scripts and replacing them with a pre-prepared Docker image (https://hub.docker.com/r/timberio/ci_image) for selective tests. 
2. Replacing docker-compose for some tests.

This cuts an average of 4-5 minutes off the run of the targeted tests. In a couple of cases, not enough however, it's 10-15 minutes.

The following tests have been migrated:

1. All of the linting suite.
2. All of the bench suite.
3. The Linux - Unit tests
4. The GCP, Loki, and Docker integration tests.

The other tests (and release functions) require some more thought. The use of docker-compose should be replaceable with GH Actions service containers but it seems their networking behavior is non-standard and I need to work out exactly what's happening. This interim step seems to have tangible value though so I am pushing it out for review.


.github/workflows/lints.yml
```diff
@@ -30,6 +30,8 @@ jobs:
     runs-on: ubuntu-20.04
     if: |
       !contains(github.event.pull_request.labels.*.name, 'ci-condition: skip')
+    container:
+      image: timberio/ci_image:latest
```
Hoverbear, line 34: So it was my general plan to build the environment via `make environment` in a pre-step of the CI (when needed) and then have all further steps use this feature to use that environment. :)

jamtur01: Per workflow or per job?

Hoverbear: Per workflow.

Since the long term goal is to have our release and test process be similar, this would suggest having all our testing pipeline on one workflow and having the `make environment` step create the initial cached image for later workflow tasks.

jamtur01: Yeah - I considered that approach and wondering if a scheduled build of the image (nightly?) was a better async approach so we're not building the image for each run.

Hoverbear: I worry that isn't effective, since our dependencies might change between PRs and render the builds for the day failing.

(For example, we introduce a dependency on `cmake` or `openssl` or `sasl`, which has happened before)

.github/workflows/tests.yml
```diff
@@ -202,6 +202,15 @@ jobs:
   test-integration-gcp:
     name: Integration - Linux, GCP
     runs-on: ubuntu-20.04
+    container:
```
Hoverbear, line 205: I'm so happy you did this! Yes!

jamtur01: I cannot work out why service containers are not working for some of the jobs though. It's very weird. They are failing on network calls to localhost. And GH's docs are woeful. I am going to have to ask some GH folks I think unless inspiration strikes today.

.github/workflows/tests.yml
```diff
@@ -266,10 +283,9 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - run: sudo bash scripts/environment/bootstrap-ubuntu-20.04.sh
-      - run: bash scripts/environment/prepare.sh
       - run: make slim-builds
-      - run: make test-integration-loki
+      - run: |
+          cargo test --no-default-features --features loki-integration-tests ::loki:: -- --nocapture
```
Hoverbear, line 288: You can still use the makfile here and set `AUTOSPAWN=false`

.github/workflows/tests.yml
Hoverbear: :cry: Want to pair a bit on this?

jamtur01: I've found the issue. :( The service container syntax is a half-bake implementation of something Docker Compose lite/like/ish. Passing options to it doesn't work like you'd expect and so it's not doing any ENTRYPOINT overriding which means most of the services aren't starting right.

jamtur01:
After much banging of my head on Github Actions, it's become apparent that their implementation of service containers is too immature. It currently does not allow us to replace bootstrapping with running in a container with integration services being provided by service containers, which would have substantially sped up CI runs.

Missing pieces for future explorers:

* Service containers can only pass `options` and can't provide ENTRYPOINT/CMD overrides.
* Heath checks are hard-coded - they need to be overridden if the check doesn't return something from an open port or the startup time is longer than expected.
* There are some weird quirks around network port mapping, the containers appear to be on the same network, ports appear to be mapped, but often the main container and the service container just can't communicate. As there is zero introspection in the implementation I cannot work out why.

I think this will need to wait to be revisited as Actions evolves.

PR #3570, opened 2020-08-25
chore: simplify validate subcommand behavior

lukesteensen:
I've pulled this out of my larger set of changes because I think it warrants a review of its own.

The basic issue is that the current `validate` subcommand tries to be very fine-grained about where in the config loading process errors occur. It also attempts to validate config files one-by-one, which is not the way that we do it when running vector normally (they're combined and the result is validated). This is making it very difficult to consolidate and make that loading process more extensible, so I've opted here to simplify it and bring it more in line with how configs get validated at runtime.

The main differences in behavior here are:

1. There are no more distinct subsections of config errors in the `validate` output (e.g. file loading vs deserialization vs macro expansion)
2. Instead of validating files independently one-by-one, we behave the same as running vector normally and combine them into one config that's validated as a whole

We may want to make another pass after this to ensure the `validate` output is still as pretty as it was intended to be, but my priority is to unblock the rest of my own work right now.

There's also probably some similar simplifications to be made around the unit testing subcommand because it also treats files independently, but that wasn't on the critical path for my own changes.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,8 +14,8 @@ use futures01::Future;
 use std::cmp::max;
 use tokio::select;
 use vector::{
-    config::{self, Config, ConfigDiff},
-    event, generate, heartbeat,
+    config::{self, ConfigDiff},
+    generate, heartbeat,
     internal_events::{VectorReloaded, VectorStarted, VectorStopped},
     list, metrics, runtime,
     signal::{self, SignalTo},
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -119,7 +120,8 @@ fn main() {
             std::process::exit(exitcode::CONFIG);
         });
 
-        let result = topology::start_validated(config, diff, pieces, opts.require_healthy).await;
+        let result =
+            topology::start_validated(config.clone(), diff, pieces, opts.require_healthy).await;
```
ktff, line 124: Why is this cloned?

src/validate.rs
```diff
@@ -92,67 +92,16 @@ fn validate_config(opts: &Opts, fmt: &mut Formatter) -> Result<Config, Option<Co
...
+            Ok(config)
         }
-
-        if let Err(errors) = full_config.append(config) {
-            sub_failed(format!("Failed to merge config {:?}", config_path), errors);
-            continue;
+        Err(errors) => {
+            fmt.title(format!("Failed to load {:?}", paths));
+            fmt.sub_error(errors);
+            Err(None)
```
ktff, line 103: We can now just return `Option<Config>` instead of `Result<Config, Option<Config>>` from this function.

ktff (commented):
Output of validate will be fine with these changes. 

Shame that we lost on granularity, but we can mostly offset it by enhancing error messages coming from config if needed.

src/main.rs
lukesteensen: Good catch, leftover from a previous version.

PR #3580, opened 2020-08-26
chore(tests): reduce compile units for integration tests

fanatid:
We do not need to build tests from `tests` folder for run integration tests. Reduce integration CI job on ~45s.


Hoverbear:
Really great job. :)

PR #3581, opened 2020-08-26
chore: Draft host-metrics RFC

jamtur01:

### DIFF
--- a/rfcs/2020-08-26-3191-host-metrics.md
+++ b/rfcs/2020-08-26-3191-host-metrics.md
@@ -137,6 +137,8 @@ Or, if we think its feasible, using globbing like so:
 network.devices = [ "eth*" ]
 ```
 
+And, if feasible, have syntax for exclusion of resources.
+
 - We'd also add a guide for doing this on Docker (similar to https://github.com/prometheus/node_exporter#using-docker).
 
 ## Rationale
--- a/rfcs/2020-08-26-3191-host-metrics.md
+++ b/rfcs/2020-08-26-3191-host-metrics.md
@@ -84,7 +84,7 @@ We'd use one of these to collect the following metrics:
 
 Users should be able to limit the collection of metrics to specific classes, here: `cpu`, `memory`, `disk`, `filesystem`, `load`, and `network`.
 
-Metrics will also be labeled with:
+Metrics will also be tagged with:
 
 - `host`: the host name of the host being monitored.
 
--- a/rfcs/2020-08-26-3191-host-metrics.md
+++ b/rfcs/2020-08-26-3191-host-metrics.md
@@ -1,6 +1,6 @@
 # RFC 3191 - 2020-08-26 - Collecting host-based metrics
 
-This RFC is to introduce a new metrics source to consume host-based metrics. The high level plan is to implement one (or more?) sources that collect CPU, disk, and memory metrics from Linux-based hosts.
+This RFC is to introduce a new metrics source to consume host-based metrics. The high level plan is to implement one (or more?) sources that collect CPU, disk, network, and memory metrics from Linux-based hosts.
 
 ## Scope
 
@@ -10,6 +10,7 @@ This RFC will cover:
   - CPU
   - Memory
   - Disk
+  - Network
 - Collection on Linux, Windows, OSX.
 
 This RFC will not cover:
@@ -72,8 +73,16 @@ We'd use one of these to collect the following metrics:
 - `memory_swapped_out_bytes_total` (gauge)
 - `memory_total_bytes` (gauge)
 - `memory_wired_bytes` (gauge)
+- `network_receive_bytes_total` by device (counter)
+- `network_receive_errs_total` by device (counter)
+- `network_receive_multicast_total` by device (counter)
+- `network_receive_packets_total` yb device (counter)
+- `network_transmit_bytes_total` by device (counter)
+- `network_transmit_errs_total` by device (counter)
+- `network_transmit_multicast_total` by device (counter)
+- `network_transmit_packets_total` by device (counter)
 
-Users should be able to limit the collection of metrics to specific classes, here: `cpu`, `memory`, `disk/filesystem`, `load`.
+Users should be able to limit the collection of metrics to specific classes, here: `cpu`, `memory`, `disk`, `filesystem`, `load`, and `network`.
 
 Metrics will also be labeled with:
 
@@ -83,6 +92,8 @@ Metrics will also be labeled with:
 - `filesystem` for filesystem based metrics
 - `memory` for memory based metrics.
 - `load` for load based metrics.
+- `network` for network based metrics.
+
 
 ## Doc-level Proposal
 
@@ -94,6 +105,7 @@ The following additional source configuration will be added:
   collecting = [ "all"] # optional, defaults collecting all metrics.
   filesystem.mountpoints = [ "all" ] # optional, defaults to all mountpoints.
   disk.devices = [ "all" ] # optional, defaults to all to disk devices.
+  network.devices = [ "all" ] # optional, defaults to all to network devices.
   scrape_interval_secs = 15 # optional, default, seconds
   namespace = "host" # optional, default is "host", namespace to put metrics under
 ```
@@ -101,14 +113,14 @@ The following additional source configuration will be added:
 For `collecting` (name open to discussion) we'd specify an array of metric collections:
 
 ```toml
-collecting = [ "cpu", "memory" ]
+collecting = [ "cpu", "memory", "network" ]
 ```
 
 - We'd also add a guide for doing this on Docker (similar to https://github.com/prometheus/node_exporter#using-docker).
 
 ## Rationale
 
-CPU, Memory, and Disk are the basic building blocks of host-based monitoring. They are considered "table stakes" for most metrics-based monitoring of hosts. Additionally, if we do not support ingesting metrics from them, it is likely to push people to use another tool
+CPU, Memory, Disk, and Network are the basic building blocks of host-based monitoring. They are considered "table stakes" for most metrics-based monitoring of hosts. Additionally, if we do not support ingesting metrics from them, it is likely to push people to use another tool
 
 As part of Vector's vision to be the "one tool" for ingesting and shipping
 observability data, it makes sense to add as many sources as possible to reduce
@@ -119,11 +131,13 @@ the likelihood that a user will not be able to ingest metrics from their tools.
 - https://github.com/influxdata/telegraf/tree/master/plugins/inputs/cpu
 - https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mem
 - https://github.com/influxdata/telegraf/tree/master/plugins/inputs/disk
+- https://github.com/influxdata/telegraf/tree/master/plugins/inputs/net
 - https://github.com/elastic/beats/tree/master/metricbeat
 - https://github.com/prometheus/node_exporter
 - https://docs.fluentbit.io/manual/pipeline/inputs/cpu-metrics
 - https://docs.fluentbit.io/manual/pipeline/inputs/disk-io-metrics
 - https://docs.fluentbit.io/manual/pipeline/inputs/memory-metrics
+- https://docs.fluentbit.io/manual/pipeline/inputs/network-io-metrics
 
 ## Drawbacks
 
--- a/rfcs/2020-08-26-3191-host-metrics.md
+++ b/rfcs/2020-08-26-3191-host-metrics.md
@@ -111,7 +111,7 @@ The following additional source configuration will be added:
 ```toml
 [sources.my_source_id]
   type = "host_metrics" # required
-  collector = [ "all"] # optional, defaults collecting all metrics.
+  collectors = [ "all"] # optional, defaults collecting all metrics.
   filesystem.mountpoints = [ "*" ] # optional, defaults to all mountpoints.
   disk.devices = [ "*" ] # optional, defaults to all to disk devices.
   network.devices = [ "*" ] # optional, defaults to all to network devices.
@@ -122,7 +122,7 @@ The following additional source configuration will be added:
 For `collector` (name open to discussion) we'd specify an array of metric collectors:
 
 ```toml
-collector = [ "cpu", "memory", "network" ]
+collectors = [ "cpu", "memory", "network" ]
 ```
 
 For disk and network devices or filesystem mountpoints the default is to collect for all ("*") devices and mountpoints. Or you can configure Vector to only collect from specific devices, for example:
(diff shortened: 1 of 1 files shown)


rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Rationale
+
+CPU, Memory, and Disk are the basic building blocks of host-based monitoring. They are considered "table stakes" for most metrics-based monitoring of hosts. Additionally, if we do not support ingesting metrics from them, it is likely to push people to use another tool
+
+As part of Vector's vision to be the "one tool" for ingesting and shipping
+observability data, it makes sense to add as many sources as possible to reduce
+the likelihood that a user will not be able to ingest metrics from their tools.
+
+## Prior Art
```
jszwedko, line 108: Some more prior art:

* https://docs.fluentbit.io/manual/pipeline/inputs/cpu-metrics
* https://docs.fluentbit.io/manual/pipeline/inputs/disk-io-metrics
* https://docs.fluentbit.io/manual/pipeline/inputs/memory-metrics

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,157 @@
...
+- `memory_swapped_in_bytes_total` (gauge)
+- `memory_swapped_out_bytes_total` (gauge)
+- `memory_total_bytes` (gauge)
+- `memory_wired_bytes` (gauge)
+
+Users should be able to limit the collection of metrics to specific classes, here: `cpu`, `memory`, `disk/filesystem`, `load`.
+
+Metrics will also be labeled with:
+
+- `host`: the host name of the host being monitored.
```
jszwedko, line 80: I think we'll probably need some more labels for some of the metrics to be useful:

* `device` for disk based metrics
* `cpu` for CPU based metrics
* `filesystem` for filesystem based metrics

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,157 @@
...
+## Doc-level Proposal
+
+The following additional source configuration will be added:
+
+```toml
+[sources.my_source_id]
+  type = "host_metrics" # required
+  collecting = [ "all"] # optional, defaults collecting all metrics.
+  scrape_interval_secs = 15 # optional, default, seconds
+  namespace = "aspace" # optional, default, namespace to put metrics under
```
jszwedko, line 91: We could leave this as a second pass, but I think it'd be useful to limit the metrics collection to specific devices, filesystems, etc.

Example from fluentbit:

2020-08-26-163531_806x340_scrot

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,157 @@
...
+Telegraf.
+
+> You SHOULD use Vector to replace Logstash, Fluent*, Telegraf, Beats, or
+> similar tools.
+
+If users are already running Telegraf or Node Exporter though, they could opt for this path.
+
+## Outstanding Questions
+
+- One source or many? Should we have `host_metrics` or `cpu_metrics`, `mem_metrics`, `disk_metrics`, or `load_metrics`?
```
jszwedko, line 146: I think this is the big one.

I personally like your approach of a single source though. I think we could allow for fine grained control over the metric "families" via the table-based TOML configuration; like:

```
[sources.my_source_id]
  type = "host_metrics"

  disk.devices = ["/dev/sda"]

  filesystem.mountpoints = ["/home"]
```

To exclude trying to collect all of them (though you could also do this in a `filter` transform).

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,157 @@
...
+
+On the same page, it is mentioned that Vector should be a replacement for
+Telegraf.
+
+> You SHOULD use Vector to replace Logstash, Fluent*, Telegraf, Beats, or
+> similar tools.
+
+If users are already running Telegraf or Node Exporter though, they could opt for this path.
+
+## Outstanding Questions
```
jszwedko, line 144: This isn't really an outstanding question, but just noting that we'll need to take consideration for documenting how to use this source within a `docker` container (similar to https://github.com/prometheus/node_exporter#using-docker).

rfcs/2020-08-26-3191-host-metrics.md
jszwedko, line 91: I might suggest a default of `system` or `host` for the namespace.

jamtur01:
@jszwedko Thanks! Updated to reflect your feedback.

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,169 @@
...
+- `disk_write_time_seconds_total` by disk (counter)
+- `disk_writes_completed_total` by disk (counter)
+- `disk_written_bytes_total` by disk (counter)
+- `disk_written_sectors_total` by disk (counter)
+- `filesystem_avail_bytes` labeled with device, filesystem type, and mountpoint (gauge)
+- `filesystem_device_error` labeled with device, filesystem type, and mountpoint (gauge)
+- `filesystem_total_file_nodes` labeled with device, filesystem type, and mountpoint (gauge)
+- `filesystem_free_file_nodes` labeled with device, filesystem type, and mountpoint (gauge)
+- `filesystem_free_bytes` labeled with device, filesystem type, and mountpoint (gauge)
+- `filesystem_size_bytes` labeled with device, filesystem type, and mountpoint (gauge)
```
bruceg, line 62: What is the distinction between `avail` and `free` or `size`? Also, there seems to be an inconsistency between avail/free/size for bytes, but free/total for file nodes.

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,169 @@
...
+## Doc-level Proposal
+
+The following additional source configuration will be added:
+
+```toml
+[sources.my_source_id]
+  type = "host_metrics" # required
+  collecting = [ "all"] # optional, defaults collecting all metrics.
+  filesystem.mountpoints = [ "all" ] # optional, defaults to all mountpoints.
+  disk.devices = [ "all" ] # optional, defaults to all to disk devices.
```
bruceg, line 107: I'd recommend against magic words within the list to represent collecting all. We could:
1. not allow wildcard configuration like that, just say the default is all metrics;
2. use an actual wildcard to represent all, like `"*"`, which may be useful to allow for glob like `"filesystem_read*"`; or
3. use a custom serializer to allow either `"ALL"` outside of the list for all, or a list to enumerite items (ie `collecting = "ALL"` vs `collecting = [ "a", "b" ]`)

rfcs/2020-08-26-3191-host-metrics.md
bruceg: From the point of view of user experience, I think a single source is better. It will end up being a large in terms of source code, but we can manage that with sub-modules internally.

rfcs/2020-08-26-3191-host-metrics.md
jamtur01: I've drawn these from Prometheus' approach specifically.

- `filesystem_avail_bytes` = Filesystem space available to non-root users in bytes.
- `filesystem_free_bytes` = Filesystem free space in bytes.
- `filesystem_size_bytes` = Filesystem size in bytes.

See https://www.robustperception.io/filesystem-metrics-from-the-node-exporter. I'll add an explainer to the RFC.

rfcs/2020-08-26-3191-host-metrics.md
jamtur01: I like the globbing. +1.

rfcs/2020-08-26-3191-host-metrics.md
bruceg: My comment about the inconsistency was to suggest, for example, we could use avail/free/total for both:
```
filesystem_avail_bytes
filesystem_free_bytes
filesystem_total_bytes
filesystem_avail_file_nodes
filesystem_free_file_nodes
filesystem_total_file_nodes
```
Unless, of course, there is a preference for sticking to Prometheus' terminology.

jamtur01: That's an interesting question - I replicated Prometheus' naming as a lot of folks are familiar with that. Also if folks have dashboards/alerts/etc that calculate things, like free space on a filesystem, then they don't need to rewrite them with new metric names.

rfcs/2020-08-26-3191-host-metrics.md
jszwedko: Sorry, I'm realizing I was unclear. I think you've addressed this above though with the update to the metrics list. I had meant, for example, that disk based metrics should be labeled (or tagged; not sure which term we prefer) with the device the metric is associated with (e.g. `device=/dev/sda`).

I think I like these "collector" labels too though. I think they'd be better as simply `collector` so you'd have things like `collector=disk`.

rfcs/2020-08-26-3191-host-metrics.md
jszwedko: :thinking: I think the naming you (and prometheus) have feels natural to me. I'm used to referring to filesystems as having a size, probably because of my association with a physical disk and the fact that files themselves are referred to as having "sizes" (and not "total bytes"). Inodes, on the other hand feels more like a "discrete, countable" thing so total makes sense to me for the cap.

I can see an argument for consistency though.

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,201 @@
...
+
+The Heim crate has a useful comparison doc showing the available tools and their capabilities.
+
+- https://github.com/heim-rs/heim/blob/master/COMPARISON.md
+
+For this implementation we're recommending the Heim crate based on platform and feature coverage.
+
+We'd use one of these to collect the following metrics:
+
+- `cpu_seconds_total` by mode (idle, nice, system, user) per CPU. (counter)
```
binarylogic, line 44: @jszwedko given the Prometheus-style naming here. Does it make you want to adopt the same for the Apache metrics source for consistency? Just floating the question while we can.

jszwedko: I could go either way on it. I think one difference here is there is no "provider name" for the metrics so we need to come up with one whereas with `apache_metrics` it is scraping already named metrics in Apache.

These is something nice about standardizing metric names in vector though. I agree that we should settle it before releasing the `apache_metrics` source though.

I'm wondering if @lukesteensen has any thoughts.

rfcs/2020-08-26-3191-host-metrics.md
jamtur01: Ah. Yeah - this is a terminology mismatch. I'll update to say tagged and make it clearer.

rfcs/2020-08-26-3191-host-metrics.md
binarylogic: Right. You can see the Prometheus `apache_exporter` [here](https://github.com/Lusitaniae/apache_exporter), notice how they follow the naming guidelines, which is what I'm proposing.

jszwedko: Linking over to https://github.com/timberio/vector/pull/3519#issuecomment-681985323 for more discussion details specific to the apache metrics source.

rfcs/2020-08-26-3191-host-metrics.md
lukesteensen: Definitely agree on having a single source

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,204 @@
...
+
+Users want to collect, transform, and forward metrics to better observe how their hosts are performing.
+
+## Internal Proposal
+
+Build a single source called `host_metrics` (name to be confirmed) to collect host/system level metrics.
+
+I've found a number of possible Rust-based solutions for implementing this collection, cross-platform.
+
+- https://crates.io/crates/heim (Rust)
```
lukesteensen, line 31: `heim` looks very promising and would likely be a good starting point.

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,204 @@
...
+- Other host metrics.
+- Other platforms
+
+## Motivation
+
+Users want to collect, transform, and forward metrics to better observe how their hosts are performing.
+
+## Internal Proposal
+
+Build a single source called `host_metrics` (name to be confirmed) to collect host/system level metrics.
```
binarylogic, line 27: I'm 👍 on `host_metrics` unless others disagree. An alternative could be `node_metrics`, but that seems less precise to me. I'm curious why Prometheus adopted this nomenclature.

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,204 @@
...
+
+
+## Doc-level Proposal
+
+The following additional source configuration will be added:
+
+```toml
+[sources.my_source_id]
+  type = "host_metrics" # required
+  collector = [ "all"] # optional, defaults collecting all metrics.
```
binarylogic, line 114: I would make this plural (`collectors`) since it accepts an array.

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,204 @@
...
+  disk.devices = [ "*" ] # optional, defaults to all to disk devices.
+  network.devices = [ "*" ] # optional, defaults to all to network devices.
+  scrape_interval_secs = 15 # optional, default, seconds
+  namespace = "host" # optional, default is "host", namespace to put metrics under
+```
+
+For `collector` (name open to discussion) we'd specify an array of metric collectors:
+
+```toml
+collector = [ "cpu", "memory", "network" ]
```
binarylogic, line 125: Same here.

rfcs/2020-08-26-3191-host-metrics.md
```diff
@@ -0,0 +1,204 @@
...
+  namespace = "host" # optional, default is "host", namespace to put metrics under
+```
+
+For `collector` (name open to discussion) we'd specify an array of metric collectors:
+
+```toml
+collectors = [ "cpu", "memory", "network" ]
+```
+
+For disk and network devices or filesystem mountpoints the default is to collect for all ("*") devices and mountpoints. Or you can configure Vector to only collect from specific devices, for example:
```
jszwedko, line 128: I'm wondering if will be useful to also allow users to specify a denylist rather than allowlist, in the future. Maybe we could model this like `network.devices = [ "!eth0", "*"]` to monitor everything but eth0? We could expand this pattern generally to this type of "filter" config.

PR #3585, opened 2020-08-26
chore(deps): Run a blanket cargo update

Hoverbear:
This rather boldly just runs `cargo update` and asks "Will the tests still pass on all platforms?"

This is partially motivated by our discussion about #2840 regarding the use of the https://docs.rs/string/0.2.1/string/index.html crate. During adoption I found we still had multiple copies of the `bytes` crate in tree. (We still have one thanks to `elastic_responses`, but it's only in the dev tree)

While attempting to hack at the forest I decided to just go "Let's see what happens" and it looks like this might do a lot of good.

I updated our lucet version.


Hoverbear:
Before
master-graph

After
graph

binarylogic:
Lol at that dependency graph. Can anyone tell the difference between the before and after?

Hoverbear:
I can! 
image

binarylogic:
On a more serious note. What are the risks of blindly do this? Thinking specifically about the items [here](https://github.com/timberio/vector/blob/master/REVIEWING.md#dependencies).

Hoverbear:
So this is what dependabot does one-by-one already, just all done at once!

Since this is only upgrading our existing dependencies up to their semantic-versioning-based compatible version, this should only create failures that we'd see in CI.

Hoverbear:
Fascinating:
image

jamtur01:
That's a surprisingly small amount of fail. 👍

Hoverbear:
Let's see if this is enough to coax it into working.

Hoverbear:
... Wow.

Hoverbear:
Small hiccup: 
image

Going to roll back zstd and try more

Hoverbear:
Okay! This is gold!

PR #3596, opened 2020-08-27
chore(tests): fix scheduler issues

### LINKED ISSUE #3543: Investigate issues with k8s reflector tests after switching to tokio::test macro
Follow up after https://github.com/timberio/vector/pull/3505/files/07b6d6b049588a6888805587d63fbf2c7007f5ba?diff=unified&w=1#r475513553.

fanatid:
Part of #3478.

@MOZGIII I'm not sure what I fixed (or not fixed), but kubernetes works fine without `threaded_scheduler` :laughing:

### DIFF
--- a/src/kubernetes/reflector.rs
+++ b/src/kubernetes/reflector.rs
@@ -535,7 +535,7 @@ mod tests {
     }
 
     /// Test that the delayed delete works accordingly.
-    #[tokio::test(threaded_scheduler, core_threads = 2)]
+    #[tokio::test(threaded_scheduler)]
     async fn test_delayed_deletes() {
         trace_init();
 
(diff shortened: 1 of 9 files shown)


MOZGIII (approved):
Yeah, I created #3543 to do a proper investigation, and I'd like to find the root cause. It's important to ensure the proper quality of the test suite. If you don't mind, let's keep the #3543 open. This can be merged in the meantime.

I have a suspicion this has something to do with the amount of effectively running threads changing somehow after the removal of the `tokio-compat`. Though, knowing the internals of `tokio-compat` a bit I'm not sure how it's possible.

fanatid:
@MOZGIII I cherry-picked commit with kubernetes changes (`threaded_scheduler` removing) on top of `master` after #3505 merge, and everything works..... Sorry for the false alarm, I was needed double check scheduler changes before submit / comment anything :disappointed: 

@lukesteensen Yes, we use `threaded` scheduler, while test macro use `basic` (`threaded` if `core_threads` more than 1). Ideally, there should not be any difference between them, but since tests hang up with `basic` I wanted to confirm that vector code not block anything and problem indeed in tests.

Issue #2841, opened 2020-06-17
Add UX improvements on `LogEvent`, particularly turning JSON into or from `LogEvent`.


Hoverbear:
Step 3 of https://github.com/timberio/vector/blob/c4707947bd876a0ff7d7aa36717ae2b32b731593/rfcs/2020-05-25-more-usable-logevents.md#sales-pitch.

This calls for general implementations of `From` and `Into` for things `serde_json::Value`, `HashMap`, `BTreeMap`. Please **do not cover TOML Tables** or **component configs**.

Include tests for data that is shaped like https://github.com/timberio/vector/blob/c4707947bd876a0ff7d7aa36717ae2b32b731593/rfcs/2020-05-25-more-usable-logevents.md#a-motivatingly-complex-example

PR #3601, opened 2020-08-27
chore(deps): Remove some From implemementations from Event

Hoverbear:
Part of #2841 

This removes a few extra implementations of these which are kind of not idiomatic since we should encourage the caller to choose the allocation time of the values.

I allowed this in tests on `&str` since it's just so, so usable.


Hoverbear:
It looks like one of the CI jobs experienced a transient failure in package management.

PR #3607, opened 2020-08-27
enhancement(docker source): Multiline support

MOZGIII:
This PR adds multiline aggregation support to the docker source.

I'm not a big fan of what I did here. I've found some potential optimizations, however, they require a bit of refactoring to be applied cleanly. My thinking currently is we can ship this to quickly solve the pain point of the people waiting, and then apply the refactoring at a slower pace.

## Refactoring plan

- This source is too big to live in a single mod. We should split it into multiple mods, and add unit tests for individual pieces in addition to the integration tests.

- We should decouple handing `Err` from the `Ok` `Result<LogOutput>` handling, and short-circuit error case early on, so that we have the ability to stack logic on top without the consideration of the error condition (simplified mental model with the same effective outcome).

- The event construction should be split into multiple steps, so that we can parse the `LogOutput` into a minimal intermediate format suitable for further processing (like line aggregation) but without the full overhead of constructing an `Event`.

- Delay the `Event` creation and additional annotation to a later stage, reducing the amount of work that might then be thrown away.

- Apply the logic of the partial event merging and line aggregation logic more cleanly.


src/sources/docker.rs
```diff
@@ -580,8 +592,21 @@ impl EventStreamBuilder {
                 }
             })
             .take_while(|v| future::ready(v.is_ok()))
-            .filter_map(|v| future::ready(v.transpose()))
```
fanatid, line 583: Why `transponse()` changed to `unwrap()`?

ktff (approved):
It definitely needs some simplifications. Looking forward to the refactored version.

src/sources/docker.rs
ktff: The `take_while` before it ends the stream if `v` is Err, so we will always have `Ok` here. Right @MOZGIII?

MOZGIII: We need to get rid of the wrapping to pass the event to the `line_agg_adapter`. It doesn't really matter where we do it (could be done inside the adapter). It's ok because of the previous line - no `Err` items will ever be at the `unwrap()`. This is one of the things I'd refactor to be less cognitively complex.

fanatid: Indeed, `unwrap()` will always successful because we have filter on previous line. Thanks.

fanatid (approved):

MOZGIII:
Thx for the feedback, I've finished this PR by adding docs and tests.

This should be ready to merge.

PR #3610, opened 2020-08-27
enhancement(templating): adds new Helm DaemonSet template variable for pods labels

### LINKED ISSUE #3568: Add `podLabels` template variable to DaemonSet template for Helm
### Current Vector Version
`0.11.0-nightly-2020-08-18` via Helm

### Use-cases
The [current Helm template for the Vector DaemonSet](https://github.com/timberio/vector/blob/master/distribution/helm/vector/templates/daemonset.yaml) does not appear to let a user installing Vector into a kubernetes cluster via Helm to modify pod labels. It does, however, already [allow a user to modify pod annotations](https://github.com/timberio/vector/blob/master/distribution/helm/vector/values.yaml#L36). For any number of reasons, users may like to label all nodes belonging to a given DaemonSet.

### Attempted Solutions
Our current workaround is to patch the DaemonSet created by Helm immediately after it is created. This works, but is clunky

### Proposal
Add a new `podLabels` section to the Vector Helm DaemonSet template, in the same way that [you already have for pod annotations](https://github.com/timberio/vector/blob/master/distribution/helm/vector/templates/daemonset.yaml#L16). This will allow users to add pod labels via a values.yaml file passed to Helm.

nkinkade:
The small PR add the ability for a user of the Helm template to add arbitrary pod labels to the DaemonSet spec template via a values.yaml file when deploying the package.

Resolves #3568.

### DIFF
--- a/distribution/helm/vector/templates/daemonset.yaml
+++ b/distribution/helm/vector/templates/daemonset.yaml
@@ -18,6 +18,9 @@ spec:
         {{- toYaml . | nindent 8 }}
     {{- end }}
       labels:
+        {{- with .Values.podLabels }}
+          {{- toYaml . | nindent 8 }}
+        {{- end }}
         {{- include "vector.selectorLabels" . | nindent 8 }}
         vector.dev/exclude: "true"
     spec:
--- a/distribution/helm/vector/values.yaml
+++ b/distribution/helm/vector/values.yaml
@@ -35,6 +35,9 @@ serviceAccount:
 # Annotations to add to the `Pod`s managed by `DaemonSet`.
 podAnnotations: {}
 
+# Labels to add to the `Pod`s managed by `DaemonSet`.
+podLabels: {}
+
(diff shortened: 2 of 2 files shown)


MOZGIII:
Thx, the change looks solid!
You mind signing off the code? See https://github.com/timberio/vector/blob/master/CONTRIBUTING.md#signing-off

We can merge it once it has a proper DCO signature.

MOZGIII (changes requested):
Needs DCO

MOZGIII:
Turns out `make signoff` is currently broekn, please use `scripts/signoff.sh` directly instead.

nkinkade:
@MOZGIII: Sorry, it should be signed now. I actually had signed the commit in my local clone, but _after_ I had pushed (using `git commit --amend`), and I failed to push the modification.

MOZGIII (approved):

PR #3614, opened 2020-08-28
chore: Generate Kubernetes distribution YAMLs from Helm Chart

### LINKED ISSUE #3576: K8s: generate distribution YAML files from Helm Chart
We currently have YAML configs for Kubernetes distribution and a Helm Chart.

YAML configs are currently manually written, but it's possible to simply generate them from the Helm Chart.

If we transition to only using Helm Chart as the source of truth, it will reduce the test surface, and also the maintenance weight.

One more detail: having two files was confusing to some users, so, while we're at it, we can switch to having just a single YAML file with everything in it.

Relevant issues:
- #3097 is better to after (or together with) this one, since it's going to be much easier to support only a single test flow.

MOZGIII:
This PR switches to generating YAML config for Kubernetes from Helm Chart.

It also changes the Helm Chart to be more like the generated YAMLs.

Closes #3576.

To do:

- [x] ~allow hiding Helm-specific annotations from the generated config, and enable it for our generated YAML config~ discarded - an unnsecessary complication.


Hoverbear (approved):
This seems to make sense to me, though I lack much context on kubernetes.

PR #3618, opened 2020-08-28
Check if playground is enabled statically

MOZGIII:
A follow up from https://github.com/timberio/vector/pull/3514#discussion_r479192068
It's the same idea, but lifted to the `graphql_playground` var.
Do you like it?


leebenson (approved):
This is great, thanks @MOZGIII!

PR #3619, opened 2020-08-28
chore: Split up event

Hoverbear:
Part of #2841, continues in the spirit of #3601.

This splits up the `event.rs` file into `log_event.rs`, `event.rs`, and `log_schema.rs`.

This also migrates `log_schema` into `crate::config::log_schema`.

This doesn't actually change any behavior, just reorganization.

PR #3621, opened 2020-08-28
chore: Add some additional Event impls

Hoverbear:
More of #2841 goals following #3619


fanatid (approved):

src/event/log_event.rs
```diff
@@ -92,6 +93,48 @@ impl Into<BTreeMap<String, Value>> for LogEvent {
...
+        LogEvent {
+            fields: map.into_iter().collect(),
+        }
+    }
+}
+
+impl Into<HashMap<String, Value>> for LogEvent {
+    fn into(self) -> HashMap<String, Value> {
+        let Self { fields } = self;
+        fields.into_iter().collect()
```
JeanMertz, line 107: Out of curiosity, (how) does this differ from `self.fields.into_iter().collect()`?

JeanMertz (approved):
sweet 👍

src/event/log_event.rs
Hoverbear: In this case we destructure the `Self` first, so we own the `fields`, as opposed to borrowing them.

PR #3626, opened 2020-08-29
chore: update runtime in benches

fanatid:
Part of #3478. Update `runtime` helper function in `test_util` and adjust benches for tokio 0.2 runtime.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -7,12 +7,13 @@ use rand::{
     prelude::*,
 };
 use std::convert::TryFrom;
-use vector::config::{self, TransformConfig, TransformContext};
-use vector::event::Event;
-use vector::test_util::{
-    next_addr, runtime, send_lines, start_topology, wait_for_tcp, CountReceiver,
+use vector::{
+    config::{self, TransformConfig, TransformContext},
+    event::Event,
+    sinks, sources,
+    test_util::{next_addr, runtime, send_lines, start_topology, wait_for_tcp, CountReceiver},
+    transforms,
 };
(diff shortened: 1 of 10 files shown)


fanatid:
Looks like splunk fail related to with docker image:
```
$ docker logs -f --tail 10 vector_splunk_1 
Index "custom_index" added.
An index with name=custom_index already exists
tail: cannot open '/opt/splunk/var/log/splunk/splunkd_stderr.log' for reading: No such file or directory
tail: no files remaining
```

benches/bench.rs
```diff
@@ -241,10 +242,12 @@ fn benchmark_simple_pipe_with_many_writers(c: &mut Criterion) {
                             .collect::<Vec<_>>()
                             .await;
                         future::try_join_all(sends).await.unwrap();
-                        std::thread::sleep(std::time::Duration::from_millis(100));
 
                         topology.stop().compat().await.unwrap();
-                        assert_eq!(num_lines * num_writers, output_lines.await.len());
+
+                        // TODO: shutdown after flush
```
Hoverbear, line 248: Oh no! What happened here?

fanatid: Number of lines in `output_lines` always less than `num_lines`, because socket from `send_lines` not fully flushed before topology stop. Everything fine if add some timeout, but then this will not correct benchmark. From vector side all code looks good, and I think flushing issue related with kernel buffers but not sure.

Hoverbear: Ok. :)

MOZGIII (approved):

PR #3636, opened 2020-08-30
fix(statsd sink): Sink should emit event on invalid metric

### LINKED ISSUE #3635: StatsD sink should create an internal event when sent a metric it can't handle.
I made a recent change to the StatsD sink that issues a warning when it receives a metric type it can't handle (https://github.com/timberio/vector/blob/master/src/sinks/statsd.rs#L189). 

As per the conversation with Bruce [here](https://github.com/timberio/vector/pull/3501#discussion_r477610861), really it should be an internal event so that metrics can count how often this happens.

FungusHumungus:
On receiving an invalid event, the StatsD sink should emit an internal event rather than a warning.

This fixed #3635 

### DIFF
--- a/src/internal_events/statsd.rs
+++ b/src/internal_events/statsd.rs
@@ -26,12 +26,14 @@ impl InternalEvent for StatsdEventReceived {
     }
 }
 
+#[cfg(feature = "sources-statsd")]
 #[derive(Debug)]
 pub struct StatsdInvalidRecord<'a> {
     pub error: crate::sources::statsd::parser::ParseError,
     pub text: &'a str,
 }
 
+#[cfg(feature = "sources-statsd")]
 impl InternalEvent for StatsdInvalidRecord<'_> {
     fn emit_logs(&self) {
         error!(message = "Invalid packet from statsd, discarding.", error = %self.error, text = %self.text);
--- a/src/internal_events/statsd.rs
+++ b/src/internal_events/statsd.rs
@@ -117,7 +117,7 @@ impl<'a> InternalEvent for StatsdInvalidMetric<'a> {
         counter!(
             "processing_errors", 1,
             "component_kind" => "sink",
-            "component_type" => "statsd_metrics",
+            "component_type" => "statsd",
             "error_type" => "invalid_metric",
         );
     }
--- a/src/internal_events/statsd.rs
+++ b/src/internal_events/statsd.rs
@@ -96,12 +96,12 @@ impl<T: std::fmt::Debug + std::fmt::Display> InternalEvent for StatsdSocketError
 }
 
 #[derive(Debug)]
-pub struct StatsdInvalidMetric {
-    pub value: MetricValue,
-    pub kind: MetricKind,
+pub struct StatsdInvalidMetric<'a> {
+    pub value: &'a MetricValue,
+    pub kind: &'a MetricKind,
 }
 
-impl InternalEvent for StatsdInvalidMetric {
+impl<'a> InternalEvent for StatsdInvalidMetric<'a> {
     fn emit_logs(&self) {
         warn!(
             message = "Invalid metric sent; dropping event.",
--- a/src/internal_events/statsd.rs
+++ b/src/internal_events/statsd.rs
@@ -1,4 +1,5 @@
 use super::InternalEvent;
+use crate::event::metric::{MetricKind, MetricValue};
 use metrics::counter;
 
 #[derive(Debug)]
@@ -90,6 +91,32 @@ impl<T: std::fmt::Debug + std::fmt::Display> InternalEvent for StatsdSocketError
             "socket_errors", 1,
             "component_kind" => "source",
             "component_type" => "statsd",
+        );
+    }
+}
+
+#[derive(Debug)]
+pub struct StatsdInvalidMetric {
+    pub value: MetricValue,
+    pub kind: MetricKind,
+}
+
+impl InternalEvent for StatsdInvalidMetric {
+    fn emit_logs(&self) {
+        warn!(
+            message = "Invalid metric sent; dropping event.",
+            value = ?self.value,
+            kind = ?self.kind,
+            rate_limit_secs = 30,
+        )
+    }
+
+    fn emit_metrics(&self) {
+        counter!(
+            "processing_errors", 1,
+            "component_kind" => "sink",
+            "component_type" => "statsd_metrics",
+            "error_type" => "invalid_metric",
         );
     }
 }
--- a/src/sinks/statsd.rs
+++ b/src/sinks/statsd.rs
@@ -3,7 +3,7 @@ use crate::{
     config::{DataType, SinkConfig, SinkContext, SinkDescription},
     event::metric::{Metric, MetricKind, MetricValue, StatisticKind},
     event::Event,
-    internal_events::StatsdInvalidMetric,
+    internal_events::StatsdInvalidMetricReceived,
     sinks::util::{encode_namespace, BatchConfig, BatchSettings, BatchSink, Buffer, Compression},
 };
 use futures::{future, FutureExt};
@@ -182,7 +182,7 @@ fn encode_event(event: Event, namespace: Option<&str>) -> Option<Vec<u8>> {
             }
         }
         _ => {
-            emit!(StatsdInvalidMetric {
+            emit!(StatsdInvalidMetricReceived {
                 value: &metric.value,
                 kind: &metric.kind,
             });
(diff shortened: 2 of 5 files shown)


src/internal_events/statsd.rs
```diff
@@ -93,3 +94,29 @@ impl<T: std::fmt::Debug + std::fmt::Display> InternalEvent for StatsdSocketError
         );
     }
 }
+
+#[derive(Debug)]
+pub struct StatsdInvalidMetric {
+    pub value: MetricValue,
+    pub kind: MetricKind,
```
lukesteensen, line 101: Could these just be references? It'd be nice to avoid the `clone` if it's not too much trouble, but I'm not hugely worried about it since it's an error path.

FungusHumungus: Yes, of course that makes sense. It might be an error path, but there is still the potential for it to be called many times per second.

I wonder though if it would be better for MetricKind to be Copy since it is such a simple struct? (https://github.com/timberio/vector/blob/master/src/event/metric.rs#L21)

FungusHumungus: I have made them both references. It just feels more aesthetic with both fields being references rather than one Copy and the other a reference!

src/internal_events/statsd.rs
```diff
@@ -93,3 +96,29 @@ impl<T: std::fmt::Debug + std::fmt::Display> InternalEvent for StatsdSocketError
         );
     }
 }
+
+#[derive(Debug)]
+pub struct StatsdInvalidMetric<'a> {
```
JeanMertz, line 101: I note that [our RFC](https://github.com/timberio/vector/blob/eaf56397175a54e6a18cbc62966d43ecb926b917/rfcs/2020-03-17-2064-event-driven-observability.md) describes:

> Both event struct names and metric keys follow the `{noun}_{verb}` naming scheme. This keeps things simple, and we can rely on tags/labels for more context.

(for structs, this would follow Rust's conventions as `{Noun}{Verb}`)

I see that we aren't applying that rule universally, as we also have an `StatsdInvalidRecord` event.

I guess we could name this `StatsdInvalidMetricSent`? Although from what I understand, this event actually describes the sink _receiving_ an invalid event, and then not _sending_ it to Statsd, so that name also doesn't describe what happened.

You could name it `StatsdInvalidMetricReceived`, but then you'd potentially be getting in the way of the statsd source.

This makes me wonder if we should either:

1. distinguish between sources and sinks of the same type by using `Statsd{Sink,Source}{...}` event names.
2. use the same event, but have `component_kind` be dynamically set to either `sink` or `source`

I'd vote for the first solution (unless there is another, better one), as the second one can get quite messy if you need to add different labels for the source and sink of the same type.

cc @lukesteensen

src/internal_events/statsd.rs
```diff
@@ -93,3 +96,29 @@ impl<T: std::fmt::Debug + std::fmt::Display> InternalEvent for StatsdSocketError
...
+    pub value: &'a MetricValue,
+    pub kind: &'a MetricKind,
+}
+
+impl<'a> InternalEvent for StatsdInvalidMetric<'a> {
+    fn emit_logs(&self) {
+        warn!(
+            message = "Invalid metric sent; dropping event.",
```
JeanMertz, line 109: This message is somewhat ambigious:

1. "dropping event" is used in other events to indicate no further components in the pipeline will receive this event. In this case, since this is a sink, there is no way for other components to receive it anyway.

2. having said that, it is valid to say that the event got dropped _before_ sending it to the external Statsd service, but in that case, the "invalid metric sent" part is confusing, as I already noted in an earlier comment.

(the original `warn` log solved this by mentioning "invalid metric sent to statsd sink", which makes it explicit that we're talking about the event sent to the sink, not the eventual metric sent to the external statsd service)

cc @lukesteensen

JeanMertz (changes requested):
Thanks for this @FungusHumungus!

I left a couple of remarks regarding ambiguity that we probably want to address first.

I tagged @lukesteensen for his opinion, as he currently has the most experience with our internal events system.

src/internal_events/statsd.rs
FungusHumungus: I think the first solution makes more sense to me, sources and sinks are completely different concepts, and there is often very little shared between them, as they both handle different sides of the same protocol.

I was also thinking it may be worth separating the events into separate modules, so a `internal_events/statsd_source.rs` and a `internal_events/statsd_sink.rs` or similar. It is certainly possible to compile Vector with just the statsd sink and not the source.  The extra feature flags I have had to put into the module suggests that lumping it all in one module is a bit of an anti pattern.

JeanMertz: Yeah, I agree with your insight regarding the feature flags. Having a separate module for each component does feel like the right solution 👍

It is still somewhat orthogonal to the name of the events though, depending on how we want to use these internal events:

```rust
use internal_event::statsd_source::StatsdSourceInvalidMetricReceived;
use internal_event::statsd_sink::StatsdSinkInvalidMetricReceived;

emit!(StatsdSourceInvalidMetricReceived)
emit!(StatsdSinkInvalidMetricReceived)
```

vs.

```rust
use internal_event::{statsd_source, statsd_sink};

emit!(statsd_source::InvalidMetricReceived)
emit!(statsd_sink::InvalidMetricReceived)
```

I'd vote for the latter, although I can also see how that could potentially interfere with any actual non-internal-event `statd_source` module. A solution to that would be to always re-export internal events from the actual component modules. But I think as long as we can get away with it, without creating name conflicts, I'd vote to keep it simple and use the second example from above.

src/internal_events/statsd.rs
```diff
@@ -93,3 +96,29 @@ impl<T: std::fmt::Debug + std::fmt::Display> InternalEvent for StatsdSocketError
...
+            kind = ?self.kind,
+            rate_limit_secs = 30,
+        )
+    }
+
+    fn emit_metrics(&self) {
+        counter!(
+            "processing_errors", 1,
+            "component_kind" => "sink",
+            "component_type" => "statsd_metrics",
```
juchiast, line 120: I think this should be just "statsd" instead of "statsd_metrics". 

https://github.com/timberio/vector/blob/c9317c5893ac87f3ba1845ad23908b7ef102e5df/src/sinks/statsd.rs#L78-L80

juchiast (changes requested):

src/internal_events/statsd.rs
FungusHumungus: At the risk of complicating things further by throwing too many options in the mix, what about something like:

```rust
emit!(InvalidMetricReceived(Source("statsd")))
emit!(InvalidMetricReceived(Sink("statsd")))
```

That way there is only one InvalidMetricRecieved event, and it is tagged with further information to indicate the source of the event. That way there is no code duplication, so we can ensure things stay consistent.

Admittedly, it would be more work to change all the events in line with that approach.

src/sinks/statsd.rs
```diff
@@ -186,10 +187,10 @@ fn encode_event(event: Event, namespace: &str) -> Option<Vec<u8>> {
...
         }
         _ => {
-            warn!(
-                "invalid metric sent to statsd sink ({:?}) ({:?})",
-                metric.kind, metric.value
-            );
+            emit!(StatsdInvalidMetric {
+                value: &metric.value,
+                kind: &metric.kind,
+            });
```
juchiast, line 193: We should also return `None` here. I wonder if a `None` would stop the sink?

juchiast (changes requested):

src/sinks/statsd.rs
FungusHumungus: I tested this out and threw a load of events at it, and the Sink was not stopped when this returned None, any ensuing events were sent out just fine. I'm guessing the `flat_map` prevents any `None`, if it was multiple events we were encoding here, a None might prevent the rest in the batch from being sent - but in this case it is just one event at a time.

juchiast (approved):

src/internal_events/statsd.rs
lukesteensen: I'm not terribly worried about the name of the event itself since that's not user-facing, but I agree it'd be nice to be unambiguous. Since only sources receive things, `StatsdInvalidMetricReceived` seems fine to me.

src/internal_events/statsd.rs
lukesteensen: Yes, the original wording here was poor. I think a simple change from `sent` to `received` should clear it up.

src/internal_events/statsd.rs
JeanMertz: Wait, now I'm confused;

> Since only sources receive things, `StatsdInvalidMetricReceived` seems fine to me.

This would imply to me that you'd be fine with that name, _if_ we used it for the statsd source, right? Because right now, we're using this event in the sink, which still seems confusing to me.

JeanMertz (approved):

PR #3644, opened 2020-08-31
chore(external docs): Typo fixes

leebenson:
Misc. typo fixes, in internal/external comments and the odd variable name.

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -15,7 +15,7 @@ privacy now and in the future.
 1. [Vector Community](#vector-community)
    1. [Vector Repository](#vector-repository)
    1. [Vector Chat](#vector-chat)
-   1. [Vecotr Mailinglist](#vecotr-mailinglist)
+   1. [Vector Mailing_List](#vector-mailing-list)
 
 <!-- /MarkdownTOC -->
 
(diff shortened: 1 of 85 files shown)


PRIVACY.md
```diff
@@ -15,15 +15,15 @@ privacy now and in the future.
 1. [Vector Community](#vector-community)
    1. [Vector Repository](#vector-repository)
    1. [Vector Chat](#vector-chat)
-   1. [Vecotr Mailinglist](#vecotr-mailinglist)
+   1. [Vector Mailing_List](#vector-mailing-list)
```
bruceg, line 18: Underscore?

distribution/helm/vector/values.yaml
```diff
@@ -55,16 +55,16 @@ securityContext: {}
 # Extra env vars to pass to the `vector` container.
 env: []
 
-# Various tweakables for the `Pod`s managed by `DaemonSet`.
+# Various tweaks for the `Pod`s managed by `DaemonSet`.
```
bruceg, line 58: The original might have been right—these are values that are able to be tweaked. @MOZGIII?

MOZGIII: I saw this, it looks like the semantics for the two would be identical in this case. I intended to use "tweakables" there, i.e. it's not an error per se - but it may be odd or unconventional wording, while simple and relatively more common "tweaks" seem to deliver the same idea.
So, it is a slight change in meaning, but I don't mind if this form is more appealing to native speakers.

PRIVACY.md
leebenson: good catch, thanks @bruceg

distribution/helm/vector/values.yaml
leebenson: 👍 Swapped it back to 'tweakables'. Jumped out at me initially, but makes sense now.

PR #3648, opened 2020-08-31
docs(rfcs): GraphQL API communications

### LINKED ISSUE #3645: GraphQL as an API comms protocol RFC
Opening an issue to propose GraphQL as a communications protocol for the upcoming API / UI layer. Pre-cursor to RFC.

leebenson:
RFC per #3514. Closes #3645.


rfcs/2020-08-31-3645-graphql-api.md
```diff
@@ -0,0 +1,258 @@
...
+
+* A known spec with 5 years of history.
+* Type safe.
+* Dual HTTP + WebSockets model suits one-time and streaming queries.
+* Production use at companies with heavy API workloads. Article refs: [Github](https://github.blog/2016-09-14-the-github-graphql-api/), [Facebook](https://www.apollographql.com/blog/graphql-at-facebook-by-dan-schafer-38d65ef075af/), [Shopify](https://shopify.dev/concepts/graphql), [Intuit](https://medium.com/intuit-engineering/graphql-intuits-path-to-one-api-system-b8495e4dd281), [Airbnb](https://medium.com/airbnb-engineering/how-airbnb-is-moving-10x-faster-at-scale-with-graphql-and-apollo-aa4ec92d69e2), [Trello / Atlassian](https://www.atlassian.com/engineering/a-look-at-trello-adopting-graphql-and-apollo-in-a-legacy-application).
+* Rich ecosystem of tooling, for server/clients.
+* Trivial debugging story.
+* No proxy/middle-tier required.
+* Works over plain HTTP.
+* Optional TLS.
```
jamtur01, line 81: What potential auth is available?

leebenson: We can use whatever is available to us in the intersection of HTTP + WebSockets.

With browsers locking down headers on WS `Upgrade` requests, in Alloy we passed a JWT in the initial connection, which was then scoped to the life of the subscription.

A `query` is just a HTTP POST request, so we'd typically pass that in an `Authorization: Bearer <jwt>` header.

In both cases, we validated the JWT, looked up the session in the DB, and inserted the session in request context so it could be referenced by GraphQL resolvers.

I played with this a little over the weekend using a sqlx backend, and async-graphql's [Context](https://async-graphql.github.io/async-graphql/en/context.html) to feed a 'user' record through context, and it worked out to be pretty terse. I didn't attempt WebSocket authentication yet, but I anticipate it following the same flow.

jamtur01: Sorry - I should have been clearer. :) I think we should put something in the RFC covering auth - the above ^^^ is pretty close.

leebenson: Will do 👍

leebenson: Added in 927e7d06d03e905dae4d5fdbd91a90f05aa2fa31

rfcs/2020-08-31-3645-graphql-api.md
```diff
@@ -0,0 +1,279 @@
...
+Essentially any browser-compatible protocol could be used, and any text/binary format. This could include other libs such as Protobuf, which might form of a hybrid of the aforementioned approaches.
+
+The maturity of tooling would need further investigation, as well as the development experience of working with any given format.
+
+## Outstanding Questions
+
+* What are the performance characteristics we care about?
+* How many clients do we assume we will be serving across the total of all `vector top` and Vector UI requests?
+* Is JSON an acceptable format for data exchange?
+* What's the GraphQL Client tooling like in Rust, to enable UI-like comms with `vector top`?
```
lukesteensen, line 271: This is a big one for me. Obviously the browser-based frontend is a priority, but it's not the only client we'll want to make sure is robust and fully-featured.

rfcs/2020-08-31-3645-graphql-api.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Alternatives
+
+Essentially any browser-compatible protocol could be used, and any text/binary format. This could include other libs such as Protobuf, which might form of a hybrid of the aforementioned approaches.
+
+The maturity of tooling would need further investigation, as well as the development experience of working with any given format.
+
+## Outstanding Questions
+
+* What are the performance characteristics we care about?
```
lukesteensen, line 268: I'm most curious about data streaming performance. Subscriptions seem like a reasonable abstraction, but I haven't dug into how they're implemented what kind of performance we should be able to expect.

Again, this is probably fine for the browser-based UI but if we're going to commit to a general protocol for getting data out of vector, I'd like to make sure it's going to hold up to other use cases. For example, our current `vector` source and sink use protocol buffers over TCP. It'd be very cool if we could replace that with something more general that allows filtering streams, etc, but I imagine that falls pretty far outside of the normal GraphQL use case. What kinds of issues might we run into if we tried to go down that path? Could we swap out WebSockets for normal sockets when going backend-to-backend?

rfcs/2020-08-31-3645-graphql-api.md
```diff
@@ -0,0 +1,279 @@
...
+* How many clients do we assume we will be serving across the total of all `vector top` and Vector UI requests?
+* Is JSON an acceptable format for data exchange?
```
lukesteensen, line 270: Is JSON usually exposed directly by GraphQL client libraries or is it generally a hidden implementation detail?

Is there support for compression on the WebSocket streams? JSON can be wordy, but compression usually handles that just fine even for high-volume workloads.

rfcs/2020-08-31-3645-graphql-api.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Rationale
+
+We have built a decent body of experience with GraphQL at Timber, albeit on non-Vector projects.
+
+The tooling I am proposing here represents the same stack (save for swapping Apollo for Urql.)
+
+In a previous Timber project, we used [gqlgen](https://github.com/99designs/gqlgen) as our server library, written in Go. The tooling with async-graphql is similar, though Rust's language features enable a more composible 'code first' approach using macros.
+
+I believe the type of data we are consuming benefits from a strongly typed interface, and that compile-time client generation will significantly reduce the time-to-market of what will be already be a very complex front-end web app.
```
lukesteensen, line 245: Could you expand on the workflow here? If the interface types are generated at compile time from the rust code in vector, how do we then consume that in a strongly typed way from the UI? Is there a compilation output from Vector that we need to keep in sync with vector-ui? Do we lose type safety if the UI is on a different version from vector itself? Are there patterns for making changes without breaking compatibility?

I'm sure these are all well understood for people with GraphQL experience, but it'd be nice to lay the answers out a bit more explicitly for those of us who haven't used it.

leebenson: I've added a section titled ["Development Workflow", here.](https://github.com/timberio/vector/blob/093543af0243ee2c4c58b9da1b1e545bbfa4a08d/rfcs/2020-08-31-3645-graphql-api.md#development-workflow) (this points to a commit to permalink it; future viewers may wish to view this on master in case of changes.)

I've included a 16 minute live coding session showing how the UI interacts with the API. This is intended to augment the example code blocks already in the RFC, which I felt were easy to overlook and weren't very indicative of the benefits the front-end team inherits opting for GraphQL. The only way to appreciate it, IMO, is to write code querying live data, so hopefully this video makes it clearer where the potential speed-ups are vs. opting for an untyped / DIY approach.

rfcs/2020-08-31-3645-graphql-api.md
leebenson: > Is JSON usually exposed directly by GraphQL client libraries or is it generally a hidden implementation detail?

Most GraphQL clients deserialize the JSON payload into something meaningful to the language. In Javascript, the payload becomes an object with exactly the same shape. In C# clients, I've seen this be an instance of a class, or an anonymous record.

Whether the client chooses to make the raw JSON payload available is up to the client; in [urql](https://formidable.com/open-source/urql/), we can use middleware to get the JSON, or the JS stdlib to turn it back into JSON:

```js
JSON.stringify(data, null, 2); // <-- JSON with formatting
```

I do this in the [development workflow video](https://www.loom.com/share/9ca38feb21bb488f92d729df0148e029) to show a JSON representation in React.

> Is there support for compression on the WebSocket streams? JSON can be wordy, but compression usually handles that just fine even for high-volume workloads.

There's `permessage-deflate` and custom extensions available to compress JSON payloads over WS if we feel we need it. Payloads are typically a couple of kb or less, depending on the query; we'd have to jam a decent volume of them down the wire before it affected the typical end-client.

async-graphql's `graphql_subscription` middleware is a thin wrapper around Warp's WS connection handling:

```rust
 warp::any()
        .and(warp::ws())
        .and(warp::any().map(move || schema.clone()))
        .map(
            |ws: warp::ws::Ws, schema: Schema<Query, Mutation, Subscription>| {
                ws.on_upgrade(move |websocket| {
                    let (mut tx, rx) = websocket.split();
                    let (mut stx, srx) =
                        schema.subscription_connection(WebSocketTransport::default());

                    let mut rx = rx.fuse();
                    let mut srx = srx.fuse();

                    async move {
                        loop {
                            select! {
                                bytes = srx.next() => {
                                    if let Some(bytes) = bytes {
                                        if let Ok(text) = String::from_utf8(bytes.to_vec()) {
                                            if tx.send(Message::text(text)).await.is_err()
                                            {
                                                return;
                                            }
                                        }
                                    } else {
                                        return;
                                    }
                                }
                                msg = rx.next() => {
                                    if let Some(Ok(msg)) = msg {
                                        if msg.is_text() {
                                            if stx.send(Bytes::copy_from_slice(msg.as_bytes())).await.is_err() {
                                                return;
                                            }
                                        }
                                    } else {
                                        return;
                                    }
                                }
                            }
                        }
                    }
                })
            },
        ).map(|reply| {
        warp::reply::with_header(reply, "Sec-WebSocket-Protocol", "graphql-ws")
    })
        .boxed()
```

What we send to the `tx` channel is up to us; we could swap this out with compressed bytes, or any other format.

The caveat is that the payload is then a DIY implementation. A connecting client would need to be aware of that divergence from the norm, and interpret accordingly. I think a more practical implementation would be passing  headers assert compression capabilities, and selectively compressing (as is typical with `Accept: br`, etc), but I'd consider that a tweak we can kick down the road until we know it's actually warranted.

Warp might even have some helpers there. I know it negotiates Brotli and Gzip for standard HTTP headers, but don't know if it offers anything special for WS. We can always roll our own.

rfcs/2020-08-31-3645-graphql-api.md
leebenson: > I'm most curious about data streaming performance. Subscriptions seem like a reasonable abstraction, but I haven't dug into how they're implemented what kind of performance we should be able to expect.

async-graphql adds very little to subscriptions; it's mostly Warp (in our implementation.)

There's some bytes -> utf8 copying, which could perhaps be improved:

```rust
select! {
    bytes = srx.next() => {
        if let Some(bytes) = bytes {
            if let Ok(text) = String::from_utf8(bytes.to_vec()) {
                if tx.send(Message::text(text)).await.is_err()
                {
                    return;
                }
            }
        } else {
            return;
        }
    }
    msg = rx.next() => {
        if let Some(Ok(msg)) = msg {
            if msg.is_text() {
                if stx.send(Bytes::copy_from_slice(msg.as_bytes())).await.is_err() {
                    return;
                }
            }
        } else {
            return;
        }
    }
}
```

The `tx` channel is the default futures-util `SplitSink` that Warp provides on WS upgraded connections. Aside from iterating over a stream and sending it down the pipe, there's no real magic.

If warp is [too slow](https://www.techempower.com/benchmarks/), there's an [actix-web](https://crates.io/crates/async-graphql-actix-web) integration. Or we could create a TCP wrapper of our own.

Based on [Hashrocket's WS shoot-out](https://github.com/hashrocket/websocket-shootout/blob/master/results/round-02.md#rust---ws) (and all things being equal), we'd likely have to exceed 5,000 - 10,000+ connected clients for WS transmission to bust the 100ms mark. Given we're targeting single users (or at most, a handful) and metrics that currently tick over at 2 second intervals... even with dozens of multi-KB payloads multiplexed over the same connection, the constraint is likely to be network I/O before it's anything instrinsically non-performant about WebSockets.

This is completely unscientific, and running on a relatively beefy 8-core i9 laptop with 32gb RAM, but here's a quick screencast showing 9 connected Chrome clients observing a single Vector instance, with events generated linearly at at 5ms intervals, aggregrated at 100ms intervals (per client), with all the overhead of 9x Chrome windows running 9x v8 stacks with 9x React and 9x urql clients on 9x local Websocket connections...

https://www.loom.com/share/b94f952fd520491bb2bec30a8dd07c34

Pause the video on any frame. Event numbers are exactly in sync, +/- 99ms to account for variations between intervals.

> Again, this is probably fine for the browser-based UI but if we're going to commit to a general protocol for getting data out of vector, I'd like to make sure it's going to hold up to other use cases. 

GraphQL is ideally suited to a general API, because it's just typed JSON over HTTP. Every language has a stdlib or de facto HTTP client that would work with it. This applies to every query/mutation (i.e. GET/POST-type query)

Streaming data is always going to require a bit more thought, by virtue of its opinionated protocols and inherently longer-lived connections. I don't think the choice between GraphQL or something has necessarily has to be binary. If we want to offer an alternative protocol, I see at least two options:

1. Use the GraphQL spec. Like the Warp WS snippet, this is just a case of iterating over a stream and sending it down whatever channel backs into the connected client. Same I/O otherwise.

2. Ignore the GraphQL spec. gRPC might be a better choice for some clients (IMO it's a poor choice for the web, for the reasons outlined in the RFC, but probably preferable for service-to-service connections.) In which case, the gRPC would drive the interface for that particular gateway, and likely implement its own abstraction over our internal types, auth, and whatever else is specific to gRPC.

As a 'general purpose' API though... one that has the least baggage from a simple protocol to handle both one-time queries and streaming data... plain ol' HTTP and WebSockets with JSON payloads is probably as simple and therefore as universally applicable as any other choice we could make.

sunli829: If necessary, I will improve.

rfcs/2020-08-31-3645-graphql-api.md
leebenson: If the client can send JSON payloads over HTTP, then it can handle GraphQL queries and mutations. If it can connect HTTP and handle WS upgraded and longer lived TCP connections, it can handle streaming. If it has a dedicated GraphQL client that can do those things _and_ add semantic meaning to the data it's sending and receiving -- then it can benefit from the type hinting espoused in the [development workflow video](https://www.loom.com/share/9ca38feb21bb488f92d729df0148e029).

Anything else we need, we can either bolt on, or work around as an aside. Namely:

- Compression. Add `permessage-deflate` or a DIY compression via `Sec-WebSocket-Extensions`.
- Message serialization. Transform JSON to binary or something else.
- gRPC. Have a separate endpoint for it to work alongside. Doesn't have to be either/or.

For `vector top` specifically, there's https://github.com/graphql-rust/graphql-client. I haven't dug into it much yet, but evidently it has some form of subscription support and even works as WASM in the browser. There's probably others-- I've only looked superficially but will dive in more once _how_ we're observing topology is solved.

rfcs/2020-08-31-3645-graphql-api.md
leebenson: > Is there a compilation output from Vector that we need to keep in sync with vector-ui?

@sghall and I talked about this a while back, and there was a brief discussion in the internal pre-RFC with some ideas, but I don't think we had enough context back then (and maybe still don't) to reach a firm consesus.

The ideas were roughly:

1. Embedding the `index.html` equivalent in Vector with some dynamic values injected to set-up the target endpoint, and then loading the rest via a `<script>` tag pointing to a canonical URL. This would free us up from having to embed a (potentially quite weighty) set of JS/CSS files into Vector, and would simplify the story around local hosting. The downside is the requirement for an active Internet connection for the end-user (i.e. the person loading the admin UI, and _not_ Vector itself-- which could still be locked down.)

2. Embed everything in Vector, so the UI and Vector are always versioned, and always in sync. The downside is the extra 2-5mb added to the binary, which will likely keep growing for a while.

3. Embed nothing (or just index.html) and then have a separate `--ui-path` to point to the UI files. Pros: No drag on the binary. Cons: Difficult versioning/upgrade story; more complex config.

Would be interested to get some feedback on the above.

> Do we lose type safety if the UI is on a different version from vector itself? Are there patterns for making changes without breaking compatibility?

Yes. This is more insidious than just loosing type safety-- queries will error if fields or types have changed. There are various best practices that have been kicked around / are quasi-established for handling versioning (@jszwedko is probably the most clued up, here-- he worked through a guide relatively recently which tackled some of the more advanced concerns that can crop up), but the general take-away is to try to avoid breaking changes where they're not necessary, and that the client should be in sync. A similar concern domain as NoSQL (although the analogy breaks down with mutations, which would also error if the query didn't match.)

I consider this a good thing, because - like protobufs - we're guarding against undefined behavior in the type system before even attempting to parse a query. But it does mean the UI has to match a given version of Vector (or at least a given version of its schema.)

rfcs/2020-08-31-3645-graphql-api.md
sghall: I think @leebenson did a great job explaining above.  I just want to amplify one piece that he mentioned in passing: multiplexing.  GraphQL subscriptions are meant to be light weight.  In a UI backed by GraphQL they can be created and destroyed at will as the user is moving around in the app (when FB introduced subscriptions, they explained that they were using these to subscribe and unsubscribe users to "likes" on every piece of content as they were scrolling through their feed).  GraphQL subscriptions are a way to multiplex lots of messages over a single connection.  To me, this goes right at the the performance question.  This paradigm provides a way for clients to subscribe and unsubscribe over a single multiplexed connection to exactly the content they want with an expressive query language with dramatically less over-fetching.  The single connection also means there is not all the setup and tear down cost of an HTTP request for each of these interactions.

rfcs/2020-08-31-3645-graphql-api.md
sghall: Yeah,  coordinating the server and client pieces is a little tricky.  The UI is pulling the types from a running instance of the GraphQL server and so the process needs some thought.  Seems like we're going to need a CI process that will type check the UI code vs the schema to get 100% assurance (?).  Something like pulls the UI by tag (?) from GitHub (?) and checks the types and builds it.  As far as what exactly it builds and where it goes, like to get some guidance on that.   Big question obviously on whether we can assume an internet connection.  IMHO that will drive usage down if they need to go out to the web from a secure environment.

rfcs/2020-08-31-3645-graphql-api.md
sghall: Wondering if we're (possibly) talking about having different types of transports supported for different types of clients (eventually).   Not sure what that would mean implementation wise but I don't think GraphQL itself is dictating how the transport of messages occurs.  Usually, when you're talking about GraphQL you're talking about doing stuff over HTTP between a client (browser) and a server.  That's the most common use case and really the only one I have any experience with.  I went back today was watching this video from Lee Byron (one of the main people that developed the GraphQL spec) where he introduced subscriptions.  This is a much more abstract idea about events streams that is not necessarily tied to any particular transport method.  He talks about using messaging pub-sub type protocols like MQTT (or ZeroMQ) inside of FB and even web hook callbacks.  

If you've never watched this video, it may be worth taking a look...
https://www.youtube.com/watch?v=bn8qsi8jVew

rfcs/2020-08-31-3645-graphql-api.md
leebenson: >  It'd be very cool if we could replace that with something more general that allows filtering streams, etc, but I imagine that falls pretty far outside of the normal GraphQL use case.

If you mean 'take a single stream, and return from it only what the client asked for', then that's very much in GraphQL's wheelhouse and a common use-case for subscriptions.

There are two ways we can leverage GraphQL for this:

### Define specific queries

Instead of a catch-all `metrics` subscription, we could have multiple `subscription` definitions which filter on the same underlying stream, but only return the relevant type.

For example -- imagining a 'bytes' and an 'events' metric, the queries could look like:

```gql
{
  bytesMetrics {
    bytesProcessed
  }
}
```

and

```gql
subscription {
  eventsMetrics {
    numberOfEvents
  }
}
```

An example of this approach is in #3514. I have a func that returns a 'global' controller, like this...

```rust
/// Returns a stream of `Metric`s, collected at the provided millisecond interval
fn get_metrics(interval: i32) -> impl Stream<Item = Metric> {
    let controller = get_controller().unwrap();
    let mut interval = tokio::time::interval(Duration::from_millis(interval as u64));

    stream! {
        while let _ = interval.tick().await {
            for ev in capture_metrics(&controller) {
                if let Event::Metric(m) = ev {
                    yield m
                }
            }
        }
    }
}
```

... and a `metrics` subscription which (currently) only returns 'events_processed' metrics, like this:

```rust
#[Subscription]
impl MetricsSubscription {
    /// Returns all Vector metrics, aggregated at the provided millisecond interval
    async fn metrics(
        &self,
        #[arg(validator(IntRange(min = "100", max = "60_000")))] interval: i32,
    ) -> impl Stream<Item = MetricType> {
        get_metrics(interval).filter_map(|m| match m.name.as_str() {
            "events_processed" => Some(MetricType::EventsProcessed(m.into())),
            _ => None,
        })
    }
}
```

For additional metrics types, we'd just create other funcs representing the parts of the stream the user firing that subscription cares about. Everything else would get mapped to `None`, and therefore would be ignored.

### Using parameters

We could also use input params to control the stream. So by default, a stream would return everything -- but could be filtered to 'subscribe' to just the streams it cares about.

Pseudo:

```rust
enum SubscribeToMetric {
    Events,
    Bytes,
}

#[Subscription]
impl MetricsSubscription {
    /// Returns all Vector metrics, aggregated at the provided millisecond interval
    async fn metrics(
        &self,
        #[arg(validator(IntRange(min = "100", max = "60_000")))] interval: i32,
        subscribe_to: SubscribeToMetric,
    ) -> impl Stream<Item = MetricType> {
        get_metrics(interval).filter_map(|m| match (m.name.as_str(), subscribe_to) {
            ("events_processed", SubscribeToMetric::Events) => {
                Some(MetricType::EventsProcessed(m.into()))
            }
            ("bytes_processed", &SubscribeToMetric::Bytes) => {
                Some(MetricType::EventsProcessed(m.into()))
            }
            _ => None,
        })
    }
}
```

In this case, both the events and the params would need to match, for the metric to be propagated to the WebSocket client.

---

This doesn't have to be limited to a single metric. We could take an array of metric types, and return them as part of the same subscription. Or break them all up, and let the client multiplex per @sghall's comments. A combination of well-defined subscriptions and flexible input filters could yield whatever we need.

sunli829: Haha, you found an unnecessary memory copy, I optimized it, and also optimized some other places about subscriptions. 😁

leebenson: Thanks @sunli829, that's awesome!

rfcs/2020-08-31-3645-graphql-api.md
leebenson: >  Seems like we're going to need a CI process that will type check the UI code vs the schema to get 100% assurance

Yeah, I think that kinda thing is going to be the right move. I'd treat Vector as the compilation target. Whether it's embedding in the binary or some other co-location strategy, the UI should be bundled against a specific version of Vector and tested via the same `yarn workspace vortex graphql`-type live introspection in GH Actions. 

If it's remote, we could have `<script src="https://cdn.vector.dev/js/main.<vector_version>.js">` to keep them in sync. If it's bundled, then some CI approach that builds from [vector-ui](https://github.com/timberio/vector-ui).

cc @Hoverbear, @lukesteensen - do you have any initial thoughts on how you foresee static UI files integrating with Vector? The loading strategy and how we handle cross-repo compilation should be driven by those conclusions, IMO.

leebenson:
@JeanMertz, I've some meat to "Prior Art". I'm not sure how satisfying this will be; I've not been able to find any public RFCs of similar internal UIs settling on GraphQL. The ones I've come across tend to use JSON-- Rancher does this over WebSockets (untyped, AFAICT), Kubernetes uses OpenAPI. [Types are defined here](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/typings/backendapi.ts). Didn't look too closely as to whether these were auto-generated, or hand-spun. Given some of the inline comments, I'd assume the latter, but haven't qualified this further.

The k8s `/metrics` endpoint returns 4mb+ of data against one of the dev servers I have Kubernetes installed at. It doesn't seem particularly efficient, but perhaps it doesn't need to be.

If you wouldn't mind reviewing this please, I'll get this merged given the agreement during stand-up to move ahead with GraphQL for now and get some real experience with it.

JeanMertz (approved):
This looks great @leebenson! 👍 Very detailed, precise, and well-structured.

I left a couple of comments and a few nits that you should feel free to ignore, especially given we've already agreed to merge this RFC.

The only thing I would note is that it would be in the interest of future readers to resolve the outstanding questions _within the RFC itself_ instead of the GitHub comments. Once this PR is merged, if I were to read the RFC locally, I would get the idea that there are still many unanswered questions that should have precluded this RFC from being accepted, while that's not actually the case.

This applies in general to questions raised through RFC comments; I believe it's best to resolve them by amending the document with new commits, and then pointing to the specific commits that address the concern/question via a comment reply, instead of replying directly (unless, of course, a discussion needs to happen first, before a commit can be made with the right solution to the concern).

leebenson:
> The only thing I would note is that it would be in the interest of future readers to resolve the outstanding questions within the RFC itself instead of the GitHub comments. Once this PR is merged, if I were to read the RFC locally, I would get the idea that there are still many unanswered questions that should have precluded this RFC from being accepted, while that's not actually the case.

👍 I agree with this. I think there's generally going to be discussion to be had first, though, which is relevant in GH comments. Some threads had multiple participants; I think before merging, it makes sense to aggregate those comments and dump the links in a Q&A section (which I'll do for this one, too), but I think it's worth teasing out the discussion first in case a point is unsettled or introduces ideas that materially changes the content.

I think that should be a final step before merging, and perhaps outside of the review process, given that discussion has already happened and it forms a conclusion to the RFC, rather than the RFC itself.

rfcs/2020-08-31-3645-graphql-api.md
```diff
@@ -0,0 +1,297 @@
...
+This RFC focuses on the web UI, but applies equally to the CLI client due to the possibility of observing a remote Vector instance.
+
+### Protocol
+
+This section summarizes various communications protocols, and their disadvantages for Vector observability:
+
+**REST:**
+
+* It's not suited for streaming data.
+* It requires polling and/or long-lived connections.
```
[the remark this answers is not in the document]
leebenson: This is perhaps lacking context and needs some extra explanation. The issue with HTTP2 isn't that it's long-lived (WebSockets are long-lived, too, after all-- and I'm pro WS). It's that there's less control in the browser to take advantage of that. With WS, we get a clear path to full duplex messaging, which GraphQL subscriptions take advantage of. AFAIK, the lack of granular HTTP2 control in the browser is why gRPC-Web is necessary and gRPC can't be used directly.

rfcs/2020-08-31-3645-graphql-api.md
```diff
@@ -0,0 +1,297 @@
...
+* Data will initially be read-only; many HTTP verbs don't apply to us.
```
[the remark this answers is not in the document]
leebenson: My point here is that our data needs don't particularly fit the REST paradigm, and that I feel they're more closely aligned with the GraphQL concepts of `query` and `subscription`. I'll change that to be clearer.

rfcs/2020-08-31-3645-graphql-api.md
```diff
@@ -0,0 +1,297 @@
...
+
+* Asynchronous data payloads that are suited to streaming data, but harder to reconcile for single requests.
+* Related: The need to match up a 'request ID' with a 'response ID', to determine which payload belongs to a given request.
+* Long-lived connections that require explicit (re)connection logic.
+* Not intrinsically type-safe for the client to consume; high degree of runtime validation.
+* Non-trivial for non-web clients to interact with; not a common protocol outside of the web.
+
+## Internal Proposal
+
+I propose using GraphQL for API communications.
```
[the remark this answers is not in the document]
leebenson: I get that point, although I don't know whether that reflects the nature of an RFC.

I take an RFC to mean, literally, a request for comments. I'm hoping to take my singular ideas and spitballings, and refine them through collaboration. By the end, the _conclusion_ of the RFC is collaborative and improved with feedback. But at the point of writing it, it's just one person's opinions/assumptions. Writing it as a team masks that evolution, IMO - I'd rather see "This is what I propose..." followed by "This is what we wound up with".

Others may disagree with this. I think it depends who we intend the audience to be. IMO, the features that fall out of an RFC and the resulting docs / blog posts, are for the public. The RFC is for one person proposing, and others improving.

rfcs/2020-08-31-3645-graphql-api.md
lukesteensen: > For `vector top` specifically, there's https://github.com/graphql-rust/graphql-client. I haven't dug into it much yet, but evidently it has some form of subscription support and even works as WASM in the browser.

Yeah, this is something that'd be nice to dig into a bit. I'm sure we can flesh out support ourselves if needed, but it's going to be important to have a first-class, ergonomic, performant Rust interface to this API.

> Wondering if we're (possibly) talking about having different types of transports supported for different types of clients (eventually).

That's exactly what I'm curious about. If we're using a non-browser client it could be nice to swap out WebSockets for plain old TCP, but it may not really matter.

rfcs/2020-08-31-3645-graphql-api.md
lukesteensen: Makes sense. As a backend person, WebSockets definitely stick out as an "extra" thing I would never normally use outside of the browser, but the difference in overhead between them and normal TCP sockets is likely not high.

> We could also use input params to control the stream. So by default, a stream would return everything -- but could be filtered to 'subscribe' to just the streams it cares about.

This would definitely be an interesting route. Allowing a client to fetch a catalog of the data that's available and then send a "query" that returns of stream of just the data matching that query would be really cool.

rfcs/2020-08-31-3645-graphql-api.md
lukesteensen: > I consider this a good thing, because - like protobufs - we're guarding against undefined behavior in the type system before even attempting to parse a query. But it does mean the UI has to match a given version of Vector (or at least a given version of its schema.)

This is actually one of the benefits of protocol buffers, in that there are a specific set of rules for how you can evolve their definition that allows clients on different versions to continue to interoperate. For example, adding new fields is always compatible and older clients will simply ignore fields they don't know about.

Since it sounds like GraphQL is pretty much just JSON, I'd imagine the specific behavior there could depend on the implementation. I just want to be realistic and understand that UI and backend versions won't always be exactly the same, so the way we handle mismatches is going to matter. It could be as simple as forcing a refresh if we detect an incompatible change, but it'd be nice if there were some known safe migration process for GraphQL types.





> do you have any initial thoughts on how you foresee static UI files integrating with Vector?

I think we'd take the static files as an input to our release process and bundle them with our distribution artifacts in some fashion. Vector itself would then act as a static file server for them. I don't want to have to assume any connectivity other than to the vector process itself.

rfcs/2020-08-31-3645-graphql-api.md
```diff
@@ -0,0 +1,297 @@
...
+
+The tooling I am proposing here represents the same stack (save for swapping Apollo for Urql.)
+
+In a previous Timber project, we used [gqlgen](https://github.com/99designs/gqlgen) as our server library, written in Go. The tooling with async-graphql is similar, though Rust's language features enable a more composible 'code first' approach using macros.
+
+I believe the type of data we are consuming benefits from a strongly typed interface, and that compile-time client generation will significantly reduce the time-to-market of what will be already be a very complex front-end web app.
+
+## Prior Art
+
+From relatively superficial searching, I've not been able to find a comparable approach to using GraphQL for internal metrics / observability. However, there are [plenty of large companies](https://graphql.org/users/) using GraphQL for public API and very possibly for internal machinery that isn't public facing.
```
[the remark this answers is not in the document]
lukesteensen: Agreed, and especially in this case where the primary question is around the specific use case. I have no doubt that GraphQL is robust and well supported for its designed use case. The risk for us is whether that's also true for streaming data from an infrastructure tool instead of a more standard web app.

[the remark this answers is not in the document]
leebenson: I agree, but in the absence of companies using it in exactly the way we intend, I also think it's better than nothing.

I'm not sure what detail we're hoping to glean from prior art, or to a large extent, even the practical benefit. Do we care if another company is using JSON or Protobufs or GraphQL, if there are equally as many companies using the alternatives? This is also making a large assumption that other companies have chosen wisely/correctly, and that their circumstances apply to us.

I don't put a lot of stock into this research, unless it very obviously (and relatively recently) overlaps with our use-case.

lukesteensen (approved):
I think I've seen enough to be comfortable moving forward, so thanks @leebenson for your patience and thorough explanations :smile: 

The biggest things I'd like to de-risk as we move forward are non-web clients and our schema evolution story. If we can tackle those first and sort out any issues then hopefully it'll be pretty smooth sailing from there.

rfcs/2020-08-31-3645-graphql-api.md
leebenson: > Since it sounds like GraphQL is pretty much just JSON, I'd imagine the specific behavior there could depend on the implementation.

A GraphQL query will always be valid, assuming the requested fields exist, and substituted input parameters match the expected type.

This is by design; one of the main intentions of GraphQL is to prevent over-fetch, which allows a query to request a subset of fields and not the full available set. Given that GraphQL data can be highly nested (the 'graph' in graphql), it wouldn't be practical or even possible to request every field, anyway -- it supports infinite recursion.

For example, a query could look like this:

```gql
query SourcesToSinks($beginsWith: String!){
  sources {
    id
    name
    transforms(filter: { name: { beginsWith: $beginsWith}}) {
      id
      name
      sinks {
        id
        name
      }
    }
  }
}
```

Note we don't specify the types. We just expect on the client, given prior introspection, that they will match expectations. Reasons this query could fail include:

- One or more of the fields don't exist
- The 'transforms' filter input isn't valid
- The nested `beginsWith` param isn't of type `String!`
- The client requesting doesn't have permission to request a given field
- A nesting/recursion/complexity limit has been defined in the server, which the query violates

Otherwise, it will pass.

This would also pass:

```gql
query JustSources {
  sources {
    id
    name
  }
}
```

Note in this case, we don't requested the related `transforms` (or sinks) at all. By design, we can request just what we need for a given query.

The practical effect to this is backwards compatibility for any added fields. If the query was written when `transforms` didn't exist, it would still work. Queries are validated at runtime by virtue of not receiving an error.

This is the same for all GraphQL clients, regardless of language or implementation.

A breaking change would be if we requested, say, `name`, and name didn't exist... or if `beginsWith` was modified to a type other than `String!` (in this case.)

leebenson: > I think we'd take the static files as an input to our release process and bundle them with our distribution artifacts in some fashion. Vector itself would then act as a static file server for them. I don't want to have to assume any connectivity other than to the vector process itself.

👍 sounds good. Do you foresee these being embedded in the binary itself? Or saved in a separate dir, which `vector` could be configured to point to? I don't have context re: existing distribution artifacts, so not sure if there's already a concept of third-party file deps we could leverage.

sunli829:
@leebenson 

‘Vector top’ can try this. It looks good, but I don't know if it is stable. I haven't used it before.

https://crates.io/crates/cynic

leebenson:
Added a ToC, and links to relevant discussion points. I'll get this merged since it's dragged on a bit, but happy to tackle further amendments either by way of a more granular RFC, or editing this one later.

PR #3650, opened 2020-08-31
chore(ci): Helm lint

MOZGIII:
This PR adds running `helm lint` at CI.

It is very simple, yet it poses an important question - is this the right way to manage Helm installation? Should we, maybe, keep it as part of our environment? If so, how is should look like?
CC @Hoverbear 

We'll also use Helm to generate YAML for Kubernetes - there's an upcoming PR.


Hoverbear (approved):

Hoverbear:
Given that most tests don't want or need helm, this seems fine. We can change it later if it ends up not being right.

PR #3661, opened 2020-09-01
fix(tests): Fix Splunk integration test

leebenson:
Attempts to fix the Splunk integration test, by removing the dupe `custom_index` index which is later re-added.

Added `wait_for_duration` + `wait_for_tcp_duration` helper helpers to test_utils, to specify a custom `Duration` to await a TCP port being reachable (8089, in this case.) Added a defensive 30s wait time. Shouldn't generally be the case, but since these are disparate Docker services, a little extra wiggle room can't hurt.

Refactored `wait_for` and `wait_for_tcp` to use `wait_for_duration`, to avoid changing existing signatures.


fanatid (approved):
LGTM if CI is green

PR #3670, opened 2020-09-01
chore(sinks): add generic to InFlightLimit

### LINKED ISSUE #3487: Deserializer for InFlightLimit without auto and none
`InFlightLimit` option allows `auto` and number in vector config, but some sinks support only a fixed number of requests. As a result if the config will have `auto` variant vector will panic on sink building. For such cases, we can define additional deserializer which will invalidate config on parsing.

`InFlightLimit`: https://github.com/timberio/vector/blob/133d4b59cdbe2eb42a65a4b661918c103b820a80/src/sinks/util/service2.rs#L64-L69

Ref. https://github.com/timberio/vector/pull/3422#discussion_r469583796

fanatid:
Closes #3487.

@bruceg thanks a lot for generic suggestions in the issue!

### DIFF
--- a/src/sinks/util/service.rs
+++ b/src/sinks/util/service.rs
@@ -126,11 +126,11 @@ impl<'de> Deserialize<'de> for InFlightLimit {
 }
 
 pub trait InFlightLimitOption {
-    fn unwrap_with(&self, default: &Self) -> Option<usize>;
+    fn parse_in_flight_limit(&self, default: &Self) -> Option<usize>;
 }
 
 impl InFlightLimitOption for Option<usize> {
-    fn unwrap_with(&self, default: &Self) -> Option<usize> {
+    fn parse_in_flight_limit(&self, default: &Self) -> Option<usize> {
         let limit = match self {
             None => *default,
             Some(x) => Some(*x),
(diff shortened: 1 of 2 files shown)


src/sinks/util/service.rs
```diff
@@ -125,11 +125,35 @@ impl<'de> Deserialize<'de> for InFlightLimit {
     }
 }
 
+pub trait InFlightLimitOption {
+    fn unwrap_with(&self, default: &Self) -> Option<usize>;
+}
```
bruceg, line 130:
```suggestion
pub trait ParseInFlightLimit {
    fn parse_in_flight_limit(&self, default: &Self) -> Option<usize>;
}
```

fanatid: Just now I realized that instead of custom trait I can use [std::convert::Into](https://doc.rust-lang.org/std/convert/trait.Into.html).

fanatid:
> from an `Option<usize>` to an `enum Something { Auto, Fixed(usize) }`

I'm not sure :thinking: in `aws_cloudwatch_logs` `in_flight_limit` value can be number (`InFlightLimit::Fixed`) or nothing (`InFlightLimit::None`), not `auto`.

PR #3671, opened 2020-09-01
perf: Use RTT variance instead of a fixed threshold

bruceg:
Instead of using a fixed threshold, this calculates the standard deviation of the EWMA of the past RTT measurements, and uses a multiple of that as the threshold range.

### DIFF
--- a/src/sinks/util/auto_concurrency/controller.rs
+++ b/src/sinks/util/auto_concurrency/controller.rs
@@ -20,11 +20,6 @@ use tower::timeout::error::Elapsed;
 // weighting.
 const EWMA_ALPHA: f64 = 0.5;
 
-// This was picked as a reasonable default threshold ratio to avoid
-// dropping concurrency too aggressively when there is fluctuation in
-// the RTT measurements.
-const THRESHOLD_RATIO: f64 = 0.05;
-
 /// Shared class for `tokio::sync::Semaphore` that manages adjusting the
 /// semaphore size and other associated data.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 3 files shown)


src/sinks/util/auto_concurrency/controller.rs
```diff
@@ -252,23 +253,47 @@ where
...
-            None => point,
-            Some(avg) => point * EWMA_ALPHA + avg * (1.0 - EWMA_ALPHA),
+        let (average, variance) = match self.average {
+            None => (point, None),
+            Some(avg) => {
+                let delta = point - avg;
+                let variance = self.variance.unwrap_or(0.0);
+                (
+                    point * self.alpha + avg * (1.0 - self.alpha),
+                    Some((1.0 - self.alpha) * (variance + self.alpha * delta * delta)),
```
jszwedko, line 291: Would you be able to explain a bit what this is doing?

jszwedko (commented):
:tada: 

I'll try this out with the tests we have so far.

src/sinks/util/auto_concurrency/controller.rs
jszwedko: (specifically the variance bit of the calculation)

jszwedko:
I note that I see:

```
warning: method is never used: `new`
   --> src/sinks/util/auto_concurrency/controller.rs:266:5
    |
266 |     fn new(alpha: f64) -> Self {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: 1 warning emitted
```

When building.

I reran the tests (including the added "knee behavior" tests).

Some observations:
* It seems like it does better in the small variance case, but actually appears worse in the medium/large variance cases
* It seems to do better for the knee case with a higher slope

Before (424ea5e0):

all

After (6670a0e4):

all


I'm realizing it'd probably be useful to plot before/after on the same graphs (specifically the throughput) to make it easier to see differences. I opened https://github.com/timberio/http_test_server/issues/4 to that end.

bruceg:
Since this doesn't seem to have helped any, at least with our simulations, and needs some work to be usable, I will close this for now and see what happens in real deployment situations.

PR #3672, opened 2020-09-02
chore: consolidate config compilation

### LINKED ISSUE #2936: Organize config loading, parsing, checking, building into staged compiler

lukesteensen:
Closes #2936 

The overall goal here is to ensure that every `Config` instance is always valid and fully constructed. This means going through type checking, DAG validation, macro expansion, etc.

To do this, we bring all of those stages together in the `config` module and lock things down such that an invalid or incomplete `Config` can never escape the `config` module. This gives us the ability to add new stages (e.g. transform inlining) in one place, knowing they will be applied consistently.

More specifically, this PR introduces a new `ConfigBuilder` struct that takes over being {,de}serializable, clonable, test-friendly, etc, allowing us to lock down the actual `Config` struct significantly, making it harder to misuse.

There are definitely a lot of compromises here, mostly around the way that subcommands like `validate` and `test` expect to be able to use configs that would normally be invalid. I tried to make the changes relatively minimal, which means there are definitely areas that we can revisit and improve as needed in the future.

### DIFF
--- a/src/config/builder.rs
+++ b/src/config/builder.rs
@@ -1,6 +1,6 @@
 use super::{
-    default_data_dir, Config, GlobalOptions, SinkConfig, SinkOuter, SourceConfig, TestDefinition,
-    TransformConfig, TransformOuter,
+    compiler, default_data_dir, Config, GlobalOptions, SinkConfig, SinkOuter, SourceConfig,
+    TestDefinition, TransformConfig, TransformOuter,
 };
 use crate::event;
 use indexmap::IndexMap;
@@ -34,16 +34,8 @@ impl Clone for ConfigBuilder {
 }
 
 impl ConfigBuilder {
-    // TODO: compilation here
-    pub fn build(self) -> Config {
-        Config {
-            global: self.global,
-            sources: self.sources,
-            sinks: self.sinks,
-            transforms: self.transforms,
-            tests: self.tests,
-            expansions: Default::default(),
-        }
+    pub fn build(self) -> Result<Config, Vec<String>> {
+        compiler::compile(self)
     }
 
     pub fn add_source<S: SourceConfig + 'static>(&mut self, name: &str, source: S) {
--- a/src/config/builder.rs
+++ b/src/config/builder.rs
@@ -37,11 +37,16 @@ impl ConfigBuilder {
         compiler::compile(self)
     }
 
-    pub fn add_source<S: SourceConfig + 'static>(&mut self, name: &str, source: S) {
-        self.sources.insert(name.to_string(), Box::new(source));
+    pub fn add_source<S: SourceConfig + 'static, T: Into<String>>(&mut self, name: T, source: S) {
+        self.sources.insert(name.into(), Box::new(source));
     }
 
-    pub fn add_sink<S: SinkConfig + 'static>(&mut self, name: &str, inputs: &[&str], sink: S) {
+    pub fn add_sink<S: SinkConfig + 'static, T: Into<String>>(
+        &mut self,
+        name: T,
+        inputs: &[&str],
+        sink: S,
+    ) {
         let inputs = inputs.iter().map(|&s| s.to_owned()).collect::<Vec<_>>();
         let sink = SinkOuter {
             buffer: Default::default(),
@@ -50,12 +55,12 @@ impl ConfigBuilder {
             inputs,
         };
 
-        self.sinks.insert(name.to_string(), sink);
+        self.sinks.insert(name.into(), sink);
     }
 
-    pub fn add_transform<T: TransformConfig + 'static>(
+    pub fn add_transform<T: TransformConfig + 'static, S: Into<String>>(
         &mut self,
-        name: &str,
+        name: S,
         inputs: &[&str],
         transform: T,
     ) {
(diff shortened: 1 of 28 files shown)


ktff (commented):
Nice

src/config/builder.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[serde(default)]
+    pub sinks: IndexMap<String, SinkOuter>,
+    #[serde(default)]
+    pub transforms: IndexMap<String, TransformOuter>,
+    #[serde(default)]
+    pub tests: Vec<TestDefinition>,
+}
+
+impl Clone for ConfigBuilder {
+    fn clone(&self) -> Self {
```
Hoverbear, line 25: We talked about adopting bincode (https://github.com/servo/bincode/blob/master/Cargo.toml) or cbor (https://github.com/pyfisch/cbor), maybe this is a good chance for that?

src/config/builder.rs
```diff
@@ -0,0 +1,150 @@
...
+        // This is a hack around the issue of cloning
+        // trait objects. So instead to clone the config
+        // we first serialize it into JSON, then back from
+        // JSON. Originally we used TOML here but TOML does not
+        // support serializing `None`.
+        let json = serde_json::to_vec(self).unwrap();
```
Hoverbear, line 31: Could you turn it into a `serde_json::Value` instead to save on the cost of making everything into a contiguous byte array?

src/config/builder.rs
```diff
@@ -0,0 +1,150 @@
...
+        serde_json::from_slice(&json[..]).unwrap()
+    }
+}
+
+impl ConfigBuilder {
+    pub fn build(self) -> Result<Config, Vec<String>> {
+        compiler::compile(self)
+    }
+
+    pub fn add_source<S: SourceConfig + 'static>(&mut self, name: &str, source: S) {
```
Hoverbear, line 41: API guidelines  suggest we allow the caller to determine copy and placement: https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control

How about `name: impl Into<String>`?

src/config/compiler.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    if errors.is_empty() {
+        Ok(config)
+    } else {
+        Err(errors)
+    }
+}
+
+/// Some component configs can act like macros and expand themselves into
+/// multiple replacement configs. Returns a map of components to their
```
Hoverbear, line 40: This suggests it returns the map, but it sets it in `config.expansions` instead.

src/config/unit_test.rs
```diff
@@ -1,22 +1,61 @@
...
 
-pub struct UnitTestTransform {
-    transform: Box<dyn Transform>,
-    next: Vec<String>,
+fn build_unit_tests(builder: ConfigBuilder) -> Result<Vec<UnitTest>, Vec<String>> {
+    let mut tests = vec![];
+    let mut errors = vec![];
+
+    // Don't let this escape since it's not validated
+    let mut config = Config {
```
Hoverbear, line 29: I could see this and the following line becoming `Config::build_unchecked`?

src/topology/mod.rs
```diff
@@ -71,29 +70,16 @@ pub async fn start_validated(
     Some((running_topology, abort_rx))
 }
 
+// TODO: inline this
```
Hoverbear, line 73: Got an issue for this? Or is it for this PR still?

src/topology/mod.rs
```diff
@@ -736,18 +715,15 @@ mod reload_tests {
             },
         );
 
-        let (mut topology, _crash) = start_topology(old_config, false).await;
-        assert!(!topology
-            .reload_config_and_respawn(new_config, false)
-            .await
-            .unwrap());
+        let (mut topology, _crash) = start_topology(old_config.build().unwrap(), false).await;
+        assert!(new_config.build().is_err());
```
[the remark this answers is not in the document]
lukesteensen: Great catch! Glad that this change is already finding issues like this.

src/config/builder.rs
lukesteensen: It could be, but I'm not sure the benefit of changing it at this point. JSON works fine and this is not a performance sensitive code path.

src/config/builder.rs
lukesteensen: This was all just copied but sure :smile:

Hoverbear (approved):

src/config/mod.rs
```diff
@@ -13,31 +13,29 @@ use snafu::{ResultExt, Snafu};
...
-pub use validation::check;
+pub use unit_test::build_unit_tests_main as build_unit_tests;
+pub use validation::warnings;
 
-#[derive(Deserialize, Serialize, Debug)]
-#[serde(deny_unknown_fields)]
+#[derive(Debug, Default)]
 pub struct Config {
-    #[serde(flatten)]
     pub global: GlobalOptions,
```
MOZGIII, line 34: Should we make those private now?

MOZGIII (approved):
This is great

src/topology/mod.rs
lukesteensen: I started down this path and it lead to a bit more than I want to add to this PR, so I'll just rename it to something more appropriate. We can continue the refactoring later.

src/config/mod.rs
lukesteensen: I did try putting most of these behind methods at least, but it ended up being a little more than I want to bite off right now. That's something we can do as a followup.

PR #3673, opened 2020-09-02
chore: Added back add index after Docker image fix

jamtur01:
I worked out what was wrong with the Docker image that was making the add index duplicated and fixed that. Now adding the index is the right step in the docker-compose file.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -444,7 +444,7 @@ services:
       [
         "sh",
         "-c",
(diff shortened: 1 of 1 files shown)


leebenson (approved):
LGTM. Where is `timberio/splunk-hec-test` being pulled from? I tried to go direct to source, but wasn't sure how they were being built/published.

jszwedko:
There appears to be a private repo with the Dockerfile for that. I'm trying to figure out if it could just be public.

jamtur01:
For reference - https://github.com/timberio/splunk-hec-test.

PR #3677, opened 2020-09-02
chore(rfcs): API topology RFC

leebenson:
WIP. Surfacing this up as a meantime draft.

### DIFF
--- a/rfcs/2020-09-03-3226-api-topology.md
+++ b/rfcs/2020-09-03-3226-api-topology.md
@@ -1,0 +1,141 @@
+# RFC #3226 - 2020-09-03 - API Topology
+
+This RFC discusses configuration, implementation and potential schema for the API to co-exist with, and observe, existing Vector topology.
+
+## Scope
+
+* Implementation of the API server alongside existing topology.
+* Configuration in `vector.toml` to set-up / enable the API.
+* GraphQL types to observe existing topology (pending #3648)
+
+## Motivation
+
+I implemented the first draft of a GraphQL API server in #3514, and made changes to the toplogy mod to accommodate it. @lukesteensen [commented](https://github.com/timberio/vector/pull/3514#discussion_r479517596) with a preference to keep the API out of the existing topology mod.
+
+This raises several design questions:
+
+* Where should the API live?
+* If the API is outside of topology, what are our expectations with users making API configuration changes in `vector.toml` and handling SIGHUP/reloading? (The initial draft in #3514 )
+* Which parts of the API are configurable?
+* What are our expectations for enabling/disabling the config?
+* Should the API be treated as a sink - or exist as a layer atop topology?
+
+Additionally, there are questions about observing topology:
+
+* Should the API consume `internal_metrics` as a source - or subscribe directly to the controller, to override source configuration?
+* More generally - how will the API observe sources/sinks/transforms?
+
+This RFC demonstrates various possibilities intended to solicit debate. I'll note my preferred routes where applicable, but for the most part, will defer heavily to codeowners with more experience in these code paths.
+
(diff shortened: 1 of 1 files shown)


rfcs/2020-09-03-3226-api-topology.md
```diff
@@ -0,0 +1,141 @@
...
+* Configuration in `vector.toml` to set-up / enable the API.
+* GraphQL types to observe existing topology (pending #3648)
+
+## Motivation
+
+I implemented the first draft of a GraphQL API server in #3514, and made changes to the toplogy mod to accommodate it. @lukesteensen [commented](https://github.com/timberio/vector/pull/3514#discussion_r479517596) with a preference to keep the API out of the existing topology mod.
+
+This raises several design questions:
+
+* Where should the API live?
```
lukesteensen, line 17: I think we should start with it as a standalone, independent component. We can simply check the config for whether or not it's enabled and then spawn the server in `main`.

rfcs/2020-09-03-3226-api-topology.md
```diff
@@ -0,0 +1,141 @@
...
+* If the API is outside of topology, what are our expectations with users making API configuration changes in `vector.toml` and handling SIGHUP/reloading? (The initial draft in #3514 )
```
lukesteensen, line 18: I propose that we ignore config reloading. There are already config values we don't support changing with reloading and I don't think it's terribly important that users be able to toggle this dynamically.

This also feels like something that's low risk to add in the future. If we decide it's worthwhile to hook into reloading, we can do it later and it shouldn't be any more difficult than doing it now.

rfcs/2020-09-03-3226-api-topology.md
```diff
@@ -0,0 +1,141 @@
...
+* Which parts of the API are configurable?
+* What are our expectations for enabling/disabling the config?
+* Should the API be treated as a sink - or exist as a layer atop topology?
```
lukesteensen, line 21: For now, I think it exists outside of the topology but we can evolve this moving forward. I want to start as simple and isolated as possible.

rfcs/2020-09-03-3226-api-topology.md
```diff
@@ -0,0 +1,141 @@
...
+
+Additionally, there are questions about observing topology:
+
+* Should the API consume `internal_metrics` as a source - or subscribe directly to the controller, to override source configuration?
```
lukesteensen, line 25: I think we need to extract the part of `internal_metrics` that pulls metrics from the controller and converts them to our internal format. Then we can use that component both in the source and directly from the API, so the API is not coupled to the source but does use the same data types.

rfcs/2020-09-03-3226-api-topology.md
```diff
@@ -0,0 +1,141 @@
...
+* More generally - how will the API observe sources/sinks/transforms?
```
lukesteensen, line 26: The two sources will be the internal metrics themselves and probably a shared view of the current configuration. The configuration sharing will likely be as simple as a [tokio `watch` channel](https://docs.rs/tokio/0.2.22/tokio/sync/watch/index.html) that is written to during config reloading and watched by the API server.

We may eventually want to expose more direct information from specific components that do not fit a metrics data model (e.g. the status of open connections or files we're tailing), but that can wait until we've solidified the basics.

leebenson: 👍 Thanks, that helps clarify the approach.

leebenson:
This is no longer relevant-- work that would have been covered here has already been merged by #3708 and #3778, and we agreed in #4083 to get `vector top` over the line before any formal discussion into the resulting schema. Closing.

PR #3685, opened 2020-09-02
chore: migrate RouterSink to enum

### LINKED ISSUE #2934: Migrate RouterSink type to enum
Blocked on #2687 

See [Architecture RFC](https://github.com/timberio/vector/blob/master/rfcs/2020-06-18-2625-architecture-revisit.md). Start with single variant wrapping existing sink type.

fanatid:
Closes #2934.

I changed `Sink` from RFC to `VectorSink` to exclude conflicts with `futures::Sink`.
`OldAndBad` renamed to `Futures01Sink` for better understanding of what under variant (name here is not very important since this variant should be removed in the end, right?).

### DIFF
--- a/src/sinks/mod.rs
+++ b/src/sinks/mod.rs
@@ -1,6 +1,7 @@
 use crate::Event;
 use futures::{compat::Future01CompatExt, TryFutureExt};
 use snafu::Snafu;
+use std::fmt;
 
 pub mod streaming_sink;
 pub mod util;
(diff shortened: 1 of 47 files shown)


src/sinks/mod.rs
```diff
@@ -87,3 +88,29 @@ pub enum HealthcheckError {
...
+    pub async fn run<S>(self, input: S) -> Result<(), ()>
+    where
+        S: futures01::Stream<Item = Event, Error = ()>,
+    {
+        match self {
+            Self::Futures01Sink(sink) => input.forward(sink).compat().map_ok(|_| ()).await,
+        }
+    }
+
+    pub fn into_futures01sink(
```
lukesteensen, line 102: Is this method just for tests? I can see it being useful for migration, but we will probably not want to keep it around.

lukesteensen (approved):
This looks good! Not too exciting yet but getting there.

src/sinks/mod.rs
fanatid: Not only for tests, but also as bypass for `StreamSink`.
Actually, I can use `run` method in tests instead converting to `Sink`, this will be more correct way. Will change it right now.

bruceg (approved):
LGTM

src/sinks/mod.rs
fanatid: Updated. Now used for `StreamSink`(sematext, file, console) and in 1 test (tcp_stream_detects_disconnect in socket sink).

bruceg (approved):

ktff (approved):

lukesteensen (approved):

daddyzilla33:
Yes

PR #3686, opened 2020-09-02
chore(rfcs): Add RFC for first-class metric namespaces

jszwedko:
Ref: https://github.com/timberio/vector/issues/3684

### DIFF
--- a/rfcs/2020-09-02-3684-metric-namespaces.md
+++ b/rfcs/2020-09-02-3684-metric-namespaces.md
@@ -1,0 +1,177 @@
+# RFC 3684 - 2020-09-02 - First-class Metric Namespaces
+
+This RFC proposes making `namespace` a first-class field on the internal
+`Metric` type to allow it to be set in sources, manipulated in transforms, and
+used by sinks.
+
+## Scope
+
+This RFC will cover:
+
+- Separating the `namespace` of metrics into a separate field on `Metric`
+
+## Motivation
+
+As we add `metric` sources like
+[`apache_metrics`](https://github.com/timberio/vector/blob/master/rfcs/2020-08-21-3092-apache-metrics)
+and
+[`postgresql_metrics`](https://github.com/timberio/vector/blob/master/rfcs/2020-08-27-3603-postgres-metrics.md),
+that set their own namespaces (defaulting to `apache` and `postgresql`), it is
+becoming more clear that we may want to maintain the `namespace` separate from
+the metric name to allow for:
+
+- simple manipulation in transforms (e.g. as a filter)
+- to be sent as a separate field for sinks that require it (e.g.
+  `aws_cloudwatch_metrics`)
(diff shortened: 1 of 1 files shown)


jamtur01 (approved):
LGTM.

rfcs/2020-09-02-3684-metric-namespaces.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Outstanding Questions
+
+- Do we want to have the `prometheus` source parse the `namespace` out of
+  metrics it scrapes? The [naming
+  conventions](https://prometheus.io/docs/practices/naming/) suggest that all
+  metrics should start with one word describing the domain (or namespace)
+  followed by a `_` but there is requirement that prometheus endpoints satisfy
+  this. We could make it optional directive on the source to control parsing
+  metric namespaces.
```
lukesteensen, line 167: Would this affect backwards compatibility? With some transform configs I'm sure it would, but I wonder if it affects simpler pipelines.

Also probably worth noting that this would be a breaking change for at least transforms that operate on metric names.

lukesteensen (approved):
Seems reasonable to me!

bruceg:
Note that this also requires changes to the `text` encoding of metrics in the console sink.

rfcs/2020-09-02-3684-metric-namespaces.md
jszwedko: We could make this behavior opt-in for now to preserve compatibility with existing pipelines (and also to handle cases where the scraped endpoint isn't prefix a domain).

jszwedko: I'm inclined to leave it for `remap` and just provide an example of doing it for the `prometheus` source.

rfcs/2020-09-02-3684-metric-namespaces.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Internal Proposal
+
+Add `namespace` to
+[`Metric`](https://github.com/timberio/vector/blob/75844bc0f67d24ad1b54bfa130d074810ad2aa50/src/event/metric.rs#L10-L17):
+
+```rust
+pub struct Metric {
+  pub name: String,
+  pub namespace: Option<String>, // added
```
Hoverbear, line 48: It seems to me a fairly common use case will be wanting the namespace to be the same as the component name?

Hoverbear (approved):
I think this is a sensible approach.

rfcs/2020-09-02-3684-metric-namespaces.md
jszwedko: :+1: or at least related (e.g. `apache_metrics` -> `apache`). We've been defaulting them for non-generic (generic being ones like `prometheus`) sources.

PR #3692, opened 2020-09-03
chore(tests): Retry_until

leebenson:
Adds a `test_util::retry_until` helper, for retrying an async block until an `Ok(T)` is returned per `retry: Duration` - or until `until: Duration` has elapsed, whichever comes first. The latter panics.

Updates the Splunk integration test to use this approach vs. checking for openness of port 8089, which could occasionally result in a race condition where the port was open but the service wasn't fully initialised.

This could be used in other places where `wait_for_tcp` is currently employed, but for the sake of limiting scope, I've avoided touching other code paths.

### DIFF
--- a/src/sinks/splunk_hec.rs
+++ b/src/sinks/splunk_hec.rs
@@ -393,7 +393,7 @@ mod tests {
 #[cfg(feature = "splunk-integration-tests")]
 mod integration_tests {
     use super::*;
-    use crate::test_util::wait_for_tcp_duration;
+    use crate::test_util::retry_until;
     use crate::{
         assert_downcast_matches,
         config::{SinkConfig, SinkContext},
@@ -721,15 +721,17 @@ mod integration_tests {
             .build()
             .unwrap();
 
-        // Wait for port 8089 to be reachable before firing off request
-        wait_for_tcp_duration("127.0.0.1:8089".parse().unwrap(), Duration::from_secs(30)).await;
-
-        let res = client
-            .get("https://localhost:8089/services/data/inputs/http?output_mode=json")
-            .basic_auth(USERNAME, Some(PASSWORD))
-            .send()
-            .await
-            .unwrap();
+        let res = retry_until(
+            || {
+                client
+                    .get("https://localhost:8089/services/data/inputs/http?output_mode=json")
+                    .basic_auth(USERNAME, Some(PASSWORD))
+                    .send()
+            },
+            Duration::from_millis(500),
+            Duration::from_secs(30),
+        )
+        .await;
 
         let json: JsonValue = res.json().await.unwrap();
         let entries = json["entry"].as_array().unwrap().clone();
--- a/src/test_util/mod.rs
+++ b/src/test_util/mod.rs
@@ -372,6 +372,7 @@ mod tests {
         time::Duration,
     };
 
+    #[allow(clippy::never_loop)]
     async fn retry_until_helper(count: Arc<RwLock<i32>>) -> Result<(), ()> {
         while *count.read().unwrap() < 3 {
             let mut c = count.write().unwrap();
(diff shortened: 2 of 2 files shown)


src/test_util/mod.rs
```diff
@@ -348,6 +348,64 @@ where
...
+
+    #[allow(clippy::never_loop)]
+    async fn retry_until_helper(count: Arc<RwLock<i32>>) -> Result<(), ()> {
+        while *count.read().unwrap() < 3 {
+            let mut c = count.write().unwrap();
+            *c += 1;
+            return Err(());
+        }
+        Ok(())
+    }
```
fanatid, line 383: Why while-loop used here? :thinking:

src/test_util/mod.rs
```diff
@@ -348,6 +348,64 @@ where
...
+{
+    let started = Instant::now();
+    while started.elapsed() < until {
+        match f().await {
+            Ok(res) => return res,
+            Err(_) => tokio::time::delay_for(retry).await,
+        }
+    }
+    panic!("Timeout")
+}
```
fanatid, line 365: Would it possible to replace `wait_for_duration` by `retry_until`?

src/test_util/mod.rs
leebenson: Thanks, this is result of some copying and pasting from an earlier inline version that I didn't replace. It's redundant. Will remove.

src/test_util/mod.rs
leebenson: It should be possible to replace most instances of `wait_for_tcp` and `wait_for_tcp_duration`, but I didn't want to make too many changes yet. Can follow-up in another PR if this one is approved.

fanatid: Sounds good :+1:

fanatid (approved):

src/sinks/splunk_hec.rs
```diff
@@ -721,15 +721,17 @@ mod integration_tests {
...
-            .basic_auth(USERNAME, Some(PASSWORD))
-            .send()
-            .await
-            .unwrap();
+        let res = retry_until(
+            || {
+                client
+                    .get("https://localhost:8089/services/data/inputs/http?output_mode=json")
+                    .basic_auth(USERNAME, Some(PASSWORD))
+                    .send()
```
MOZGIII, line 729: Does this `.send` call fail if the remote returns status code that's `>= 400`? We might want to add an accompanying wrapper to treat those as errors for the purposes of this functionality to reduce the possibility of race even further - as there are some services that start listening on the socket and responding to the HTTP requests before properly initializing the internal routes - and thus responding with 404 in a first few moments before after boot.

leebenson: According to [`send` docs](https://docs.rs/reqwest/0.9.21/reqwest/struct.RequestBuilder.html#method.send):

> This method fails if there was an error while sending request, redirect loop was detected or redirect limit was exhausted.

There's [.error_for_status](https://docs.rs/reqwest/0.10.8/src/reqwest/async_impl/response.rs.html#337-343), which turns a 400+ response into an error.

MOZGIII: Great, I'm not familiar with `reqwest`, but it should've guessed they have a built-in helper for that :smile: 
Shall we use in this case?

lukesteensen (approved):

jamtur01 (approved):
LGTM.

src/sinks/splunk_hec.rs
leebenson: I'll hold off for now, to scope this PR to just the retry logic, but I think it's worth doing a pass for HTTP tests in general, to ensure we're handling `404`s, etc, as you mentioned.

Issue #3702, opened 2020-09-03
Refactor sources that scrape an endpoint on an interval


jszwedko:
With the upcoming introduction of the `apache_metrics` and `nginix_metrics` sources, we'll have three sources that operate in a very similar manner. Namely:

* Scrape an HTTP endpoint on an interval
* Parse the results to generate metrics

As such, I think it'd be useful to refactor them to share internal components similar to how the HTTP-client-based sinks (`humio`, `splunk`, `http`) do. This should make it easier to add more of these types of sources in the future.

bruceg:
This could also be useful for the host metrics source, except that it won't be scraping a remote endpoint but rather local sources. The process will be much the same otherwise.

PR #3704, opened 2020-09-03
feat(apache_metrics source): Initial implementation

### LINKED ISSUE #3092: New `apache_metrics` source

jszwedko:
**EDIT** This is ready for review now. See comments below.

Closes #3092 

I am just pushing this up for any preliminary feedback before I get back to it on Tuesday. I'm definitely still on the upclimb of the Rust learning curve so feedback on Rust idioms and patterns especially welcome.

This is roughly based on the implementation of the `prometheus` source which operates in a similar manner: scraping an HTTP endpoint and parsing the payload to generate metrics.

I left some notes on where I think the architecture could be improved, below, but I think it'd be best to hold off until https://github.com/timberio/vector/issues/3702 to do a larger refactor of how this and the `prometheus` source work.

The namespace handling will likely change as a part of #3686 

TODOs:

- [x] Tests in `apache_metrics/mod.rs`
- [x] Testing parsing failures
- [x] Applying `host` and `endpoint` tags
- [x] Handling request errors and publishing `apache_up=0`
- [x] Documentation


src/sources/apache_metrics/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    urls: Vec<http::Uri>,
+    interval: u64,
+    namespace: String,
+    shutdown: ShutdownSignal,
+    out: Pipeline,
+) -> super::Source {
+    let out = out
+        .sink_map_err(|e| error!("error sending metric: {:?}", e))
+        .sink_compat();
+    let task = tokio::time::interval(Duration::from_secs(interval))
```
jszwedko, line 86: I'd really like to break up the below into separate functions but figured I'd wait for #3702 . This largely matches the `prometheus` source in its current form.

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,852 @@
...
+/// - `payload` - the mod_status output
+/// - `namespace` - the namespace to put the generated metrics in
+/// - `now` - the time the payload was fetched
+/// - `tags` - any base tags to apply to the metrics
+pub fn parse(
+    payload: &str,
+    namespace: &str,
+    now: DateTime<Utc>,
+    tags: Option<&BTreeMap<String, String>>,
+) -> (Vec<Metric>, Vec<ParseError>) {
```
jszwedko, line 24: I'm curious to get feedback or thoughts on how to model this better where the parse function wants to return metrics _and_ errors rather than simply one or the other. This was a design choice I made to still publish metrics that we could parse even if there was an error with one of them.

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,852 @@
...
+    }
+}
+
+impl Error for ParseError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        Some(self.err.as_ref())
+    }
+}
+
+enum LineResult {
```
jszwedko, line 78: Is this sort of `enum` idiomatic?

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,852 @@
...
+                    kind: MetricKind::Absolute,
+                    value: MetricValue::Gauge { value: 74.0 },
+                },
+            ]
+        );
+        assert_eq!(errors.len(), 0);
+    }
+
+    #[test]
+    fn test_extended() {
```
jszwedko, line 600: I'll update with code comments, but the difference between this and the above test is that the output is different depending on whether `ExtendedStatus` is enabled or not.

https://httpd.apache.org/docs/2.4/mod/core.html#extendedstatus

src/sources/apache_metrics/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+        .take_until(shutdown.compat())
+        .map(move |_| stream::iter(urls.clone())) // TODO remove clone?
```
jszwedko, line 89: I'd like to get rid of this clone, but was unable to figure out how without annoying the borrow checker.

src/sources/apache_metrics/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+                .expect("error creating request");
+
+            let mut tags: BTreeMap<String, String> = BTreeMap::new();
+            tags.insert("endpoint".into(), url.to_string());
+            if let Some(host) = url.host() {
+                tags.insert("host".into(), host.into());
+            }
+
+            let start = Instant::now();
+            let namespace = namespace.clone(); // TODO remove?
```
jszwedko, line 106: Same here.

src/sources/apache_metrics/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+        .flatten()
+        .forward(out)
+        .inspect(|_| info!("finished sending"));
+
+    Box::new(task.boxed().compat())
+}
+
+#[cfg(feature = "sources-apache_metrics")]
+#[cfg(test)]
+mod test {
```
jszwedko, line 197: These tests just test high-level behavior; there are separate tests in `apache_metrics/parser.rs` that test metric generation behavior.

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,837 @@
...
+            let value = parts.next().map(|s| s.trim());
+            match (key, value) {
+                (Some(k), Some(v)) => Some((k, v)),
+                _ => None,
+            }
+        })
+        .collect::<Vec<_>>();
+
+    // mod_status has BusyWorkers/IdleWorkers repeated
+    // TODO better way to do this without .collect()ing? Do we care?
```
jszwedko, line 40: Thoughts here? I did this to dedupe the lines which required sorting.

jszwedko (commented):
I think this is ready for review now!

Again, thoughts on patterns and idioms especially welcome. I tried to call out a couple of places where I had a question. I'll drop the `TODOs` from the code before merging this after I see people's thoughts on them.

jamtur01 (commented):
Core stuff looks good to me. I am skeptical I can add value reviewing any of the actual code. :)

src/sources/apache_metrics/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+                .flatten()
+        })
+        .flatten()
+        .forward(out)
+        .inspect(|_| info!("finished sending"));
+
+    Box::new(task.boxed().compat())
+}
+
+#[cfg(feature = "sources-apache_metrics")]
```
JeanMertz, line 195: I don't believe this is needed, since the whole module is behind the same feature flag.

.github/semantic.yml
```diff
@@ -102,6 +102,7 @@ scopes:
   - sematext provider
 
   # sources
+  - apache_metrics source
```
JeanMertz, line 105: Was this auto-generated from `.github/semantic.yml.erb`? (and if so, how did you do that 👀).

I need to add one for the `remap` transform as well, but I haven't looked into where the actual `.metadata` data comes from when it generates this file.

src/sources/apache_metrics/mod.rs
```diff
@@ -0,0 +1,394 @@
+use crate::{
```
JeanMertz, line 1: nit: since this module consists of a single file, you can put it in `src/sources/apache_metrics.rs`.

Basically, these two are exactly the same:

* `src/sources/apache_metrics.rs`
* `src/sources/apache_metrics/mod.rs`

The advantage of the second one is if you have multiple modules that you want to namespace under the `apache_metrics` namespace (and potentially re-export the most common public interface at that level).

JeanMertz: Whoops, never mind, there's actually a `parser.rs` file that was collapsed in the GH UI, so I missed it.

All good 😄

src/sources/apache_metrics/parser.rs
JeanMertz: I wonder if an iterator-based approach would work well?

The iterator would return `Option<Result<Metric, ParseError>>`. The caller can continue iterating, even if an error is returned, until no more metrics are available.

You can't have it return a metric _and_ error simultaneously, but I don't believe that's needed?

Regardless, what you have here works nicely, so I wouldn't worry too much about changing this.

src/sources/apache_metrics/parser.rs
JeanMertz: I'm not quite sure what the purpose of this de-duplication is.

Is it our job to filter out those duplicate status events, if it's what Apache is reporting? What happens if I call this function twice, it would filter out duplicates for each individual call, but I would still end up with two similar events since it doesn't filter duplicates across calls, right?

src/sources/apache_metrics/parser.rs
JeanMertz: This seems similar to `Option<Result<Vec<Metric>, ParseError>>`?

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,837 @@
...
+
+fn line_to_metrics(
+    key: &str,
+    value: &str,
+    namespace: &str,
+    now: DateTime<Utc>,
+    tags: Option<&BTreeMap<String, String>>,
+) -> LineResult {
+    match key {
+        "ServerUptimeSeconds" => match value.parse::<f64>() {
```
JeanMertz, line 92: Perhaps we could move these keys into a (documented) enum?

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,837 @@
...
+                tags: tags.map(|tags| tags.clone()),
+                kind: MetricKind::Absolute,
+                value: MetricValue::Counter { value }, // TODO verify unit
+            }]),
+            Err(err) => LineResult::Error(ParseError {
+                key: key.to_string(),
+                err: err.into(),
+            }),
+        },
+        "CPUUser" => match value.parse::<f64>() {
```
JeanMertz, line 146: I see a reasonable amount of duplication in the next couple of match arms.

Perhaps use a single arm, and then split up where needed? e.g.

```rust
"CPUUser " | "CPUSystem" | /* ... */ => {
	// ...

	match { /* ... */ }
}
```

Can't say for certain how much clean-up that would provide, but might be worth a try.

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,837 @@
...
+                ('K', "keepalive"),
+                ('D', "dnslookup"),
+                ('C', "closing"),
+                ('L', "logging"),
+                ('G', "finishing"),
+                ('I', "idle_cleanup"),
+                ('.', "open"),
+            ]
+            .into_iter()
+            .collect();
```
JeanMertz, line 363: Seems like we should be able to move this into a constant to avoid the allocations on each cycle?

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,837 @@
...
+
+            LineResult::Metrics(
+                scoreboard
+                    .iter()
+                    .map(|(c, name)| to_metric(name, scores.get(c).unwrap_or(&0u32)))
```
JeanMertz, line 368: `unwrap_or_default()` should work here as well I believe.

JeanMertz: Although that would probably require something similar to `(*scores.get(c)).unwrap_or_default()` to dereference the `u32`.

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,837 @@
...
+                    .collect::<Vec<_>>(),
+            )
+        }
+        _ => LineResult::None,
+    }
+}
+
+fn encode_namespace(namespace: &str, name: &str) -> String {
+    if !namespace.is_empty() {
```
JeanMertz, line 377: Maybe swap the cases, to avoid the double negative?

JeanMertz (approved):
Sweet! Left a couple of comments, but didn't dive too deep into the Apache metrics logic itself. I can circle back later, but let's wait if someone with more knowledge comes along to review 😄

src/sources/apache_metrics/parser.rs
jszwedko: I'll try to improve the comment there, but basically there is a bug in mod_status where, in the text output, `IdleWorkers` and `BusyWorkers` appear twice. If I don't dedupe we end up publishing duplicate metrics, which is probably ok, but seems a little wonky.

Apache bug: https://bz.apache.org/bugzilla/show_bug.cgi?id=63300

src/sources/apache_metrics/parser.rs
jszwedko: :thinking: yeah it is basically that. I'll try swapping to remove the type.

.github/semantic.yml
jszwedko: I actually missed that this was a generated file. I'm actually not sure how to regenerate it. I assume @binarylogic knows? Does this repo need a `generate.rb` like `vector-website` has?

src/sources/apache_metrics/parser.rs
jszwedko: Hmm, yeah, I think I like that better. I'll give it a shot.

src/sources/apache_metrics/parser.rs
jszwedko: Agree :+1:

src/sources/apache_metrics/parser.rs
jszwedko: It is a bit duplicative, but I actually like the simplicity of just mapping each string rather than parsing the key names. If you feel strongly though, I'll give it a shot.

src/sources/apache_metrics/parser.rs
jszwedko: Agreed :+1:

src/sources/apache_metrics/parser.rs
jszwedko: :+1: I ended up with `scores.get(c).copied().unwrap_or_default()`.

jszwedko:
@JeanMertz thanks for the thorough review and recommendations! I think I implemented them all in https://github.com/timberio/vector/pull/3704/commits/5e06a5273c512210b88d8ef1910949e28fa6a91d if you want to take a look. I think it is cleaner now.

jszwedko:
**Note** I'll still drop the TODOs I had in the code before merging this; I just put them there to call out questions I had.

src/sources/apache_metrics/parser.rs
JeanMertz: Ah, right 👍

I guess one way to avoid having to collect is to keep track of the field through a boolean, and return `None` in the `filter_map` above as soon as we've flipped the boolean to `true`?

jszwedko: :+1: This made me realize I could also just use a `HashMap` which would take care of making it unique: 8e3d7539 . I'm curious what you think of that.

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,919 @@
...
+        },
+    }
+}
+
+fn encode_namespace(namespace: &str, name: &str) -> String {
+    if namespace.is_empty() {
+        name.to_string()
+    } else {
+        format!("{}_{}", namespace, name)
+    }
```
bruceg, line 436:
```suggestion
    match namespace {
        "" => name.to_string(),
        _ => format!("{}_{}", namespace, name),
    }
```

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,919 @@
...
+            };
+            acc
+        })
+        .into_iter()
+}
+
+#[derive(Debug)]
+pub struct ParseError {
+    key: String,
+    err: Box<dyn Error>,
```
bruceg, line 215: Can this have a concrete type, or are there multiple types of error handled here? If multiple types, did you consider an `enum ParseError`?

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,919 @@
...
+    }
+}
+
+fn line_to_metrics(
+    key: &str,
+    value: &str,
+    namespace: &str,
+    now: DateTime<Utc>,
+    tags: Option<&BTreeMap<String, String>>,
+) -> Option<Result<Vec<Metric>, ParseError>> {
```
bruceg, line 236: Could this return an iterator in the place of the `Vec`? Everything but the `Scoreboard` has a single value and could use `std::iter::once`.

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,919 @@
...
+    StatusFieldStatistic::from_key_value(key, value).map(|result| {
+        result.map(|statistic| match statistic {
+            StatusFieldStatistic::ServerUptimeSeconds(value) => vec![Metric {
+                name: encode_namespace(namespace, "uptime_seconds_total"),
+                timestamp: Some(now),
+                tags: tags.map(|tags| tags.clone()),
+                kind: MetricKind::Absolute,
+                value: MetricValue::Counter { value },
+            }],
```
bruceg, line 245: This looks like there would be some opportunity for refactoring the duplication, but I see there is a few different things going on with `tags` and `value`.

bruceg (commented):
Looks pretty good from an idiomatic Rust point of view. There's definite room for refactoring, but I understand you want to hold off on some parts of that.

src/sources/apache_metrics/parser.rs
jszwedko: I believe there are at least two types of parse errors, one for int and one for float. I think I did, at one point, have it as an enum, but it didn't seem worth it since all we were doing was just propagating it up without handling.

I guess, generally, do you feel it's better to be specific with the error types even if the caller doesn't care?

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,919 @@
...
+    ]
+    .into_iter()
+    .collect();
+}
+
+/// enum of mod_status fields we care about
+enum StatusFieldStatistic<'a> {
+    ServerUptimeSeconds(f64),
+    TotalAccesses(f64),
+    TotalKBytes(f64),
```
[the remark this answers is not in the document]
jszwedko: I thought about that, but figured since they are just passed through directly to the `Metric` type which wants a `f64`s anyway, that converting back and forth didn't really give as much. However, I can see the argument for modeling the data as it is. I'll make that change.

src/sources/apache_metrics/parser.rs
bruceg: Generally, I think it's better to be specific with _all_ types, even if the caller doesn't care, as it avoids invoking more "mechanism" at run time (allocations, dynamic dispatch, etc). However, it doesn't matter as much when it isn't the hot path and, by definition, exceptional errors are not the hot path. In the case of errors, though, adding a layer of wrapping adds context as to why the error happened. So it's partly a judgment call.

src/sources/apache_metrics/parser.rs
```diff
@@ -0,0 +1,919 @@
...
+        value: &'a str,
+    ) -> Option<Result<StatusFieldStatistic<'a>, ParseError>> {
+        match key {
+            "ServerUptimeSeconds" => match value.parse::<f64>() {
+                Ok(value) => Some(Ok(StatusFieldStatistic::ServerUptimeSeconds(value))),
+                Err(err) => Some(Err(ParseError {
+                    key: key.to_string(),
+                    err: err.into(),
+                })),
+            },
```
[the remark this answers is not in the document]
jszwedko: This feels cleaner, thanks for the suggestion!

The compiler did walk me into quite the type signature:

```rust
fn parse_value<T: std::str::FromStr>(key: &str, value: &str) -> Result<T, ParseError>                     
where                                                                                                     
    <T as std::str::FromStr>::Err: std::error::Error,                                                     
    <T as std::str::FromStr>::Err: 'static,                                                               
{                                                                                                         
    value.parse::<T>().map_err(|err| ParseError {                                                         
        key: key.to_string(),                                                                             
        err: err.into(),                                                                                  
    })                                                                                                    
}  
```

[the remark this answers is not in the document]
bruceg: Even though the compiler recommends it, I think you can simplify the `where` clause since you already have `T: FromStr`:
```rust
fn parse_value<T>(key: &str, value: &str) -> Result<T, ParseError>
where
    T: FromStr,
    T::Err: std::error::Error + 'static
```

jszwedko:
@bruceg thanks for all of the thorough feedback! It was helpful. I think I implemented all of the suggestions aside from the general refactoring one. I tried to keep the commits atomic so you could just review them in order from the time of your review.

For the general refactoring comment (https://github.com/timberio/vector/pull/3704#discussion_r488839729), I definitely agree that it could use some love. I still feel like it might be better to wait for #3702 to spend a lot of time on it, but, if you feel strongly about refactoring some sooner, would you be up for pairing on it? I think that'd help me work through it a bit more quickly. Maybe we could sketch out what #3702 might look like too.

src/sources/apache_metrics/parser.rs
jszwedko: :+1: that makes sense. It seems more flexible too in that if we add another caller that would care, the specificity would already be there. Thanks for the explanation.

src/sources/apache_metrics/parser.rs
```diff
@@ -424,6 +376,17 @@ fn line_to_metrics(
     })
 }
 
+fn parse_numeric_value<T: std::str::FromStr>(key: &str, value: &str) -> Result<T, ParseError>
+where
+    <T as std::str::FromStr>::Err: std::error::Error,
+    <T as std::str::FromStr>::Err: 'static,
```
bruceg: These can combine, and I'm not sure the `as FromStr` is necessary:
```suggestion
    T::Err: std::error::Error + 'static,
```

bruceg (approved):
I'm kinda sorry I suggested looking into this path. This turned a relatively simple error struct into something pretty verbose, all to avoid a boxed trait. I had forgotten you would be working with both `ParseFloatError` and `ParseIntError`.

jszwedko:
> I'm kinda sorry I suggested looking into this path. This turned a relatively simple error struct into something pretty verbose, all to avoid a boxed trait. I had forgotten you would be working with both ParseFloatError and ParseIntError.

It's definitely more verbose, but I do like that there is a tighter contract on the error types. It feels like this would be a common pattern for types that can return multiple types of errors?

Thanks again for the reviews!

bruceg:
> It feels like this would be a common pattern for types that can return multiple types of errors?

Definitely, and that is what our use of the snafu crate is designed to help with. It might (have) helped with this, though your usage here is a little different than other places.

jszwedko:
Opened https://github.com/timberio/vector/issues/4003 for the flakey mac failure.

The shutdown linux failure (`remap_function_now`) doesn't seem to be flakey, but is on master already. I pinged Jean about it since he was working in that area.

jamtur01:
Boo ya! 🥳 🎆 🍾 

Awesome stuff.

src/sources/apache_metrics/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+        .map(move |url| {
+            let https = HttpsConnector::new().expect("TLS initialization failed");
+            let client = Client::builder().build(https);
+
+            let request = Request::get(&url)
+                .body(Body::empty())
+                .expect("error creating request");
+
+            let mut tags: BTreeMap<String, String> = BTreeMap::new();
+            tags.insert("endpoint".into(), url.to_string());
```
fanatid, line 106: From [rfc/2020-08-21-3092-apache-metrics-source.md](https://github.com/timberio/vector/blob/3d11ab33f6a3ae3de6e2b1a73f97ca23879720a8/rfcs/2020-08-21-3092-apache-metrics-source.md):
> `endpoint` the full endpoint (sans any basic auth credentials)

Should we remove credentials here?

jszwedko: Yes we should :smile: Thanks for catching that!

jszwedko: I started to fix this, but will have to pick it back up later. It wasn't as straightforward as I thought it might be.

jszwedko: https://github.com/timberio/vector/pull/4410

PR #3708, opened 2020-09-04
feat(observability): API GraphQL, round 2

### LINKED ISSUE #1073: Add HTTP ready check

### LINKED ISSUE #3225: Real-time communication R&D and POC

leebenson:
Picks up from #3514. 

Closes #1073. Closes #3225.

Much of the readme there is relevant, with a few notable changes per feedback:

- No longer touches the `topology` mod.
- The API server is initialised in main.
- It doesn't reload on config changes.
- Simplified API init via a single `make_server` func.
- Removed the `balanced_or_tree!` macro, pending benchmarks.

As before, to enable, add this to `vector.toml`:

```toml
[api]
  enabled = true # default: false
  playground = true # optional. Default: true
```

As before, use http://localhost:8686/playground to browse GraphQL schema and test queries/subscriptions.

### Example

**vector.toml**

```toml
[api]
  enabled = true

[sources.generator]
  type = "generator" # required
  lines = ["Random line", "And another"] # required
  batch_interval = 0.05 # optional, no default

[sinks.my_sink_id]
  # General
  type = "console" # required
  inputs = ["generator"] # required
  target = "stdout" # optional, default

  # Encoding
```

**And then in http://localhost:8686/playground, run:**

```gql
subscription Metrics {
  metrics(interval: 1000) {
    timestamp
    ... on EventsProcessed {
      eventsProcessed
    }
  }
}
```

### TODO

- [x] Give the code another pass
- [x] Check all GH actions pass (lint + clippy might not yet)
- [x] Add some ~documentation~ commentary (docs to follow when this is public?)


src/internal_events/generator.rs
```diff
@@ -0,0 +1,18 @@
+use super::InternalEvent;
```
binarylogic, line 1: Should we split the generator changes out?

leebenson: Can do, although since this PR currently only observes the `events_processed` types and the least opinionated way to produce them is via the generator source, I think folks might benefit from this being included to execute the test subscription noted in the PR.

Either way is cool by me though.

jamtur01: I vote leave it in.

src/api/schema/metrics.rs
```diff
@@ -0,0 +1,155 @@
...
+
+#[derive(Default)]
+pub struct MetricsSubscription;
+
+#[Subscription]
+impl MetricsSubscription {
+    /// Metrics for how long the Vector instance has been running
+    async fn uptime_metrics(
+        &self,
+        #[arg(default = 1000, validator(IntRange(min = "100", max = "60_000")))] interval: i32,
```
leebenson, line 96: Looks like the `default` arg was fixed, and this now gets encoded as `Int` (optional) vs `Int!` (required)

tests/api.rs
```diff
@@ -0,0 +1,83 @@
...
+    use vector::test_util::{next_addr, retry_until};
+
+    // provides a config that enables the API server, assigned to a random port. Implicitly
+    // tests that the config shape matches expectations
+    fn api_enabled_config() -> Config {
+        let mut config = Config::builder();
+        config.add_source("in1", source().1);
+        config.add_sink("out1", &["in1"], sink(10).1);
+        config.api.enabled = true;
+        config.api.bind = Some(next_addr());
```
leebenson, line 20: `next_addr()` assumes port 1234 is available. Ran into a situation where I had a running instance of [Parcel](https://parceljs.org/), which uses this same port by default. Although this is probably uncommon, it led to some head stratching.

I think it's worth switching this to a real, free port -- will tackle in a separate PR.

tests/api.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        config.build().unwrap()
+    }
+
+    // spawns the API server on the provided port; optionally enables playground
+    fn start_api_server(addr: SocketAddr, playground: bool) {
+        tokio::spawn(async move {
+            let _ = api::make_server(addr, playground).await;
```
leebenson, line 28: This will be spawned independently for each test. Could drop this at the end of each test but not sure it's worth explicitly handling, since these tests run for all of 300ms.

leebenson (commented):
This should be ready for review.

### Subscriptions

I've broken out subscription queries, to demonstrate different (basic) use-cases, i.e:

**Uptime metrics:**

```gql
subscription {
  uptimeMetrics {
    timestamp
    seconds
  }
}
```

**Events processed:**

```gql
subscription {
  eventsProcessedMetrics {
    timestamp
    eventsProcessed
  }
}
```

**Bytes processed:**

```gql
subscription {
  bytesProcessedMetrics {
    timestamp
    bytesProcessed
  }
}
```

**All metrics (just the above 3 for now-- will build on this):**

```gql
subscription {
  metrics {
    __typename
    timestamp
    ... on EventsProcessed {
      eventsProcessed
    }
    ... on Uptime {
      seconds
    }
    ... on BytesProcessed {
      bytesProcessed
    }
  }
}
```

I'll add to these in follow-up PRs.

### Docs

@Hoverbear, you mentioned in #3514 the requirement for better docs. I wasn't sure whether you meant inline comments, or actual public-facing docs.

If it's the latter, my expectation is that this would remain an unadvertised feature for now, until the schema has settled and some useful functionality is exposed. @binarylogic / @jamtur01 may have more specific thoughts there.

If you mean general commentary or other requirements, please advise and I'll adjust as needed.

### Tests

Currently, API tests assert the following are reachable:

- `/health`
- `/playground` - enabled/disabled

In a follow-up PR, I'll add the Rust GraphQL client I've been playing with. It requires an explicit GraphQL schema file, which can be generated from the runtime async-graphql schema.

I think this would be useful as a separate bin util so that it can be asserted via CI that it matches, and have tests run against it.

Hoverbear:
Noting binary size differences (release, unstripped):

```
# main branch
   Compiling rdkafka v0.24.0
    Finished release [optimized] target(s) in 3m 22s
ana@autonoma:~/git/timberio/vector$ ls -lah target/release/vector
-rwxrwxr-x 2 ana ana 96M Sep  9 09:34 target/release/vector
```

```
# this branch
ana@autonoma:~/git/timberio/vector$ ls -lah target/release/vector
-rwxrwxr-x 2 ana ana 101M Sep  9 09:37 target/release/vector
```

src/api/handler.rs
```diff
@@ -0,0 +1,8 @@
+use serde_json::json;
+use warp::reply::json;
+use warp::{Rejection, Reply};
+
+// health handler, responds with { ok: true }
```
Hoverbear, line 5: Doc comments are `///` :)

leebenson:
> Noting binary size differences (release, unstripped):

Is this something we need to be concerned with? 96mb already strikes me as pretty hefty; a 5% increase for the API interface seems reasonable at first glance.

Are there any plans to provide a 'customise your download' page, with features that can be turned on/off? This is all behind an `api` flag so would be a good candidate for shaving where a user doesn't want/need it.

src/api/handler.rs
leebenson: I didn't intend for this to be a doc comment, but since it's public, it probably should be.

leebenson: actually scratch that, it's only public to the `api` own mod. Should we be doc-commenting all functions, public and otherwise?

tests/api.rs
Hoverbear: You can have the OS assign you a free port by asking for port `0`.

Hoverbear:
@leebenson It's something our marketing is concerned with:

image


See the bottom. :)

It's not a blocker or an issue! I just wanted to flag it.

tests/api.rs
leebenson: That's actually what portpicker does (after attempting a random port first). It momentarily spawns a TCP listener, copies the port, and then drops it.

src/config/builder.rs
```diff
@@ -76,6 +81,10 @@ impl ConfigBuilder {
     pub fn append(&mut self, with: Self) -> Result<(), Vec<String>> {
         let mut errors = Vec::new();
 
+        // API
```
Hoverbear, line 84: I don't think this comment is necessary. :)

leebenson:
@Hoverbear, what do you think about the idea of customising binary downloads via a feature selection page, a la [Caddy](https://caddyserver.com/download)?

Is that a feasible strategy? Or would the number of possible configurations and the compilation time waiting for uncached builds make this impractical?

On the surface, having a friendly one-pager feels less of an ask to users who may not want to go down the DIY compile route.

Interested to hear your thoughts.

src/main.rs
```diff
@@ -127,6 +129,23 @@ fn main() {
...
 
+        // API
+        #[cfg(feature = "api")]
+        if config.api.enabled {
+            let opt = config.api; // copy to move to spawn
+            tokio::spawn(async move {
+                let addr = opt.bind.unwrap();
+                let playground = &*format!("http://{}:{}/playground", addr.ip(), addr.port());
+
+                info!(
```
Hoverbear, line 140: Below, on line 155, we `emit!(VectorStarted)`, maybe here would be a good place to mimic this?

tests/api.rs
Hoverbear: Cool!

tests/api.rs
Hoverbear: Our entire test suite can take some time though, and having active API servers could definitely impact testing. We run our tests on fairly slow CI boxes, so it's probably safer for reliability reasons to properly reap this.

Hoverbear (commented):
This looks great @leebenson ! wow! :)

I think if we can make sure we're not leaving API servers running during testing, this should be good!

src/api/handler.rs
Hoverbear: You're right, I'm sorry. :) I was being overly pedantic.

Hoverbear:
@leebenson I'd love to do that, it's kind of a pipe dream of mine, I think we need to make more progress against our cross compiling and releases first.

We'd probably want to have only a few "options" but we could definitely do like vim does and have like a `minimal` and `full` build or something without toooo much pain.

tests/api.rs
leebenson: That's a good point. I'll make these tear-downable 👍

src/main.rs
leebenson: Added an `ApiStarted` event in 80bd6615e0290a75855a22bae2e1d404f0e2f8eb

tests/api.rs
leebenson: Updated the API implementation to allow for `server.stop()`, which shuts down the API gracefully. Updated tests to use it in 80bd6615e0290a75855a22bae2e1d404f0e2f8eb

jamtur01 (approved):
LGTM

tests/api.rs
leebenson: Argh, no I didn't. It's actually in https://github.com/timberio/vector/pull/3778/commits/9e1a79b40ec4214292c8e9eaea86620f9f495893 -- the GraphQL Client branch. I'll bring those changes over.

leebenson:
@Hoverbear - I implemented your feedback (thanks!). Would you mind giving this a final pass to approve?

@JeanMertz, @bruceg, @lukesteensen -- you're tagged as codeowners. I've tried to keep the scope of this PR relatively small to iterate on the API schema in stages. #3778 has the GraphQL Client counterpart. I'd appreciate if you could look this over so I can get it in master, and I can then move away from GraphQL as a lib and more into designing the schema.

Thanks!

src/api/handler.rs
```diff
@@ -0,0 +1,8 @@
+use serde_json::json;
+use warp::reply::json;
+use warp::{Rejection, Reply};
+
+// Health handler, responds with { ok: true }
```
Hoverbear, line 5: `///` would let this show up in our IDE hints. :)

Hoverbear (approved):

src/api/handler.rs
leebenson: Thanks, that's a good point. I've generally been using comments for private functions, and doc comments for public. I don't know if that's idiomatic. But if there's tooling benefit to doc comment everywhere, might it make sense to adopt that for all funcs/methods?

bruceg (approved):
I didn't do a deep dive into this code, but what I saw looked fine.

leebenson:
The two failing tests are noted in #4003 (which I'll try to fix tomorrow) and #4004 (fixed in master), so this should be good to merge.

PR #3743, opened 2020-09-07
chore: remove compat from sinks healthchecks

fanatid:
Each sink already uses new futures but adds an extra compatibility layer for nothing. This PR removes compat layer from sinks healthchecks.


lukesteensen (approved):
Nice!

PR #3765, opened 2020-09-08
fix(ci): add deny.toml

fanatid:
CI fix for https://github.com/timberio/vector/issues/3764


juchiast (commented):
net2, spin and failure are warnings so they are not strictly needed.

PR #3772, opened 2020-09-08
chore(tests): test_util::next_addr -- pick available port

leebenson:
Replaces the static port 1234 used in `test_util::next_addr()` with [portpicker](https://github.com/Dentosal/portpicker-rs), to return an available TCP/UDP port.

I ran into a situation where I actually had a running service on 1234, and asserts of the response body were failing. This prevents that (probably quite uncommon, but hey-- it happened) scenario.

The [source](https://github.com/Dentosal/portpicker-rs/blob/master/src/lib.rs) looks fairly benign. A random port between 15000 - 25000 is tried (up to 10x) before deferring to the OS to pick a port by binding to localhost:0 and the attempting to listen on that port (again, up to 10x.) If the total 20 iterations fail, nothing is returned and the tests fail.

Should be a drop-in replacement, unless we're relying on hardcoded ports (which I'll scan for in tests.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3120,6 +3120,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "portpicker"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5b497d05c16fe00939445c00a4fe2fa4f3d3dfc9c0401a3ab5c577afda2debb9"
+dependencies = [
+ "rand 0.6.5",
+]
+
+[[package]]
 name = "ppv-lite86"
 version = "0.2.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


fanatid (approved):

lukesteensen (approved):

Hoverbear:
@leebenson This PR broke master. Please don't merge PRs with broken CI jobs unless you are confident they are a flaky test and you've tracked it!

jszwedko:
I think in this case it was a silent merge conflict as the CI doesn't indicate this failure. I'm assuming that it was based off an older master commit.

We could try to integrate some sort of "rebase + run tests + merge" flow for PRs via bot, but with CI as flakey as it is, I think that would be a little rough.

fanatid:
CI was good here. Master is broken only because https://github.com/timberio/vector/pull/3699 was merged after this PR was opened.. Both pull requests independently are ok, but together produce the broken master :(

leebenson:
@Hoverbear, the failing tests here were exactly the same tests that you were talking with @JeanMertz about in Discord. I believe he's already tracking them.

Hoverbear:
Ahhh okay!

PR #3774, opened 2020-09-09
chore(observability): Make counters naming consistent

juchiast:
We are using both "processing_error" and "processing_errors" in counters.

This PR renames all to "processing_error**s**".


JeanMertz (approved):
Sweet! Thanks @juchiast

We've actually been discussing making an enum out of these key names, or at least use a constant variable to reference the key names. I don't believe we acted on that discussion so far, but it would certainly help avoid a situation like this in the future.

PR #3793, opened 2020-09-10
docs: Add JMX metrics RFC

jamtur01:


rfcs/2020-07-28-3642-jmx_rfc.md
```diff
@@ -0,0 +1,239 @@
...
+
+```text
+This exporter is intended to be run as a Java Agent, exposing a HTTP server and serving metrics of the local JVM. It can be also run as an independent HTTP server and scrape remote JMX targets, but this has various disadvantages, such as being harder to configure and being unable to expose process metrics (e.g., memory and CPU usage). Running the exporter as a Java Agent is thus strongly encouraged.
+```
+
+We don't necessarily concur but it may require some testing as we proceed.
+
+The query will return these metrics by parsing the query results and converting them into metrics:
+
+- `jmx_up` -> Used as an uptime metric (0/1) ? - merits a broader discussion.
```
jszwedko, line 40: :+1: I think these up/down metrics are useful. It lets you know if the process is even listening.

jszwedko (approved):
In general, this looks good, but I think we should validate whether it is preferable to run a local agent vs. scrape the metrics remotely. It seems like both Prometheus and Telegraf take the agent approach which makes me feel like perhaps it is preferable for some reasons.

If we do decide an agent is preferable, we could even opt to point people to run the Prometheus exporter and use the `prometheus` source.

Assuming that we do want to support remote scraping of JMX with vector, and given there is a first-class prometheus plugin, I think we should aim to keep the metrics as close as possible to it following the guidance in https://github.com/timberio/vector/pull/3519#issuecomment-682096987 regarding the matching of our host metrics with the prometheus `node_exporter`.

PR #3798, opened 2020-09-10
enhancement(prometheus source): Prometheus source to support TLS config and auth token

langecode:
Adding support for the standard Vector TLS configuration as well as a custom authentication token. This allows for scraping Kubernetes control plane components (#3797 )

### DIFF
--- a/src/sources/prometheus/mod.rs
+++ b/src/sources/prometheus/mod.rs
@@ -5,6 +5,7 @@ use crate::{
         PrometheusParseError, PrometheusRequestCompleted,
     },
     shutdown::ShutdownSignal,
+    tls::{tls_connector_builder, TlsOptions, TlsSettings},
     Event, Pipeline,
 };
 use futures::{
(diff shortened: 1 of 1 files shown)


src/sources/prometheus/mod.rs
```diff
@@ -47,7 +55,31 @@ impl crate::config::SourceConfig for PrometheusConfig {
...
         }
-        Ok(prometheus(urls, self.scrape_interval_secs, shutdown, out))
+
+        let mut headers = IndexMap::new();
+        if let Some(bearer_token_file) = &self.bearer_token_file {
+            let mut text = Vec::<u8>::new();
+            File::open(bearer_token_file)
+                .expect("unable to open file")
+                .read_to_end(&mut text)
+                .expect("unable to read file");
```
fanatid, line 65: It would be better to read file asynchronous. `fn build` should be changed to `fn build_async`. https://github.com/timberio/vector/blob/439a9cfdb48d976f19a0d2e5b84fc268752d93ba/src/config/mod.rs#L141

fanatid (changes requested):
Need tests

langecode:
Thanks for the comments, I will look into this - never wrote any rust code before :-)

jamtur01:
Neat. The new option also needs docs (https://github.com/timberio/vector/blob/master/.meta/sources/prometheus.toml.erb)

bruceg (changes requested):
Looks good, especially for your first steps in Rust! I agree it needs tests and docs updates though.

src/sources/prometheus/mod.rs
fanatid: @langecode if you will have any problem with changing file reading to non-blocking let us know, we will help :slightly_smiling_face:

langecode:
Thanks a lot for the feedback!

I will close this and await a solution to  #3800 / #3801 instead. The scrape of workloads inside of Kubernetes requires a level dynamic configuration pretty hard to handle within Vector but already present in Prometheus. So we would like to have Prometheus do the advanced scraping and then pass the metrics to Vector.

PR #3807, opened 2020-09-10
enhancement(tokenizer transform): add internal events

### LINKED ISSUE #3410: Define events for the `tokenizer` transform
As part of finishing https://github.com/timberio/vector-product/issues/14, we need to define internal events for the `tokenizer` transform.

## Defining events

## Requirements

- [ ] Add a `TokenizerEventProcessed` event.

lukesteensen:
Closes #3410

Pretty straightfoward. Snuck in one small cleanup to get rid of an unnecessary conversion and align better with other `MissingField` errors.

### DIFF
--- a/src/internal_events/mod.rs
+++ b/src/internal_events/mod.rs
@@ -61,6 +61,8 @@ mod syslog;
 #[cfg(feature = "transforms-tag_cardinality_limit")]
 mod tag_cardinality_limit;
 mod tcp;
+#[cfg(feature = "transforms-tokenizer")]
+mod tokenizer;
 mod unix;
 mod vector;
 #[cfg(feature = "wasm")]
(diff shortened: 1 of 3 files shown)


src/internal_events/mod.rs
```diff
@@ -126,6 +128,8 @@ pub use self::syslog::*;
 #[cfg(feature = "transforms-tag_cardinality_limit")]
 pub(crate) use self::tag_cardinality_limit::*;
 pub use self::tcp::*;
+#[cfg(feature = "transforms-tokenizer")]
+pub use self::tokenizer::*;
```
bruceg, line 132: Using `pub(crate)` here and in `tokenizer.rs` will help with inlining

bruceg (approved):
Looks good except for my gripe about `pub`.

src/internal_events/mod.rs
lukesteensen: Adding here, but do we need to propagate it down? I assumed not, but now I'm less sure.

Separately, I was also thinking of trying to make the `internal_events` module itself `pub(crate)`, which seems like a good thing to do if propagation works the way I assumed.

bruceg: Yes, it would need to be on both the actual structs and the re-import, as without the structs being restricted, they are still fully `pub`.

Making the module itself `pub(crate)` is an good idea try, and it should achieve the same effect as restricting everything inside. I tested with the playground, and it seems to achieve the desired effect.

lukesteensen: Ah, right, there is still a fully `pub` path to the item. Will fix.

JeanMertz (approved):

PR #3822, opened 2020-09-11
fix: Avoid calling `peer_addr`

### LINKED ISSUE #2731: TCP source: "Socket not connected"

ktff:
Ref #2731.

The issue with vector shutting down when we fail to accept `tcp` connenction was fixed with #3449. This PR addresses the source of the error which was a call to `peer_addr`. There was probably a race, between us receiving the socket and calling `peer_addr`, and other side closing the channel. Although I wasn't able to trigger this race, the PR adds a fix which completely eliminates the need for calling `peer_addr` and so eliminates any race that we can avoid.

The bug was effecting multiple sources and sinks  

### Todo

- [x] Ask for original reporters to verify that this PR fixes the issue.

### DIFF
--- a/src/sinks/socket.rs
+++ b/src/sinks/socket.rs
@@ -242,7 +242,7 @@ mod test {
         // Only accept two connections.
         let jh = tokio::spawn(async move {
             let tls = MaybeTlsSettings::from_config(&config, true).unwrap();
-            let mut listener = tls.bind(&addr).await.unwrap();
+            let listener = tls.bind(&addr).await.unwrap();
             listener
                 .accept_stream()
                 .take(2)
--- a/src/sinks/socket.rs
+++ b/src/sinks/socket.rs
@@ -244,7 +244,7 @@ mod test {
             let tls = MaybeTlsSettings::from_config(&config, true).unwrap();
             let mut listener = tls.bind(&addr).await.unwrap();
             listener
-                .incoming()
+                .accept_stream()
                 .take(2)
                 .for_each(|connection| {
                     let mut close_rx = close_rx.take();
--- a/src/sources/splunk_hec.rs
+++ b/src/sources/splunk_hec.rs
@@ -120,12 +120,12 @@ impl SourceConfig for SplunkConfig {
             .or_else(finish_err);
 
         let tls = MaybeTlsSettings::from_config(&self.tls, true)?;
-        let mut listener = tls.bind(&self.address).await?;
+        let listener = tls.bind(&self.address).await?;
 
         let fut = async move {
             let _ = warp::serve(services)
                 .serve_incoming_with_graceful_shutdown(
-                    listener.incoming(),
+                    listener.accept_stream(),
                     shutdown.clone().compat().map(|_| ()),
                 )
                 .await;
--- a/src/sources/util/http.rs
+++ b/src/sources/util/http.rs
@@ -126,10 +126,10 @@ pub trait HttpSource: Clone + Send + Sync + 'static {
 
         let tls = MaybeTlsSettings::from_config(tls, true)?;
         let fut = async move {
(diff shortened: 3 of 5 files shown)


bruceg (changes requested):
Once this clears the tests (lints and component tests), looks good to go.

ktff:
@bruceg I'm curious, why is `peer_addr` behind `#[cfg(feature = "listenfd")]`? Can we remove it now?

bruceg:
I'm not entirely sure, those were added by @fanatid

fanatid (approved):
I added `#[cfg(feature = "listenfd")]` to `peer_field` because there were warnings with some features (`check-component-features` script).

`MaybeTlsIncomingStream::new` can be rewritten with [cfg-if](https://crates.io/crates/cfg-if) crate for removing duplicated code:
```rust
    pub(super) fn new(
        stream: TcpStream,
        peer_addr: SocketAddr,
        acceptor: Option<SslAcceptor>,
    ) -> Self {
        let state = match acceptor {
            Some(acceptor) => StreamState::Accepting(
                async move { tokio_openssl::accept(&acceptor, stream).await }.boxed(),
            ),
            None => StreamState::Accepted(MaybeTlsStream::Raw(stream)),
        };

        cfg_if::cfg_if! {
            if #[cfg(feature = "listenfd")] {
                Self { peer_addr, state }
            } else {
                Self { state }
            }
        }
    }
```

ktff:
@bruceg @fanatid I think this a valid usage of conditional `#[allow(dead_code)]` so check the commit after this message. With it we can remove the specialized `new` function.

bruceg (approved):

fanatid (approved):

ktff:
@joscha-alisch @perfectayush can you both check that this branch fixes the issue? I would like to confirm it before merging it.

perfectayush:
@ktff I won't be able to verify for this. I was working on this in my previous organization. But I will let my colleagues who are still working on this for verification.

joscha-alisch:
@ktff I can test soonest on Monday, currently on holidays without the machine I had those issues with.

ktff:
@perfectayush @joscha-alisch thanks, that's great!

ktff:
@perfectayush @joscha-alisch any success?

perfectayush:
@ktff Sorry, my colleagues won't be able to test this anytime soon. We had implemented a workaround for this (supervisor restarts). And fixing/testing this isn't a high priority task at the moment.

jamtur01:
@ktff You comfortable merging this without the validation?

ktff:
@jamtur01 I am, it has worth in removing one source of errors, but I would like to keep the issue open until we hear something from @joscha-alisch

PR #3826, opened 2020-09-11
chore: introduce Stream variant in VectorSink enum

### LINKED ISSUE #2937: Migrate StreamingSink to new sink variant instead of using compat
Blocked on #2934 

See [Architecture RFC](https://github.com/timberio/vector/blob/master/rfcs/2020-06-18-2625-architecture-revisit.md).

fanatid:
This is attempt to close #2937.

While this changed introduce enum variant as trait with `run` method, I do not like:
- Boxing stream (`Pin` added for supporting `next` and other methods from `StreamExt`) in `StreamSink` signature. This required because trait with generics (stream in our case) can not be put into enum variant as `Box<dyn StreamSink>`.
- Call `run` with `&mut self` instead `self` which can be more flexible. This required because the value can not be moved out of Box because the size of the value can not be statically determined.
- Manual `acker` usage in each sink, we can easily forget about this as about `apply_rules` from encoding for example. Previously ack was done in `StreamSink` which worked with delays as I understand, but at least we did not repeat ack call in each sink.

The old `StreamSink` is still here, if these changes look correct I'll update other sinks.


src/sinks/mod.rs
```diff
@@ -96,26 +96,27 @@ pub enum HealthcheckError {
...
-    pub async fn run<S>(self, input: S) -> Result<(), ()>
+    pub async fn run<S>(mut self, input: S) -> Result<(), ()>
     where
-        S: futures::Stream<Item = Result<Event, ()>> + Send,
+        S: futures::Stream<Item = Result<Event, ()>> + Send + 'static,
     {
-        self.run01(Compat::new(input.boxed())).await
+        match self {
+            Self::Futures01Sink(sink) => input.forward(sink.sink_compat()).await,
+            Self::Stream(ref mut s) => s.run(Box::pin(input)).await,
```
MOZGIII, line 110: Can't this be just `pin_mut!` instead of `Box::pin`?

src/sinks/util/sink.rs
```diff
@@ -129,6 +133,16 @@ impl<T: Sink> Sink for StreamSink<T> {
     }
 }
 
+// === StreamSink2 ===
+
+#[async_trait]
+pub trait StreamSink2 {
+    async fn run(
+        &mut self,
+        input: Pin<Box<dyn futures::Stream<Item = Result<Event, ()>> + Send>>,
```
MOZGIII, line 142: How do you feel about using [`BoxStream`](https://docs.rs/futures/0.3.5/futures/stream/type.BoxStream.html) here?

fanatid: Missed this, it will be great to use `BoxStream` here!

MOZGIII: I recently realized this type is also useful to bind the generics.

Instead of:

```rust
Box::pin(myval) as BoxStream<_>
```

we can do

```rust
BoxStream::pin(myval)
```

I think it's pretty neat, but somehow I didn't figure it before. Just wanted to share this small tip :smile:

src/sinks/mod.rs
fanatid: I'm not sure that this is possible?
`run` signature:
```rust
async fn run(
        &mut self,
        input: Pin<Box<dyn futures::Stream<Item = Result<Event, ()>> + Send>>,
    ) -> Result<(), ()>;
```

I need `Box<dyn Trait>` because generic can not be used...
```bash
error[E0038]: the trait `sinks::util::sink::StreamSink2` cannot be made into an object
   --> src/sinks/mod.rs:71:12
    |
71  |     Stream(Box<dyn util::StreamSink2 + Send>),
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `sinks::util::sink::StreamSink2` cannot be made into an object
    | 
   ::: src/sinks/util/sink.rs:139:11
    |
139 | pub trait StreamSink2 {
    |           ----------- this trait cannot be made into an object...
140 |     async fn run<S: futures::Stream<Item = Result<Event, ()>> + Send>(
    |              --- ...because method `run` has generic type parameters
    |
    = help: consider moving `run` to another trait

For more information about this error, try `rustc --explain E0038`.
```

We can use generic, but only if add `Self: Sized`, but then we it will not be possible call `run` method from `VectorSink` (because we have trait, not object) :disappointed:

src/sinks/util/sink.rs
fanatid: Wow, I do not see this in the docs :thinking: 
`BoxStream` is `Pin<alloc::boxed::Box<dyn Stream<Item = T> + Send + 'a>>` and `Pin` do not have method `pin`, how it's possible?

MOZGIII: Hmm, nvm then :D

src/sinks/mod.rs
MOZGIII: Ah, right. Well, this isn't temporary though, it is the aspect of the new design, right? I don't think it's problematic, but it would be interesting to see if there are workarounds for this...

src/sinks/util/sink.rs
```diff
@@ -129,6 +133,13 @@ impl<T: Sink> Sink for StreamSink<T> {
     }
 }
 
+// === StreamSink2 ===
+
+#[async_trait]
+pub trait StreamSink2 {
+    async fn run(&mut self, input: BoxStream<'_, Result<Event, ()>>) -> Result<(), ()>;
```
MOZGIII, line 140: Why do we have `Result<Event, ()>` in this type, rather than bare `Event`? Is it meaningful to pass errors to sinks like this?

fanatid: @lukesteensen ask me about same in DM :smile: 
The problem in `futures01::Stream` which requires 2 types: `Item` and `Error`. These 2 types converted with compatibility code to `Result<Item, Error>`. I decide to leave `Result` for some time, but maybe I should delete it already and add extra code for compatibility with futures01.

MOZGIII: I'd suggest we map the `Err(())` of `futures01::Stream` to `None`, and unwrap the `Ok`. It's a universally applicable conversion, meaningful if we bound the error type to `()` (which we do).

I suppose the sooner we sweep this `Result` out the better. It will allow us to get rid of the odd (and potentially erroneous) logic at the sink level.

fanatid: Yeah, I think you are right. Sooner is better, will change :+1:

src/sinks/file/mod.rs
```diff
@@ -181,12 +180,16 @@ impl FileSink {
...
-        pin_mut!(input);
+    async fn run(&mut self, mut input: BoxStream<'_, Result<Event, ()>>) -> crate::Result<()> {
         loop {
             tokio::select! {
                 event = input.next() => {
                     match event {
+                        Some(Err(())) => {},
+                        Some(Ok(event)) => {
+                            self.process_event(event).await;
+                            self.acker.ack(1);
```
MOZGIII, line 191: We can't ack at this level, as this would ignore the error conditions.

I think if we make the `process_event` return `bool`, or `Result<(), ()>` it would be possible to lift this up. Better yet: return an `enum ShouldAck { Ack, NoAck }` fore more explicitness.

Alternatively, we could move the logic of issuing the `self.acker.ack(1);` down the line - but it might hurt the clarity of the code. The return type approach would guarantee a signle pass to the caller (and calling the `self.acker.ack(1);`).

MOZGIII: I think it is. It was wrong before as well though.
This implementation would ack even if there was an error. I might be getting the concept of our acks wrong though...

lukesteensen: We have to ack to continue to make progress, so it's more a way of saying "I've done all I'm going to do with this event" than "this was processed successfully". I'd like to make that better in the future but for now I think we can just match the existing behavior.

src/sinks/util/sink.rs
```diff
@@ -129,6 +133,13 @@ impl<T: Sink> Sink for StreamSink<T> {
     }
 }
 
+// === StreamSink2 ===
+
+#[async_trait]
+pub trait StreamSink2 {
```
lukesteensen, line 139: It'd be nice to get rid of the old `StreamSink` and not introduce a `2` here, but I'm not sure how much more work that'd be. Easy to do as a follow up.

lukesteensen (approved):
Need to get CI green before merging, but the shape of this is looking good.

src/sinks/util/sink.rs
fanatid: I'd like not to introduce `StreamSink2` too, but I did this just for the demonstration of the new `StreamSink`. I'll finish update in this PR.

MOZGIII: In case we can't cut the old `StreamSink` just yet, I propose renaming it to `StreamSinkOld` in this PR.
I figured starting the phasing out with renaming helps a lot with the merges/rebases in the parallel PRs - as the step for that massive rename is pushed to the colder code, rather than the hotter one.

bruceg (commented):
I'm with @lukesteensen, this looks fine but needs to pass CI.

Hoverbear (approved):

MOZGIII (approved):

src/sinks/util/sink.rs
fanatid: Renamed to `StreamSinkOld`. Will rid in other PR. These changes already reviewed/approved, with new changes process will need to be repeated and I also not sure that this will be done before I off for a week.

PR #3827, opened 2020-09-11
chore(data model): Event fixture tests and `is_$TYPE()` functions

Hoverbear:
Polishes off #2841 so it can be closed.


Hoverbear:
@lukesteensen I'd love to sit down and make these into a bench suite as well if you thing it's worth the time.

src/event/log_event.rs
```diff
@@ -177,3 +177,58 @@ impl Serialize for LogEvent {
...
+    //   * Ensures the EventLog parsed from bytes and turned into a serde_json::Value are equal to the
+    //     item being just plain parsed as json.
+    //
+    // Basically: This test makes sure we aren't mutilating any content users might be sending.
+    #[test]
+    fn json_value_to_vector_log_event_to_json_value() {
+        crate::test_util::trace_init();
+        const FIXTURE_ROOT: &str = "tests/data/fixtures/log_event";
+
+        tracing::trace!(?FIXTURE_ROOT, "Opening.");
```
lukesteensen, line 208: nitpick: Any reason for `tracing::trace!` here instead of just `trace!`? Everywhere else we use the latter.

src/event/value.rs
```diff
@@ -142,7 +143,66 @@ impl From<serde_json::Value> for Value {
...
+            Value::Map(v) => Ok(serde_json::to_value(v)?),
+            Value::Array(v) => Ok(serde_json::to_value(v)?),
+            Value::Null => Ok(serde_json::Value::Null),
+            Value::Timestamp(v) => Ok(serde_json::Value::from(timestamp_to_string(&v))),
+        }
+    }
+}
+
 impl Value {
+    pub fn is_boolean(&self) -> bool {
```
lukesteensen, line 164: We're already using the `is_enum_variant` derive for this elsewhere, so could probably use it here too.

lukesteensen (approved):

PR #3840, opened 2020-09-12
feat(platforms): rough draft of "service" deployment for vector

spencergilbert:
As I'm drafting an internal proposal to migrate our existing logging pipeline to use vector I needed a deployment method that would replace Logstash for doing transformations across all events and index them appropriately into Elasticsearch.

Currently Vector provides a chart for a "daemon" installation with a DaemonSet resource, and I converted those resources to use a StatefulSet to handle the "service" strategy described [here](https://vector.dev/docs/setup/deployment/strategies/#service).

This is still a WIP, but is currently working in my current experiments.

Related: #3227

### DIFF
--- a/distribution/helm/vector-service/.helmignore
+++ b/distribution/helm/vector-service/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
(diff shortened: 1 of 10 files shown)


MOZGIII:
Thanks for sharing this with us! I looked through the chart, and it looks good so far.

We'd need a couple more things to properly integrate this into our project - E2E tests, support for more configurations (see below), sharing things with the other chart, ensure it can autoconfigure if run together with DaemonSet one, release mechanism and etc. Don't worry about all that, I'll take care of all that when I get to it. This is definitely useful work, thx!

Regarding the chart itself, looking at this PR, I see a couple of things we'd want to include with our chart for service deployment, things like:
- allowing the use of existing PVC instead of using a volume claim template;
- allowing to set a non-headless `Service` alongside with a headless `Service` for the `StatefulSet` to enable k8s-level balancing.

You don't have to worry about these either, but maybe these concerns would be useful for your use case too. :smile: 

The overall process for this work would most likely be similar to the `DaemonSet` - we'll have this PR up without merging until we close it with our own broader integration. I appreciate your contribution, the chart itself is neat and our final one will most likely effectively do very similar things!

spencergilbert:
Yep - definitely, I'll probably have time to add some of that as I continue testing and debugging

MOZGIII:
Closing in favor of #4404. Thanks again @spencergilbert, I'm basing our implementation on the code from this PR!

PR #3842, opened 2020-09-12
chore: Added workflow_dispatch for the env workflow

jamtur01:

### DIFF
--- a/.github/workflows/environment.yml
+++ b/.github/workflows/environment.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
 
 env:
   VERBOSE: true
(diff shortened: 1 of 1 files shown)


Hoverbear (approved):

Hoverbear (changes requested):
Can you make it not publish?

.github/workflows/environment.yml
```diff
@@ -18,4 +19,3 @@ jobs:
       - run: docker login https://docker.pkg.github.com -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }}
       - uses: actions/checkout@v2
       - run: make environment-prepare
-      - run: docker push docker.pkg.github.com/timberio/vector/environment:${{ github.sha }}
```
MOZGIII, line 21: Randomly spot this, and wondering is this correct?
Maybe an `if: github.event_name != 'workflow_dispatch'` would be a better fit?

PR #3862, opened 2020-09-14
chore: Add back .PHONY

juchiast:
I have local scripts named `build`, `test`, `check`, etc. which need these targets to be phony.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-# .PHONY: $(MAKECMDGOALS) all
+.PHONY: $(MAKECMDGOALS)
 .DEFAULT_GOAL := help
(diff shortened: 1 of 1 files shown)


Hoverbear:
@juchiast Sorry this caused you issues! You're right, all of these targets https://github.com/timberio/vector/blob/master/Makefile#L175-L193

Need to be `.PHONY`. I'll go through and do this one-by-one because we no longer want to make all commands `.PHONY`.

juchiast:
@Hoverbear np

I'm only defining `$(MAKECMDGOALS)` as phony in this PR, which seems to be make all targets returned by `make` phony. Maybe you can double check it.

Hoverbear:
@juchiast Does #3865 work for you? I'll close this now but feel free to reopen if you feel this is better.

Issue #3843, opened 2020-09-12
Flaky `test_max_size` test


ktff:
Test `test_max_size` fails from time to time with:

### Output
```
2020-09-11T16:41:25.5887695Z thread 'main' panicked at 'assertion failed: `(left == right)`
2020-09-11T16:41:25.5888306Z   left: `500`,
2020-09-11T16:41:25.5888893Z  right: `494`', tests/buffering.rs:231:9
```

### References
https://github.com/timberio/vector/pull/3822/checks?check_run_id=1102732219
https://github.com/timberio/vector/pull/3824/checks?check_run_id=1102711898

jszwedko:
Another example failure: https://github.com/timberio/vector/pull/3895/checks?check_run_id=1120060759

PR #3865, opened 2020-09-14
chore: Add PHONY per target in makefile

Hoverbear:
This takes a more conservative approach to #3862 and adds `.PHONY` to the targets where appropriate.


juchiast:
@Hoverbear this works! Thanks!

Hoverbear:
This is hit by #3843

PR #3867, opened 2020-09-14
chore: Added back docker push to environment workflow

jamtur01:

### DIFF
--- a/.github/workflows/environment.yml
+++ b/.github/workflows/environment.yml
@@ -25,4 +25,8 @@ jobs:
           sudo apt clean
           docker rmi $(docker image ls -aq)
           df -h
-      - run: make environment-prepare
+      - name: prepare environment
+        run: make environment-prepare
+      - name: push image
+        if: github.event_name != 'workflow_dispatch'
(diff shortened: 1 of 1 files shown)


.github/workflows/environment.yml
```diff
@@ -25,4 +25,8 @@ jobs:
           sudo apt clean
           docker rmi $(docker image ls -aq)
           df -h
-      - run: make environment-prepare
+      - name: prepare environment
+        run: make environment-prepare
+      - name: push image
+        if: github.event_name != 'workflow_dispatch'
```
Hoverbear, line 31: So we're adding a dispatch later? https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event

jamtur01: there's also a workflow_dispatch in there.

Hoverbear: Amazing.

Hoverbear (approved):

.github/workflows/environment.yml
jamtur01: :)

PR #3657, opened 2020-08-31
chore(platforms): Migrate  x86 gnu target to cross, add new aarch64 gnu target (no sasl)

Hoverbear:
Builds a bit alongside the ideas of  #3537.

Closes #3638.
Closes #3811. 
Causes #3805.

PR #3876, opened 2020-09-14
chore(windows platform): Support tarball generation on Windows

Hoverbear:
Supports tarball generation on Windows after #3657.

Sadly Windows generally packages busybox make or bsdmake, which don't support `--transform`

This also removes our use of the `vector-target` mount.


jamtur01 (approved):
LGTM.

PR #3890, opened 2020-09-15
chore(rpm platform): Fixup rpm builds to support new tarballs

Hoverbear:
In #3657 we changed the tarball layout a bit and so the `--strip-components=` flag had to be updated.


jamtur01 (approved):
LGTM.

PR #3936, opened 2020-09-16
chore: Do not run http_headers_auth_conflict() on OSX

jamtur01:
This test fails for reasons I do not understand - only on GH Actions macos-latest - I suspect some SSL inconsistency. I'll marking it as "do not run on osx" as a result. Happy to remove the test but it works fine on all the other platforms so this seemed to be a better outcome.

### DIFF
--- a/src/sinks/http.rs
+++ b/src/sinks/http.rs
@@ -367,6 +367,8 @@ mod tests {
         );
     }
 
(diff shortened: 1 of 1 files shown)


lukesteensen (approved):
Should we open an issue to track this?

jamtur01:
@lukesteensen #3937.

PR #3938, opened 2020-09-16
chore: Added update grub to fix bootstrap

jamtur01:

### DIFF
--- a/scripts/environment/bootstrap-ubuntu-20.04.sh
+++ b/scripts/environment/bootstrap-ubuntu-20.04.sh
@@ -3,6 +3,7 @@ set -e -o verbose
 
 export DEBIAN_FRONTEND=noninteractive
 
+apt-get install -y grub-efi
 update-grub
 
 apt update -y -qq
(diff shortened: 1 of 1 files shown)


jamtur01:
In reference to https://github.com/actions/virtual-environments/issues/1605.

scripts/environment/bootstrap-ubuntu-20.04.sh
```diff
@@ -3,7 +3,11 @@ set -e -o verbose
 
 export DEBIAN_FRONTEND=noninteractive
 
+apt-get install -y grub-efi
```
jszwedko, line 6: Not a big deal in this particular case, but I generally prefer to use the longer form of flags, that is `--yes`, in scripts to make them easier to read without pulling up the man page.

scripts/environment/bootstrap-ubuntu-20.04.sh
```diff
@@ -3,7 +3,11 @@ set -e -o verbose
...
+update-grub
+
 apt update -y -qq
+apt upgrade -y -qq
```
jszwedko, line 10: Did you end up needing this? I see it wasn't mentioned in https://github.com/actions/virtual-environments/issues/1605

jamtur01: Yeah - that's the original step that was failing. The grub-efi packages was dying during the apt upgrade.

jszwedko (approved):

PR #3962, opened 2020-09-16
chore: Fix grub to work in both CI and Docker containers

jamtur01:
This is a workaround for GH https://github.com/actions/virtual-environments/issues/1605
This fix will be removed when GH addresses the issue.
What's happening here? Well, we use this script inside CI and Docker containers.
It'll run fine in CI because update-grub can find a root partition.
Inside a container, it'll fail. Either way, we don't care about the outcome of the command
so we ignore its exit.

This will fix our currently failing `environment-prepare` Make task.

### DIFF
--- a/scripts/environment/bootstrap-ubuntu-20.04.sh
+++ b/scripts/environment/bootstrap-ubuntu-20.04.sh
@@ -3,10 +3,20 @@ set -e -o verbose
 
 export DEBIAN_FRONTEND=noninteractive
 
+apt update --yes
+
+# This is a workaround for GH https://github.com/actions/virtual-environments/issues/1605
+# This fix will be removed when GH addresses the issue.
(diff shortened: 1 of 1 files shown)


scripts/environment/bootstrap-ubuntu-20.04.sh
```diff
@@ -3,10 +3,20 @@ set -e -o verbose
 
 export DEBIAN_FRONTEND=noninteractive
 
+apt update --yes
```
Hoverbear, line 6: How about we just update the packages we care about and leave the rest?

jamtur01: I feel like its a) good hygiene and b) there's no guarantee this won't be triggered by a future package install if a dependency changes.

Hoverbear (approved):

PR #3985, opened 2020-09-16
chore: Fix strip components call for creating rpms

Hoverbear:
I am not sure why #3938 broke things, but it did!


jamtur01:
It can't have been #3938. The 4441cc48e commit changed stripe components from 2 to 1 was the last chance to this.

jamtur01 (approved):
Okay. Let's give this another spin.

PR #3989, opened 2020-09-17
feat(remap transform): accept more timestamp patterns in `to_timestamp`

JeanMertz:
Based on comment in https://github.com/timberio/vector/issues/3741#issuecomment-692075031.

### DIFF
--- a/src/mapping/query/functions.rs
+++ b/src/mapping/query/functions.rs
@@ -190,12 +190,14 @@ impl Function for ToTimestampFn {
 }
 
 fn to_timestamp(value: Value) -> Result<Value> {
-    use chrono::{DateTime, TimeZone, Utc};
+    use chrono::{TimeZone, Utc};
 
     match value {
-        Value::Bytes(b) => DateTime::parse_from_rfc3339(&String::from_utf8_lossy(&b))
-            .map(|dt| dt.with_timezone(&Utc))
-            .map_err(|_| "cannot parse string as RFC3339".to_owned()),
+        Value::Bytes(b) => {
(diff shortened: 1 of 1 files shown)


src/mapping/query/functions.rs
```diff
@@ -190,12 +190,14 @@ impl Function for ToTimestampFn {
...
 
     match value {
-        Value::Bytes(b) => DateTime::parse_from_rfc3339(&String::from_utf8_lossy(&b))
-            .map(|dt| dt.with_timezone(&Utc))
-            .map_err(|_| "cannot parse string as RFC3339".to_owned()),
+        Value::Bytes(b) => {
+            return Conversion::Timestamp
+                .convert(Value::Bytes(b))
+                .map_err(|e| e.to_string())
+        }
```
bruceg, line 200: Does this work?
```suggestion
        Value::Bytes(b) => Conversion::Timestamp
                .convert(value)
                .map_err(|e| e.to_string()),
```

bruceg (approved):

StephenWakely (approved):
I concur with Bruce's statement.

src/mapping/query/functions.rs
JeanMertz: It doesn't, because `convert` returns a `Result<Value>`, whereas the other arms return `Result<DateTime>`.

bruceg: Oh, I see, I missed that in the reduced context of the diff.

Given the other two arms that return a value are pretty simple, I think removing the map would simplify things here.

PR #4004, opened 2020-09-17
fix: remove failing remap behavior test

JeanMertz:
I believe the nanoseconds aren't padded, and so the length equality check doesn't always succeed.

I've removed it for now, will see if we can make the numbers padded again in a follow-up PR, as I think that's a nice property to have actually.


jszwedko (approved):
Thanks!

I agree that it'd be nice to pad the nanoseconds out.

PR #4013, opened 2020-09-18
chore: Fix fix_retry_until_after_timeout by adding panic message

jamtur01:

### DIFF
--- a/src/test_util/mod.rs
+++ b/src/test_util/mod.rs
@@ -340,7 +340,7 @@ mod tests {
     }
 
     #[tokio::test]
-    #[should_panic]
+    #[should_panic](expected = "Timeout")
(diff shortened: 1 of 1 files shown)


JeanMertz (approved):
I'm happy to merge this, but I'm not sure how this fixes anything.

From what I understand, the `expected` parameter _narrows_ the accepted panics, so any test using this function would fail more often, if this function panic'ed with any message that doesn't contain `Timeout`, correct?

leebenson:
I'm confused by this too. I'm also not clear why this only shows up on macOS tests, and doesn't happen locally for me on macOS.

jamtur01:
Yeah I am unclear too. :(

jamtur01:
And I try and run it again this morning and it fails. 😂

PR #4014, opened 2020-09-18
Create k8s-kafka-to-loki.yaml

imdingtalk:
a quick start on k8s for kafka to loki.


MOZGIII:
Thank you for the PR, @imdingtalk.

We're thinking about creating a wiki where users would be able to share the configurations, and this kind of example should probably belong to a wiki page.

In the meantime I'll close this PR, as adding this example increases the maintenance weight - we'll have to keep an eye on it to keep it up to date with the changes - but at this time we don't have the automation or process to do it. A potentially outdated example in our repo is not a good thing, so we have to reject changes like this at this time.

PR #4018, opened 2020-09-18
enhancement(kubernetes_logs): Include container_name in kubernetes_logs

### LINKED ISSUE #3696: K8s: add container name to log event metadata
A follow up from https://github.com/timberio/vector/issues/3674#issuecomment-686560848.

### LINKED ISSUE #4017: Kubernetes Logs Container Name
Aware this is technically an unreleased source, so I may be jumping the gun a bit here, but the current kubernetes_log source doesn't provide an easy way to get the container_name from which it is pulling the logs.

One can parse the file event field in order to obtain this, but this feels a little bit fragile and doesn't have the best UX.

tustvold:
Closes #3696

### DIFF
--- a/src/sources/kubernetes_logs/pod_metadata_annotator.rs
+++ b/src/sources/kubernetes_logs/pod_metadata_annotator.rs
@@ -2,7 +2,7 @@
 
 #![deny(missing_docs)]
 
-use super::path_helpers::parse_log_file_path;
+use super::path_helpers::{parse_log_file_path, LogFileInfo};
 use crate::{
     event::{LogEvent, PathComponent, PathIter},
     kubernetes as k8s, Event,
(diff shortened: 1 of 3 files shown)


src/sources/kubernetes_logs/pod_metadata_annotator.rs
```diff
@@ -56,10 +58,15 @@ impl PodMetadataAnnotator {
...
+        let file_info = parse_log_file_path(file)?;
+        let guard = self.pods_state_reader.get(file_info.pod_uid)?;
         let entry = guard.get_one()?;
         let pod: &Pod = entry.as_ref();
+
+        log.insert(
+            &self.fields_spec.container_name,
+            file_info.container_name.to_owned(),
+        );
         annotate_from_metadata(log, &self.fields_spec, &pod.metadata);
```
JeanMertz, line 70: We might want to rename this `annotate_from_pod_metadata`, since we're not also annotating from non-pod metadata, which happens outside this function.

JeanMertz (approved):
Sweet! This looks solid to me, but curious what @MOZGIII thinks.

src/sources/kubernetes_logs/pod_metadata_annotator.rs
MOZGIII: That's not really a concern here - there are no other candidates that have metadata in this code. What's specializes this fn here is the fact that it reads from the metadata. We'll add other fns in the future - for instance, to populate the container image we'd need one to `annotate_from_spec`. In all cases `Pod` is assumed here.

The mod name itself might need to be altered, but then it can be considered as mentioning "metadata" in a broader sense.

Either way, I don't think this PR should be doing this.

MOZGIII (changes requested):
Yep, this looks great!
One issue though is the unit tests - they should be updated accordingly to test that the container name is populated as expected.
I saw E2E tests are updated - this is good, but unit tests are also required.

MOZGIII:
I suggest extracting the added logic into a new function:

```rust
fn annotate_from_file_info(
    log: &mut LogEvent,
    fields_spec: &FieldsSpec,
    file_info: &LogFileInfo<'_>,
) {
    log.insert(
        &fields_spec.container_name,
        file_info.container_name.to_owned(),
    );
}
```

and adding the test cases for it similar to how we have the `test_annotate_from_metadata`.

MOZGIII:
I ended up just doing it to prove it's the right concept :smile: https://github.com/tustvold/vector/pull/1

tustvold:
Err... I had hoped it would squash - will do manually

MOZGIII (approved):
Great! The Semantic Pull Request fail is fine, something's WIP with metadata.

MOZGIII:
I think we can merge this, @tustvold ready to proceed?

MOZGIII:
@tustvold could you merge the `master` in or rebase please?

PR #3869, opened 2020-09-14
chore: correct linger misbehavior

Hoverbear:
It's my hope this will resolve #3830, but I have no proof of it.

Regardless, this corrects a misbehavior I found with @LucioFranco and @fanatid during debugging. 

You can see in #3830 @spencergilbert hit:

```rust
  12:          0x1f193b7 - <futures_util::future::future::NeverError<Fut> as core::future::future::Future>::poll::h7fb471e0dc82775f
  13:          0x2203da9 - <futures_util::compat::compat03as01::Compat<Fut> as futures::future::Future>::poll::hd3a935b220d64df3
  14:          0x1c1510a - <vector::sinks::util::sink::BatchSink<S,B,Request> as futures::sink::Sink>::poll_complete::h2ae3a9d473642464
```

Which leads us to here:

https://github.com/timberio/vector/blob/4d139851e441cac846414fe604c039d98461c7ff/src/sinks/util/sink.rs#L231-L232

As you can see, below, we use it here:

https://github.com/timberio/vector/blob/4d139851e441cac846414fe604c039d98461c7ff/src/sinks/util/sink.rs#L274-L277

But in the case where it returns a `Ready` (such that we are not allowed to poll it again) it is not reset.

This probably started due to the 0.1 to 0.2 migration of Tokio, since delay changed:

https://github.com/timberio/vector/blob/1b270eb591fbc08743c466164d36ba7879bdd21a/src/sinks/util/batch.rs#L132-L134

Issue #3830, opened 2020-09-12
thread 'tokio-runtime-worker' panic


spencergilbert:
### Vector Version

```
vector 0.11.0 (g50a9c3a x86_64-unknown-linux-musl 2020-09-10)
```

### Vector Configuration File

```toml
data_dir = "/vector-data-dir"
[log_schema]
  host_key = "host"
  message_key = "message"
  source_type_key = "source_type"
  timestamp_key = "timestamp"
[sources.internal_metrics]
  type = "internal_metrics"
[sources.kafka]
  type = "kafka"
  bootstrap_servers = "logging-kafka-bootstrap.logging-kafka:9092"
  group_id = "test-vector"
  topics = ["test-vector"]
  
[transforms.message_parser]
  type = "json_parser"
  inputs = ["kafka"]
  field = "message"
  drop_invalid = true
  
[sinks.elasticsearch]
  type = "elasticsearch"
  inputs = ["message_parser"]
  host = "http://logging-es-http.logging:9200"
  index = "logstash-vector-test"
  auth.strategy = "basic"
  auth.password = "${ELASTICSEARCH_PASSWORD}"
  auth.user = "${ELASTICSEARCH_USER}"
  # https://github.com/timberio/vector/pull/3716
  #encoding.except_fields = ["index"]
  
[sinks.prometheus]
  type = "prometheus"
  inputs = ["internal_metrics"]
  address = "0.0.0.0:9598"
  namespace = "vector"
```

### Debug Output

https://gist.github.com/spencergilbert/52f59f015769e19f4723a08da9c29f2f

### Expected Behavior

No panic!

### Actual Behavior

Panic!

spencergilbert:
The build might have been a few days old, I'm trying to reproduce with the nightly from 9/11

MOZGIII:
I suspect this has something to do with the HTTP stuff, might've been fixed already

spencergilbert:
@MOZGIII I'll try the latest again today, and also verify this wasn't happening with 0.10.0

I think I ran the stable release for long enough originally and didn't see any issues but it would be good to confirm there's no regression

MOZGIII:
I checked the history and didn't find the relevant mentions of a fix too. I just saw a mention of the fix in Discord, so take it with a grain of salt. This definitely needs to be properly verified.

jszwedko:
Noting this was also reported as https://github.com/timberio/vector/issues/3863

spencergilbert:
[...]

PR #4024, opened 2020-09-18
fix(sinks): instrument and guard against double poll

lukesteensen:
Ref #3830

It's still not clear to me how we could be getting a double-poll after #3869, but this should plug the gap rather thoroughly and give us some instrumentation to help explain. It also addresses a (very slight, likely never hit) behavior issue introduced there, along with a few small cleanups and explanatory comments.

The clue system can likely be removed once we figure out what is going on, but it should help us continue to track down the proble while mitigating user impact in the meantime.

### DIFF
--- a/src/sinks/util/sink.rs
+++ b/src/sinks/util/sink.rs
@@ -193,15 +193,16 @@ where
     }
 
     fn linger_elapsed(&mut self) -> bool {
-        let clue = self.capture_clue();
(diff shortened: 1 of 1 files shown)


Hoverbear (approved):
Not a very satisfying fix, but this appears to accomplish what we discussed in DMs. I think it's a way forward.

PR #4025, opened 2020-09-18
fix(windows platform): Update tokio signal to avoid unnecessary shutdowns.

Hoverbear:
This fixes the shutdown issue reported by @oktal today on Windows by upgrading to `tokio::signal`.

```rust
   Compiling vector v0.11.0 (C:\Users\ana\CLionProjects\vector)
    Finished dev [unoptimized + debuginfo] target(s) in 29.69s
     Running `target\debug\vector.exe --config .\vector.toml`
Sep 18 13:56:13.021  INFO vector: Log level "info" is enabled.
Sep 18 13:56:13.025  INFO vector: Loading configs. path=["vector.toml"]
Sep 18 13:56:13.041  INFO vector::sources::stdin: Capturing STDIN.
Sep 18 13:56:13.042  INFO vector::topology: Running healthchecks.
Sep 18 13:56:13.042  INFO vector::topology::builder: Healthcheck: Passed.
Sep 18 13:56:13.042  INFO vector::topology: Starting source "source0"
Sep 18 13:56:13.043  INFO vector::topology: Starting sink "sink0"
Sep 18 13:56:13.043  INFO vector: Vector has started. version="0.11.0" git_version="v0.9.0-702-g26cd7ae" released="Fri, 18 Sep 2020 17:28:57 +0000" arch="x86_64"
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SpawnError { is_shutdown: true }', C:\Users\ana\scoop\persist\rustup-msvc\.cargo\registry\src\github.com-1ecc629
9db9ec823\tokio-executor-0.1.10\src\global.rs:172:5
Sep 18 13:56:13.045  INFO source{name=source0 type=stdin}: vector::sources::stdin: Finished sending
error: process didn't exit successfully: `target\debug\vector.exe --config .\vector.toml` (exit code: 101)
```

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -1,6 +1,4 @@
-use futures::{
-    FutureExt, Stream,
-};
+use futures::{FutureExt, Stream};
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub enum SignalTo {
@@ -15,8 +13,8 @@ pub enum SignalTo {
 /// Signals from OS/user.
 #[cfg(unix)]
 pub fn signals() -> impl Stream<Item = SignalTo> {
-    use tokio::signal::unix::{signal, SignalKind};
     use futures::future::{select_all, BoxFuture};
+    use tokio::signal::unix::{signal, SignalKind};
 
     let mut sigint = signal(SignalKind::interrupt()).expect("Signal handlers should not panic.");
     let mut sigterm = signal(SignalKind::terminate()).expect("Signal handlers should not panic.");
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -34,6 +34,7 @@ pub fn signals() -> impl Stream<Item = SignalTo> {
 /// Signals from OS/user.
 #[cfg(windows)]
 pub fn signals() -> impl Stream<Item = SignalTo> {
+    use futures::future::FutureExt;
     let ctrl_c = tokio::signal::ctrl_c();
 
     ctrl_c.map(|_| SignalTo::Shutdown).into_stream()
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -1,5 +1,4 @@
 use futures::{
-    future::{select_all, BoxFuture},
     FutureExt, Stream,
 };
 
@@ -17,6 +16,7 @@ pub enum SignalTo {
 #[cfg(unix)]
 pub fn signals() -> impl Stream<Item = SignalTo> {
     use tokio::signal::unix::{signal, SignalKind};
+    use futures::future::{select_all, BoxFuture};
 
     let mut sigint = signal(SignalKind::interrupt()).expect("Signal handlers should not panic.");
     let mut sigterm = signal(SignalKind::terminate()).expect("Signal handlers should not panic.");
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -1,6 +1,5 @@
 use futures::{
-    future::{select_all, BoxFuture},
-    FutureExt, Stream,
+    Stream,
 };
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
@@ -15,7 +14,7 @@ pub enum SignalTo {
 
 /// Signals from OS/user.
 #[cfg(unix)]
-pub fn signals() -> impl Stream<Item = SignalTo> {
+pub fn signals() -> impl Stream<Item = SignalTo> + 'static {
     use tokio::signal::unix::{signal, SignalKind};
 
     let mut sigint = signal(SignalKind::interrupt()).expect("Signal handlers should not panic.");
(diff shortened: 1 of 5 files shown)


oktal:
Hey, thanks for this !

So I've tried merging your PR to my branch, and I can confirm that it now runs correctly on Windows.

I just get a couple of warnings that can be easily fixed by conditionally importing components I guess

```
λ cargo build --no-default-features --features default-msvc
warning: unused imports: `BoxFuture`, `select_all`
 --> src\signal.rs:2:14
  |
2 |     future::{select_all, BoxFuture},
  |              ^^^^^^^^^^  ^^^^^^^^^
  |

warning: variable does not need to be mutable
   --> src\app.rs:173:17
    |
173 |             let mut signals = signal::signals();
    |                 ----^^^^^^^
    |                 |
    |                 help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default
```

Hoverbear:
Ah, dangit! Let me fix these. :)

src/signal.rs
```diff
@@ -14,35 +13,30 @@ pub enum SignalTo {
...
+    let set: Vec<BoxFuture<SignalTo>> = vec![
+        Box::pin(async move { sigint.recv().map(|_| SignalTo::Shutdown).await }),
+        Box::pin(async move { sigterm.recv().map(|_| SignalTo::Shutdown).await }),
+        Box::pin(async move { sigquit.recv().map(|_| SignalTo::Quit).await }),
+        Box::pin(async move { sighup.recv().map(|_| SignalTo::Reload).await }),
+    ];
+
+    let selection = select_all(set.into_iter()).map(|(val, _, _)| val);
+
+    selection.into_stream()
```
lukesteensen, line 33: It looks like we might have changed the behavior here a bit. Before, we had a combined continuous stream of these signals. Now it seems like we have a future of only the first to be received, which is then converted to a single-element stream.

lukesteensen (changes requested):

src/signal.rs
Hoverbear: Thank you for noticing this. :) I fixed it (I think?) and learnt some great new tricks while doing it.

src/signal.rs
```diff
@@ -14,35 +13,29 @@ pub enum SignalTo {
...
-    let ctrl_c = tokio_signal::ctrl_c().flatten_stream();
+    use futures::future::FutureExt;
+    let ctrl_c = tokio::signal::ctrl_c();
 
-    ctrl_c
-        .map(|_| SignalTo::Shutdown)
-        .compat()
-        .into_stream()
-        .map(|result| result.expect("Shouldn't error"))
+    ctrl_c.map(|_| SignalTo::Shutdown).into_stream()
```
lukesteensen, line 40: Do we need this to be more than a single element stream too? It looks like `ctrl_c()` is just an async function, so a single future that we're `to_stream()`ing here.

src/signal.rs
```diff
@@ -14,35 +13,29 @@ pub enum SignalTo {
...
-        .into_stream()
-        .map(|result| result.expect("Neither stream errors"))
+    async_stream::stream! {
+        let signal = tokio::select! {
+            _ = sigint.recv() => SignalTo::Shutdown,
+            _ = sigterm.recv() => SignalTo::Shutdown,
+            _ = sigquit.recv() => SignalTo::Quit,
+            _ = sighup.recv() => SignalTo::Reload,
+        };
+        yield signal;
```
lukesteensen, line 30: Nice! Pleasantly clean too.

src/signal.rs
Hoverbear: With windows this will be shutdown, which I don't think we need to handle multiple times, since it's turning off? I'll update it anyways. :)

Hoverbear: Should be fixed up now. :)

src/signal.rs
```diff
@@ -14,35 +13,32 @@ pub enum SignalTo {
...
+    use futures::future::FutureExt;
+    let ctrl_c = tokio::signal::ctrl_c();
 
-    ctrl_c
-        .map(|_| SignalTo::Shutdown)
-        .compat()
-        .into_stream()
-        .map(|result| result.expect("Shouldn't error"))
+    async_stream::stream! {
+        let signal = ctrl_c.map(|_| SignalTo::Shutdown).await;
```
lukesteensen, line 41: `ctrl_c` is still a future here, not a stream, so I'm pretty sure this is the same as before. That said, if we don't need to handle it more than once here then this is fine. We handle it twice on linux because the first is "shutdown nicely please" and the second is "shutdown immediately".

Hoverbear: I tried 

```rust
    async_stream::stream! {
        let mut signal;
        loop {
            tokio::signal::ctrl_c().map(|_| SignalTo::Shutdown).await;
        }
        yield signal;
    }
}
```

But that left me with a binary that didn't handle ctrl  C.

Hoverbear: Yeah I thought I actually wrote a loop...

lukesteensen (approved):

src/signal.rs
```diff
@@ -14,35 +13,33 @@ pub enum SignalTo {
...
-        .into_stream()
-        .map(|result| result.expect("Neither stream errors"))
+    async_stream::stream! {
+        let signal = tokio::select! {
+            _ = sigint.recv() => SignalTo::Shutdown,
+            _ = sigterm.recv() => SignalTo::Shutdown,
+            _ = sigquit.recv() => SignalTo::Quit,
+            _ = sighup.recv() => SignalTo::Reload,
+        };
+        yield signal;
```
ktff, line 30: @Hoverbear this should also be a loop, I tested the master with this PR and after the first signal we stopped accepting following ones.

ktff: It seams the shutdown tests failing because of this.

Hoverbear: Let me update this.

PR #4033, opened 2020-09-21
fix: Wrong inventory::submit

juchiast:


jszwedko (approved):
Doh, thanks for catching that.

bruceg:
In the future, please use a more specific semantic tag, as we use them to build out the release announcements. This should read `fix(apache_metrics source): Wrong inventory::submit`

PR #4035, opened 2020-09-21
fix(startup): init script / rm output option

emuesuenu:
function daemon does not have an output option, so delete it.

I couldn't find the init.d processor that has the output option, so I just want to delete it.
https://github.com/fedora-sysv/initscripts/blob/master/etc/rc.d/init.d/functions#L239-L272


jamtur01 (approved):
LGTM

PR #4049, opened 2020-09-21
enhancement: AWS EKS Web Identity support

tustvold:
Rebased version of #3468 with some minor tweaks.

Tested in EKS and working as expected

### DIFF
--- a/src/sinks/util/rusoto.rs
+++ b/src/sinks/util/rusoto.rs
@@ -20,7 +20,7 @@ use rusoto_credential::{
     StaticProvider,
 };
 use rusoto_signature::{SignedRequest, SignedRequestPayload};
-use rusoto_sts::{StsAssumeRoleSessionCredentialsProvider, StsClient};
+use rusoto_sts::{StsAssumeRoleSessionCredentialsProvider, StsClient, WebIdentityProvider};
 use snafu::{ResultExt, Snafu};
 use std::{
     fmt, io,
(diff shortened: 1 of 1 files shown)


MOZGIII:
What if we implement a custom [`ProvideAwsCredentials`](https://docs.rs/rusoto_credential/0.45.0/rusoto_credential/trait.ProvideAwsCredentials.html), similar to [`ChainProvider`](https://docs.rs/rusoto_credential/0.45.0/rusoto_credential/struct.ChainProvider.html) that would do what chain does + try web identity?

I'm proposing this because, after thinking about it more, I really like the idea of not having the extra configuration parameter.

It's a shame there are crate relation issues preventing this from being implemented at upstream mentioned at https://github.com/timberio/vector/pull/3468#issuecomment-694722906, but I guess this would be the next best thing.

tustvold:
Sounds good to me, I'll have a look at implementing this, time permitting, in the next couple of days

tustvold:
The graphql failures in the tests don't appear to be related to the changes I've made...

MOZGIII (approved):
LGTM

MOZGIII:
Yes, we have some ongoing problems with those tests, they shouldn't be related.

The code looks good, I'm ready to proceed with the merge if there are no plans to add further changes.

src/sinks/util/rusoto.rs
```diff
@@ -44,9 +44,50 @@ enum RusotoError {
...
+            web_provider: WebIdentityProvider::from_k8s_env(),
+        }
+    }
+
+    pub fn set_timeout(&mut self, duration: Duration) {
+        self.chain_provider.set_timeout(duration);
+    }
+}
+
+impl Default for CustomChainProvider {
```
fanatid, line 67: As I understand this forced by [clippy::new_without_default](https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default) rule, so maybe it worth to remove `CustomChainProvider::new` in favor of `Default` only?

fanatid (approved):

src/sinks/util/rusoto.rs
tustvold: I have no strong opinions either way, although you are completely correct about the origin of the default implementation :sweat_smile: . For what it is worth the rusoto ChainProvider implements both, but if you'd prefer just the one I'm happy to change it

MOZGIII: Let's just copy what the upstream does for this one.

leebenson:
@tustvold I can confirm the GraphQL test failure is a known issue and unrelated to this PR

Issue #4003, opened 2020-09-17
Flakey test on OSX: retry_until_after_timeout


jszwedko:
Failure: https://github.com/timberio/vector/pull/3704/checks?check_run_id=1129767532

```
2020-09-17T17:16:01.2446470Z thread 'main' panicked at 'Timeout', src/test_util/mod.rs:310:5
2020-09-17T17:16:01.2446980Z stack backtrace:
2020-09-17T17:16:01.6290410Z error: test failed, to rerun pass '--lib'
2020-09-17T17:16:01.6290800Z 
2020-09-17T17:16:01.6291020Z Caused by:
2020-09-17T17:16:01.6292180Z   process didn't exit successfully: `/Users/runner/work/vector/vector/target/debug/deps/vector-30d59d8cd75cf082  --nocapture` (signal: 6, SIGABRT: process abort signal)
2020-09-17T17:16:01.6318310Z make: *** [test] Error 101
2020-09-17T17:16:01.6323890Z test test_util::tests::retry_until_after_timeout ... 
2020-09-17T17:16:01.6328800Z ##[error]Process completed with exit code 2.
```
Ref: https://github.com/timberio/vector/issues/2978

leebenson:
Hmm. `test_util::tests::retry_until_after_timeout` is marked with `#[should_panic]`, so this is expected behavior. Wonder why the panic is slipping through?

jszwedko:
Closing since this test doesn't exist anymore.

PR #4050, opened 2020-09-21
chore: Remove retry until after timeout test

jamtur01:
Removes test until #4003 can be fixed.


leebenson (approved):
Removal looks fine for now. It's possible other `#[should_panic]` tests could suffer this same fate on macOS until the cause is rooted out. Tracking in #4056.

Possibly related: https://github.com/rust-lang/rust/issues/74820

PR #4062, opened 2020-09-22
chore: Updated cargo test to add --no-fail-fast flag

jamtur01:
This PR enables the cargo test flag: `--no-fail-fast`.

> --no-fail-fast
 Run all tests regardless of failure. Without this flag, Cargo will exit after the first executable fails. The Rust test harness will run all tests within the executable to completion, this flag only applies to the executable as a whole.

This will allow us to see all failed tests rather than the cargo test run failing after the first failed test.


MOZGIII (approved):

JeanMertz (approved):
Sweet, this will definitely help 👍

Hoverbear (approved):

bruceg (approved):
Makes sense

PR #4068, opened 2020-09-23
enhancement(influxdb_metrics sink): Add a tags configuration options to add user-defined tags

oktal:
Fixes https://github.com/timberio/vector/issues/4057

The Pull Request adds a new `tags` configuration option to the `influxdb_metrics` sink so that the user can specify an additional list of tags to be added to every point sent to InfluxDB


src/sinks/influxdb/metrics.rs
```diff
@@ -144,6 +146,7 @@ impl Service<Vec<Metric>> for InfluxDBSvc {
             self.protocol_version,
             items,
             self.config.namespace.as_deref(),
+            self.config.tags.clone(),
```
Hoverbear, line 149: I *imagine* you can make this work without requiring a clone here. Below, on l180, you do another clone after an `.into_iter()`. I wonder if you can make this a borrow and that inner clone will be enough?

src/sinks/influxdb/metrics.rs
```diff
@@ -796,6 +820,42 @@ mod tests {
         );
         assert_eq!("1542182950000000011", line_protocol.3);
     }
+
+    #[test]
+    fn test_encode_with_some_tags() {
```
Hoverbear, line 825: Can you add `crate::test_util::trace_init()` here? :)

oktal: Should I add it to every test also ?

Hoverbear: Just the ones you add is fine. 😓

Hoverbear (approved):

Hoverbear:
Very happy to merge once we get the `crate::test_util::trace_init()` call added to the test you added! :)

oktal:
That's done @Hoverbear :)

Hoverbear (approved):

PR #4069, opened 2020-09-23
enhancement(buffer): Record a metric when dropping an event

oktal:
The `Buffer` `DropWhenFull` variant will only log a message in debug.

This PR adds a metric so that the user can alert on it and tweak the size of the channel.

Note that this PR is a rough first draft. From what I understand from the code, every sink can have a buffering feature. Ideally, I would like the counter to be tagged on the `component_type`  so that I can easily figure out which part of my pipeline is too slow


src/internal_events/buffer.rs
```diff
@@ -0,0 +1,18 @@
+use crate::internal_events::InternalEvent;
+use metrics::counter;
+
+#[derive(Debug)]
+pub struct BufferEventDropped;
+
+impl InternalEvent for BufferEventDropped {
```
Hoverbear, line 7: Based on our current structure, we have most events per-component. So if you wanted to track say, `GcpBufferEventDropped` that would be its own event. (https://github.com/timberio/vector/blob/5111f3a5b522fa06ca2ddf321e668246a78dc9fa/src/internal_events/elasticsearch.rs#L15)

This means quite some duplication though, so it would  be nice to either find another way, or you could create a small `macro_rules` and use that to do this: https://doc.rust-lang.org/reference/macros-by-example.html

Hoverbear: You may also be able to add some metadata to structs like `DiskBuffer` that contain the component name (`Option<&'static str>`) and have a builder `.with_component_name()` for opt in settings?

src/internal_events/buffer.rs
```diff
@@ -0,0 +1,18 @@
...
+    fn emit_logs(&self) {
+        debug!(
```
Hoverbear, line 9:
```suggestion
        info!(
```

src/internal_events/buffer.rs
```diff
@@ -0,0 +1,18 @@
...
+            message = "Shedding load; dropping event.",
+            rate_limit_secs = 10
```
Hoverbear, line 11: I wonder if we should log the message on `trace` logging?

jamtur01:
Hi @oktal Just wanted to check in on this one?

jamtur01:
hey @oktal just following up here.

oktal:
Hi,

I'll update the PR shortly

jamtur01:
Hi @oktal - just checking in.

oktal:
Hi, I'm closing this PR for now, as we've switched back to blocking.

I'll open a new one if needed in a near feature

PR #4077, opened 2020-09-24
feat(logdna sink): allow inclusion of env field in request and default_env in config

### LINKED ISSUE #4076: Support specifying the environment in the LogDNA sink config
As you can see in the [LogDNA API reference](https://docs.logdna.com/reference), it allows you to include an `env` field that can be used to specify the environment (staging, acceptance, production, etc.) from which an event originated.

Currently, the Vector configuration for the LogDNA sink does not seem to support this, and setting an `env` field to an event manually only includes it in the `meta` object.

Ideally, an `env` field that is set manually should be properly included in the LogDNA request body, and there should probably also be a `default_env` config option for events which do not include such a field, very similar to the current `default_app` option.

Considering that this is a pretty small change I can submit a quick PR for this if so desired.

romatthe:
This PR fixed #4076 by allowing both the specification of a `default_env` option in the LogDNA sink config as well as including `env` fields in the LogDNA request body.


jamtur01 (approved):
LGTM

binarylogic (approved):

src/sinks/logdna.rs
```diff
@@ -125,6 +130,14 @@ impl HttpSink for LogdnaConfig {
             map.insert("file".to_string(), json!(file));
         }
 
+        if !map.contains_key("env") {
+            if let Some(default_env) = &self.default_env {
+                map.insert("env".to_string(), json!(default_env.as_str()));
+            } else {
+                map.insert("env".to_string(), json!("production"));
+            }
```
bruceg, line 138:
```suggestion
            map.insert("env".to_string(), json!(self.default_env.as_ref().map(|s| s.as_str()).unwrap_or("production")));
```

bruceg (approved):
Looks good, just one minor suggestion.

PR #4080, opened 2020-09-24
chore(remap transform): split up functions into separate modules

### LINKED ISSUE #4079: Separate remap functions into their own modules/files
The `functions.rs` file is becoming difficult to read, causes merge conflicts and has all remap function tests grouped together, which makes it harder to know if all potential cases are covered.

We're going to split them up into separate modules, such as `function::to_string` and `function::uuid_v4`.

JeanMertz:
This splits up the `functions` module into one module per function, re-exported at the top-level `function` module. There's also a new `function::prelude` module, which exports the main types used in each function implementation.

ref #3784 
closes #4079


src/mapping/query/function/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+    uuid_v4 => UuidV4Fn,
+    md5 => Md5Fn,
+    sha1 => Sha1Fn,
+    now => NowFn,
+    truncate => TruncateFn,
+    parse_json => ParseJsonFn,
+}
+
+/// Commonly used types when building new functions.
+mod prelude {
```
FungusHumungus, line 89: Minute nitpick - I think this would be better at the top of the file.

JeanMertz: done

FungusHumungus (approved):
Nice!

Hoverbear (approved):

PR #4081, opened 2020-09-24
chore: Add vector to systemd-journal on Debian and RPM

### LINKED ISSUE #3531: Tutorial on https://vector.dev/docs/setup/installation/operating-systems/ubuntu/ failed
The tutorial on:

https://vector.dev/docs/setup/installation/operating-systems/ubuntu/

Here's what went wrong:

`sudo dpkg -i vector-amd64.deb` installs a new systemd unit which executes `/usr/bin/vector` as user `vector`. By default that user has no read access to `journald` in order to read logs like in the installation example. It also cannot access `/var/log/syslog` and other logs for the same reason.

```
$ cat <<-VECTORCFG > /etc/vector/vector.toml
[sources.in]
  type = "journald" # required
...
```

```
$ cat /etc/systemd/system/vector.service
[Service]   
User=vector 
Group=vector
```

```
$ sudo journalctl -u vector
Aug 23 11:02:13 test-xenial-1 vector[9207]: No journal files were opened due to insufficient permissions.
```

jamtur01:
Closes #3531

### DIFF
--- a/distribution/debian/scripts/postinst
+++ b/distribution/debian/scripts/postinst
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -e
 
+usermod -aG systemd-journal vector
+
 if command -v systemctl >/dev/null 2>&1
 then
   systemctl daemon-reload
--- a/distribution/debian/scripts/postinst
+++ b/distribution/debian/scripts/postinst
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e
 
-usermod -aG systemd-journal vector
+usermod -aG systemd-journal vector || true
 
 if command -v systemctl >/dev/null 2>&1
 then
--- a/distribution/rpm/vector.spec
+++ b/distribution/rpm/vector.spec
@@ -45,6 +45,7 @@ getent passwd %{_username} > /dev/null || \
   useradd -r -d %{_sharedstatedir}/%{_name} -g %{_username} -s /sbin/nologin \
   -c "Vector observability data router" %{_username}
 chown %{_username} %{_sharedstatedir}/%{_name}
+usermod -aG systemd-journal %{_username}  || true
 
 %clean
 rm -rf %{buildroot}


Hoverbear (approved):

binarylogic (approved):
Good catch, added https://github.com/timberio/vector-website/issues/105 to make this more clear.

distribution/debian/scripts/postinst
```diff
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -e
 
+usermod -aG systemd-journal vector || true
```
jszwedko, line 4: I prefer the long form of flags in scripts to make it less likely I have to go lookup what a flag does. In this case, that'd be: `usermod --append --groups systemd-journal vector`.

I take it this group doesn't always exist on these systems given the `|| true`? Do we need different permissions for `vector` to be able to use the `journald` source on those OSes?

jamtur01: This isn't user-facing and a short form is consistently used elsewhere in our packaging so I am going to keep it. 

I added the || true in the unlikely event someone tried to install the packages on a platform without systemd (people do force installs etc on older platforms for example).

jszwedko: > This isn't user-facing and a short form is consistently used elsewhere in our packaging so I am going to keep it.

I think we should be preferring long form flags elsewhere in this script (and generally in checked-in scripts). I realize this could be bike shedding a little, but I think there is a concrete practical benefit to preferring long-form flags when they exist: namely that it is less likely that the reader of a script will need to lookup what a short-flag does if they are unfamiliar with it. In this particular example, `--append` and `--groups` are clear enough that, even if I wasn't familiar with `usermod`, I'd probably be able to guess what it does.

I think the short-form flags are useful when used in an interactive terminal to save keystrokes, but for checked-in scripts that are likely to be read often, I think we should prefer legibility over typing speed.

I'd propose that all new shell script code use the long form flags, when they exist, and that we switch to them as we modify existing code.

> I added the || true in the unlikely event someone tried to install the packages on a platform without systemd (people do force installs etc on older platforms for example).

:+1:

PR #4088, opened 2020-09-25
fix(add_fields transform): Emit overwrite logs as debug

### LINKED ISSUE #3803: add_fields transform: Don't error if fields not overwritten when overwrite = false
https://github.com/timberio/vector/blob/439a9cfdb48d976f19a0d2e5b84fc268752d93ba/src/internal_events/add_fields.rs#L71

This should be a `trace` call.

spencergilbert:
Updates logs related to overwriting from the `add_fields` transformer. Since these logs are related to Vector _correctly_ carrying out the configured function they shouldn't be logged as errors.

closes https://github.com/timberio/vector/issues/3803

### DIFF
--- a/src/internal_events/add_fields.rs
+++ b/src/internal_events/add_fields.rs
@@ -57,7 +57,7 @@ pub struct AddFieldsFieldOverwritten<'a> {
 
 impl<'a> InternalEvent for AddFieldsFieldOverwritten<'a> {
     fn emit_logs(&self) {
-        error!(message = "Field overwritten.", %self.field, rate_limit_secs = 30);
+        debug!(message = "Field overwritten.", %self.field, rate_limit_secs = 30);
     }
 }
 
@@ -68,6 +68,6 @@ pub struct AddFieldsFieldNotOverwritten<'a> {
 
 impl<'a> InternalEvent for AddFieldsFieldNotOverwritten<'a> {
     fn emit_logs(&self) {
-        error!(message = "Field not overwritten.", %self.field, rate_limit_secs = 30);
+        debug!(message = "Field not overwritten.", %self.field, rate_limit_secs = 30);
     }
 }


juchiast:
`rename_fields` logs should be changed too, and I think we should add an `overwrite` config there.

spencergilbert:
Do we want that included in this PR or standalone?

juchiast:
Probably standalone.

JeanMertz (approved):
Good catch!

JeanMertz (changes requested):
@spencergilbert I see the [original issue](https://github.com/timberio/vector/issues/3803) asked for these to be emitted at the `trace` level, which I think makes sense, given the volume of messages these can produce.

Could you update `debug!` to `trace!`? Thanks!

spencergilbert:
Sorry @JeanMertz I forgot to mention in the issue/or that I had been pairing with @Hoverbear and we decided debug would be a good level - happy to set it to whatever everyone feels is appropriate though.

JeanMertz (approved):
Ah no, if you and @Hoverbear concluded `debug` was appropriate, then I'm happy with that as well 👍

(since you two paired on this, I'll wait for her to approve and merge the PR)

src/internal_events/add_fields.rs
```diff
@@ -57,7 +57,7 @@ pub struct AddFieldsFieldOverwritten<'a> {
 
 impl<'a> InternalEvent for AddFieldsFieldOverwritten<'a> {
     fn emit_logs(&self) {
-        error!(message = "Field overwritten.", %self.field, rate_limit_secs = 30);
+        debug!(message = "Field overwritten.", %self.field, rate_limit_secs = 30);
```
Hoverbear, line 60: We need to consider 2 possibilities here, each of which is of different importance.

* The field exists, and should have been overwritten. This should be `debug`. (A user would appreciate this information while debugging)
* The field does not exist and should have been overwritten. This should be `trace` (we did not need to take action, but tracing that we did check if we needed to take action is good)

src/internal_events/add_fields.rs
```diff
@@ -68,6 +68,6 @@ pub struct AddFieldsFieldNotOverwritten<'a> {
 
 impl<'a> InternalEvent for AddFieldsFieldNotOverwritten<'a> {
     fn emit_logs(&self) {
-        error!(message = "Field not overwritten.", %self.field, rate_limit_secs = 30);
+        debug!(message = "Field not overwritten.", %self.field, rate_limit_secs = 30);
```
Hoverbear, line 71: (Same as above but the points are different)

We need to consider 2 possibilities here, each of which is of different importance.

 * The field exists, and we didn't overwrite because we shouldn't have overwritten. This should be debug. (A user would appreciate this information while debugging)
 * The field does not exist, and we did write in. This should be trace (we did take an action which mutated the event)

As I mentioned in #3803 (comment), we might need to pass that variable into this event.

Hoverbear (approved):
I'm fine with this as is. :)

PR #4089, opened 2020-09-25
fix(file source): more robust handling of split reads

### LINKED ISSUE #2992: File source have problems with reading files under high load
Hi,
I'm trying vector 0.9.1 in docker (image - timberio/vector:0.9.1-debian)
My docker-compose has 3 services - nginx, vector and kafka.

vector has such configuration
```
[sources.some_log]
  # General
  type = "file" # required
  include = ["/opt/logs/some.log"] # required
  start_at_beginning = true # optional, default
  host_key = "host"
  data_dir = "/tmp"

  # Priority
  oldest_first = true # optional, default
  max_read_bytes = 1048576 #1mb

[transforms.parse_json]
  type = "json_parser" # required
  inputs = ["hit_log"] # required
  drop_field = true # optional, default
  drop_invalid = true # required
  field = "message" # optional, default
  target_field = "src"

[transforms.remove_file]
  type = "remove_fields" # required
  inputs = ["parse_json"] # required
  fields = ["source_type"] # required

[sinks.hit_output_kafka]
  # General
  type = "kafka" # required
  inputs = ["remove_file"] # required
  bootstrap_servers = "kafka:9092" # required
  compression = "none" # optional, default
  key_field = "user_id" # required
  topic = "hits" # required

```

I do benchmark with calling 20-30k qps to nginx. Nginx log every request as json. 
And for 2kk requests made in benchmark I see 2kk lines in nginx log and 1999998 messages in kafka.
When I turn on debug logging in vector I see few json parsing errors. 

I`ve tried two more tests:
1. Turn off vector while nginx is being under benchmark load. And after bench is finished - I see no errors and no messages lost.
2. Restart vector with empty state after few messages were lost. For example - under load I have made 2kk messages and got 1999998 message in kafka. I restart vector with empty source offsets (by defining new data_dir location in vector.toml). And in such case all messages are processed. I see 3999998 messages in kafka (1999998 for first run and 2000000 for second).

I`ve done some research and looks like found cause of such behaviour:
https://github.com/timberio/vector/blob/master/lib/file-source/src/file_watcher.rs#L257-L261

Look like under high load another process or kernel didn't flushed buffers to disk in 1ms interval. 
As for me vector should not threat EOF as line delimiter, or such behavior should be configurable.

lukesteensen:
Fixes #2992 

This is an initial minimal-ish fix for the file source being too willing to return lines that did not end in a newline. Our original fix in #1236 was to wait up to a millisecond for the rest of a line to show up before considering it complete and returning it anyway. It turns out that in high-throughput situations, it's not uncommon for partial log lines to be visible for periods of time longer than this. This causes Vector to return a single logical line split across multiple events, which obviously breaks many kinds of downstream processing.

The problem with the sleep approach is that we can't know how long we need to sleep to see the rest of the line, and any amount of time we sleep is time we're not doing useful work, potentially lowering our overall throughput. This PR eliminates the sleep, instead opting to make a single logical line read operation able to be suspended and resumed across calls. The core of that change is to switch from a single line buffer to one that's specific to each file watcher. This means the intermediate state won't get overwritten between calls and we're safe to return control to the main loop.

This opens up a couple of edge cases that will require a larger change to handle, but they generally seem very unlikely to cause issues. Mostly, we are less likely to return legitimately interrupted writes as their own event. These are likely _extremely_ uncommon, and will now be prepended to the following line, just as they appear if you read the file normally. If there is no following line, we won't flush the partial message until the file is deleted, which may not happen. If Vector is shut down before it sees another newline, it doesn't have a mechanism to flush this intermediate state.

There's some further refactoring here that should happen, but I cut it short in favor of getting the fix out sooner:

1. We should read directly into the `BytesMut` instead of double buffering via `BufReader`
2. `read_until_with_max_size` should be inlined into `read_line` because the boundary doesn't really make sense anymore

Both these refactoring and plugging any edge cases can happen as part of the larger file source cleanup that should be coming up soon.

### DIFF
--- a/lib/file-source/src/file_server.rs
+++ b/lib/file-source/src/file_server.rs
@@ -71,7 +71,6 @@ where
         C: Sink<(Bytes, String)> + Unpin,
         <C as Sink<(Bytes, String)>>::Error: std::error::Error,
     {
-        let mut line_buffer = Vec::new();
         let mut fingerprint_buffer = Vec::new();
 
         let mut fp_map: IndexMap<FileFingerprint, FileWatcher> = Default::default();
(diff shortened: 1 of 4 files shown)


lib/file-source/src/file_watcher.rs
```diff
@@ -144,24 +149,32 @@ impl FileWatcher {
     /// This function will attempt to read a new line from its file, blocking,
     /// up to some maximum but unspecified amount of time. `read_line` will open
```
ktff, line 150: Does this function still block?

ktff (approved):
Seams great, and the shutdown tests are failing because of other PR, so they can be ignored for this one.

jamtur01:
+1.

lib/file-source/src/file_server.rs
```diff
@@ -204,52 +203,53 @@ where
...
-                            }
-                        }
-
+                while let Ok(Some(line)) = watcher.read_line() {
+                    if line.is_empty() {
                         break;
                     }
+
+                    let sz = line.len();
+                    trace!(
```
Hoverbear, line 212: This could shift into an internal event perhaps?

Hoverbear (approved):

lib/file-source/src/file_watcher.rs
lukesteensen: It still does multiple blocking filesystems reads, so yes. This comment predates the addition of the sleep.

lib/file-source/src/file_server.rs
lukesteensen: The way we do events here is a little different since it's its own crate, but not a bad idea. I'll make a note to do that in the next set of changes.

mikhno-s:
Can somebody trigger container building the version with this fix, please? The last one in docker hub dated 5 days ago (24 sep)
Thanks

jamtur01:
@mikhno-s Apologies! We're working on an issue with the nightly builds. Hopefully, it will be fixed tomorrow.

PR #4095, opened 2020-09-25
fix: Fix tokio signal behavior to handle a loop of signals

Hoverbear:
Fix the fix of #4025 on Unix so it handles signals in a look as well. Thanks @ktff for finding this.


jszwedko (approved):
Good catch and fix!

ktff (approved):

PR #4101, opened 2020-09-25
feat(new source): Firehose source with CloudWatch Logs subscription transform

### LINKED ISSUE #3566: New `aws_kinesis_firehose` source to read forwarded AWS CloudWatch Logs
This issue represents the introduction of a new source that can be used as the HTTP endpoint for an [AWS Kinesis Firehose destination](https://docs.aws.amazon.com/firehose/latest/dev/create-destination.html#create-destination-http).

This source will specifically be designed to read forwarded AWS CloudWatch Logs first ([format](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html), but could be expanded to other Firehose sources in the future.

I recommend a spike into this to discover and answer open questions and, possibly, an RFC if there appear to be more decisions to be made.

Other work that will probably need to happen as part of this:

* Creating a guide on vector.dev for setting this up
* Streamlining the AWS configuration via cloudformation, terraform module, or something else

jszwedko:
This PR represents the initial implementation of the the Firehose source.

It also introduces an `aws_cloudwatch_logs_subscription_parser` transform to handle the transformation of CloudWatch Log events. I opted not to split this off into a separate PR because I think its existence is relevant to the overall discussion of this source implementation as a goal is to facilitate getting CloudWatch Logs to vector.

Closes #3566

I tried to keep the commits relatively atomic (aside from the first one) so it may be useful to review them in order. Otherwise I'd recommend reviewing the the source first and then the transform.

This source ingests arbitrary data forwarded by AWS Kinesis Firehose via their [HTTP forwarder](https://docs.aws.amazon.com/firehose/latest/dev/create-destination.html#create-destination-http).

TODO:
- [ ] Deploy vector somewhere and verify it works end-to-end. I currently simulated this by capturing and replaying Firehose requests to a local vector, but I'd still like to check.
- [ ] Write a guide specifically for ingesting CloudWatch Logs (and maybe a cloudformation template). I'd like to split this out into a separate issue once this PR has been reviewed and we are satisfied with the approach.

Some design decisions:

I opted to implement the Firehose source using `warp` directly rather than using the `HttpSource` trait primarily because of Firehose's requirement of a [specific HTTP response structure](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html). I modeled it pretty closely after warp's [todo example](https://github.com/seanmonstar/warp/blob/master/examples/todos.rs) which I think resulted in pretty good modeling.

I was initially going to make parsing of CloudWatch Log subscription events part of the Firehose source (implementing it as a `codec`), but, after starting in on this, it became clear to me that Firehose is capable of forwarding arbitrary payloads, which makes it more like the `kafka` source. Up until now, any decoding and parsing in `vector` has been delegated downstream to transforms so, instead, I introduced an `aws_cloudwatch_logs_subscription_parser` that is able to parse out the [JSON structure of subscription events](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html#LambdaFunctionExample). An advantage of this approach is that this transform could be used in other circumstances where `vector` is ingesting CloudWatch Logs subscription events.

Specifically for the scenario where a user wants to ingest CloudWatch Logs via Firehose, the steps would look look like:

1. Deploy `vector` somewhere using the added source. Note there is a caveat that the endpoint be exposed as HTTPS. The configuration would look something like:

```toml
[sources.firehose]
  # General
  type = "aws_kinesis_firehose"
  address = "127.0.0.1:9000"
  access_key = "secret"

[transforms.cloudwatch]
  type = "aws_cloudwatch_logs_subscription_parser"
  inputs = ["firehose"]

[transforms.json]
  type = "json_parser"
  inputs = ["cloudwatch"]

[sinks.console]
  type = "console"
  inputs = ["json"]
  encoding.codec = "json"
```

I think this is good fodder for thinking about what configuration macros in vector might look like.

2. Run through setting up a Firehose delivery stream and CloudWatch Logs subscription based on our guide. AWS has [a guide](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html#FirehoseExample) that we can loosely follow; instead of forwarding to S3, they'd forward via HTTP to their deployed vector instance.


src/event/value.rs
```diff
@@ -89,9 +89,12 @@ impl From<BTreeMap<String, Value>> for Value {
     }
 }
 
-impl From<Vec<Value>> for Value {
-    fn from(value: Vec<Value>) -> Self {
-        Value::Array(value)
+impl<T> From<Vec<T>> for Value
```
jszwedko, line 92: I'm curious to get feedback on this change. I added it to make it easier to create a `Value` from a `Vec<String>`, but it ended up having the side-effect of also implementing it for `Vec<Value>` and, I think?, will result in extra allocations for that specific case where `into()` really isn't needed.

src/sources/aws_kinesis_firehose.rs
```diff
@@ -0,0 +1,714 @@
+use crate::{
```
jszwedko, line 1: Make sure to expand this file :smile:

src/sources/aws_kinesis_firehose.rs
```diff
@@ -0,0 +1,714 @@
...
+
+        if let Some(e) = err.find::<RequestError>() {
+            message = format!("{}", e);
+            code = e.status();
+            request_id = e.request_id();
+        } else if let Some(e) = err.find::<warp::reject::MissingHeader>() {
+            code = StatusCode::BAD_REQUEST;
+            message = format!("Required header missing: {}", e.name());
+            request_id = None;
+        } else {
```
jszwedko, line 214: I'm not terribly happy about catching the other errors here and marshaling as their debug format. The intention is to transform those errors into the expected JSON response structure, but there doesn't appear to be a way to generically handle them: https://github.com/seanmonstar/warp/issues/451. Instead, I tried to just catch the common errors and let the rest fall through to here.

Theoretically AWS wouldn't be sending requests that would fail for any reason other than the access key not matching, but I could see users hitting the header errors if they are just testing out the source via 'curl' or the like.

src/sources/aws_kinesis_firehose.rs
```diff
@@ -0,0 +1,714 @@
...
+            Ok(events) => {
+                let request_id = request_id.clone();
+                out.send_all(futures01::stream::iter_ok(events))
+                    .compat()
+                    .map_err(|err| {
+                        let err = RequestError::ShuttingDown {
+                            request_id: request_id.clone(),
+                            source: err,
+                        };
+                        // can only fail if receiving end disconnected, so we are shutting down,
```
jszwedko, line 268: I copied this from the `HttpSource` trait.

src/test_util/mod.rs
```diff
@@ -50,6 +50,20 @@ macro_rules! assert_downcast_matches {
     }};
 }
 
+#[macro_export]
+macro_rules! log_event {
```
jszwedko, line 54: Also curious to get feedback on this addition of this helper macro. I found it useful in tests to easily construct events to assert against.

src/transforms/aws_cloudwatch_logs_subscription_parser.rs
```diff
@@ -0,0 +1,260 @@
...
+            &log_schema().message_key()
+        };
+        AwsCloudwatchLogsSubscriptionParser {
+            field: field.clone(),
+        }
+    }
+}
+
+impl Transform for AwsCloudwatchLogsSubscriptionParser {
+    fn transform(&mut self, _event: Event) -> Option<Event> {
```
jszwedko, line 68: This one is weird. This transform needs to create multiple events from a single input event, which only seems possible through `transform_into`, but I still have to implement `transform` even though it'll never be called.

cc/ @lukesteensen as this might be relevant to transforms part of the architecture revisit https://github.com/timberio/vector/blob/master/rfcs/2020-06-18-2625-architecture-revisit.md#transforms

jszwedko:
cc/ @binarylogic to validate the approach.

jszwedko:
@fanatid I tagged you for review since this is similar to the Splunk HEC source, but let me know if you'd prefer someone else take a look.

src/event/value.rs
JeanMertz: https://github.com/timberio/vector/pull/4093#discussion_r494991793

😄

jszwedko: Aha! :smile: Thanks for digging into the assembly. I'm curious what @bruceg thinks of it.

src/test_util/mod.rs
lukesteensen: Seems like a reasonable addition to me! I know @Hoverbear has thoughts around event usability as well.

src/transforms/aws_cloudwatch_logs_subscription_parser.rs
```diff
@@ -0,0 +1,260 @@
...
+
+impl From<AwsCloudwatchLogsSubscriptionParserConfig> for AwsCloudwatchLogsSubscriptionParser {
+    fn from(
+        config: AwsCloudwatchLogsSubscriptionParserConfig,
+    ) -> AwsCloudwatchLogsSubscriptionParser {
+        let field = if let Some(field) = &config.field {
+            field
+        } else {
+            &log_schema().message_key()
+        };
```
lukesteensen, line 60: This would be a good use for `Option::or_else`.

src/transforms/aws_cloudwatch_logs_subscription_parser.rs
lukesteensen: Yes, this is something I'd like to clean up by having more specific types of transforms. For now, this is the right workaround.

lukesteensen (approved):
Looks good to me! Left a few small notes but nothing blocking.

binarylogic:
> It also introduces an aws_cloudwatch_logs_subscription_parser transform to handle the transformation of CloudWatch Log events.

@jszwedko there's no indicatation, at all, that the event came from CloudWatch? I would much prefer to autodetect this format and parse it in the source if we can. In general, we're moving that way when the format of the data is part of the protocol, which in this case it feels like it is.

JeanMertz:
> In general, we're moving that way when the format of the data is part of the protocol, which in this case it feels like it is.

Is it, though? Firehose can be used for many shapes of data, and simultaneously CloudWatch logs subscription data could be ingested via other sources, for example if the customer stores their data on S3 and then later reads it through the S3 source.

Having said that, I do feel like our encoding/decoding story has been somewhat in flux the last couple of months/weeks, with many ideas on how to decouple that from sources/sinks and introduce some top-level configuration option to handle the task.

Because of that, it might be worth it not to expose this transform just yet, but keep it as an internal implementation detail that is directly called by the Firehose source for now (assuming we can, as @binarylogic mentioned, detect that we're dealing with CW logs)?

binarylogic:
> Is it, though? Firehose can be used for many shapes of data, and simultaneously CloudWatch logs subscription data could be ingested via other sources, for example if the customer stores their data on S3 and then later reads it through the S3 source.

That's a good point, and we can still offer a transform, but we should still auto-detect the type if it's within the scope of the source. In my opinion it's like saying the `http` source shouldn't decode `Content-Type: application/json` because we have a `json_parser` transform.

jszwedko:
> > It also introduces an aws_cloudwatch_logs_subscription_parser transform to handle the transformation of CloudWatch Log events.
> 
> @jszwedko there's no indicatation, at all, that the event came from CloudWatch? I would much prefer to autodetect this format and parse it in the source if we can. In general, we're moving that way when the format of the data is part of the protocol, which in this case it feels like it is.

@binarylogic It seems like Kinesis Firehose is very much an extension of Kinesis in that it just ships arbitrary bytes around and relies on the consumer/producer to agree on what those bytes mean. You can see an example request from Firehose here: https://gist.github.com/jszwedko/46bf9338c737cd3b1ac5f5cd39c48daa (the request is encoded as JSON to show headers). However, users can set custom key/values that will be passed along with Firehose requests in a `X-Amz-Firehose-Common-Attributes` header so we could do something like instruct users to set a custom `Content-Type` field to `application/cloudwatch-logs-subscription+json` and then have special handling in `vector`. If we do want to do that, I'd prefer to punt and do it as a follow-up to not block this and so we can think about and discuss it a bit more.

I think the configuration macro idea could help here too.

src/event/value.rs
jszwedko: This actually got merged in in another PR (#4066).

src/sources/aws_kinesis_firehose.rs
```diff
@@ -0,0 +1,714 @@
...
+    fn source_type(&self) -> &'static str {
+        "aws_kinesis_firehose"
+    }
+}
+
+inventory::submit! {
+    SinkDescription::new_without_default::<AwsKinesisFirehoseConfig>("aws_kinesis_firehose")
+}
+
+mod filters {
```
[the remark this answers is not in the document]
jszwedko: Updated!

src/sources/aws_kinesis_firehose/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+inventory::submit! {
+    SinkDescription::new_without_default::<AwsKinesisFirehoseConfig>("aws_kinesis_firehose")
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::shutdown::ShutdownSignal;
```
JeanMertz, line 79: nit: looks like you're re-importing things already imported using `super::*`

src/sources/aws_kinesis_firehose/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+    use crate::{
+        config::{GlobalOptions, SourceConfig},
+        event::{Event, LogEvent},
+        log_event,
+        test_util::{collect_ready, next_addr, wait_for_tcp},
+        Pipeline,
+    };
+    use chrono::{DateTime, SubsecRound, Utc};
+    use flate2::{read::GzEncoder, Compression};
+    use futures::compat::Future01CompatExt;
```
JeanMertz, line 89: Not sure if they are accurate, but seems there are some lints still failing or giving warnings that would be good to fix?

src/transforms/aws_cloudwatch_logs_subscription_parser.rs
```diff
@@ -0,0 +1,259 @@
...
+use chrono::{serde::ts_milliseconds, DateTime, Utc};
+use serde::{Deserialize, Serialize};
+use std::iter;
+use string_cache::DefaultAtom as Atom;
+
+#[derive(Deserialize, Serialize, Debug, Clone, Derivative)]
+#[serde(deny_unknown_fields, default)]
+#[derivative(Default)]
+pub struct AwsCloudwatchLogsSubscriptionParserConfig {
+    pub field: Option<Atom>,
```
JeanMertz, line 19: I was believing that we're moving away from using `Atom`, might be good to avoid introducing it in new components if that's the case.

src/transforms/aws_cloudwatch_logs_subscription_parser.rs
```diff
@@ -0,0 +1,259 @@
...
+    fn transform_into(&mut self, output: &mut Vec<Event>, event: Event) {
+        let log = event.as_log();
+
+        emit!(AwsCloudwatchLogsSubscriptionParserEventProcessed);
+
+        let message = log
+            .get(&self.field)
+            .map(|s| s.as_bytes())
+            .and_then(|to_parse| {
+                serde_json::from_slice::<AwsCloudWatchLogsSubscriptionMessage>(to_parse.as_ref())
```
JeanMertz, line 81: nit:

```suggestion
                serde_json::from_slice::<AwsCloudWatchLogsSubscriptionMessage>(&to_parse)
```

unless there's a reason this won't work?

JeanMertz (approved):
This looks great. The implementation itself looks good to me. I left a few nits w.r.t. code formatting/conventions, but nothing blocking.

src/sources/aws_kinesis_firehose/errors.rs
```diff
@@ -0,0 +1,87 @@
+use crate::event::Event;
+use snafu::Snafu;
+use warp::http::StatusCode;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub")]
+pub enum RequestError {
+    #[snafu(display("X-Amz-Firehose-Access-Key required for request: {}", request_id))]
```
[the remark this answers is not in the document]
jszwedko: This particular error indicates that the incoming HTTP request did not have the access key (passed as `X-Amz-Firehose-Access-Key`) rather than that `vector` didn't have the access key configured. I can see what you are saying about it being an implementation detail though. I'll update the language (though I may still mention `X-Amz-Firehose-Access-Key` for end-user debugging purposes).

src/sources/aws_kinesis_firehose/filters.rs
```diff
@@ -0,0 +1,157 @@
...
+    Pipeline,
+};
+use bytes::{buf::BufExt, Bytes};
+use chrono::Utc;
+use flate2::read::GzDecoder;
+use snafu::ResultExt;
+use std::convert::Infallible;
+use std::io;
+use warp::http::StatusCode;
+use warp::Filter;
```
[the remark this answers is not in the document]
jszwedko: :+1: from these suggestions, I take it that you like to group anything that is under the same top-level module?

src/sources/aws_kinesis_firehose/filters.rs
```diff
@@ -0,0 +1,157 @@
...
+    out: Pipeline,
+) -> impl Filter<Extract = impl warp::Reply, Error = Infallible> + Clone {
+    warp::post()
+        .and(emit_received())
+        .and(authenticate(access_key))
+        .and(warp::header("X-Amz-Firehose-Request-Id"))
+        .and(warp::header("X-Amz-Firehose-Source-Arn"))
+        .and(
+            warp::header("X-Amz-Firehose-Protocol-Version")
+                .and_then(|version: String| async move {
```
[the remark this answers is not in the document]
jszwedko: `warp::header` is capable of extracting anything that implements `FromStr` so, from what I can tell, it needs to know what you want. I could rewrite this as `warp::header::<String>("X-Amz-Firehose-Protocol-Version")` though if you think that is preferable.

https://docs.rs/warp/0.1.0/warp/filters/header/fn.header.html

src/sources/aws_kinesis_firehose/filters.rs
```diff
@@ -0,0 +1,157 @@
...
+///
+/// Firehose can be configured to gzip compress messages so we handle this here
+fn parse_body() -> impl Filter<Extract = (FirehoseRequest,), Error = warp::reject::Rejection> + Clone
+{
+    warp::any()
+        .and(warp::header::optional::<String>("Content-Encoding"))
+        .and(warp::header("X-Amz-Firehose-Request-Id"))
+        .and(warp::body::bytes())
+        .and_then(
+            |encoding: Option<String>, request_id: String, body: Bytes| async move {
```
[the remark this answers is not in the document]
jszwedko: Same note as https://github.com/timberio/vector/pull/4101#discussion_r501039664 . I think either these annotations are needed, or I need to provide the type parameters to `warp::header` and the like. I'm happy either way though depending on your preference.

src/sources/aws_kinesis_firehose/filters.rs
```diff
@@ -0,0 +1,157 @@
...
+) -> impl Filter<Extract = impl warp::Reply, Error = Infallible> + Clone {
+    warp::post()
+        .and(emit_received())
+        .and(authenticate(access_key))
+        .and(warp::header("X-Amz-Firehose-Request-Id"))
+        .and(warp::header("X-Amz-Firehose-Source-Arn"))
+        .and(
+            warp::header("X-Amz-Firehose-Protocol-Version")
+                .and_then(|version: String| async move {
+                    match version.as_str() {
```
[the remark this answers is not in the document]
jszwedko: This didn't quite work as it gives me a `&String` when I want to compare it to a `&str`. Is there something I'm missing?

src/sources/aws_kinesis_firehose/filters.rs
```diff
@@ -0,0 +1,157 @@
...
+        .and_then(move |request_id: String, access_key: Option<String>| {
+            let configured_access_key = configured_access_key.clone();
+            async move {
+                match (access_key, configured_access_key) {
+                    (_, None) => Ok(()),
+                    (Some(configured_access_key), Some(access_key))
+                        if configured_access_key == access_key =>
+                    {
+                        Ok(())
+                    }
```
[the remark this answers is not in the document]
jszwedko: :smile: yeah, `rustfmt` seems to want it like that. I tried without the `{}`s, but it formatted to back that way, I believe due to the line length.

src/sources/aws_kinesis_firehose/mod.rs
jszwedko: Oops :sweat_smile:

src/transforms/aws_cloudwatch_logs_subscription_parser.rs
jszwedko: Looking at the PR for that it looks like `Atom` is still being used for component configuration, just not internally in the `Event`. Example: https://github.com/timberio/vector/pull/4255/files#diff-25a17f1694d8da276ec6ca5083fb832c . If that is the case, I guess this should stay as-is?

jszwedko:
Just noting that I'll need to update the docs to use the new cue setup.

jszwedko:
@JeanMertz thanks for all of the feedback! I believe I've either addressed all of it ("resolved the conversation") or left a comment or question.

src/sources/aws_kinesis_firehose/filters.rs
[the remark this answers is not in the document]
fanatid: This not works because the compiler can not apply [Deref](https://doc.rust-lang.org/std/string/struct.String.html#deref) in `match`. So [String::as_str](https://doc.rust-lang.org/std/string/struct.String.html#method.as_str) need to be used or slice with reference `&version[..]`, but `as_str` looks better :slightly_smiling_face:

fanatid (approved):

src/sources/aws_kinesis_firehose/filters.rs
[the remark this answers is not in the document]
JeanMertz: > This not works because the compiler can not apply [Deref](https://doc.rust-lang.org/std/string/struct.String.html#deref) in `match`

Thanks @fanatid, do you have a reference that mentions Deref not working in `match`? I didn't expect that not to work, so that's an interesting titbit to learn.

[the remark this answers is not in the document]
fanatid: I do not have reference where `match` mentioned with `Deref`, it's my assumption from the documentation on [Deref](https://doc.rust-lang.org/std/ops/trait.Deref.html) page.

> If T implements Deref<Target = U>, and x is a value of type T, then:
>    Values of type &T are coerced to values of type &U

From this, if the compiler can calculate output type `Deref` will work, example: `let str_type: &str = &string_type;` or when we call function like `bar(&string_type)` with signature `bar(_: &str)` -- in both cases output type can be calculated. But looks like type calculation not works for `match` in our case (I guess because we have `_` in arm).

jszwedko:
Opened issue for unrelated test failure: https://github.com/timberio/vector/issues/4456

PR #4176, opened 2020-09-28
chore(tests)!: Use Helm chart at the K8s E2E tests

### LINKED ISSUE #3097: Kubernetes: test Helm at the E2E tests
This is self-explanatory - we have a need to test Helm Chart just like the kubectl manifests in the E2E tests to ensure the Helm Chart operates properly.

This work is, however, blocked by #2702 and #2871 - until those both are merged it would be extremely difficult to maintain the base branch for the changes. I'll just wait until we move forward.

MOZGIII:
This PR switches the E2E tests to using Helm chart instead of YAML config for greater configurability.

It also changes some of the Helm chart defaults, hence the `!`:
- moves master node toleration to the chart default, previously it was a YAML config-specific setting

Closes #3097.

It also unlocks writing more advanced tests, as now we can pass Helm values instead of just custom resources. This gives more possibilities to tweak the deployed Helm chart, and thus, to test more code paths.

### DIFF
--- a/scripts/deploy-kubernetes-test.sh
+++ b/scripts/deploy-kubernetes-test.sh
@@ -58,6 +58,12 @@ up() {
 
   HELM_VALUES=()
 
+  HELM_VALUES+=(
+    # Set a reasonable log level to avoid issues with internal logs
+    # overwriting console output.
+    --set "env[0].name=LOG,env[0].value=info"
+  )
+
   if [[ -n "$CUSTOM_HELM_VALUES_FILE" ]]; then
     HELM_VALUES+=(
       --values "$CUSTOM_HELM_VALUES_FILE"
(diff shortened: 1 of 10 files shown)


scripts/deploy-kubernetes-test.sh
```diff
@@ -31,16 +31,19 @@ NAMESPACE="${2:?"Specify the namespace as the second argument"}"
...
 
-# TODO: replace with `helm template | kubectl apply -f -` when Helm Chart is
-# available.
-
-templated-config() {
-  cat < "distribution/kubernetes/vector.yaml" \
-    | sed "s|^    namespace: vector|    namespace: $NAMESPACE|"
+split-container-image() {
+  local INPUT="$1"
+  CONTAINER_IMAGE_REPOSITORY="${INPUT%:*}"
```
Hoverbear, line 45: Yess I love all this bash substitution use!!!

Hoverbear (approved):

ktff (approved):

PR #4187, opened 2020-09-29
fix(tests): Fix uptime_metrics GraphQL/API tests

### LINKED ISSUE #4175: Fix flaky `uptime_metrics` tests for macOS
https://github.com/timberio/vector/pull/3778#issuecomment-698622017 is still occasionally an issue in macOS CI tests. The interval fix I suggested in that thread seems to have fixed it locally, so I think this probably hints at it being a race condition of some kind.

leebenson:
Closes #4175 (hopefully.)

Test with:

```bash
RUST_TEST_THREADS=X cargo test --test api -- --nocapture
```

Where 'X' is the number of threads to test. CI was balking at `1`; my local machine panicked with `2`. If both pass, it probably works.


sghall (approved):
Looks good @leebenson 👍   Didn't get any failures locally with this.

PR #4191, opened 2020-09-29
enhancement(observability): Initial GraphQL topology

### LINKED ISSUE #4172: GraphQL schema for topology
Prefacing #3211, the initial GraphQL schema to represent topology needs designing for. This should represent sources, transforms and sinks, and be exposed in the following ways:

### Direct queries

Via `{ sources }` and the equivalent for transforms/sinks. This returns 'flat' concrete types, filtered for each resource type. A use-case is when we want to return specific parts of the configuration.

### Flat queries

Via `{ topology }`, an interface for sources, transforms and sinks. This provides an unordered 'flat' hierarchy that is simple to traverse in clients. A use-case is when we want to return _all_ config resources in a single pass, and arrange them in the client.

### Graph relationships

For each resource type, related data should be exposed through individual fields.

For example, a source may be defined as (partially):

```gql
type Source {
  transforms: [Transform!]!
  sinks: [Sink!]!
}
```

This enables traversal where both the resource and the query depth is known. A use-case is querying for a particular resource, and its direct relationships to known, type-specific resources.

leebenson:
This PR adds initial observability for `Config` inside of GraphQL, a proxy for the currently active topology.

Closes #4172 

`Config` is used in two places:

1. `topology::RunningTopology`, as before. I added a `.config()` method returning `&Config`, to borrow for reading config, since topology is typically in scope wherever the API server is being started/updated.

2. In `api::schema::topology::update_config`. This is exposed publically via `Server.update_config()`, on the running server instance. The config is borrowed and the perinent topology data is extracted into a static ref wrapped as `Arc<RwLock<Vec<Source>>>`, where `Source` is the GraphQL wrapper.

This represents the pre-allocated GraphQL format required to transmit to clients. As queries come in, the underlying data structs are cloned to provide a unique copy to each reader.

A new sources query can be executed via playground at http://localhost:8686/playground (by default), which currently returns a sample `Source` type containing a couple of the fields derived from `dyn SourceConfig`:

```gql
{
  sources {
    name
    outputType
  }
}
```

Returning:

Screenshot 2020-10-01 at 20 16 51

From:

```toml
[api]
  enabled = true

[sources.generator]
  type = "generator"
  batch_interval = 1

[sinks.my_sink_id]
  # General
  type = "console"
  inputs = ["generator"]
  target = "stdout"

  # Encoding
  encoding.codec = "json"
```

I've broken out transforms/sinks into a separate PR to keep this relatively light.


leebenson:
Updated the original PR commentary to reflect the updated approach, and taken this out of draft. Ready to review.

src/api/schema/topology.rs
```diff
@@ -0,0 +1,65 @@
...
+impl Source {
+    /// Source name
+    async fn name(&self) -> String {
+        self.0.name.clone()
+    }
+
+    /// The output type given by the source
+    async fn output_type(&self) -> SourceOutputType {
+        match self.0.output_type {
+            DataType::Metric => SourceOutputType::Metric,
```
Hoverbear, line 36: You can do `impl Into<SourceOutputType> for DataType {}` to get a similar, but more idiomatic API. :)

src/api/schema/topology.rs
```diff
@@ -0,0 +1,65 @@
+use crate::config::{Config, DataType};
+use async_graphql::{Enum, Object};
+use lazy_static::lazy_static;
+use std::sync::{Arc, RwLock};
+
+#[Enum]
+pub enum SourceOutputType {
```
Hoverbear, line 7: Is it the intention that later on our core types such as `Value` will be graphql types? That'd be sweet!

Hoverbear (approved):

src/api/schema/topology.rs
leebenson: Do you mean the TOML `Value` types?

I'm using `MetricValue` in https://github.com/timberio/vector/blob/master/src/api/schema/metrics.rs#L18-L24, which gives you a rough idea how we can 'overlay' existing types and make them 'GraphQL types'.

My intention is to support the full range of config options and have them represented in GraphQL, so I think this will wind up wrapping most of those existing types.

src/api/schema/topology.rs
leebenson: Thanks! I actually used `From<T>` [here](https://github.com/timberio/vector/blob/master/src/api/schema/metrics.rs#L51-L55), and then proceeded to forget that was an option :-) I used `From<T>` to get `Into<T>` for free, although perhaps that happens the other way round too?

Async-graphql has a new [Enum derive](https://async-graphql.github.io/async-graphql/en/define_enum.html), which implements `Into<T>` by specifying an existing enum to map. Unfortunately, it's only in the newer version, which requires a 1.46 toolchain since some related functionality is unstable in 1.44, so we're blocked from upgrading.

In the meantime, this might be a good candidate for an macro to forego the manual matching of two disparate but otherwise identical enums. Or just a good excuse for me to explore macros a bit more!

leebenson:
Failing tests here look like they'll be fixed by https://github.com/timberio/vector/pull/4259

lukesteensen (approved):

PR #4194, opened 2020-09-29
chore(architecture): unify build methods

### LINKED ISSUE #3161: Unify component build methods
Currently, we have both `fn build` and `async fn build_async` for each of sources, transforms, and sinks. This was a minimally invasive way to achieve the needed functionality, but the resulting API is not the clearest. Once the dust has settled a bit, we should circle back and clean this up.

For context, the existing `build` method is used into two place. First, the default implementation of `build_async` simply delegates to it. Second, and more involved, is the various component tests that invoke it. Since the tests are the vast majority of the code, we'll focus mostly on those.

There are (at least) two ways we could go about this. The simplest would be to remove `fn build` and use `async fn build_async` (renamed to `build`) everywhere. This is probably preferred due to the simplicity, but it does have the downside of requiring async stuff where it may not be needed (e.g. almost no transforms need it). Much of this downside could be alleviated by `#[tokio::test]` when we get to the point where that is usable and we don't require `tokio-compat`.

The other route would be to split out a better-named `Sync{Source,Transform,Sink}Config` trait and have the `fn build` live there. This would get rid of the confusion around which method to implement but complicates our model a bit by introducing a new type of thing. On the upside, we could easily `impl SourceConfig for SyncSourceConfig` and avoid having to change all the tests.

fanatid:
Closes #3161


fanatid:
Looks like `clippy` have false-positive errors for `async_trait` :disappointed:

benches/bench.rs
```diff
@@ -426,13 +426,19 @@ fn benchmark_regex(c: &mut Criterion) {
...
             b.iter_with_setup(
                 || {
-                    let parser =transforms::regex_parser::RegexParserConfig {
-                        // Many captures to stress the regex parser
-                        patterns: vec![r#"^(?P<addr>\d+\.\d+\.\d+\.\d+) (?P<user>\S+) (?P<auth>\S+) \[(?P<date>\d+/[A-Za-z]+/\d+:\d+:\d+:\d+ [+-]\d{4})\] "(?P<method>[A-Z]+) (?P<uri>[^"]+) HTTP/\d\.\d" (?P<code>\d+) (?P<size>\d+) "(?P<referrer>[^"]+)" "(?P<browser>[^"]+)""#.into()],
-                        field: None,
-                        drop_failed: true,
-                        ..Default::default()
-                    }.build(TransformContext::new_test()).unwrap();
+                    let mut rt = runtime();
```
ktff, line 429: Runtime creation should probably be outside of benchmark loop.

benches/bench.rs
```diff
@@ -835,18 +841,22 @@ fn benchmark_remap(c: &mut Criterion) {
...
-               [types]
-               number = "int"
-               bool = "bool"
-               timestamp = "timestamp|%d/%m/%Y:%H:%M:%S %z"
-               "#,
-        )
-        .unwrap()
-        .build(TransformContext::new_test())
-        .unwrap();
+        let mut rt = runtime();
```
ktff, line 844: Also this one.

benches/lua.rs
```diff
@@ -94,12 +95,16 @@ fn field_filter(c: &mut Criterion) {
         Benchmark::new("native", move |b| {
             b.iter_with_setup(
                 || {
-                    transforms::field_filter::FieldFilterConfig {
-                        field: "the_field".to_string(),
-                        value: "0".to_string(),
-                    }
-                    .build(TransformContext::new_test())
-                    .unwrap()
+                    let mut rt = runtime();
```
ktff, line 98: Also this one.

ktff (approved):
Looks good, and regarding requiring `async` in tests that don't need it, it seams some ~50 tests are affected but just using `async build` seems like a clean option, so I'm ok with it.

PR #4215, opened 2020-09-30
enhancement(json_parser transform): add field's value in warn message when failing to parse

oktal:
When debugging json_parser transform, either in tests or classic run, it might be hard to understand why the parser fails.

The PR adds the value of the field that was attempted to be parsed in the warn message to help the user figuring out why the parsing failed.

Example of an output:

```
Sep 30 17:33:54.573  WARN vector::internal_events::json_parser: Event failed to parse as JSON. field=json_string value={this: is not: how json: looks} self.error=key must be a string at line 1 column 2 rate_limit_secs=30
```

### DIFF
--- a/src/internal_events/json_parser.rs
+++ b/src/internal_events/json_parser.rs
@@ -22,6 +22,7 @@ impl InternalEvent for JsonParserEventProcessed {
 #[derive(Debug)]
 pub(crate) struct JsonParserFailedParse<'a> {
     pub field: &'a Atom,
+    pub value: &'a str,
     pub error: Error,
 }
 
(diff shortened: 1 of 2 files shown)


src/internal_events/json_parser.rs
```diff
@@ -30,6 +31,7 @@ impl<'a> InternalEvent for JsonParserFailedParse<'a> {
         warn!(
             message = "Event failed to parse as JSON.",
             field = %self.field,
+            value = %self.value,
             %self.error,
```
Hoverbear, line 35: Can you make this `error = %self.error` to trim out the `self` in the output?

Hoverbear (approved):

PR #4219, opened 2020-09-30
feat(remap transform): add `split` function

### LINKED ISSUE #3745: New `split` remap function

### LINKED ISSUE #4091: Add support for non-Value types in remap function signatures
The `remap` transform language has support for functions that accept arguments.

These arguments can be either literals, or dynamic references to other fields within an event.

An example of this is:

```rust
.foo = contains(.foo, substring = .bar, case_sensitive = true)
```

In this case, the first (positional) argument dynamically refers to the `foo` field. This field can contain any variant of the `Value` enum. The same applies to the `substring` argument. The `case_sensitive` argument is set to the literal `true` boolean value, but even there, any supported literal value has to be (and is) converted to (in this case) the `Value::Boolean` enum variant.

However, [the issue](https://github.com/timberio/vector/issues/3838) describing the technical requirements of the `contains` function explicitly mentions it should _also_ support regular expression patterns, e.g.:

```rust
.foo = contains(.foo, pattern = /foo/)
```

This is currently **not supported**, because there is no sane `Value` enum variant to which this pattern can be converted, and all arguments _have to_ convert to one of the defined `Value` variants.

(as an aside; the reason why argument types have to match one of the `Value` variants is so that their values can uniformly be dynamically supplied, by providing a path `.foo` instead of a literal `"foo"`, such that function arguments can be constructed using other fields within the event)

To support non-Value argument types, we need to add support for this both in the parser and the function definitions. A discussion on how we can go about doing so can be found here: https://github.com/timberio/vector/pull/3927#discussion_r492707337.

The TL;DR is:

---

We're going to update the Pest `argument` rule, something like:

```diff
argument = { positional_item | keyword_item }

- positional_item = { query_arithmetic }
+ positional_item = { argument_item }

- keyword_item = { ident ~ "=" ~ query_arithmetic }
+ keyword_item = { ident ~ "=" ~ argument_item }

+ argument_item = { query_arithmetic | regex }

+ regex = { "/" ~ /* TODO */ ~ "/" }
```

Then we'll change the parser to convert the input into an enum:

```rust
enum ArgumentKind {
	Value(event::Value),
	Regex(regex::Regex),
}
```

(in actuality, `ArgumentKind::Value` would contain a `Box<dyn Function>`, as it needs to be resolved at runtime to a `Value`)

And finally, we update function signatures to have them either accept or reject `ArgumentKind::Regex`.

If a parameter accepts a regex, it cannot be provided by a dynamic field lookup, which will result in a compile-time error.

The only thing I'm still on the fence about is whether we should allow _dynamic regex_ by default on anything that accepts a regex, or if each function should build its own support for dynamic regexes, by accepting both `Value::Bytes` and `Regex` as an input value for a single parameter.

Basically, should we do something like this:

```rust
enum ArgumentKind {
	Value(event::Value),
	Regex(Regex),
}

enum Regex {
	Static(regex::Regex),
	Dynamic(bytes::Bytes),
}
```

And have the parser accept a string as regex input, and then have it compile the regex at runtime by implementing `Function` for `Regex` and call `regex::Regex::new` on the slice of bytes.

---

There's still some design choices to be made here. It _might_ be worth writing an RFC first, or we could just go ahead and spike an initial implementation to see how well it fits the language design.

The first iteration of this would be built to support the regular expression case mentioned before, but it should be flexible enough to support other argument types that can't be transformed into a `Value`.

FungusHumungus:
Fixes #3745 
Closes #4091 

This adds the `split` remap function. Since split takes a regex parameter it also adds in the ability for remap functions to handle regexes.

It's very possible this needs some tidy up, I want to post it here to get some feedback on the approach taken to implement regex parameters. I just have this vague notion in my head that this can be improved.

I have created an `ArgumentKind` struct that can either be a query or a regex. The `Argument` then stores this rather than the query directly.

Now, because the `Argument` is not necessarily a function, it can no longer implement `Function` in order to ensure that the result resolves to the type specified by `Parameter`. Instead, I have created a new struct, `TypedArgument` that does. It is a little bit unsatisfying because now both `Argument` and `TypedArgument` contains a `Parameter` field - `Argument` because it needs the `Parameter` name and `TypedArgument` because it needs the `Parameters` type. 

### DIFF
--- a/src/mapping/parser/mod.rs
+++ b/src/mapping/parser/mod.rs
@@ -7,7 +7,10 @@ use crate::{
             self,
             arithmetic::Arithmetic,
             arithmetic::Operator,
-            function::{Argument, ArgumentList, FunctionSignature, NotFn},
+            function::{
+                Argument, ArgumentKind, ArgumentList, FunctionSignature, NotFn, Parameter,
+                RemapRegex, TypedArgument,
+            },
             path::Path as QueryPath,
             Literal,
         },
@@ -19,6 +22,7 @@ use pest::{
     iterators::{Pair, Pairs},
     Parser,
 };
+use regex::RegexBuilder;
 use std::str::FromStr;
 
 // If this macro triggers, it means the parser syntax file (grammar.pest) was
@@ -286,20 +290,64 @@ fn positional_item_from_pair(
     index: usize,
     signature: &FunctionSignature,
 ) -> Result<()> {
-    let resolver = query_arithmetic_from_pair(pair.into_inner().next().ok_or(TOKEN_ERR)?)?;
-
     let parameter = signature.parameters().get(index).cloned().ok_or(format!(
         "unknown positional argument '{}' for function: '{}'",
         index,
         signature.as_str()
     ))?;
 
+    let resolver = argument_item_from_pair(
+        parameter.clone(),
+        pair.into_inner().next().ok_or(TOKEN_ERR)?,
+    )?;
     let keyword = parameter.keyword.to_owned();
     let argument = Argument::new(resolver, parameter);
 
     list.push(argument, Some(keyword));
 
     Ok(())
+}
+
+fn argument_item_from_pair(parameter: Parameter, pair: Pair<Rule>) -> Result<ArgumentKind> {
+    let inner = pair.into_inner().next().ok_or(TOKEN_ERR)?;
+    match inner.as_rule() {
+        Rule::query_arithmetic_boolean => query_arithmetic_from_pair(inner)
+            .map(|resolver| ArgumentKind::Value(Box::new(TypedArgument::new(resolver, parameter)))),
+        Rule::regex => regex_from_pair(inner).map(ArgumentKind::Regex),
+        _ => Err(TOKEN_ERR.into()),
+    }
+}
+
+fn regex_from_pair(pair: Pair<Rule>) -> Result<RemapRegex> {
+    match pair.as_rule() {
+        Rule::regex => {
+            let mut inner = pair.into_inner();
+            let regex = inner.next().ok_or(TOKEN_ERR)?;
+            let flags = inner.next().map(|flags| flags.as_str()).unwrap_or("");
+
+            let mut global = false;
+            let mut insensitive = false;
+            let mut multi_line = false;
+            for flag in flags.chars() {
+                if flag == 'i' {
+                    insensitive = true;
+                } else if flag == 'g' {
+                    global = true;
+                } else if flag == 'm' {
+                    multi_line = true;
+                }
+                // Invalid flags shouldn't get picked up by the parser.
+            }
+
+            RegexBuilder::new(&regex.as_str().replace("\\/", "/"))
+                .case_insensitive(insensitive)
+                .multi_line(multi_line)
+                .build()
+                .map(|regex| RemapRegex { regex, global })
+                .map_err(|err| format!("invalid regex {}", err))
+        }
+        _ => Err(TOKEN_ERR.into()),
+    }
 }
 
 fn keyword_item_from_pair(
@@ -309,7 +357,6 @@ fn keyword_item_from_pair(
 ) -> Result<()> {
     let mut pairs = pair.into_inner();
     let keyword = pairs.next().ok_or(TOKEN_ERR)?.as_span().as_str();
-    let resolver = query_arithmetic_from_pair(pairs.next().ok_or(TOKEN_ERR)?)?;
 
     let parameter = signature
         .parameters()
@@ -322,6 +369,7 @@ fn keyword_item_from_pair(
         ))?
         .clone();
 
+    let resolver = argument_item_from_pair(parameter.clone(), pairs.next().ok_or(TOKEN_ERR)?)?;
     let argument = Argument::new(resolver, parameter);
 
     list.push(argument, Some(keyword.to_owned()));
@@ -507,8 +555,8 @@ mod tests {
     use super::*;
     use crate::mapping::query::function::{
         ContainsFn, DowncaseFn, FormatTimestampFn, Md5Fn, NowFn, ParseJsonFn, ParseTimestampFn,
-        Sha1Fn, SliceFn, StripWhitespaceFn, ToBooleanFn, ToFloatFn, ToIntegerFn, ToStringFn,
-        ToTimestampFn, TokenizeFn, TruncateFn, UpcaseFn, UuidV4Fn,
+        Sha1Fn, SliceFn, SplitFn, StripWhitespaceFn, ToBooleanFn, ToFloatFn, ToIntegerFn,
+        ToStringFn, ToTimestampFn, TokenizeFn, TruncateFn, UpcaseFn, UuidV4Fn,
     };
 
     #[test]
@@ -1178,6 +1226,20 @@ mod tests {
                 Mapping::new(vec![Box::new(Assignment::new(
                     "foo".to_string(),
                     Box::new(TokenizeFn::new(Box::new(QueryPath::from("foo")))),
+                ))]),
+            ),
+            (
+                r#".foo = split(.bar, /\b/i)"#,
+                Mapping::new(vec![Box::new(Assignment::new(
+                    "foo".to_string(),
+                    Box::new(SplitFn::new(
+                        Box::new(QueryPath::from("bar")),
+                        ArgumentKind::Regex(RemapRegex {
+                            regex: regex::Regex::new(r#"\b"#).unwrap(),
+                            global: false,
+                        }),
+                        None,
+                    )),
                 ))]),
             ),
         ];
--- a/src/mapping/query/function/mod.rs
+++ b/src/mapping/query/function/mod.rs
@@ -12,7 +12,9 @@ use std::str::FromStr;
 
 /// Commonly used types when building new functions.
 mod prelude {
-    pub(super) use super::{is_scalar_value, ArgumentList, Parameter};
+    #[cfg(test)]
+    pub(super) use super::RemapRegex;
+    pub(super) use super::{is_scalar_value, ArgumentKind, ArgumentList, Parameter};
     pub(super) use crate::event::{Event, Value};
     pub(super) use crate::mapping::query::Function;
     #[cfg(test)]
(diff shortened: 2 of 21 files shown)


src/mapping/parser/mod.rs
```diff
@@ -302,14 +308,55 @@ fn positional_item_from_pair(
     Ok(())
 }
 
+fn argument_item_from_pair(parameter: Parameter, pair: Pair<Rule>) -> Result<ArgumentKind> {
+    let inner = pair.into_inner().next().ok_or(TOKEN_ERR)?;
+    match inner.as_rule() {
+        Rule::query_arithmetic_boolean => query_arithmetic_from_pair(inner)
+            .map(|resolver| ArgumentKind::Value(Box::new(TypedArgument::new(resolver, parameter)))),
+        Rule::regex => regex_from_pair(inner).map(ArgumentKind::Regex),
+        _ => Err(TOKEN_ERR.into()),
```
JeanMertz, line 317:
```suggestion
        _ => unexpected_parser_sytax!(inner),
```

src/mapping/query/function/slice.rs
```diff
@@ -81,9 +81,12 @@ impl TryFrom<ArgumentList> for SliceFn {
...
     fn try_from(mut arguments: ArgumentList) -> Result<Self> {
-        let query = arguments.required("value")?;
-        let start = arguments.required("start")?;
-        let end = arguments.optional("end");
+        let query = arguments.required("value")?.into_value()?;
+        let start = arguments.required("start")?.into_value()?;
+        let end = arguments
+            .optional("end")
+            .map(|v| v.into_value())
+            .transpose()?;
```
JeanMertz, line 89: Maybe just add a separate `required_value` and `optional_value` to `ArgumentList`, to make it less tedious to implement new functions?

src/mapping/query/function/split.rs
```diff
@@ -0,0 +1,183 @@
...
+
+                let iter: Box<dyn Iterator<Item = _>> = match limit {
+                    Some(limit) => Box::new(string.splitn(limit, &pattern)),
+                    None => Box::new(string.split(&pattern)),
+                };
+
+                Ok(Value::Array(
+                    iter.map(|sub| Value::Bytes(sub.to_string().into()))
+                        .collect(),
+                ))
```
JeanMertz, line 55: Seems like we should be able to pull this (and below) out of the match statement to reduce duplication.

src/mapping/query/function/split.rs
```diff
@@ -0,0 +1,183 @@
...
+    fn parameters() -> &'static [Parameter] {
+        &[
+            Parameter {
+                keyword: "value",
+                accepts: |v| matches!(v, Value::Bytes(_)),
+                required: true,
+            },
+            Parameter {
+                keyword: "pattern",
+                accepts: |v| matches!(v, Value::Bytes(_)),
```
JeanMertz, line 82: Am I correct in assuming that the `RemapRegex` argument kind isn't part of the validation check right now? I could see that be problematic, as that means you can now basically pass in a regex to any parameter of any function?

I assume that's not the case, so I'll go dig for where this is prevented, but it does seem like we should potentially integrate your changes with the validation logic, just so that we can keep this centralized in the exposed parameter signatures.

JeanMertz: Right, so I believe `ArgumentKind::into_value` now returns an error if you pass in a regex to a parameter that expects a value, correct?

I haven't thought through how to do it, but I do think it makes sense to consider updating the `accepts` check here to have it support checking for all `ArgumentKind` variants. We can optionally add some kind of macro to make the check easier to describe, e.g.:

```rust
Parameter {
	accepts: |v| accepts!(v, Regex, Value::Integer(_))
}
```

JeanMertz (changes requested):
This look ssolid, thanks @FungusHumungus!

I don't have the time today any more to do a more in-depth review, but I left a few comments that stood out to me.

src/mapping/query/function/split.rs
FungusHumungus: Yes, that is correct. It was something I wasn't totally sure about, your suggestion feels more consistent so all parameter checks happen in one place.

src/mapping/query/function/mod.rs
```diff
@@ -120,6 +122,70 @@ build_signatures! {
     contains => ContainsFn,
     slice => SliceFn,
     tokenize => TokenizeFn,
+    split => SplitFn,
+}
+
+pub(in crate::mapping) struct TypedArgument {
```
JeanMertz, line 128: Maybe `DynamicArgument` is a better name? Since the other arguments (Regex, but others as well) are also "typed", they just don't resolve to a `Value`.

JeanMertz (commented):
I wonder if it would be simpler if we kept the top-level `Argument` type, which has both a `parameter` field and a `argument_kind` field.

The `kind` would be the enum for `Value` and `Regex`.

We can then still implement `Function` for `Argument`, but have it return an error if it tries to resolve the `Regex` variant.

We could then have `Argument` expose some kind of way to either get the `Box<dyn Function>` or the `Regex` type.

The functions then fetch one of the two in `try_from`, and store it in their struct for future use.

It _seems_ that we should be able to use most of the existing plumbing by going that route, including the runtime type validation.

JeanMertz:
I've been thinking some more on this, and I think it might make sense to still support implementing `Function` for all remap-specific types.

We can easily do so, by converting "complex" (e.g. non-`Value`) types to `Value::Map`. We can use Serde's [internally-tagged enum representation](https://serde.rs/enum-representations.html#internally-tagged) to represent these complex types.

For `Regex`, we can represent it as:

```json
{
	"type": "Regex",
	"pattern": "foo",
	"flags": ["i"]
}
```

_(there are other ways to represent the regex, we'll have to experiment with what is most flexible, while still being fairly simple)_

So one could for example do:

```rust
.foo = "/foo/i"
.foo_regex = parse_regex(.foo)
```

Now `.foo_regex` can be used as a dynamic field in any parameter that takes a `RemapValue::Regex` argument kind.

If one were to convert the above event to JSON, it would look as such:

```json
{
	"foo": "/foo/i",
	"foo_regex": {
		"type": "regex",
		"pattern": "foo",
		"flags": ["i"]
	}
}
```

Now, any function that explicitly takes a regex will accept `.foo_regex` (doing regex object validation at runtime), and the user can also still dynamically manipulate the regex before using it by manipulating the paths, such as `.foo_regex.flags`.

Functions can also still accept _either a string or regex_ since the regex is a `Value::Map` of a specific shape, whereas a string is just a `Value::Bytes`.

---

Using this, we'd gain multiple advantages:

* Complex remap value types are still just regular `Value` types.
* We give the maximum amount of control to users to shape the types as they wish, by allowing them to inspect and manipulate the type properties.
* We keep our implementation as simple as possible, since `remap::Regex` can implement `Function` and thus be dynamically accessible, and can be used in our existing type-validation infrastructure.

---

If we go this route, I would like to do the same for other types that we can't represent as `Value`, but are currently using in one way or another. For example [a `URL` type](https://github.com/timberio/vector/issues/3762) to easily access URL components, or [a `Duration` type](https://github.com/timberio/vector/pull/4186).

src/mapping/parser/mod.rs
```diff
@@ -302,14 +308,55 @@ fn positional_item_from_pair(
...
+            let mut insensitive = false;
+            let mut multi_line = false;
+            for flag in flags.chars() {
+                if flag == 'i' {
+                    insensitive = true;
+                } else if flag == 'g' {
+                    global = true;
+                } else if flag == 'm' {
+                    multi_line = true;
+                }
```
bruceg, line 338:
```suggestion
                match flag {
                    'i' => insensitive = true,
                    'g' => global = true,
                    'm' => multi_line = true,
                    _ => (),
                }
```

bruceg: Could also do this with `let (global, insensitive, multi_line) = flags.chars().fold(…)` but it's probably not worth it.

src/mapping/query/function/mod.rs
```diff
@@ -120,6 +122,70 @@ build_signatures! {
...
+}
+
+/// Because the Regex object doesn't contain a global flag
+/// (this is determined by calling either replace or replace_all)
+/// we need to wrap the object and store this flag when we parse it.
+#[derive(Debug, Clone)]
+pub(in crate::mapping) struct RemapRegex {
+    pub regex: regex::Regex,
+    pub global: bool,
+}
```
bruceg, line 173: I'm not sure `RemapRegex` is the best name for this, as on my first glance I wondered what it was remapping. I suggest `RegexArgument` similar to `DynamicArgument` above.

src/mapping/query/function/mod.rs
```diff
@@ -120,6 +122,70 @@ build_signatures! {
...
+    Regex(RemapRegex),
+}
+
+impl ArgumentKind {
+    pub fn into_value(self) -> Result<Box<dyn Function>> {
+        if let ArgumentKind::Value(value) = self {
+            Ok(value)
+        } else {
+            Err("expected value".into())
+        }
```
bruceg, line 187: Match all things :grin: 
```suggestion
        match self {
            ArgumentKind::Value(value) => Ok(value),
            _ => Err("expected value".into()),
        }
```

src/mapping/query/function/split.rs
bruceg: Seconded. I think all of the iterators should even be the same type thanks to the box, so everything except the split could move up.

bruceg (changes requested):
I agree with @JeanMertz's concerns, and have a few questions to add.

FungusHumungus:
After rethinking things a bit with @JeanMertz, we will probably be killing this pull in favour of a solution more in line with #4091.

src/mapping/query/function/mod.rs
FungusHumungus: I so much prefer the first version, using `if let` when just matching against a single variant.

What is the reason you would prefer to match here?

bruceg: I have come to prefer `match` over `if let … else`. It seems more idiomatic (matching the state of an enum), easier to add cases later, and takes fewer lines to boot in cases like this.

FungusHumungus: Making it easier to add cases later is a very good point. I'm sold!

src/mapping/query/function/split.rs
FungusHumungus: Unfortunately, this gives me lifetime issues. The pattern and regex are both dropped as we leave the `match`, but the iterator still needs them to be around as it iterates through the results.

The best I can do is extract the repeated code out to a closure and call that from each branch.

PR #4226, opened 2020-09-30
chore(aws_s3 sink): Fix flakey s3_gzip test

### LINKED ISSUE #4223: chore: Replace sinks::aws_s3::integration_tests::s3_gzip test
Test removed in #4222 due to this failure:

https://github.com/timberio/vector/pull/4195/checks?check_run_id=1188483854

```rust
    #[tokio::test]
    async fn s3_gzip() {
        let cx = SinkContext::new_test();

        let config = S3SinkConfig {
            compression: Compression::Gzip,
            filename_time_format: Some("%S%f".into()),
            ..config(10000).await
        };

        let prefix = config.key_prefix.clone();
        let client = config.create_client(cx.resolver()).unwrap();
        let sink = config.new(client, cx).unwrap();

        let (lines, events) = random_lines_with_stream(100, 500);
        sink.run(events).await.unwrap();

        let keys = get_keys(prefix.unwrap()).await;
        assert_eq!(keys.len(), 6);

        let response_lines = stream::iter(keys).fold(Vec::new(), |mut acc, key| async {
            assert!(key.ends_with(".log.gz"));

            let obj = get_object(key).await;
            assert_eq!(obj.content_encoding, Some("gzip".to_string()));

            acc.append(&mut get_gzipped_lines(obj).await);
            acc
        });

        assert_eq!(lines, response_lines.await);
    }
```

jszwedko:
The test was using `%S%f` for the time format for the object name, where
`%S` is "Second number (00--60), zero-padded to 2 digits."
(https://docs.rs/chrono/0.4.11/chrono/format/strftime/index.html#specifiers).
When the test ran close to the minute mark, this caused it write later
batches as objects that appear _earlier_ in the ListObjectsV2 call than
objects written before it.

This changes the filename_time_format to be `%s%f` where `%s` is the unix
timestamp so objects written later should always appear after.

Supersedes https://github.com/timberio/vector/pull/4222
Closes #4223

### DIFF
--- a/src/sinks/aws_s3.rs
+++ b/src/sinks/aws_s3.rs
@@ -617,7 +617,7 @@ mod integration_tests {
 
         let config = S3SinkConfig {
             compression: Compression::Gzip,
-            filename_time_format: Some("%S%f".into()),
+            filename_time_format: Some("%s%f".into()),
             ..config(10000).await
         };
 


jamtur01 (approved):
Nice catch. LGTM!

PR #4230, opened 2020-09-30
chore: An internal entry API for LogEvent

### LINKED ISSUE #2844: Add an Entry style API to LogEvent
Step 7 from https://github.com/timberio/vector/blob/c4707947bd876a0ff7d7aa36717ae2b32b731593/rfcs/2020-05-25-more-usable-logevents.md#sales-pitch.

Support an `Entry` style API in `LogEvent`.

Remove any `try_insert` or similar calls that are obsolete.

Hoverbear:
Close #2844.

This notably is a bit simpler than one might hope: **It does not support indexing**.

This is because there is no good definition of how to approach the situation of something like this:

```json
{
    "a": [],
}
```

Okay, now consider `entry(Lookup::from_str("a[99999999]"))`. What does this give? Does this create `99999998` null entries in the array and then return to the caller a mutable borrow of the 99999999th element? This seems pretty reasonable.

However, the `Vec` type doesn't have an `Entry` API, so we'd have to re-implement the `entry()` API for it internally. This broadly expands the work for this feature, so I left it for a later date.

## What works:

Normal nested lookups on existing fields.

```rust
let lookup = Lookup::from_str("map.map.buddy").unwrap();
let entry = event.entry(lookup).unwrap();
trace!(?entry);
let fallback = json!(
    "If you see this, the `LogEvent::entry` API is not working on existing, multi-segment lookups."
);
entry.or_insert(fallback.clone().into());
let json: serde_json::Value = event.clone().try_into().unwrap();
assert_eq!(json.pointer("/map/map/buddy"), Some(&serde_json::Value::Number((-1).into())));
```

Nested lookups on non-existing fields at the tail.

```rust
let lookup = Lookup::from_str("map.map.non-existing").unwrap();
let entry = event.entry(lookup).unwrap();
trace!(?entry);
let fallback = json!(
    "If you don't see this, the `LogEvent::entry` API is not working on non-existing multi-segment lookups."
);
entry.or_insert(fallback.clone().into());
let json: serde_json::Value = event.clone().try_into().unwrap();
assert_eq!(json.pointer("/map/map/non-existing"), Some(&fallback));
```

Nested lookups on non-existing fields in the middle.

```rust
let lookup = Lookup::from_str("map.non-existing.map").unwrap();
let entry = event.entry(lookup).unwrap();
trace!(?entry);
let fallback = json!(
    "If you don't see this, the `LogEvent::entry` API is not working on a non-existing middle field in multi-segment lookups."
);
entry.or_insert(fallback.clone().into());
let json: serde_json::Value = event.clone().try_into().unwrap();
assert_eq!(json.pointer("/map/non-existing/map"), Some(&fallback));
```

Tracing is also integrated throughout:

```rust
Sep 30 15:34:22.293 TRACE entry{lookup=map-non-existing.non-existing.non-existing}: vector::event::log_event: Seeking to entry.
Sep 30 15:34:22.304 TRACE entry{lookup=map-non-existing.non-existing.non-existing}: vector::event::log_event: Seeking segment. segment=map-non-existing index=0
Sep 30 15:34:22.304 TRACE entry{lookup=map-non-existing.non-existing.non-existing}: vector::event::log_event: Seeking next segment. segment=non-existing index=1
Sep 30 15:34:22.304 TRACE entry{lookup=map-non-existing.non-existing.non-existing}: vector::event::log_event: Injecting empty map. segment=non-existing index=1
Sep 30 15:34:22.304 TRACE entry{lookup=map-non-existing.non-existing.non-existing}: vector::event::log_event: Seeking next segment. segment=non-existing index=2
Sep 30 15:34:22.304 TRACE entry{lookup=map-non-existing.non-existing.non-existing}: vector::event::log_event: Injecting empty map. segment=non-existing index=2
Sep 30 15:34:22.304 TRACE entry{lookup=map-non-existing.non-existing.non-existing}: vector::event::log_event: Result. entry=Entry(VacantEntry("non-existing"))
```


Hoverbear:
I want to validate this won't create entries unless it needs to.

src/event/log_event.rs
```diff
@@ -78,6 +80,41 @@ impl LogEvent {
...
+    #[instrument(skip(self, lookup), fields(lookup = %lookup), err)]
+    fn entry(&mut self, lookup: Lookup) -> crate::Result<Entry<String, Value>> {
+        trace!("Seeking to entry.");
+        let mut walker = lookup.into_iter().enumerate();
+
+        let mut current_pointer = if let Some((index, Segment::Field(segment))) = walker.next() {
+            trace!(%segment, index, "Seeking segment.");
+            self.fields.entry(segment)
+        } else {
+            return Err("Lookups should have at least one segment.".into());
```
lukesteensen, line 93: Could we enforce this on `Lookup` creation and not worry about it here?

src/event/log_event.rs
```diff
@@ -78,6 +80,41 @@ impl LogEvent {
...
+                    Value::Map(map) => map.entry(field),
+                    v => return Err(format!("Looking up field on a non-map value: {:?}", v).into()),
+                },
+                (Segment::Field(field), Entry::Vacant(entry)) => {
+                    trace!(segment = %field, index, "Met vacant entry.");
+                    return Err(format!(
+                        "Tried to step into `{}` of `{}`, but it did not exist.",
+                        field,
+                        entry.key()
+                    )
```
lukesteensen, line 109: What would cause this to happen exactly? As a user of the API I'm not exactly sure how I would handle errors like this.

lukesteensen (commented):
This looks good! My only question is around returning `Result` from `entry`. It'd be really great if we could make those error cases impossible, because it's not clear to me what should happen when they're hit.

src/event/log_event.rs
Hoverbear: Say I said `foo.bar.baz.bat` but only `foo.bar` existed, this would tell you "Tried to step into `baz` of `bar` but it did not exist", so you could diagnose.

Hoverbear:
@lukesteensen That's such a great challenge. I'm on board. :) Let me try.

Hoverbear:
We chatted a bit and I implemented the length invariant, we opted to leave the rest for the future.

PR #4236, opened 2020-10-01
enhancement(docker platform): Added distroless-libc and distroless-static docker container bases

rrichardson:
This is a simple PR which adds 2 docker container targets that use the [distroless](https://github.com/GoogleContainerTools/distroless) Container base.  

distroless/static is basically `FROM scratch` plus CA certificates.  
distroless/cc is distroless/static plus libc and related dependencies, so that dynamic binaries may run.  

I've successfully tested the musl build within the distroless/static,  and tested the debian package binary using the distroless/cc platform.  

The final containers are 86MB and 89MB respectively.

### DIFF
--- a/distribution/docker/README.md
+++ b/distribution/docker/README.md
@@ -81,6 +81,16 @@ which is a smaller, more compact variant of the [`debian` image][urls.docker_deb
 docker run timberio/vector:0.10.0-debian
 ```
 
+#### distroless-*
+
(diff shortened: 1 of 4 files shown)


jamtur01 (approved):
I'm okay with this. I do like the attraction of a very light-weight image. LGTM

Hoverbear:
I really really like this!

Issue #4238, opened 2020-10-01
Prometheus sink doesn't bind/listen until it receives events


rrichardson:
```
vector 0.11.0 (gb031ead x86_64-unknown-linux-gnu 2020-09-30) 
(Current master as of 2020-09-30) 
```

### Vector Configuration File

```toml
[sinks.prometheus]
  type = "prometheus"
  inputs = ["log_to_metric"] # required
  address = "0.0.0.0:9091" # required
  namespace = "vector" # required
```

### Expected Behavior

Vector should begin listening on port 9091 for inbound requests from prometheus. 

### Actual Behavior

Vector does not begin listening on port 9091 for inbound requests from prometheus. 

### Additional Context

Looking at the code that was refactored 3 days ago, the `start_server_if_needed()` function is now executed at the top of a while loop that is processing incoming events. 
I think it should be at the top of the 

### References

The change was made here: 

#3933

rrichardson:
Note that if a prometheus sink receives events consistently, it might not experience this problem. However, I am running vector as a daemonset in k8s, and, in some cases, it won't be running on the same node as the services which generate logs which spawn these events.  So what I get is Prometheus alerting me in the middle of the night that some of it's endpoints cannot be scraped.

PR #4239, opened 2020-10-01
fix(prometheus sink): Ensure server starts immediately

rrichardson:
This addresses #4238 

This assumes that the server will run forever (until shutdown) so the function only needs to be run once (but it won't hurt if it's run multiple times)


ktff (approved):
Looks good. The change to move `start_server_if_needed` to inner loop was probably seen as an optimization, but there is value in starting the server as soon as possible so that the other end knows this is running and also any issue with the connection can be detected more promptly.

PR #4242, opened 2020-10-01
fix(tests): Fix GraphQL heartbeat test

### LINKED ISSUE #4241: Fix GraphQL heartbeat test
Occasionally files (usually on macOS) due to a similar timing bug that previously affected `uptime_metrics`. It's generally redundant to assert specific timing, since the practical use-case isn't very sensitive to timing anyway.

leebenson:
Getting this up to see the effect on CI, since this is passing for me locally and is flaky only on CI. It might need a few adjustments along the way, hence the draft.

Closes #4241

### DIFF
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -198,7 +198,7 @@ mod tests {
                 .utc
                 - now;
 
-            assert!(diff.num_milliseconds() > mul as i64 * interval);
+            assert!(diff.num_milliseconds() >= mul as i64 * interval);
         }
 
         // Stream should have stopped after `num_results`


leebenson:
First all-green test I've seen in a while, but I can't be 100% sure this actually fixes the problem we saw before.

It basically just swaps `>` to `>=` on the heartbeat check, so a subsequent heatbeat that fires exactly `interval` milliseconds also qualifies.

I'm surprised this didn't fail more often, especially locally, since the process of marshalling this into resulting JSON and firing it back out to a local client is often a sub-ms operation, which is another reason I'm skeptical this will fix the issue we saw... but it's the more correct behaviour, in any case.

jszwedko:
> I'm surprised this didn't fail more often, especially locally, since the process of marshalling this into resulting JSON and firing it back out to a local client is often a sub-ms operation, which is another reason I'm skeptical this will fix the issue we saw... but it's the more correct behaviour, in any case.

I'm wondering, if we assume that operation takes 0.8ms that, we'd expect it to pass unless it ran in the first 0.2ms of a given millisecond. That could account for it mostly passing.

Agreed that this seems like the more "correct" assertion.

jszwedko (approved):

PR #4243, opened 2020-10-01
chore(journald source): remove spawn_blocking 

### LINKED ISSUE #3616: Update spawned process in journald source
On the time when journald source was introduced (https://github.com/timberio/vector/pull/702) `tokio:0.2` was not released yet, as a result `std::process::Command` was used. Because this we need a dedicated thread for reading output from spawned process: https://github.com/timberio/vector/blob/53c8e006773115c54597d06d4bde89213c6330dc/src/sources/journald.rs#L185

I think we should able to migrate to [tokio::process::Command](https://docs.rs/tokio/0.2.22/tokio/process/struct.Command.html) and remove blocking code.

fanatid:
close #3616 

~I did not figure out how I can use [AsyncBufReadExt::read_until](https://docs.rs/tokio/0.2.22/tokio/io/trait.AsyncBufReadExt.html#method.read_until) inside `poll_next` due to lifetime issues, so I created stream with [async-stream](https://crates.io/crates/async-stream).~ fixed, thanks to @juchiast


src/sources/journald.rs
```diff
@@ -287,12 +298,21 @@ impl JournalSource for Journalctl {
         }
 
         let mut child = command.spawn().context(JournalctlSpawn)?;
-        let stdout = child.stdout.take().unwrap();
-        let stdout = BufReader::new(stdout);
+        let mut stdout = BufReader::new(child.stdout.take().unwrap());
+        let jstream = Box::pin(async_stream::stream! {
```
juchiast, line 302: How about using `lines`? https://docs.rs/futures/0.3.5/futures/io/trait.AsyncBufReadExt.html#method.lines

```diff
diff --git a/src/sources/journald.rs b/src/sources/journald.rs
index a0d83948f..3115e7d66 100644
--- a/src/sources/journald.rs
+++ b/src/sources/journald.rs
@@ -298,17 +298,8 @@ impl JournalSource for Journalctl {
         }

         let mut child = command.spawn().context(JournalctlSpawn)?;
-        let mut stdout = BufReader::new(child.stdout.take().unwrap());
-        let jstream = Box::pin(async_stream::stream! {
-            loop {
-                let mut line = Vec::new();
-                yield match stdout.read_until(b'\n', &mut line).await {
-                    Ok(0) => break,
-                    Ok(_) => Ok::<String, io::Error>(String::from_utf8_lossy(&line).into()),
-                    Err(err) => Err::<String, io::Error>(err),
-                };
-            }
-        });
+        let stdout = BufReader::new(child.stdout.take().unwrap());
+        let jstream = stdout.lines().boxed();

         let pid = Pid::from_raw(child.id() as i32);
         Ok((
```

juchiast: It is not the same as `from_utf8_lossy`, though.

fanatid: Oh, I did not noticed that `lines` exists... but yeah there `from_utf8` used instead `from_utf8_lossy` =\

juchiast: hmm we can use `split(b'\n')`, https://docs.rs/tokio/0.2.22/tokio/io/trait.AsyncBufReadExt.html#method.split

fanatid: Great! I'll change it now.

bruceg (approved):

src/sources/journald.rs
```diff
@@ -303,20 +317,20 @@ impl JournalSource for Journalctl {
...
-            Ok(0) => None,
-            Ok(_) => Some(Ok(String::from_utf8_lossy(&line).into())),
-            Err(err) => Some(Err(err)),
-        }
+impl Stream for Journalctl {
+    type Item = io::Result<String>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
+        Poll::Ready(match ready!(self.project().inner.poll_next(cx)) {
+            Some(Ok(segment)) => Some(Ok(String::from_utf8_lossy(&segment).into())),
```
juchiast, line 325: This `.into()` add an unnecessary clone every time `segment` is valid utf8, right? I'm not sure if rust is smart enough to optimize it. 

We can get away with something like
```rust
fn from_utf8_lossy(bytes: Vec<u8>) -> String {
    use std::borrow::Cow;
    match String::from_utf8_lossy(&bytes) {
        Cow::Owned(s) => s,
        _ => unsafe { String::from_utf8_unchecked(bytes) },
    }
}
```
but it add an unsafe block.

juchiast (approved):
A comment on minor performance issue, other than that it is fine.

src/sources/journald.rs
fanatid: Yeah, that's looks correct @juchiast, `.into()` will cause extra copy for each valid utf8. I do not think that this copy super critical but it would be nice to change if not unsafe block :disappointed:

bruceg: It is possible without a clone, but does the UTF-8 validation twice (assumes `segment: Vec<u8>`):
```rust
String::from_utf8(segment).unwrap_or_else(|error| String::from_utf8_lossy(error.as_bytes()).into())
```
This is *probably* more efficient, as the majority of the time the source will be valid UTF-8, so it avoids the allocation and copy, and if it isn't valid, it's likely to fail early anyways.

src/sources/journald.rs
```diff
@@ -331,21 +344,18 @@ where
...
-            let mut at_end = false;
             let mut cursor: Option<String> = None;
 
             for _ in 0..self.batch_size {
-                let text = match self.journal.next() {
+                let text = match self.journal.next().await {
                     None => {
-                        at_end = true;
-                        break;
+                        let _ = self.shutdown.compat().await;
```
juchiast, line 357: @fanatid As I understand, `shutdown` is a signal to stop all sources, I'm not sure if this source should wait for it instead of terminating immediately?

fanatid: It's a good question. I only removed `spawn_blocking` and did not change logic. I think we should return on `None` and I also do not see that we actually use the `shutdown` signal.

juchiast: This wait on `shutdown` seem to be used as backoff originally.

https://github.com/timberio/vector/blob/0f72a2b1669a97e4838d3ca852d2f68a878915f4/src/sources/journald.rs#L171-L175

We use `shutdown` in `select(...)` so should be safe to remove it.

https://github.com/timberio/vector/blob/521711ee44b1dd83426fd8f4ae9f0b90e75769ab/src/sources/journald.rs#L200-L202

Issue #4056, opened 2020-09-22
Investigate `#[should_panic]` flakiness in macOS CI tests


leebenson:
Surfaced in: https://github.com/timberio/vector/issues/4003

Workaround for now: https://github.com/timberio/vector/pull/4050

Possibly related: https://github.com/rust-lang/rust/issues/74820

jszwedko:
Closing since tests have been deleted.

Issue #4196, opened 2020-09-29
OSX unit tests segfault


jszwedko:
Discussed in discord: https://discordapp.com/channels/742820443487993987/746070604192415834/760546620559065148

Example: https://github.com/timberio/vector/pull/4193/checks?check_run_id=1183596467

Appears to have been introduced by https://github.com/timberio/vector/pull/4187

leebenson:
FWIW, I'm pretty sure I saw this even before #4187. It's especially hard to diagnose because it doesn't always happen, it happens even after _all_ tests have already passed, and I can't reproduce it locally in macOS 10.15.6.

There are similar issues on rust-lang:

- [FFI to C on MacOS causes segmentation fault (works on Ubuntu)](https://github.com/rust-lang/rust/issues/59202)
- [No issues on Linux, segfault on OS X (but only when building with `–release`)](https://users.rust-lang.org/t/no-issues-on-linux-segfault-on-os-x-but-only-when-building-with-release/17327/21)
- [https://users.rust-lang.org/t/segfaults-under-mac-os-with-multithreaded-code/18724](https://users.rust-lang.org/t/segfaults-under-mac-os-with-multithreaded-code/18724)
- [https://users.rust-lang.org/t/without-a-single-line-of-unsafe-code-i-am-getting-segfault-and-i-cant-figure-out-why-since-i-cant-stack-trace/40017](https://users.rust-lang.org/t/without-a-single-line-of-unsafe-code-i-am-getting-segfault-and-i-cant-figure-out-why-since-i-cant-stack-trace/40017)
- [https://github.com/rust-lang/rust/issues/73030](https://github.com/rust-lang/rust/issues/73030) (perhaps not directly related, but we were bitten by this in #4056 and removed related tests in #4050)
- [RUST_BACKTRACE on macOS can trigger crashes](https://github.com/rust-lang/rust/issues/71397)
- [Spurious LLDB segfault (signal 11)](https://github.com/rust-lang/rust/issues/45230)

Some of these are still open. Others were fixed in earlier Rust versions. Some may not be relevant. Some may be compiler issues, or due to external libs.

When using Travis, we could SSH into a test and perform some more in-depth testing againt the environment. I'm not sure what the options are for GH Actions. It's really hard to test because `make test` _never_ segfaults locally, with an identical toolchain.

[...]

PR #4251, opened 2020-10-01
chore(tests): Disable flaky tests on MacOS

ktff:
Ref. #2978, #4056

I'll just keep disabling tests for Mac in this PR until we get a consistently passing check, and after that make a triage of them. This to avoid accumulation of broken tests behind a couple, hopefully, of them at moment 

Tests disabled on Mac:
* `tcp_stream_detects_disconnect`
* `file_update`
* `topology` tests

### DIFF
--- a/src/config/watcher.rs
+++ b/src/config/watcher.rs
@@ -117,8 +117,7 @@ fn add_paths(watcher: &mut RecommendedWatcher, config_paths: &[PathBuf]) -> Resu
     Ok(())
 }
 
-#[cfg(unix)]
-#[cfg(test)]
+#[cfg(all(test, unix))]
 mod tests {
     use super::*;
     use crate::test_util::{temp_file, trace_init};
--- a/src/config/watcher.rs
+++ b/src/config/watcher.rs
@@ -136,6 +136,7 @@ mod tests {
         tokio::time::timeout(timeout, signal.recv()).await.is_ok()
     }
 
+    #[cfg(not(macos))] // Issue https://github.com/timberio/vector/issues/2978
     #[tokio::test]
     async fn file_update() {
         trace_init();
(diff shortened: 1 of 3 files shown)


jamtur01:
cc @leebenson

leebenson:
Even if all tests pass, there's https://github.com/timberio/vector/issues/4196 to contend with. I added some additional commentary to that issue this morning.

I'm not sure how to probe it further, without being able to SSH into an active runner and do some more extensive stack tracing against the compiled binary. I can't recreate this locally at all. I've seen segfaults in other environments where memory pressure was clearly the issue (i.e. actual "out of memory" errors in the console), so _possibly_ this is a memory pressure thing. Not sure. It could also be a genuine segfault due to `unsafe` code in a lib somewhere, but why then would this only show up in macOS and not my own macOS environment at that, I don't know.

Not sure how to probe this any further without choosing a 'proper' CI where we have more control over the underlying hardware, OS and can SSH into the environment.

leebenson:
It could also be due to something I've done in GraphQL tests if, like @jszwedko commented, this was indeed introduced in https://github.com/timberio/vector/pull/4187 (although I'm pretty sure I've come across this a time or two before; feels older than a few days.)

jszwedko:
@leebenson It's a bit hacky, but I've actually used https://github.com/mxschmitt/action-tmate before to get SSH access to a running Github Action job for debugging a similar type of issue that only happens on CI.

leebenson:
Thanks @jszwedko, I was looking for something like that! Will give it a shot.

ktff:
@leebenson 
> Even if all tests pass, there's #4196 to contend with

That sounds like a tough issue. Then I would at least like to either confirm or eliminate the possibility that some smallish subset of tests can cause #4196. It's quite possible that some specific tests are corrupting the memory which then causes SEGFAULTs at random times. Although I'm now aware that this approach in PR may not work.

ktff:
@leebenson as you reported in #4196 SEGFAULT started to happen after the tests where whole test groups can be the cause, and with recent bors upgrade this isn't viable approach so I'll close the PR.

PR #4252, opened 2020-10-01
chore: Better tracing in lookups.

Hoverbear:
Adds a tad more hygenic tracing to our new lookup.

### DIFF
--- a/src/event/lookup/segment.rs
+++ b/src/event/lookup/segment.rs
@@ -30,7 +30,8 @@ impl Segment {
     #[tracing::instrument(skip(segment))]
     pub(crate) fn from_lookup(segment: Pair<'_, Rule>) -> crate::Result<Vec<Segment>> {
         let rule = segment.as_rule();
-        tracing::trace!(segment = segment.as_str(), ?rule, action = %"enter");
+        let full_segment = segment.as_str();
+        tracing::trace!(segment = %full_segment, ?rule, action = %"enter");
         let mut segments = Vec::default();
         for inner_segment in segment.into_inner() {
             match inner_segment.as_rule() {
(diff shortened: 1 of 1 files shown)


jszwedko (approved):
This change looks fine to me! I am interested to hear more about its connection to hygiene though, for my own edification. I've heard the term before, but hadn't previously looked it up until now (primarily reading https://en.wikipedia.org/wiki/Hygienic_macro).

As far as I can tell, it means that the macro expansion doesn't pollute the scope of wherever the macro is being used?

Hoverbear:
I more meant that it removes some unnecessary quotes from the logs

leebenson:
@jszwedko, to expand the definition of 'pollution' a little: it's not just that variables won't be injected into scope; interaction isn't allowed either. The following won't compile:

```rust
macro_rules! test {
    () => {
        a = 2;
    }
}

fn main() {
    let mut a = 1;
    test!();
    dbg!(a);
}
```

Error:

> error[E0425]: cannot find value `a` in this scope
>   --> src/main.rs:3:9
>    |
> 3  |         a = 2;
>    |         ^ not found in this scope
> ...
> 11 |     test!();
>    |     -------- in this macro invocation

Likewise, as you'd probably expect per normal shadowing (although in this case, it's a separate scope), this won't modify `a`:

```rust
macro_rules! test {
    () => {
        let mut a = 2;
    }
}

fn main() {
    let mut a = 1;
    test!();
    dbg!(a);
}
```

Prints:

> Finished dev [unoptimized + debuginfo] target(s) in 0.45s
>   Running `target/debug/playground`
> [src/main.rs:13] a = 1

jszwedko:
Makes sense, thanks for the additional detail @leebenson !

jszwedko:
@Hoverbear I think we are supposed to be using `bors merge` for this now.

PR #4254, opened 2020-10-01
fix(startup): Fix 101 exit code on `api.enabled = true`

leebenson:
Fixes the 101 exit code caused by adding `api.enabled = true` to `vector.toml`, due to the server being started outside of the tokio runtime.

Moved the API startup into the `rt.block_on()` block.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -184,6 +184,8 @@ impl Application {
         let mut config_paths = self.config.config_paths;
 
         let opts = self.opts;
+
+        #[cfg(feature = "api")]
         let api_config = self.config.api;
 
         rt.block_on(async move {
--- a/src/app.rs
+++ b/src/app.rs
@@ -183,24 +183,25 @@ impl Application {
 
         let mut config_paths = self.config.config_paths;
 
-        #[cfg(feature = "api")]
-        // assigned to prevent the API terminating when falling out of scope
-        let _api = if self.config.api.enabled {
-            emit!(ApiStarted {
-                addr: self.config.api.bind.unwrap(),
-                playground: self.config.api.playground
-            });
-
-            Some(api::Server::start(self.config.api))
-        } else {
-            None
-        };
-
         let opts = self.opts;
+        let api_config = self.config.api;
 
         rt.block_on(async move {
             emit!(VectorStarted);
             tokio::spawn(heartbeat::heartbeat());
+
+            #[cfg(feature = "api")]
+            // assigned to prevent the API terminating when falling out of scope
+            let _api = if api_config.enabled {
+                emit!(ApiStarted {
+                    addr: api_config.bind.unwrap(),
+                    playground: api_config.playground
+                });
+
+                Some(api::Server::start(api_config))
+            } else {
+                None
+            };
 
             let signals = signal::signals();
             tokio::pin!(signals);


src/app.rs
```diff
@@ -183,25 +183,26 @@ impl Application {
...
-                playground: self.config.api.playground
-            });
-
-            Some(api::Server::start(self.config.api))
-        } else {
-            None
-        };
-
         let opts = self.opts;
+        let api_config = self.config.api;
```
jszwedko, line 187: Seems like this `let` needs `#[cfg(feature = "api")]` too? Maybe it could be moved into the below block?

src/app.rs
```diff
@@ -183,25 +183,26 @@ impl Application {
...
 
         rt.block_on(async move {
             emit!(VectorStarted);
             tokio::spawn(heartbeat::heartbeat());
 
+            #[cfg(feature = "api")]
+            // assigned to prevent the API terminating when falling out of scope
+            let _api = if api_config.enabled {
```
jszwedko, line 195: Something feels a bit weird here. Is there somewhere else we could put this handle? I'm not familiar with the runtime stuff much yet. Maybe @Hoverbear will have a better suggestion.

I see this was already like this though so shouldn't block this PR :smile:

jszwedko (approved):
I feel like I don't really understand the fix, but it works :smile: I tried it out locally.

Hoverbear:
I think this is from #2896 :(

src/app.rs
Hoverbear: As long as we don't let it drop, it should be fine.

src/app.rs
leebenson: Thanks, `#[cfg(feature = "api")]` would prevent an unnecessary copy when it's not enabled 👍 

It can't be moved into the block, unfortunately, since that'd take ownership of `self`. The type here is `config::api::Option`, which derives `Copy`, which is why I'm able to assign this despite it looking like something that could be moved.

jszwedko: :+1: it stood out because the integration tests failed (presumably because they don't enable this feature so `self.config` had no `api` field).

leebenson: Just noticed that too, thanks!

src/app.rs
leebenson: @jszwedko - It does look a bit weird, but it might make more sense in the context of https://github.com/timberio/vector/blob/43e6ee0b93921f2a327a5b41c5854fccf3e6df20/src/app.rs#L228-L231

Here, `_api` vs `_` is important, because of the `Drop` implementation. If it goes out of scope, a shutdown signal fires which terminates the spawned API listener. So having this block outside of `rt.block_on()` (which itself takes an async func), keeps the server alive and inside the active tokio runtime.

In the LsoC referenced above, I'm making use of `_api` (renamed to `api_server`), to update reloaded config. It's wrapped as `Option<Server>` to accommodate the scenario where the API _feature_ is enabled (which it is by default), but the API _itself_ isn't (also default.)

It looks a bit odd in this PR where the result of the if statement looks like it's returning an unnecessary wrapper, bound to a variable that isn't consumed, but looks more obvious in the follow-up PR, I think.

leebenson: (Seeing the brief snippet of code inline per the above doesn't make a lot of sense-- would recommend viewing the full file for context: https://github.com/timberio/vector/blob/43e6ee0b93921f2a327a5b41c5854fccf3e6df20/src/app.rs)

jszwedko: :thinking: yeah, I think I see. It just all seems a little implicit (necessitating the explanatory comment), but I'm not sure of a better way to structure it.

leebenson: It definitely is implicit, but I think that's a useful trait when a thread is spawned and we want some way to associate the parent value dropping out of scope with a shutdown signal on an inner thread.

An alternative could be to have an explicit `server.disconnect()` (or similar), but forgetting to call it could introduce subtle bugs.

It seemed simpler to keep the server alive when it's in scope, and kill it when it's not. There may be more idiomatic patterns for this. I took inspiration from [Tokio's select! docs](https://tokio.rs/tokio/tutorial/select), from the _Cancellation_ section:

>  ... sometimes an asynchronous operation will spawn background tasks or start other operation that run in the background. For example, in the above example, a task is spawned to send a message back. Usually, the task will perform some computation to generate the value.
> 
> Futures or other types can implement Drop to cleanup background resources. Tokio's oneshot::Receiver implements Drop by sending a closed notification to the Sender half. The sender half can receive this notification and abort the in-progress operation by dropping it.

Hoverbear (approved):

Hoverbear:
@leebenson Sorry I merged this by accident out of old habit. :|

src/app.rs
jszwedko: Yeah, it's probably just a pattern I have to get used to :smile: Thanks for the explanation!

PR #4255, opened 2020-10-01
chore: Log Schema uses strings

Hoverbear:
Teach the `LogSchema` to store `String` instead of `Atom`, and shift all atoms out into their callers.

This is a continuation of work @lukesteensen started some time ago to shift our event API away from atoms.


src/event/mod.rs
```diff
@@ -388,7 +388,7 @@ impl From<Bytes> for Event {
             .insert(log_schema().message_key(), message);
         event
             .as_mut_log()
-            .insert(log_schema().timestamp_key(), Utc::now());
+            .insert(Atom::from(log_schema().timestamp_key()), Utc::now());
```
jszwedko, line 391: Why is this `Atom::from` needed here and other places we are using `log_schema()` fields with `.insert`?

jszwedko (commented):
This looks good to me though I'm not sure I completely understand the context for these changes. Is there other background for this change aside from https://github.com/timberio/vector/issues/1891 ?

Do we expect to continue removing uses of `Atom` within `LogEvent`? It seems odd to me that `get()` and `remove()` still require Atoms even though `insert()` just requires `AsRef<str>`.

I noticed some places where `insert` is used are using `Atom::from()`. Is this intended? I expected them to just be using `String`s (especially when referencing fields from `log_schema()`).

src/event/mod.rs
Hoverbear: The goal is to remove `DefaultAtom`/`Atom` completely, but doing it in one chunk would be a very large change, this preps the change a bit. :)

Hoverbear:
@jszwedko 😅 That should be fine for now. Since Atoms are cheap to reference like strings are (They're just expensive to drop) it won't have an impact as far as I can tell. I'm following up with a PR as soon as I can that removes this outright.

src/event/mod.rs
jszwedko: :+1: why would it be needed for `timestamp_key()` but not `message_key()` (appearing above).

PR #4266, opened 2020-10-02
chore: Update nightly Helm charts to a single nightly

jamtur01:

### DIFF
--- a/scripts/release-helm.sh
+++ b/scripts/release-helm.sh
@@ -14,9 +14,8 @@ CHANNEL="${CHANNEL:-"$(scripts/util/release-channel.sh)"}"
 VERSION="${VERSION:-"$(scripts/version.sh)"}"
 
 if [[ "$CHANNEL" == "nightly" ]]; then
-  DATE="${DATE:-"$(date -u +%Y-%m-%d)"}"
-  APP_VERSION="nightly-$DATE" # matches the version part of the image tag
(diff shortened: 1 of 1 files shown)


jamtur01:
@MOZGIII This is my proposed fix. A single nightly instead of dated nightly. There are obviously pros and cons here. Happy to discuss this.

MOZGIII:
Helm chart release is tightly coupled to the underlying configuration, so we need to be able to helm chart and the docker image together. So this solution, unfortunately, will be unsound.

Why don't we just rollback the change to the nightlies?

The lack of the ability to pinpoint a particular version is very problematic for people running Vector off of nightlies in Kubernetes even outside of the context of the Helm charts. With a setup like this, people won't be able to update the image by changing the tag (aka do not deploy with `lastest` tag problem), and they'll then have to set `ImagePullTag` to `Always`. This in turn will break the consistency of the deployment in case they use autoscaling or just manually scale the replicaset the next day after the initila rollout.

In essence, without the ability to pinpoint the individual nightly image, the nightlies are pretty much unusable in the K8s context. If we didn't have that ability before, I'd have to create an issue and make that happen, but things were in a good shape.

I'm sorry I didn't catch this up during the review process of the nightlies change PR, might've been I wasn't tagged. I'll make sure to add myself to all the docker-relates stuff at the CODEWONDERS to avoid this in the future.

PR #4288, opened 2020-10-03
[Merged by Bors] - enhancement(console sink): Improve error handling

juchiast:
- Don't panic.
- Emit a warning on missing field, previously it just prints an empty line.

### DIFF
--- a/src/sinks/console.rs
+++ b/src/sinks/console.rs
@@ -123,20 +123,16 @@ struct WriterSink {
 impl StreamSink for WriterSink {
     async fn run(&mut self, mut input: BoxStream<'_, Event>) -> Result<(), ()> {
         while let Some(event) = input.next().await {
-            let buf = match encode_event(event, &self.encoding) {
-                Some(mut buf) => {
-                    buf.push('\n');
-                    buf
-                }
-                None => continue,
-            };
-            self.output
-                .write_all(buf.as_bytes())
-                .await
-                .map_err(|error| {
-                    error!("Error writing to output: {}.", error);
-                })?;
             self.acker.ack(1);
+            if let Some(mut buf) = encode_event(event, &self.encoding) {
+                buf.push('\n');
+                if let Err(error) = self.output.write_all(buf.as_bytes()).await {
+                    // Error when writing to stdout/stderr is likely irrecoverable,
+                    // so stop the sink.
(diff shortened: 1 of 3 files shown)


src/sinks/console.rs
```diff
@@ -124,9 +123,19 @@ struct WriterSink {
 impl StreamSink for WriterSink {
     async fn run(&mut self, mut input: BoxStream<'_, Event>) -> Result<(), ()> {
         while let Some(event) = input.next().await {
-            write_event_to_output(&mut self.output, event, &self.encoding)
+            let buf = match encode_event(event, &self.encoding) {
+                Some(mut buf) => {
+                    buf.push('\n');
+                    buf
+                }
+                None => continue,
```
fanatid, line 131: `self.acker.ack(1)` should be called anyway

fanatid (changes requested):

src/sinks/console.rs
juchiast: moved `ack` to the top.

fanatid:
What will happen with transforms connected to sink if sink will finish with error? They will stop produce items because the pipeline will be full? Can sinks return a result if no signal outside? (It's more about topology, maybe @ktff will able to answer)

juchiast:
@fanatid A test shows that vector just stops if `console` sink return Err.

```
Oct 03 15:58:15.112 ERROR sink{topology_component_name=console topology_component_type=console}: vector::topology: An error occurred that vector couldn't handle.
```

fanatid (approved):
lgtm
@juchiast can we wait Luke or Kruno answer regarding no panic behavior? I do not remember that any other sinks return error result.

juchiast:
@fanatid yes we should wait! 

Btw why does the second commit only trigger 5 checks, seem weird.

fanatid:
Strange :thinking: any ideas @jamtur01?

jamtur01:
@fanatid because we turned on Bors. When you're ready to merge follow the instructions in contributing.

ktff:
> What will happen with transforms connected to sink if sink will finish with error?

When any sink panics or finishes with an error we start a graceful shutdown of the process.

ktff:
So replacing panic with a returned Error, doesn't change anything regarding topological behavior, although returning error is preferred.

juchiast:
bors merge

PR #4290, opened 2020-10-03
feat(remap transform): add `split` function 

### LINKED ISSUE #3745: New `split` remap function

FungusHumungus:
Closes #3745 

This is the first stage of changes from [this comment](https://github.com/timberio/vector/issues/4091#issuecomment-702066049)

Split now works with a Regex that is represented as a `Value::Map`. No code to the actual parser has been changed yet.

This is just stage 1. Next work will need to be done to integrate this into the parser so that regexes can become first class objects.

### DIFF
--- a/src/mapping/query/dynamic_regex.rs
+++ b/src/mapping/query/dynamic_regex.rs
@@ -1,0 +1,95 @@
+use crate::{event::Value, mapping::Result};
+use bytes::Bytes;
+use std::{collections::BTreeMap, convert::TryFrom};
+
+#[derive(Debug, Clone)]
+pub struct DynamicRegex {
+    pattern: String,
+    multiline: bool,
+    insensitive: bool,
+    global: bool,
+    compiled: Option<Result<regex::Regex>>,
+}
+
+impl DynamicRegex {
+    pub fn new(pattern: String, multiline: bool, insensitive: bool, global: bool) -> Self {
+        Self {
+            pattern,
+            multiline,
+            insensitive,
+            global,
+            compiled: None,
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn is_global(&self) -> bool {
+        self.global
+    }
+
+    pub fn compile(&mut self) -> Result<&regex::Regex> {
+        if self.compiled.is_none() {
+            self.compiled = Some(
+                regex::RegexBuilder::new(&self.pattern)
+                    .case_insensitive(self.insensitive)
+                    .multi_line(self.multiline)
+                    .build()
+                    .map_err(|err| format!("invalid regex {}", err)),
+            );
+        }
+
+        self.compiled
+            .as_ref()
+            // We know this unwrap is safe because we have just populated the Option above.
+            .unwrap()
+            .as_ref()
+            .map_err(|err| err.to_string())
+    }
+}
+
+/// Our dynamic regex equality shouldn't rely on the compiled value
+/// as this is largely an implementation detail.
+/// Plus regex::Regex doesn't implement PartialEq.
+impl PartialEq for DynamicRegex {
+    fn eq(&self, other: &Self) -> bool {
+        self.pattern == other.pattern
+            && self.multiline == other.multiline
+            && self.insensitive == other.insensitive
+            && self.global == other.global
+    }
+}
+
+impl TryFrom<BTreeMap<String, Value>> for DynamicRegex {
+    type Error = String;
+
+    /// Create a regex from a map containing fields :
+    /// pattern - The regex pattern
+    /// flags   - flags including i => Case insensitive, g => Global, m => Multiline.
+    fn try_from(map: BTreeMap<String, Value>) -> std::result::Result<Self, Self::Error> {
+        let pattern = map
+            .get("pattern")
+            .ok_or_else(|| "field is not a regular expression".to_string())
+            .and_then(|value| match value {
+                Value::Bytes(ref bytes) => Ok(String::from_utf8_lossy(bytes)),
+                _ => Err("regex pattern is not a valid string".to_string()),
+            })?
+            .to_string();
+
+        let (global, insensitive, multiline) = match map.get("flags") {
+            None => (false, false, false),
+            Some(Value::Array(ref flags)) => {
+                flags
+                    .iter()
+                    .fold((false, false, false), |(g, i, m), flag| match flag {
+                        v if v == &Value::from(Bytes::from_static(b"g")) => (true, i, m),
+                        v if v == &Value::from(Bytes::from_static(b"i")) => (g, true, m),
+                        v if v == &Value::from(Bytes::from_static(b"m")) => (g, i, true),
+                        _ => (g, i, m),
+                    })
+            }
+            Some(_) => return Err("regular expression flags is not an array".to_string()),
+        };
+
+        Ok(DynamicRegex::new(pattern, multiline, insensitive, global))
+    }
+}
--- a/src/mapping/query/dynamic_regex.rs
+++ b/src/mapping/query/dynamic_regex.rs
@@ -106,6 +106,28 @@ impl TryFrom<Value> for DynamicRegex {
     }
 }
 
+impl From<DynamicRegex> for Value {
+    fn from(regex: DynamicRegex) -> Self {
+        let mut map = BTreeMap::new();
+        map.insert("pattern".to_string(), Value::from(regex.pattern));
+
+        let mut flags = Vec::new();
+        if regex.insensitive {
+            flags.push(Value::from(Bytes::from_static(b"i")));
+        }
+        if regex.multiline {
+            flags.push(Value::from(Bytes::from_static(b"m")));
+        }
+        if regex.global {
+            flags.push(Value::from(Bytes::from_static(b"g")));
+        }
+
+        map.insert("flags".to_string(), Value::from(flags));
+
+        Value::from(map)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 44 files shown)


src/mapping/query/function/split.rs
```diff
@@ -0,0 +1,247 @@
...
+            limit,
+        }
+    }
+}
+
+lazy_static! {
+    static ref I_FLAG: Value = Value::from(Bytes::from("i"));
+    static ref G_FLAG: Value = Value::from(Bytes::from("g"));
+    static ref M_FLAG: Value = Value::from(Bytes::from("m"));
+}
```
JeanMertz, line 33: Can't we use `char`/`u8` here?

E.g., use the `Bytes::from_static` const method, and potentially not use `lazy_static` but call `Value::from` at the relevant location (the overhead would be non-existent).

src/mapping/query/function/split.rs
```diff
@@ -0,0 +1,247 @@
...
+
+/// Create a regex from a map containing fields :
+/// pattern - The regex pattern
+/// flags   - flags including i => Case insensitive, g => Global, m => Multiline.
+fn regex_from_map(mut map: BTreeMap<String, Value>) -> Result<(regex::Regex, bool)> {
```
JeanMertz, line 38: I feel like we can move this into a newtype, and have it impl `From` etc. for a more conventional internal API.

src/mapping/query/function/split.rs
```diff
@@ -0,0 +1,247 @@
...
+                        _ => (g, i, m),
+                    });
+
+            let pattern = String::from_utf8_lossy(&bytes);
+            let regex = regex::RegexBuilder::new(&pattern)
+                .case_insensitive(insensitive)
+                .multi_line(multi_line)
+                .build()
+                .map_err(|err| format!("invalid regex {}", err))?;
+            Ok((regex, global))
```
JeanMertz, line 66: If we use a newtype as suggested, we can have an `is_global(&self)` method, which removes the need to return a tuple in whatever function we create to return the compiled regex.

src/mapping/query/function/split.rs
```diff
@@ -0,0 +1,247 @@
...
+            }
+            Value::Map(pattern) => {
+                let (regex, _global) = regex_from_map(pattern)?;
+                Ok(match &limit {
+                    Some(ref limit) => to_value(Box::new(regex.splitn(&string, *limit))),
+                    None => to_value(Box::new(regex.split(&string))),
+                })
+            }
+            _ => Err("invalid pattern".to_string()),
+        }
```
JeanMertz, line 112: I think this is the part we'd eventually want to remove into a separate `impl Function for LazyRegex`, or whatever newtype we'll end up with.

src/mapping/query/function/split.rs
```diff
@@ -0,0 +1,247 @@
...
+    fn parameters() -> &'static [Parameter] {
+        &[
+            Parameter {
+                keyword: "value",
+                accepts: |v| matches!(v, Value::Bytes(_)),
+                required: true,
+            },
+            Parameter {
+                keyword: "pattern",
+                accepts: |v| matches!(v, Value::Bytes(_)),
```
JeanMertz, line 124: This would have to be updated to also accept `Value::Map`.

src/mapping/query/function/split.rs
```diff
@@ -1,8 +1,86 @@
...
+                    .multi_line(self.multiline)
+                    .build()
+                    .map_err(|err| format!("invalid regex {}", err)),
+            );
+        }
+
+        self.compiled
+            .as_ref()
+            // We know this unwrap is safe because we have just populated the Option above.
+            .unwrap()
```
JeanMertz, line 44: You should be able to use `get_or_insert` here, to achieve the same without having to unwrap.

src/mapping/query/function/split.rs
```diff
@@ -1,8 +1,86 @@
...
+            .as_ref()
+            .map_err(|err| err.to_string())
+    }
+}
+
+impl TryFrom<BTreeMap<String, Value>> for DynamicRegex {
```
JeanMertz, line 50: If we make this `TryFrom<Value> for DynamicRegex`, we can have both the `Value::Map` and `Value::Bytes` case be handled by this conversion, and have `DynamicRegex` be useable in both situations.

It seems like this can reduce a bunch of duplication, possibl?

JeanMertz (commented):
Sweet! Left two more comments, but I like where this is heading.

src/mapping/query/function/split.rs
FungusHumungus: Oh, nice!

src/mapping/query/function/split.rs
FungusHumungus: I'm wondering if we do want to create a `DynamicRegex` from a `Value::Bytes`? In a function such as `split`, if you split by a string, then that means you want to split by that exact string and not a regex.

So, 

`split("apple.pear", /./)`

will be different to

`split("apple.pear", ".")`

JeanMertz: No sorry, I didn’t mean creating a dynamic regex from a string. I meant using a singular type to represent both a dynamic and static regex, to allow consolidating some logic.

Basically a `CoW` type, where a static regex is used if one is stored, or else the dynamic regex is compiled to a static one (and served from then on if the regex is requested from the type).

src/mapping/query/dynamic_regex.rs
```diff
@@ -0,0 +1,95 @@
...
+use bytes::Bytes;
+use std::{collections::BTreeMap, convert::TryFrom};
+
+#[derive(Debug, Clone)]
+pub struct DynamicRegex {
+    pattern: String,
+    multiline: bool,
+    insensitive: bool,
+    global: bool,
+    compiled: Option<Result<regex::Regex>>,
```
JeanMertz, line 11: I would assume whatever function compiles the regex, return an error, so that we only store valid regex's here?

src/mapping/query/function/format_timestamp.rs
```diff
@@ -18,20 +18,21 @@ impl FormatTimestampFn {
...
             .execute(ctx)
-            .map(|v| match v {
+            .map(|v| match v.into() {
                 Value::Timestamp(ts) => ts,
-                _ => unexpected_type!(v),
+                v => unexpected_type!(v),
             })
             .and_then(|ts| try_format(&ts, &format))
             .map(Value::from)
+            .map(Into::into)
```
JeanMertz, line 35: Can we add an `impl<T: Into<Value>> From T for QueryValue`? That way we can directly go from any `T: Into<Value>` to `QueryValue`.

If not, perhaps we should just implement a custom `QueryValue::from_value`, which would do the same.

src/mapping/query/function/format_timestamp.rs
```diff
@@ -18,20 +18,21 @@ impl FormatTimestampFn {
 }
 
 impl Function for FormatTimestampFn {
-    fn execute(&self, ctx: &Event) -> Result<Value> {
-        let format = match self.format.execute(ctx)? {
+    fn execute(&self, ctx: &Event) -> Result<QueryValue> {
+        let format = match self.format.execute(ctx)?.into() {
```
JeanMertz, line 22: While we're updating these, I suggest converting them to `required!` or `optional!`, as that will reduce all the `into` calls we're doing right now?

Although we could also defer that to a follow-up PR.

src/mapping/query/function/mod.rs
```diff
@@ -228,16 +233,18 @@ impl Argument {
...
-                self.parameter.keyword
-            ));
+        if let QueryValue::Value(ref value) = value {
+            if !(self.parameter.accepts)(&value) {
+                return Err(format!(
+                    "invalid argument type '{}' for parameter '{}'",
+                    value.kind(),
+                    self.parameter.keyword
+                ));
+            }
```
JeanMertz, line 247: Would be good to add a `TODO` here w.r.t. also checking non `Value` kinds (e.g. `Regex`, etc).

Perhaps make the `if let` into a `match`, something like:

```rust
match &value {
	QueryValue::Value(value) => { ... },
	QueryValue::Regex => { /* TODO: ... */ },
}
```

src/mapping/query/dynamic_regex.rs
FungusHumungus: I left the `Result` in so if it has an invalid regex the error is stored here. Otherwise on an invalid regex it would just compile it every time it is accessed.

src/mapping/query/function/format_timestamp.rs
FungusHumungus: The `impl<T: Into<Value.>` keeps giving *conflicting implementations* errors unfortunately, but a `QueryValue::from_value` seems to work nicely.

FungusHumungus:
Following a conversation with @binarylogic  there is no need to allow for dynamic regexes. All regexes need to be specified in the remap script so they can be compiled at load time (#4503). 

This saves having to do further work here to allow for dynamic regexes. I will just spend a little bit of time tidying up what we have here to ensure that when a regex is created it is compiled right away. This will be done during parsing, which will ensure that it never needs compiling more than once.

JeanMertz (approved):
Great work @FungusHumungus 🙌 

I'm not entirely convinced using `QueryValue` is the right approach, as it causes a lot of indirection, and technically allows functions to return non-`Value` types (which we don't, and can't do).

But, given that this unblocks regular expression argument support — which is something we really need — and everything except the added regex Pest rules is internal logic we can change later, I'm happy to see this merged 👍

FungusHumungus:
Great, thanks. I agree, it does feel a bit clunky. I think it comes from having two different types of types. I can't think of a better way around it. If you have any ideas, I'm all ears! Maybe something will come after we have used it for a bit..

PR #4364, opened 2020-10-04
chore(ci): Run GH Actions Checks job in container

jamtur01:
This runs the Checks job in a Docker container to avoid the 6-10 minutes needed for bootstrap.

We've also moved the creation and management of the environment image (`timberio/ci_image`) to the Docker Hub. This is because of the constraint of not being able to use Docker images from GH's Package Registry inside GH Actions.

We've also addressed an annotation on the job that appeared today regarding the [security of GH Actions add_path](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).


jamtur01:
bors ping

jamtur01:
bors ping

jamtur01:
bors ping

jamtur01:
bors try

jamtur01:
bors ping

jamtur01:
bors try

.github/workflows/prs.yml
```diff
@@ -421,6 +421,8 @@ jobs:
   compute-k8s-test-plan:
     name: Compute K8s test plan
     runs-on: ubuntu-latest
+    needs: changes
+    if: ${{ needs.changes.outputs.source == 'true' }}
```
MOZGIII, line 425: We also need to run these for `needs.changes.outputs.helm`, and also for any docker or packaging configuration changes.

scripts/environment/bootstrap-macos-10.sh
```diff
@@ -6,5 +6,5 @@ brew install ruby coreutils cuelang/tap/cue
 echo "export PATH=\"/usr/local/opt/ruby/bin:\$PATH\"" >> "$HOME/.bash_profile"
 
 if [ -n "${CI-}" ] ; then
-  echo "::add-path::/usr/local/opt/ruby/bin"
+  echo "/usr/local/opt/ruby/bin" >> "$GITHUB_PATH"
```
MOZGIII, line 9: Interesting, does this work? I didn't encounter it in the docs before...

jamtur01: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

.github/workflows/environment.yml
```diff
@@ -17,11 +17,11 @@ jobs:
...
-      - run: docker login https://docker.pkg.github.com -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }}
+      - run: docker login -u ${{ secrets.DOCKER_USER }} -p ${{ secrets.DOCKER_PASSWORD }}
       - uses: actions/checkout@v2
       - run: make ci-sweep
       - name: prepare environment
         run: make environment-prepare
       - name: push image
         if: github.event_name != 'workflow_dispatch'
-        run: docker push docker.pkg.github.com/timberio/vector/environment:${{ github.sha }}
+        run: docker push timberio/ci_image:${{ github.sha }}
```
MOZGIII, line 27: What's the idea behind using the `${{ github.sha }}`? I couldn't figure it out before, and I don't get it now :(

If jobs that build and consume the image start simultaneously, and the image if scoped to the commit sha - how can it be possibly available in time for the consuming job to not fail?

I think we should either order the environment jobs before the rest of the jobs via `needs` (and move them to the same workflow file), or use some other way to stall the kick-off of the jobs that consume the image until the image is available.

A decent alternative would be to build the environment images nightly and push with a sliding tag `nightly`. Then at the consumer location, allow using the image tagged with the branch name, or, if no branch-specific image was pushed - fallback to `nightly`. This was people will be able to push their own custom `ci_image` tags with per-branchs specifics. We can also allow them building them in the cloud, via manually triggering the environment workflow via `workflow_dispatch` event, allowing to build an image off of a specific branch, and push it with the branch name instead of `nightly`.

scripts/environment/bootstrap-macos-10.sh
MOZGIII: Nice, thx!

.github/workflows/prs.yml
```diff
@@ -422,7 +424,7 @@ jobs:
     name: Compute K8s test plan
     runs-on: ubuntu-latest
     needs: changes
-    if: ${{ needs.changes.outputs.source == 'true' }}
+    if: ${{ needs.changes.outputs.source == 'true' }} || ${{ needs.changes.outputs.helm == 'true' }} || ${{ needs.changes.outputs.docker == 'true' }}
```
MOZGIII, line 427:
```suggestion
    if: needs.changes.outputs.source == 'true' || needs.changes.outputs.helm == 'true' || needs.changes.outputs.docker == 'true'
```

From the https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions:

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression. For more information about if conditionals, see "Workflow syntax for GitHub Actions."

I think the `${{ }}` and `||` like it was are a syntax error.

.github/workflows/environment.yml
jamtur01: That's exactly what we want to avoid by using a container - the time needed to bootstrap or in the case you're proposing to build the container. Github Actions will pull `timberio/ci_image:latest`. This is the last image built off `master` by the `environment` workflow. I think it's very unlikely a CI job will fail because the `ci_image` is out of date. Everything vector-related gets added by `cargo` or elsewhere.

.github/workflows/prs.yml
jamtur01: If they were a syntax error the job would have failed to run. Removing the expression syntax is optional.

.github/workflows/benches.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 env:
   VERBOSE: true
-  ENVIRONMENT_UPSTREAM: docker.pkg.github.com/timberio/vector/environment:${{ github.sha }}
+  ENVIRONMENT_UPSTREAM: timberio/ci_image:${{ github.sha }}
```
MOZGIII, line 10: Following up from https://github.com/timberio/vector/pull/4364/files#r499995758, why is this set to `timberio/ci_image:${{ github.sha }}` instead of `timberio/ci_image:latest`?

.github/workflows/environment.yml
MOZGIII: Good, I see how what you explained above can work.

But I'm still not getting why, instead of pushing `timberio/ci_image:latest` here, we push `timberio/ci_image:${{ github.sha }}`.

jamtur01: Largely because if we have a bad image we can easily roll back by deleting that tag without needing to re-roll the image.

jamtur01:
The force-push is me rewriting the commit-msg to make DCO happy.

jamtur01:
Failed checks are bootstrapping failures.

.github/workflows/environment.yml
MOZGIII: But docker doesn’t make these images accessible as `:latest`, does it? There’s no automatic rule there on a server side to exposed the last uploaded image with the `:latest` tag, is there?

PR #4381, opened 2020-10-05
chore(ci): Added clippy to all source changes

jamtur01:


MOZGIII:
I wonder if we should also add `make check` there? Afaik, `clippy` will cover that as well, but we have slightly different flags set at `Makefile` for `make check` and `make clippy`

MOZGIII (approved):

jamtur01:
I'm actually going to close this in favor of refactoring how changes are tested so we're testing for them all in one place and not between two jobs.

PR #4383, opened 2020-10-05
enhancement(observability): Bidirectional source/transform/sink GraphQL types

leebenson:
Updates the GraphQL schema to accommodate a `Topology` interface, implementing `Source` and the new types, `Transform` and `Sink`.

This enables:

- Pulling back all topology resources as a flat list
- Pulling back the relevant resources only - e.g. sources, transforms or sinks
- Starting from resource A and getting to resource B in any direction, by following its relevant input/output field types

Relationship directions follow a name rather than a type - e.g. `Transform.sources` represents _input_, but `Source.transforms` represents _output_.

### Demo

https://www.loom.com/share/8bbb5b7c478e49b29023c395654a3333

## Schema

The relevant schema now looks like this:

```gql
enum SourceOutputType {
  ANY
  LOG
  METRIC
}

type Source implements Topology {
  # Source name
  name: String!

  # Source output type
  outputType: SourceOutputType!

  # Transform outputs
  transforms: [Transform!]!

  # Sink outputs
  sinks: [Sink!]!
}

type Transform implements Topology {
  # Transform name
  name: String!

  # Source inputs
  sources: [Source!]!

  # Sink outputs
  sinks: [Sink!]!
}

type Sink implements Topology {
  # Sink name
  name: String!

  # Source inputs
  sources: [Source!]!

  # Transform inputs
  transforms: [Transform!]!
}

type Query {
  # Configured Topology (source/transform/sink)
  topology: [Topology!]!

  # Configured sources
  sources: [Source!]!

  # Configured transforms
  transforms: [Transform!]!

  # Configured sinks
  sinks: [Sink!]!
}
```

## Example queries

Run these via the playground at http://localhost:8686/playground:

### Topological resources by name + type (flat list)

Use-case: The web UI that wants to parse _all_ resources in a single pass:

```gql
{
  topology {
    __typename
    name
  }
}
```

### All resources, with resource-specific fields

Use-case: You want all resources, including fields that are relevant to each type, without issuing separate requests for that data.

In this case, if the returned resource is a source, we also get its transforms; if it's a sink, we get its sources.

```gql
{
  topology {
    __typename
    name
    ... on Source {
      outputType
      transforms {
        name
      }
    }
    ... on Sink {
      sources {
        name
      }
    }
  }
}
```

### Resource only (in this case, sources)

Use-case: If you want to return a specific resource type

```gql
{
  sources {
    name
    outputType
  }
}
```

### Nested data (either direction)

Use-case: You want a resource, and all of its inputs/outputs based on the type of resource. In this case, we're getting all transforms, an `sources` will be an input on each:

```gql
{
  transforms {
    name
    sources {
      name
      outputType
    }
  }
}
```


leebenson:
@sghall - the 'flat' case (`{ topology }`) is what I had in mind for the main UI topology page, where we don't know each of time how many resources we're dealing with, or the level of nesting we might want to do for each.

The individual `{ sources }`-type queries are the basic building blocks for narrowing down to individual sources.

Next, I'd like to add `source(input: { name })` and other 'singular' types.

Also some filtering/sorting, like we had in Alloy, although I'm not sure exactly what the requirements will be yet. If it makes sense for us to have data in a particular way as we start piecing together the relevant pages, just let me know.

The very next move will be attempting to sample stats/metrics from these types, which will form the basis of #3211

src/api/schema/topology.rs
```diff
@@ -29,44 +32,232 @@ pub struct SourceData {
...
+
+    /// Transform outputs
+    async fn transforms(&self) -> Vec<Transform> {
+        filter_topology(|(_name, topology)| match topology {
+            Topology::Transform(t) if t.0.inputs.contains(&self.0.name) => Some(t.clone()),
+            _ => None,
+        })
+    }
+
+    //noinspection DuplicatedCode
```
Hoverbear, line 55: I don't think this is required.

src/api/schema/topology.rs
```diff
@@ -29,44 +32,232 @@ pub struct SourceData {
...
+pub struct Transform(TransformData);
+
+#[Object]
+impl Transform {
+    /// Transform name
+    async fn name(&self) -> String {
+        self.0.name.clone()
+    }
+
+    //noinspection DuplicatedCode
```
Hoverbear, line 81: Similarly, not required.

src/api/schema/topology.rs
```diff
@@ -29,44 +32,232 @@ pub struct SourceData {
...
+                Some(t) => match t {
+                    Topology::Source(s) => Some(s.clone()),
+                    _ => None,
+                },
+                _ => None,
+            })
+            .collect()
+    }
+
+    //noinspection DuplicatedCode
```
Hoverbear, line 97: Ditto

src/api/schema/topology.rs
```diff
@@ -29,44 +32,232 @@ pub struct SourceData {
...
+pub struct Sink(SinkData);
+
+#[Object]
+impl Sink {
+    /// Sink name
+    async fn name(&self) -> String {
+        self.0.name.clone()
+    }
+
+    //noinspection DuplicatedCode
```
Hoverbear, line 123: Ditto

src/api/schema/topology.rs
leebenson: Thanks, wasn’t sure how we felt about lint-disabling comments.

These are IDEA specific so they’re probably not very useful to everyone. Will nix.

src/api/schema/topology.rs
```diff
@@ -29,44 +32,232 @@ pub struct SourceData {
...
+    async fn sinks(&self) -> Vec<Sink> {
+        filter_topology(|(_name, topology)| match topology {
+            Topology::Sink(s) if s.0.inputs.contains(&self.0.name) => Some(s.clone()),
+            _ => None,
+        })
+    }
+}
+
+#[derive(Clone)]
+pub struct TransformData {
```
Hoverbear, line 66: These `SinkData` and `SourceData` types seem like they'd fit well as a `Data<T>` ?

leebenson: I kept them separate as it’s possible they’ll diverge as requirements evolve, but also maybe not. I’ll unify for now and split out later if needed 👍

src/api/schema/topology.rs
```diff
@@ -29,44 +32,222 @@ pub struct SourceData {
...
+            name.to_string(),
+            Topology::Sink(Sink(InputsData {
+                name: name.to_owned(),
+                inputs: sink.inputs.clone(),
+            })),
+        );
+    }
+
+    // override the old hashmap
+    *TOPOLOGY.write().unwrap() = new_topology
```
Hoverbear, line 252: I'd also like to see this possible error have an expect statement if you could!

Hoverbear (approved):
The rest, other than some diagnostics on what should be unreachable scenarios, looks good!

src/api/schema/topology.rs
```diff
@@ -29,44 +32,222 @@ pub struct SourceData {
...
+    /// Configured sinks
+    async fn sinks(&self) -> Vec<Sink> {
+        get_sinks()
     }
 }
 
+fn filter_topology<T>(map_func: impl Fn((&String, &Topology)) -> Option<T>) -> Vec<T> {
+    TOPOLOGY
+        .read()
+        .unwrap()
```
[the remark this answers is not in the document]
leebenson: Thanks, good shout on `.expect()` in lieu of `.unwrap()`. I'll make that a habit.

leebenson:
Merging this, but noting the linting errors already on master.

PR #4402, opened 2020-10-06
chore(ci): Fix missing AUTOINSTALL: true env variable

jamtur01:


MOZGIII:
Looks like `RUST_TEST_THREADS: 1` is also gone after https://github.com/timberio/vector/pull/4400, we probably should add it too.
Same for `ENVIRONMENT_UPSTREAM: docker.pkg.github.com/timberio/vector/environment:${{ github.sha }}`, but the value is suspicious, and it might've been broken either way.

jamtur01:
We don't use the docker image environment variable in this workflow.

MOZGIII (approved):
I guess it should fix the issue, but without the CI jobs running it's hard to tell.
We probably should allow all the CI jobs to run if the CI config changes - just to make sure it works :smiley:

PR #4408, opened 2020-10-06
enhancement(windows platform): correctly handle service restart

oktal:
Now properly handle `service restart` subcommand

### DIFF
--- a/src/vector_windows.rs
+++ b/src/vector_windows.rs
@@ -113,25 +113,23 @@ pub mod service_control {
 
     pub fn control(service_def: &ServiceDefinition, action: ControlAction) -> crate::Result<()> {
         match action {
-            ControlAction::Start => start_service(&service_def).map_err(|e| Error::from(e).into()),
-            ControlAction::Stop => stop_service(&service_def).map_err(|e| Error::from(e).into()),
-            ControlAction::Restart => restart_service(&service_def).map_err(Into::into),
-            ControlAction::Install => {
-                install_service(&service_def).map_err(|e| Error::from(e).into())
-            }
-            ControlAction::Uninstall => uninstall_service(&service_def).map_err(Into::into),
+            ControlAction::Start => start_service(&service_def),
+            ControlAction::Stop => stop_service(&service_def),
+            ControlAction::Restart => restart_service(&service_def),
+            ControlAction::Install => install_service(&service_def),
+            ControlAction::Uninstall => uninstall_service(&service_def),
         }
     }
 
-    fn start_service(service_def: &ServiceDefinition) -> Result<()> {
+    fn start_service(service_def: &ServiceDefinition) -> crate::Result<()> {
         let service_access = ServiceAccess::QUERY_STATUS | ServiceAccess::START;
         let service = open_service(&service_def, service_access)?;
-        let service_status = service.query_status()?;
+        let service_status = service.query_status().map_err(Error::from)?;
 
         if service_status.current_state != ServiceState::StartPending
             || service_status.current_state != ServiceState::Running
         {
-            service.start(&[] as &[OsString])?;
+            service.start(&[] as &[OsString]).map_err(Error::from)?;
             emit!(WindowsServiceStart {
                 name: &*service_def.name.to_string_lossy(),
                 already_started: false,
@@ -146,15 +144,15 @@ pub mod service_control {
         Ok(())
     }
 
-    fn stop_service(service_def: &ServiceDefinition) -> Result<()> {
+    fn stop_service(service_def: &ServiceDefinition) -> crate::Result<()> {
         let service_access = ServiceAccess::QUERY_STATUS | ServiceAccess::STOP;
-        let service = open_service(&service_def, service_access)?;
-        let service_status = service.query_status()?;
+        let service = open_service(&service_def, service_access).map_err(Error::from)?;
+        let service_status = service.query_status().map_err(Error::from)?;
 
         if service_status.current_state != ServiceState::StopPending
             || service_status.current_state != ServiceState::Stopped
         {
-            service.stop()?;
+            service.stop().map_err(Error::from)?;
             emit!(WindowsServiceStop {
                 name: &*service_def.name.to_string_lossy(),
                 already_stopped: false,
@@ -172,13 +170,13 @@ pub mod service_control {
     fn restart_service(service_def: &ServiceDefinition) -> crate::Result<()> {
         let service_access =
             ServiceAccess::QUERY_STATUS | ServiceAccess::START | ServiceAccess::STOP;
-        let service = open_service(&service_def, service_access)?;
-        let service_status = service.query_status()?;
+        let service = open_service(&service_def, service_access).map_err(Error::from)?;
+        let service_status = service.query_status().map_err(Error::from)?;
 
         if service_status.current_state == ServiceState::StartPending
             || service_status.current_state == ServiceState::Running
         {
-            service.stop()?;
+            service.stop().map_err(Error::from)?;
         }
 
         let timeout = Duration::from_secs(10);
@@ -198,14 +196,14 @@ pub mod service_control {
             .into());
         }
 
-        service.start(&[] as &[OsString])?;
+        service.start(&[] as &[OsString]).map_err(Error::from)?;
         emit!(WindowsServiceRestart {
             name: &*service_def.name.to_string_lossy()
         });
         Ok(())
     }
 
-    fn install_service(service_def: &ServiceDefinition) -> Result<()> {
+    fn install_service(service_def: &ServiceDefinition) -> crate::Result<()> {
         let manager_access = ServiceManagerAccess::CONNECT | ServiceManagerAccess::CREATE_SERVICE;
         let service_manager = ServiceManager::local_computer(None::<&str>, manager_access)?;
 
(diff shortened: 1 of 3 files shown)


src/vector_windows.rs
```diff
@@ -121,6 +167,41 @@ pub mod service_control {
...
+    fn restart_service(service_def: &ServiceDefinition) -> std::result::Result<(), Error> {
+        let service_access =
+            ServiceAccess::QUERY_STATUS | ServiceAccess::START | ServiceAccess::STOP;
+        let service = open_service(&service_def, service_access)?;
+        let service_status = service.query_status()?;
+
+        if service_status.current_state == ServiceState::StartPending
+            || service_status.current_state == ServiceState::Running
+        {
+            service.stop()?;
```
Hoverbear, line 179: So this is doing like an `HUP` on Linux? I know on Linux we try to ensure the API server etc don't need to restart when we get this signal.

src/vector_windows.rs
```diff
@@ -152,7 +233,7 @@ pub mod service_control {
         Ok(())
     }
 
-    fn uninstall_service(service_def: &ServiceDefinition) -> Result<()> {
+    fn uninstall_service(service_def: &ServiceDefinition) -> std::result::Result<(), Error> {
```
Hoverbear, line 236:
```suggestion
    fn uninstall_service(service_def: &ServiceDefinition) -> crate::Result<()> {
```

I think this would be preferred.

src/vector_windows.rs
```diff
@@ -66,11 +108,15 @@ pub mod service_control {
         }
     }
 
-    pub fn control(service_def: &ServiceDefinition, action: ControlAction) -> Result<()> {
+    pub fn control(
+        service_def: &ServiceDefinition,
+        action: ControlAction,
+    ) -> std::result::Result<(), Error> {
```
Hoverbear, line 114: I think here a `crate::Result` is preferred still.

src/vector_windows.rs
oktal: I don't think that it maps to a particular signal on Linux. Windows does not have a notion of signal.

It actually calls the winsvc [ControlService](https://docs.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-controlservice) function with a `SERVICE_CONTROL_STOP` control code, which will then send a stop notification to the service (the event handler of the service will be notified with a `ServiceControlStop`.

It won't restart when you only send a Stop notification.

Hoverbear: So does this mean the API server would go down during the restart? I know @leebenson has been fairly careful to make sure the API server stands persisting over restarts.

oktal: Oh ok, I see what you mean.

Yes, the API server will also restart when issuing a restart. What you describe would be a reload then (the `SignalTo::Reload`).

Windows has a `SERVICE_CONTROL_PARAMCHANGE` control code that we could use to trigger a reload. I would personnally keep the behavior of `service restart` as-is, and maybe add a `service reload` sub command which would then use `SERVICE_CONTROL_PARAMCHANGE` to instruct the service to reload its parameters.

What do you think ?

leebenson: I'm not especially familiar with Windows, but if `service restarts` is intended to mean 'hard stop, then start', I think it's fine to also bring down the API. This should only happen where the user wants to tear down the entire Vector process, and start a new one in its place.

`service reload` sounds equivalent to `HUP`; in the situation where a config is being reloaded and topology needs to tear down and start again, the API should remain up. It's been specifically designed to allow connecting API clients to remain active while hot config changes are loaded. This facilitates our eventual work with a web GUI that will allow users to make live config changes, validate config, and see toplogy graphs respond to updates in real-time.

Does that fit with your model here, @oktal?

oktal: Yup, that sounds good to me.

`service restart` -> triggers a hard stop, then start (you might want to do that upon version update, and not only config change)
`service reload` -> only trigger a reload of the running topology, without bringing down the entire service

I would keep the changes separate and implement `reload` on a separate `windows-service-reload` branch if that sounds good to you @Hoverbear :)

Hoverbear: Sounds great! Let's get this merged then!

src/vector_windows.rs
```diff
@@ -21,20 +21,65 @@ pub mod service_control {
...
+    pub enum Error {
+        Service(windows_service::Error),
+        PollTimeout {
+            state: ServiceState,
+            expected_state: ServiceState,
+            timeout: Duration,
+        },
+    }
+
+    impl std::error::Error for Error {}
```
Hoverbear, line 42: We typically use `crate::Error` :)

src/vector_windows.rs
```diff
@@ -98,15 +144,15 @@ pub mod service_control {
         Ok(())
     }
 
-    fn stop_service(service_def: &ServiceDefinition) -> Result<()> {
+    fn stop_service(service_def: &ServiceDefinition) -> crate::Result<()> {
         let service_access = ServiceAccess::QUERY_STATUS | ServiceAccess::STOP;
-        let service = open_service(&service_def, service_access)?;
-        let service_status = service.query_status()?;
+        let service = open_service(&service_def, service_access).map_err(Error::from)?;
```
Hoverbear, line 149: Usually it's enough to use `?` here, since that will call `.into()` for you.

src/vector_windows.rs
```diff
@@ -121,7 +167,43 @@ pub mod service_control {
...
+        {
+            service.stop().map_err(Error::from)?;
+        }
+
+        let timeout = Duration::from_secs(10);
+        let poll_status = poll_state(
+            &service,
+            ServiceState::Stopped,
+            timeout,
+            Duration::from_secs(1),
```
Hoverbear, line 187: It seems we can use a wait hint here: https://docs.rs/windows-service/0.3.0/windows_service/service/struct.ServiceStatus.html#structfield.wait_hint ?

src/vector_windows.rs
oktal: So `crate::Error` is a `pub type Error = Box<dyn std::error::Error + Send + Sync + 'static>;` which means that I can use anything that implements `std::error::Error`, or am I missing something ?

I wanted custom and clear error message so that I created my own `Error` type.

src/vector_windows.rs
oktal: Two comments on this:

* Unfortunately, it looks like the `windows-service` crate does not let you specify a `wait_hint` when stopping the service. It uses a zeroe'd `SERVICE_STATUS` under the hood.
* I think I still need to poll the state of the service to determine when it is safe to start it back.

Note that I'm also using it in `uninstall` to wait for the service to be stopped before deleting it from the Service Control Manager.

src/vector_windows.rs
```diff
@@ -121,7 +167,43 @@ pub mod service_control {
...
+    fn restart_service(service_def: &ServiceDefinition) -> crate::Result<()> {
+        let service_access =
+            ServiceAccess::QUERY_STATUS | ServiceAccess::START | ServiceAccess::STOP;
+        let service = open_service(&service_def, service_access).map_err(Error::from)?;
+        let service_status = service.query_status().map_err(Error::from)?;
+
+        if service_status.current_state == ServiceState::StartPending
+            || service_status.current_state == ServiceState::Running
+        {
+            service.stop().map_err(Error::from)?;
```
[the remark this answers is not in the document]
oktal: Unclear, what do you think we should do when the service exits with a non-zero exit code ? Erroring out or still restarting the service ?

src/vector_windows.rs
oktal: From what I've understood, in this particular case, `.into()` will actually `box` my error type since `crate::Error` is a `Box`, but won't map the error type from `windows-service` (which is a `windows_service::Error`) to my custom `Error` type, right ?

src/vector_windows.rs
Hoverbear: I think this is a good reasoning. :)

Hoverbear: So you might want to use snafu like as shown in https://docs.rs/snafu/0.6.9/snafu/guide/the_macro/index.html.

If you give a look for `Snafu` in the vector codebase you can see how we often use it.

oktal: Thanks for that, fixed in https://github.com/timberio/vector/pull/4408/commits/ae2dd39304d0e32d4515222b9cb2800aa19fd49b

Note that I had to use a custom implementation of `Display` for the `windows_service::Error` enum since I wanted to be able to extract the Winapi error instead of the generic "IO error in Winapi call" error message.

I'm not sure if it's the right way to do it, let me know if you have a better idea :)

src/vector_windows.rs
Hoverbear: Okay! :) Very well considered reasoning, thank you.

src/vector_windows.rs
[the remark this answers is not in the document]
oktal: Makes sense. I added a `warn` message when the service stops with a non-zero exit code :+1:

Hoverbear (approved):

Hoverbear:
I think this broke our shutdown tests but these were not run by the CI. :(

Hoverbear:
Thanks for this @oktal =D

PR #4410, opened 2020-10-06
fix(apache_metrics source): Sanitize URL when tagging metrics

jszwedko:
Neglected this in the initial implementation. Thanks to @fanatid for
catching.

Resolves comment:
https://github.com/timberio/vector/pull/3704#discussion_r499801789


fanatid:
I think that sanitization step should be moved to `build` method. This will be better place for error and we will not need sanitize url on each scrape.

jszwedko:
@fanatid  I refactored the sanitization to an extension trait on `http::Uri` that simply serializes the URL as a `String` without the username/password, which is what we need in the various places we use the sanitized URL. This still happens every scrape, but avoids raising any errors. Given there is an HTTP request happening each scrape, the overhead seemed minimal. I'm curious to get feedback on this pattern though.

My initial approach to update this was just to introduce a wrapper type that contained both the original and the sanitized URL but this felt a bit awkward to work with.

fanatid (approved):

PR #4422, opened 2020-10-06
chore(external docs): Add options to `cue vet` command line

bruceg:
When a component instance is not completely specified, `cue vet`
produces the less than helpful error that `some instances are
incomplete; use the -c flag to show errors or suppress this message`.
Adding the `--concrete` option causes it to produce messages about which
specific fields were incomplete.

Additionally, the `--all-errors` causes `cue vet` to print all the
errors rather than just the first, reducing the number of
trial-and-error cycles necessary.

### DIFF
--- a/scripts/check-docs.sh
+++ b/scripts/check-docs.sh
@@ -21,7 +21,7 @@ if ! [ -x "$(command -v cue)" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


jamtur01 (approved):
LGTM

binarylogic (approved):

PR #4426, opened 2020-10-07
chore: Add settings for cue to .editorconfig

MOZGIII:


binarylogic (approved):

bruceg:
Is this really what we want? As far as I can determine, all of the existing cue files use 2 space indents and spaces instead of tabs. If I'm reading right, this change sets them up for 4 space indents and tabs only. Without going back and reformatting all the cue files, this will cause a real mess of confusion between tabs, spaces, and indent levels whenever somebody commits changes using this editor config.

While I'm not a big fan of of using tabs for indenting, it's not a strong preference. I just want the formatting to stay consistent to avoid the mess that can follow from a change like this.

MOZGIII:
`cue fmt` converts to this format, thus it's the obvious pick

bruceg:
I agree, if the standard formatting tool uses this format then it is the obvious pick. I was going off of the PRs I had already worked on.

… I see #4424 has been opened to make sure this happens in the future. My apologies for the noise.

MOZGIII:
No worries, totally reasonable!

PR #4429, opened 2020-10-07
enhancement(log_to_metric transform): Improve error messages

juchiast:
- Add field name to FieldNotFound error.
- Add field name and error detail to ParseFloatError.
- Clarify in TemplateError that we failed to parse a _template_.
- Previously, template rendering error prints "dropping event", this could be misleading because this transform tries many configs for each event, therefore other configs might successfully convert the event. Other errors in this transform do not have "dropping event".
- Fix render_tags silently ignores template rendering error. ~~I changed the code to return an error immediately, I'm not sure if this counts as a breaking change or not.~~ I emit a warning on missing keys error, and return on other errors (invalid template string in config).

### DIFF
--- a/src/transforms/log_to_metric.rs
+++ b/src/transforms/log_to_metric.rs
@@ -142,8 +142,15 @@ fn render_tags(
         Some(tags) => {
             let mut map = BTreeMap::new();
             for (name, value) in tags {
-                let tag = render_template(value, event)?;
-                map.insert(name.to_string(), tag);
+                match render_template(value, event) {
+                    Ok(tag) => {
+                        map.insert(name.to_string(), tag);
+                    }
+                    Err(TransformError::TemplateRenderError { missing_keys }) => {
+                        emit!(LogToMetricTemplateRenderError { missing_keys });
+                    }
+                    Err(other) => return Err(other),
+                }
             }
             if !map.is_empty() {
                 Some(map)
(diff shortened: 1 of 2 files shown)


juchiast:
Need to fix some tests..

src/transforms/log_to_metric.rs
```diff
@@ -113,46 +115,61 @@ impl LogToMetric {
...
+                field: field.clone(),
+            })?;
+        value
+            .to_string_lossy()
+            .parse()
+            .map_err(|error| TransformError::ParseFloatError {
+                field: field.clone(),
+                error,
+            })
+    }
```
bruceg, line 171: It doesn't appear this function closes over any of the parameters. Could it be moved out of `fn to_metric`?

src/transforms/log_to_metric.rs
```diff
@@ -258,21 +255,25 @@ fn to_metric(config: &MetricConfig, event: &Event) -> Result<Metric, TransformEr
...
+            let tags = render_tags(&set.tags, &event)?;
 
             Ok(Metric {
                 name,
                 timestamp,
                 tags,
                 kind: MetricKind::Incremental,
                 value: MetricValue::Set {
-                    values: vec![value].into_iter().collect(),
+                    values: std::iter::once(value).collect(),
```
bruceg, line 276: Good catch!

bruceg (approved):
Looks like a definite improvement! Just one question, not blocking.

src/transforms/log_to_metric.rs
juchiast: I moved it out!

juchiast:
bors merge

jamtur01:
@juchiast we've disabled Bors for now. Just merge stuff with the green button. Thanks!

juchiast:
@jamtur01  I was about to ask, thanks for the info!

PR #4431, opened 2020-10-07
feat(observability): `vector top`, v1

leebenson:
~Draft~ PR for initial plumbing for `vector top`.

In this first version, sources/transforms/sinks are dumped to the console in a table.

Follow-up work will include moving to subscriptions to auto-refresh stats in place (via `--refresh-interval` (default: 500ms)) and get metrics such as throughput.

This PR provides builds on recent GraphQL work, providing the CLI sub-command `vector top`

### Usage

```
vector top --url 0.0.0.0:8686
```

(`url` defaults to 127.0.0.1:8686 if omitted.)

Returning (for example):

```
 NAME             | TYPE      | EVENTS 
------------------+-----------+--------
 second_generator | source    |     12 
 hole             | sink      |     72 
 first_transform  | transform |     60 
 first_generator1 | source    |     60 

```

WIP toward #3211


binarylogic:
Nice to see this making progress!

> vector top --refresh-interval 200 --remote 0.0.0.0:8686 ~/Desktop/vector.toml

What's the `~/Desktop/vector.toml` part?

leebenson:
> What's the ~/Desktop/vector.toml part?

That's a path to a config file that I'm using to test (i.e. one that specifically has `api.enabled = true`). It currently uses the same `paths` option as other sub-commands, but an open question is whether we want to use the same `--config` switch as 'main'.

leebenson:
(Also worth noting that this will go away after dev, since it'll be observing an active instance and not the config directly!)

binarylogic:
Why do we need a config path at all? Shouldn't a connection to a running Vector instance be all that's required?

leebenson:
@binarylogic @Hoverbear - marked this for review, before this gets too big.

This first version dumps out the topology + aggregate number of events processed via `vector top`.

The next version will auto-update stats in place. More stats are coming.

Notes (2 mins): https://www.loom.com/share/0e4caeee762f4ca1b791d4cac998294a

Hoverbear:
Hmmm, I feel like we missed an event in the console sink. 😅 

image

Hoverbear:
I noted during testing if I remove a component from my config, it no longer appears in `top`, despite it having been existed during this run of Vector and the other stats persisting. I feel like this could lead to some confusion of where events were sourced from.

Do you think we could show `(tombstoned)` beside dead components instead?

Cargo.toml
```diff
@@ -262,6 +268,14 @@ api = [
   "async-graphql-warp",
 ]
 
+# API client
+api-client = [
+  "vector-api-client",
+  "prettytable-rs",
```
Hoverbear, line 274: Oh this crate was fun to use!

src/lib.rs
```diff
@@ -17,6 +17,10 @@ extern crate derivative;
 #[macro_use]
 extern crate pest_derive;
 
+#[cfg(feature = "api-client")]
+#[macro_use]
+extern crate prettytable;
```
Hoverbear, line 22: So you generally don't need to do this. You can do things like `#[prettytable::Thingy]` decorators and use `use prettytable::DerivableTrait` to import a custom derive. You may still need this though... Did you try without?

src/top.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl StatsWriter for LocaleWriter {
+    fn kb(&mut self, n: f64) -> String {
+        self.buf
+            .write_formatted(&(n as i64), &num_format::Locale::en);
+        self.buf.to_string()
+    }
+}
+
+fn new_formatter(humanize: bool) -> Box<dyn StatsWriter> {
```
Hoverbear, line 52: image

I am a bit curious how I can get `1.5` of an event?

src/top.rs
```diff
@@ -0,0 +1,124 @@
...
+pub struct Opts {
+    /// How often the screen refreshes (in milliseconds)
+    #[structopt(default_value = "500", short = "i", long)]
+    refresh_interval: i32,
+
+    #[structopt(short, long)]
+    url: Option<Url>,
+
+    #[structopt(short, long)]
+    humanize: bool,
```
Hoverbear, line 71: I am not sure we need to say "humanize". Seems `ls` uses `--human-readable` but I think that's quite long...

```suggestion
    human: bool,
```

Maybe?

Hoverbear (approved):
Overall this looks good, I noted a few potential improvements, but those are ok to be future work.

leebenson:
> I noted during testing if I remove a component from my config, it no longer appears in top, despite it having been existed during this run of Vector and the other stats persisting. I feel like this could lead to some confusion of where events were sourced from.
>
> Do you think we could show (tombstoned) beside dead components instead?

That's a nice idea. We can certainly do that in the client, by keeping the 'dead' topology resident in the table and just showing the last stats.

The problem currently is going to be restarting `vector top` later. If we query for topology _again_, currently the stuff that's gone away won't show up. We can fix that by making the config write-only.

But the real issue is handling renames (and I've just realised there's potentially a UX consideration here, depending how we want to treat it.)

Let's say we have `source1`. It generates 50 events. Initially, it looks like this:

```
 NAME             | TYPE      | EVENTS 
------------------+-----------+--------
 source1          | source    |     50

```

Then, we decide we like `source2` better. So we rename it. Now it looks like this:

```
 NAME             | TYPE      | EVENTS 
------------------+-----------+--------
 source1 ⚰️        | source    |     50
 source2          | source    |      0
```

So far, so good. There's no way for us to know that `source1` has just been renamed. It could be a rename. Or it could be that we deleted the old source, and just happened to use the exact same config -- albeit a different name.

Then, we decide actually, we _did_ like `source1` best after all. So after another 50 events, we rename it. Then it generates another 50 on top with its new name. The new output is:

```
 NAME             | TYPE      | EVENTS 
------------------+-----------+--------
 source1          | source    |    100
 source2 ⚰️        | source    |     50
```

Maybe this is what we want. But what if we _didn't_ rename it -- and instead, this is a brand new source that just happened to have the same name?

To the user, it looks like `source1` is the same source type, and was happily chugging away the whole time, generating 150 events. But it could really be two sources over the lifetime of the config with the same name. Or 10.

Even worse -- what if it used to be a source, but now it's a `sink` with the same name? Now it's reporting 100 events dumped out, when it might have just 100 events _ingested_. The current implementation grabs the first name that matches, so this could happen (the simple fix is to filter by `component_name` _and_ `component_type`-- I can update that easily enough.)

So currently, I think there are a couple of UX issues here:

1. The metrics `Controller` keeps around stats for old/dead topology. This is useful if we want to mark topology as dead, since we pull from this 'live' when topology is queried. But it's less useful if we rename a piece of topology, and then rename it back. Should we reset it for dead topology?

2. There's no concept of renaming, since the string name and its ID are one of the same. Renaming and starting something with an identical config aren't distinguishable. In #4581, a 'rename' means both a `topologyRemoved` + `topologyAdded`event fire. Do we need an internal concept of renaming? How would we want to tie an 'old' piece to a new one -- an ID?

3. If we _do_ reset stats on reload, then the 'dead' topology will no longer show any events or other stats when it's next queried. We'd need some way to remember the last known set of stats to be able to query them. But then how does this work with a later source that happens to re-use the same name?

---

Currently, we sidestep the latter point by overwriting the old config with the new, every time it changes. So 'dead' pieces don't show up.

Curious to hear your thoughts here.

Should we be resetting stats for old pieces of topology that no longer exist after a reload?

If so -- does this preclude showing them as dead in the console?

CC @binarylogic

binarylogic:
> Should we be resetting stats for old pieces of topology that no longer exist after a reload?

Yes, we should. Stats should only reflect the current state of the configuration. So if a component is removed, so should everything associated with it.

In the case of a rename, and then renaming it back, data would be dropped after each rename.

leebenson:
> Yes, we should. Stats should only reflect the current state of the configuration. So if a component is removed, so should everything associated with it.

Thanks @binarylogic - tracking in #4583.

If the configuration changes but the name remains, should we also delete?

src/top.rs
leebenson: 😺 Yeah, this is a bit odd. It'll be cleaned up when I do a 'humanize' sweep; I don't think `-h` needs to do anything if the value is < 1k in this case. Also, the type of `events_processed` is `MetricValue::Counter { value: f64 }`, which should probably be cast to an int in GraphQL schema to avoid having to treat this as a float downstream.

src/lib.rs
leebenson: This is for prettytable's macros: `row!`, `table!`, etc.

I did this Rust 2015-stylee because regular imports were throwing 'private mod' errors during dev (but not build), due to clashing between macro + mod names.

Screenshot 2020-10-15 at 20 40 25

`use prettytable:*` would work here too, at the cost of some import pollution.

Do we have a preference?

binarylogic:
@leebenson this is really awesome to see. Nice work so far. Agree with @Hoverbear's comments. Once you feel this is cleaned up, and the edge cases are solved, I'll take it for another spin.

leebenson:
Merging this to unblock #4581. Future work is tracked:

- Topology stats should be deleted when the component is removed #4583
- Investigate missing console events #4588
- Do a sweep for `--human` stats in `vector top` #4595

Issue #3896, opened 2020-09-15
Make `namespace` a first class field on metrics


jszwedko:
Implementation issue for https://github.com/timberio/vector/pull/3686

Also see https://github.com/timberio/vector/pull/4437#issue-499451433

ktff:
Plan of action:

- [ ] Introduce `namespace` to `struct Metric` and  proto `Metric`.

- [ ] Update sinks to combine namespace to name. 

- [ ] Update transforms to expose namespace and to not merge namespaces with names. Possible breaking change.  

- [ ] Update sources to not merge namespaces with names. Breaking change.

ktff:
Regarding metric sinks I propose that we:
* Add required `default_namespace` option, mentioned in the rfc, to those sinks that need namespace to be defined.
* Remove optional `namespace` option from all sinks.

Having both options in the sinks seems a bit too much when required `default_namespace` option for, `prometheus`, `aws_cloudwatch_metrics`, and `sematext_metrics` sinks, is all that is needed. The rest is covered with `add_tags`/`remove_tags` transforms.

Both points are a breaking change.  

### Alternative
Have both options. This would still be a breaking change.

cc. @jszwedko @binarylogic

jamtur01:
I am +1 the proposal. The alternatives seem confusing to the end user.

jszwedko:
Agreed, I like `default_namespace`.

PR #4437, opened 2020-10-07
fix(apache_metrics source): Use configured namespace for `up` metric

jszwedko:
Was previously hardcoding `apache_up` as the metric.

I was going to refactor this namespace encoding to share among metrics sources, but figured I'd punt in-lieu of #3896 which I hope to get to next sprint, which would make it a first-class field on the `Metric` type.


fanatid (approved):

jszwedko:
Opened issue for windows test failure: https://github.com/timberio/vector/issues/4439

PR #4440, opened 2020-10-08
chore(ci): Add cue formatting check

### LINKED ISSUE #4424: Add `cue fmt` to CI

jamtur01:
Closes #4424 

### DIFF
--- a/docs/reference.cue
+++ b/docs/reference.cue
@@ -7,7 +7,7 @@ package metadata
 
 #ConfigurationOptions: [Name=string]: {
 	description: string
-groups?: [...string]
+	groups?: [...string]
 	name:           Name
 	relevant_when?: string
 	required:       bool
--- a/docs/reference.cue
+++ b/docs/reference.cue
@@ -7,7 +7,7 @@ package metadata
 
 #ConfigurationOptions: [Name=string]: {
 	description: string
-	groups?: [...string]
+groups?: [...string]
 	name:           Name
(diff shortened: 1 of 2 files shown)


binarylogic (approved):
Nice

MOZGIII:
This only makes sense in a CI context, locally it can easily mess up the state or report incorrectly.

The idea is in CI it doesn't matter if the files state becomes corrupted - if it's anyhow different the check is failed, and the execution is considered a failure.

Locally, the assumption that any file changes are caused by formatting issues is false, thus such assertion is incorrect in this context. Not only that, but `fmt` at `check` is prone to disrupt people's work, as any kind of `fmt` can occasionally corrupt the file (by nature of the autoformatters).

This check would totally work though if we only keep it at CI.

jszwedko:
I'm digging into this now, but I did get an unexpected false positive from it: https://github.com/timberio/vector/pull/4101/checks?check_run_id=1226461323#step:9:15

It looks like `Cargo.toml` and `.cargo/config` changed during the CI run and so showed up in the diff.

jszwedko:
Ah, I see, `make slim-builds` modifies it: https://github.com/timberio/vector/blob/master/scripts/slim-builds.sh#L16-L49

jszwedko:
Opened https://github.com/timberio/vector/pull/4448

PR #4446, opened 2020-10-08
chore(observability): API client refactor + async-graphql bump

leebenson:
A slight rejig of API code, to separate the server + client into `api` and `api_client` packages.

This makes it a little more lightweight in cases where a running Vector instance wants to expose an API, but isn't necessarily the same binary that will `vector top` it.

Bumps async-graphql to 1.18.2, now that rust 1.46 is unblocked (thanks @Hoverbear 🎉 ). Note: async-graphql is now focusing efforts on the v2.0.0-alpha tag, which I'll tackle later. It wasn't sluggish before, but there's evidently some performance boosts and other features that I'd like to try out.


leebenson:
Updated `make_subscription` to take a `&url::Url` instead of a `SocketAddr`, for flexibility. Users will be able to define `ws://` or `wss://` URLs, to accommodate TLS, proxies, flexible routing, etc.

Aside: I noticed`test_utils::retry_until()` was using a `'static` lifetime-- changed that to `'a` to allow borrowed values in the retried closure, which was relevant for borrowing a URL (`SocketAddr` is `Copy`, so this wasn't caught before.)

Hoverbear (approved):
The CI suggests you have some lints, but the rest looks good!

leebenson:
GH's _"Unchanged files with check annotations (beta)"_ seem like a new addition that aren't picked up by `make check-clippy`.

The lints ref code paths external to this PR, so I'll try to figure out how to replicate locally, and will tackle separately 👍

PR #4448, opened 2020-10-08
chore(ci): Only check for diffs in the docs for `make check-docs`

jszwedko:
The check previously picking up changes to `Cargo.toml` and `.cargo/config` that happen as part of `make slim-builds` in CI.

Example: https://github.com/timberio/vector/pull/4101/checks?check_run_id=1226461323#step:9:15

### DIFF
--- a/scripts/check-docs.sh
+++ b/scripts/check-docs.sh
@@ -22,12 +22,12 @@ fi
 echo "Validating ${DOCS_PATH}/**/*.cue formatting."
 
 cue fmt ${DOCS_PATH}/**/*.cue
-status="$(git status --porcelain docs)"
+status="$(git status --porcelain ${DOCS_PATH})"
 
 [[ -z "$status" ]] || {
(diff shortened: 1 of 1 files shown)


jamtur01 (approved):
LGTM

MOZGIII:
We can apply a more accurate check, so it's not disrupted by any other changes. I'll try to hack together a PR over the weekend.

jszwedko:
@MOZGIII that would be great :smile: . It'd be ideal if `cue` just had a way to exit non-zero if files needed to be formatted.

MOZGIII:
We could help with pushing https://github.com/cuelang/cue/pull/247

jszwedko:
I'd be a fan of both: solving the immediate need better, but also pushing forward https://github.com/cuelang/cue/pull/247. It looks like the latter is just waiting on review comments to be resolved. I could try to poke at it next week.

PR #4480, opened 2020-10-09
enhancement(aws_ec2_metadata transform): Add internal events

### LINKED ISSUE #3390: Define events for the `aws_ec2_metadata` transform
As part of finishing https://github.com/timberio/vector-product/issues/14, we need to define internal events for the `aws_ec2_metadata` transform.

## Defining events

## Requirements

- [ ] Add a `AWSEC2MetadataEventProcessed` event.

jszwedko:
Closes #3390 

This adds a few internal events to the aws_ec2_metadata transform:

* AwsEc2MetadataEventProcessed: fired any time an event flows through
  the transform. Note that the event might not be annotated with any
  fields if the EC2 metadata refresh was never successful.
* AwsEc2MetadataRefreshComplete: fired whenever the EC2 metadata refresh
  executes. Note that this just means the refresh ran, but not that it
  was able to refresh any fields. More on this below.
* AwsEc2MetadataRequestFailed: Fired any time a EC2 metadata field fails
  to be fetched.

This also makes a couple of changes to the way errors are handled and
propagated within the metadata refresh. Previously it would swallow
requests that returned with a non-200 status code, but propagate any
other errors up (like network connection errors). It felt odd to me to
treat the HTTP status code errors differently, so I modified the
functions fetching the metadata to simply propagate up errors. I then
added wrapper functions that simply emit any errors and return an
Option<Bytes> containing the value that was requested. This feels more
consistant to me.

The refresh task part of this transform is a bit different in that it
can experience partial failures that we don't want to halt the fetching
and refresh of additional fields. For this reason, I added the
AwsEc2MetadataRequestFailed event that can be used to track individual
request failures, but there is no general refresh task failure. Instead,
I just have an AwsEc2MetadataRefreshComplete event that demarks that the
refresh task ran at all.

I ended up leaning on `crate::Error` a bunch here. I could see introducing some more specific error types, but the boilerplate felt overmuch for this use case and `crate::Error` was already being used here in-lieu of more specific errors.


jszwedko:
An alternative approach to the metadata refresh task I was considering over the weekend is to make the refresh atomic rather than updating what we can. This would have the effect of being able to have simpler pass/fail events for the refresh and would never have log events with partial annotations. The downside is that it less gracefully degrades if one of the refresh requests fails intermittently.

src/transforms/aws_ec2_metadata.rs
```diff
@@ -281,66 +294,37 @@ impl MetadataClient {
...
-            return Ok(None);
-        }
-
-        let body = body_to_bytes(res.into_body()).await?;
-
-        serde_json::from_slice(&body[..])
-            .map_err(Into::into)
-            .map(Some)
+    pub async fn get_document(&mut self) -> Option<IdentityDocument> {
+        self.get_document_error()
```
Hoverbear, line 298: You may also want to not swallow the error, so that the caller can act on it.

src/transforms/aws_ec2_metadata.rs
```diff
@@ -281,66 +294,37 @@ impl MetadataClient {
...
-
-        if StatusCode::OK != res.status() {
-            warn!(message="Metadata request failed.", status = %res.status());
-            return Ok(None);
-        }
-
-        let body = body_to_bytes(res.into_body()).await?;
-
-        Ok(Some(body))
+    pub async fn get_metadata(&mut self, path: &PathAndQuery) -> Option<Bytes> {
```
Hoverbear, line 310: Similarly, swallowing the error prevents action on it.

Hoverbear (approved):

jamtur01:
@jszwedko I think if the metric names can get updated then this is +1 to go.

jszwedko:
I've been thinking about this one off and on and am unsure if the metrics, as reported here, will be easily understandable by users.

I've been thinking about making the change mentioned in https://github.com/timberio/vector/pull/4480#issuecomment-707144043 to make the metadata refresh all or nothing. I think this behavior would be more clear. I'll aim to circle back to this soon.

jszwedko:
@Hoverbear I pushed the mentioned change to this, to have the refresh be all or nothing, to simplify the metrics and avoid partially refreshed data. Could you take another peek?

Hoverbear (approved):

PR #4485, opened 2020-10-09
chore: Remove Atoms

Hoverbear:
A component of https://github.com/timberio/vector/blob/c4707947bd876a0ff7d7aa36717ae2b32b731593/rfcs/2020-05-25-more-usable-logevents.md#sales-pitch

Precursor of #2843


src/event/log_event.rs
```diff
@@ -3,80 +3,91 @@ use serde::{Serialize, Serializer};
...
 
-    pub fn insert<K, V>(&mut self, key: K, value: V) -> Option<Value>
-    where
-        K: AsRef<str>,
-        V: Into<Value>,
-    {
+    #[instrument(skip(self, key), fields(key = %key.as_ref()))]
+    pub fn insert(
+        &mut self,
+        key: impl AsRef<str>,
```
Hoverbear, line 39: Future work we should probably upgrade this to an owned string so we can insert that allocation.

src/internal_events/elasticsearch.rs
```diff
@@ -20,11 +19,11 @@ impl InternalEvent for ElasticSearchEventReceived {
 }
 
 #[derive(Debug)]
-pub struct ElasticSearchMissingKeys {
-    pub keys: Vec<Atom>,
+pub struct ElasticSearchMissingKeys<'a> {
+    pub keys: &'a [String],
```
Hoverbear, line 23: Allowing this instead of a `&'a [&'a str]` removes the need for the code that calls this to make an extra allocation through a `by_ref()` call.

src/sources/docker.rs
```diff
@@ -42,11 +42,11 @@ const VECTOR_IMAGE_NAME: &str = "timberio/vector";
 lazy_static! {
     static ref STDERR: Bytes = "stderr".into();
     static ref STDOUT: Bytes = "stdout".into();
-    static ref IMAGE: Atom = Atom::from("image");
-    static ref CREATED_AT: Atom = Atom::from("container_created_at");
-    static ref NAME: Atom = Atom::from("container_name");
-    static ref STREAM: Atom = Atom::from("stream");
-    static ref CONTAINER: Atom = Atom::from("container_id");
+    static ref IMAGE: &'static str = "image";
```
Hoverbear, line 45: These should be const

benches/lua.rs
```diff
@@ -17,11 +17,11 @@ fn add_fields(c: &mut Criterion) {
     let key = "the key";
     let value = "this is the value";
 
-    let key_atom_native = key.into();
+    let key_atom_native = key;
```
jszwedko, line 20: Nbd, but it seems like we could remove `_atom` from these names.

src/event/log_event.rs
```diff
@@ -3,80 +3,91 @@ use serde::{Serialize, Serializer};
...
 
 #[derive(PartialEq, Debug, Clone, Default)]
 pub struct LogEvent {
     fields: BTreeMap<String, Value>,
 }
 
 impl LogEvent {
-    pub fn get(&self, key: &DefaultAtom) -> Option<&Value> {
-        util::log::get(&self.fields, key)
+    #[instrument(skip(self, key), fields(key = %key.as_ref()))]
```
jszwedko, line 16: I see you've been adding these since the introduction of the `Lookup` type. Do you recommend we add them to all methods? Or just certain ones? They seem useful for tracing.

src/mapping/query/path.rs
```diff
@@ -3,12 +3,11 @@ use crate::{
     event::{util::log::get_value, Event, PathIter, Value},
     mapping::Result,
 };
-use string_cache::DefaultAtom as Atom;
 
 #[derive(Debug)]
 pub(in crate::mapping) struct Path {
     // TODO: Switch to String once Event API is cleaned up.
```
jszwedko, line 9: Seems like this TODO could be dropped?

src/transforms/log_to_metric.rs
```diff
@@ -120,26 +119,30 @@ impl LogToMetric {
...
 fn render_template(s: &str, event: &Event) -> Result<String, TransformError> {
     let template = Template::try_from(s).map_err(TransformError::TemplateParseError)?;
     template.render_string(&event).map_err(|missing_keys| {
-        // convert to String to avoid printing Atom in Debug format
-        let missing_keys = missing_keys
-            .into_iter()
-            .map(|k| k.to_string())
-            .collect::<Vec<String>>();
+        // convert to String to avoid printing String in Debug format
+        let missing_keys = missing_keys.into_iter().collect::<Vec<String>>();
```
jszwedko, line 139: Seems like this might have been left over? `missing_keys` is already a `Vec<String>`.

jszwedko (approved):
Awesome!

I focused on `src/event` and other non-component or internal events files and just scanned the rest since it seemed to be mechanical.

benches/lua.rs
Hoverbear: Oyyy good call.

src/event/log_event.rs
Hoverbear: We haven't really made an architectural call one way or the other. What they do is populate the span: https://tracing-rs.netlify.app/tracing/attr.instrument.html, so if we did wide adoption we might be able to reap some really nice contextual rewards. It's a practice I've been doing in my own code lately. I'm not sure if @lukesteensen has an opinion though.

bruceg (commented):
Looks mostly ok, but I added a number of questions.

I am also concerned that, if we are serious about moving to either smolstr or smartstring for keys, we will end up reverting most of these changes and converting all `Atom` to `SmartString` or whatever.

src/event/log_event.rs
```diff
@@ -3,80 +3,91 @@ use serde::{Serialize, Serializer};
...
         util::log::insert_path(&mut self.fields, key, value.into())
     }
 
+    #[instrument(skip(self, key), fields(key = %key))]
     pub fn insert_flat<K, V>(&mut self, key: K, value: V)
     where
-        K: Into<String>,
-        V: Into<Value>,
+        K: Into<String> + Display,
+        V: Into<Value> + Debug,
```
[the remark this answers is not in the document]
jszwedko: My thought is that `instrument` might require them.

src/event/log_event.rs
JeanMertz: I've seen these used as well.

I agree that better instrumentation is a worthy goal, but I would prefer if we at least document how/why we use them, including some best-practices/patterns everyone should follow.

I don't mind if it's in an RFC or a "tracking issue" that contains some context, and a few relevant links.

src/event/discriminant.rs
```diff
@@ -24,7 +23,7 @@ pub struct Discriminant {
 impl Discriminant {
     /// Create a new Discriminant from the `LogEvent` and an ordered slice of
     /// fields to include into a discriminant value.
-    pub fn from_log_event(event: &LogEvent, discriminant_fields: &[Atom]) -> Self {
+    pub fn from_log_event(event: &LogEvent, discriminant_fields: &[String]) -> Self {
```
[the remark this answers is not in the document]
Hoverbear: Hmmm I think I could use `[impl AsRef<str>]` and allow both 🤔

Hoverbear:
@bruceg I feel if we move to smolstr or otherwise, we'll be able to internalize this decision making to the `Value` or `Lookup` type, or add transparent conversions, and not require such intrusive changes.

src/sinks/influxdb/metrics.rs
```diff
@@ -888,6 +888,7 @@ mod integration_tests {
                 username: None,
                 password: None,
             }),
+            tags: None,
```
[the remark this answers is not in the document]
Hoverbear: Seems this is a merge issue 🤔

src/sources/journald.rs
```diff
@@ -207,16 +207,16 @@ impl JournaldConfig {
 fn create_event(record: Record) -> Event {
     let mut log = LogEvent::from_iter(record);
     // Convert some journald-specific field names into Vector standard ones.
-    if let Some(message) = log.remove(&MESSAGE) {
+    if let Some(message) = log.remove(&*MESSAGE) {
```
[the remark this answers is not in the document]
Hoverbear: Yes! This is leftover from them being lazy statics. 😓

src/event/log_event.rs
lukesteensen: > better instrumentation is a worthy goal, but I would prefer if we at least document how/why we use them, including some best-practices/patterns everyone should follow.

I agree! I'd also note that we should strive to be as consistent as possible across the codebase. The macro is cool and useful, but right now it mostly tells me which code was written by @Hoverbear :smile: 

An easy first step would be to open an issue for setting some guidelines around thing kind of instrumentation. We'll want to cover `#[instrumented(...)]` vs `trace!` vs `emit!`, message style, etc in the eventual guidelines, but for now, maybe @Hoverbear can just summarize her current thoughts as a starting point for the conversation.

Hoverbear: I'd love to! For now, as a note `#[instrument]` just creates a span when the function is entered, with the function args (or as configured) as span items. It's dropped when the function exits. That's all it does. :)

It doesn't log. :)

PR #4508, opened 2020-10-11
fix: logdna clippy and prometheus skipped field

fanatid:
prometheus error introduced in https://github.com/timberio/vector/pull/4412
logdna in https://github.com/timberio/vector/pull/4077


ktff (approved):
As soon as the tests pass this is ready to go.

PR #4528, opened 2020-10-12
feat(remap transform): add `format_url` function

### LINKED ISSUE #3762: New URL functions for the remap syntax

JeanMertz:
Counterpart to #4428, which parses a URL-formatted string to an object. This function turns that object back into a string.

```rust
.url.scheme = "https"
.url.host = "vector.dev"

.url = format_url(.url)
```

```json
{
	"url": "https://vector.dev"
}
```

closes #3762

### DIFF
--- a/src/event/value.rs
+++ b/src/event/value.rs
@@ -227,6 +227,34 @@ impl Value {
         self.as_bytes()
     }
 
+    pub fn as_option(&self) -> Option<&Value> {
+        match self {
+            Value::Null => None,
+            value => Some(value),
+        }
+    }
+
+    pub fn as_i64(&self) -> Option<i64> {
+        match self {
+            Value::Integer(i) => Some(*i),
+            _ => None,
+        }
+    }
+
+    pub fn as_map(&self) -> Option<&BTreeMap<String, Value>> {
+        match self {
+            Value::Map(v) => Some(v),
+            _ => None,
+        }
+    }
+
+    pub fn as_bytes2(&self) -> Option<Bytes> {
+        match self {
+            Value::Bytes(b) => Some(b.clone()), // cloning a Bytes is cheap
+            _ => None,
+        }
+    }
+
     pub fn as_timestamp(&self) -> Option<&DateTime<Utc>> {
         match &self {
             Value::Timestamp(ts) => Some(ts),
--- a/src/mapping/query/function/format_url.rs
+++ b/src/mapping/query/function/format_url.rs
@@ -1,0 +1,235 @@
+use super::prelude::*;
+use std::collections::BTreeMap;
+use std::convert::TryFrom;
+use url::Url;
+
+const SCHEME: &str = "scheme";
+const USERNAME: &str = "username";
+const PASSWORD: &str = "password";
+const HOST: &str = "host";
+const PORT: &str = "port";
+const PATH: &str = "path";
+const FRAGMENT: &str = "fragment";
+const QUERY: &str = "query";
+
+#[derive(Debug)]
+pub(in crate::mapping) struct FormatUrlFn {
+    query: Box<dyn Function>,
+}
+
+impl FormatUrlFn {
+    #[cfg(test)]
+    pub(in crate::mapping) fn new(query: Box<dyn Function>) -> Self {
+        Self { query }
+    }
+}
+
+impl Function for FormatUrlFn {
+    fn execute(&self, ctx: &Event) -> Result<Value> {
+        let components = required!(ctx, self.query, Value::Map(v) => v);
+
+        components_to_url(components).map(|url| url.into_string().into())
+    }
+
+    fn parameters() -> &'static [Parameter] {
+        &[Parameter {
+            keyword: "value",
+            accepts: |v| matches!(v, Value::Map(_)),
+            required: true,
+        }]
+    }
+}
+
+impl TryFrom<ArgumentList> for FormatUrlFn {
+    type Error = String;
+
+    fn try_from(mut arguments: ArgumentList) -> Result<Self> {
+        let query = arguments.required("value")?;
+
+        Ok(Self { query })
+    }
+}
+
+fn components_to_url(components: BTreeMap<String, Value>) -> Result<Url> {
+    let scheme = components
+        .get(SCHEME)
+        .map(|v| string(SCHEME, v))
+        .transpose()?
+        .ok_or_else(|| missing(SCHEME))?;
+
+    let username = components
+        .get(USERNAME)
+        .map(|v| string(USERNAME, v))
+        .transpose()?
+        .unwrap_or_else(|| "".to_owned());
+
+    let password = components
+        .get(PASSWORD)
+        .and_then(Value::as_option)
+        .map(|v| string(PASSWORD, v))
+        .transpose()?;
+
+    let host = components
+        .get(HOST)
+        .and_then(Value::as_option)
+        .map(|v| string(HOST, v))
+        .transpose()?;
+
+    let port = components
+        .get(PORT)
+        .and_then(Value::as_option)
+        .map(|v| {
+            v.as_i64()
+                .map(|i| i as u16)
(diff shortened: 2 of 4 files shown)


src/event/value.rs
```diff
@@ -227,6 +227,34 @@ impl Value {
...
+            _ => None,
+        }
+    }
+
+    pub fn as_bytes2(&self) -> Option<Bytes> {
+        match self {
+            Value::Bytes(b) => Some(b.clone()), // cloning a Bytes is cheap
+            _ => None,
+        }
+    }
```
JeanMertz, line 256: "Unfortunately", we already have `as_bytes`, which behaves differently from what you'd expect (given the precedence set by Serde itself), as it tries to convert any `Value` type into `Bytes` by doing lossy conversions.

Happy to do a sweep-and-rename for the original `as_bytes` if others agree this should be the canonical implementation of such a method.

src/mapping/query/function/format_url.rs
```diff
@@ -0,0 +1,235 @@
...
+        .map(|v| {
+            v.iter()
+                .map(|(k, v)| (k, v.to_string_lossy()))
+                .collect::<Vec<_>>()
+        })
+        .unwrap_or_default();
+
+    // This is a really awkward way to create a valid URL from components, but
+    // it doesn't appear the `Url` API offers any better way to handle this.
+    let mut url = Url::parse("https://vector.dev").expect("valid");
```
JeanMertz, line 113: I'm really not a fan of this, but I didn't see any other option, aside from formatting manually using `format!`, but that's rather tricky to get correct, given the esoteric rules w.r.t. valid URL parsing.

I looked into using [`http::Uri`](https://docs.rs/http/0.2.1/http/uri/struct.Uri.html), which we already use in Vector, as it provides a `from_parts` function, but that type (among other issues) [doesn't support parsing the _fragment_ part of the URL](https://github.com/hyperium/http/issues/127).

Anyone have a better alternative to the current approach?

src/mapping/query/function/format_url.rs
```diff
@@ -0,0 +1,235 @@
...
+    format!("missing url component '{}'", s)
+}
+
+#[inline]
+fn invalid(component: &'static str, expected: &str, got: &str) -> String {
+    format!(
+        "invalid type for url component '{}' (expected {}, got {})",
+        component, expected, got,
+    )
+}
```
JeanMertz, line 149: I suspect at some point we'll want a more generic approach to deserialize a `Value::Map`, which we can use throughout remap functions dealing with maps, so that we don't have to re-do parsing and error handling, but I figured I'd do it manually for now, until we have a need to extract things out.

src/event/value.rs
FungusHumungus: Imo serde is wrong here. I think a better name might be something like `try_get_bytes`.

src/mapping/query/function/format_url.rs
FungusHumungus: It feels clunky, but I can't think of a better way. 

It might be worth raising an issue with `rust-url` to see if we could have some kind of `UrlBuilder` object that would allow you to build up a Url without having to parse one first? I can't imagine we are the only people who would find that useful..

src/mapping/query/function/format_url.rs
```diff
@@ -0,0 +1,235 @@
...
+use url::Url;
+
+const SCHEME: &str = "scheme";
+const USERNAME: &str = "username";
+const PASSWORD: &str = "password";
+const HOST: &str = "host";
+const PORT: &str = "port";
+const PATH: &str = "path";
+const FRAGMENT: &str = "fragment";
+const QUERY: &str = "query";
```
FungusHumungus, line 13: Will these keys be shared with ParseUrl?

src/event/value.rs
FungusHumungus: It might be nice to have a function `as_utf8`. A very common occurrence in remap is to get the bytes and then convert to utf8.

FungusHumungus (approved):

src/event/value.rs
```diff
@@ -227,6 +227,34 @@ impl Value {
         self.as_bytes()
     }
 
+    pub fn as_option(&self) -> Option<&Value> {
```
jszwedko, line 230: Not really related to this PR, but this method is making me wonder if we shouldn't just model `Value::Null` as `None` and have `LogEvent.fields` be a `BTreeMap<String, Option<Value>>`.

src/event/value.rs
jszwedko: I agree that the existing `as_bytes` method isn't well named. Given all of the other `as_X` methods you are adding here, I'd prefer to rename the existing `as_bytes` to something else to make the `as_X` methods consistent. Maybe we could call it `to_bytes`? Or something else indicating that it is serializing the data.

In my small sample audit, it looks like `as_bytes` is only used with the `message` key or other fields expected to be `Value::Bytes` so it's possible that the other enum implementations aren't even useful there and maybe we could drop it and just use your new `as_bytes` method. I'm not 100% sure though.

src/mapping/query/function/format_url.rs
jszwedko: Hmm, yeah, I'm :+1: on filing an issue for a builder with that crate.

jszwedko (commented):
I'd like to see the conversation around `as_bytes2` resolved, but otherwise this looks good!

src/mapping/query/function/format_url.rs
bruceg: This work seems to be about half of serde's deserialize operation, going from an intermediate form instead of a fully serialized form, into a struct. I wonder if you couldn't use some of the serde framework to do this, treating a `Value::Map` as the serialized form.

src/mapping/query/function/format_url.rs
```diff
@@ -0,0 +1,235 @@
...
+    macro_rules! map {
+        () => (
+            Value::Map(BTreeMap::new())
+        );
+        ($($k:tt: $v:expr),+ $(,)?) => (
+            Value::from(vec![
+                $(($k.into(), $v.into())),+
+            ].into_iter().collect::<BTreeMap<String, Value>>())
+        );
+    }
```
bruceg, line 170: AFAIK we have at least three version of this macro now. I think that would be good to consolidate them somewhere. Not necessary for this PR though.

bruceg (commented):
This looks reasonable, aside from the noted unfortunate hackery around constructing the URL. The only doubt I have is regarding the naming of `as_bytes2`.

jamtur01:
@JeanMertz What's the status here?

JeanMertz:
@jamtur01 I need to convert this to the new function set-up (see #4709). Once that PR is done, I'll rebase and update this one, and merge it 👍

jamtur01:
@JeanMertz Just pinging on this now #4709 is merged.

jszwedko (commented):
Just crawling through the PRs I've looked at to see any stale ones.

@JeanMertz I'd still like to see `as_bytes` / `as_bytes2` renamed before this gets merged. Otherwise this looks good though!

JeanMertz:
> @JeanMertz I'd still like to see `as_bytes` / `as_bytes2` renamed before this gets merged. Otherwise this looks good though!

I'll follow up with a new PR, since a lot has changed since this PR was created, including Remap having its own `Value` type (and [the associated `as_bytes` etc added to it](https://github.com/timberio/vector/blob/5f97129f8e2f5bc0e86f4e85d2d4dbf7cb2c2a09/lib/remap-lang/src/value.rs#L403-L408)).

PR #4537, opened 2020-10-13
chore(observability): Split `api_client` from Vector proper

### LINKED ISSUE #4525: Break out `api_client` into a separate workspace crate
Currently, `api_client` is part of Vector proper. Aside from tests, it's entirely ring-fenced and would therefore make an ideal candidate for breaking out into a separate `lib` package.

There are two areas which would need altering to accommodate this:

1. [API E2E tests](tests/api.rs) mash together the API server + client into a single suite of tests, since the two are intrinsically linked (clearly, we can't assert GraphQL tests if there's nothing to test against.) It probably still makes sense to keep these top-level rather than move into `lib/api_client` given this, but the exports of api_client should be altered to create a better separation for its public interface. This includes being able to make assertions on returned data.

2. Queries are produced ad hoc, using the `#[derive(GraphQL)]` annotation. This blurs the public interface somewhat, by moving query/subscription types into the caller. Ideally, this should be an implementation detail, and instead `new_XX_query()` helpers would provide a public api for specific GraphQL queries.

leebenson:
This PR moves `api_client` into a separate workspace crate, and refactors the API to offer a cleaner public surface.

Closes #4525


lib/api_client/src/client.rs
```diff
@@ -0,0 +1,39 @@
...
+        Self { url }
+    }
+
+    /// Issue a GraphQL query using Reqwest, serializing the response to the associated
+    /// GraphQL type for the given `request_body`
+    pub async fn query<T: GraphQLQuery>(
+        &self,
+        request_body: &graphql_client::QueryBody<T::Variables>,
+    ) -> QueryResult<T> {
+        let client = reqwest::Client::new();
```
leebenson, line 23: This could perhaps be stored against the `Client` instance directly, but I'm not sure if there's any interior mutability in play between requests, so I just left it as a new instance for each query.

lib/api_client/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 mod client;
+pub mod gql;
```
leebenson, line 2: Individual structs/types generated by `graphql_client` are still accessible publically, in cases where types need to be referenced in function sigs, or asserted against. Generally though, this won't be necessary; the traits implement a cleaner abstraction where those implementation details are hidden away.

src/api/schema/gen.rs
```diff
@@ -106,5 +106,6 @@ async fn main() {
     let res = schema.execute(INTROSPECTION_QUERY).await;
     let json = serde_json::to_string_pretty(&async_graphql::http::GQLResponse(res)).unwrap();
 
-    fs::write("graphql/schema.json", format!("{}\n", json)).expect("Couldn't save schema file");
+    fs::write("lib/api_client/graphql/schema.json", format!("{}\n", json))
```
leebenson, line 109: This is slightly overlapping-- the resulting `schema.json` is pushed into `lib/api_client`, rather than being top-level.

This is still generated by Vector proper (via `cargo run --bin graphql-schema`) since it's the API that's responsible for the schema; the client merely consumes it for code gen. So I think this is still the right place for it.

tests/api.rs
```diff
@@ -7,9 +7,13 @@ mod support;
 #[cfg(all(feature = "api", feature = "api_client"))]
 mod tests {
     use crate::support::{sink, source};
+    use api_client::{
+        connect_subscription_client,
+        gql::{HealthQueryExt, HealthSubscriptionExt, MetricsSubscriptionExt},
```
leebenson, line 12: These traits bring the `Client` and `SubscriptionClient` methods into scope.

lib/api_client/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "api_client"
```
MOZGIII, line 2: How about changing this to `vector-api-client`?

1. Uses underscores - this notation is kind of more wide-spread in the crates as I see, and it has a special-cased support in cargo.
2. Name is scoped with `vector` - we're unlocking potential publishing of this crate to crates.io without extra renaming steps.

leebenson: 👍 done in 7f7c7dde5319da2417c47f2169fd341f15dcfdf9

lib/api_client/src/client.rs
MOZGIII: Yes, we should definitely do this. All the interesting calls are non-mut. Furthermore, we should definitely allow passing a custom `reqwest` instance - so that users can specify thier own preferences for the connection.

I suggest adding a builder (using https://docs.rs/derive_builder/0.9.0/derive_builder/) and accepting both `url` and `reqwest::Client`.

MOZGIII (approved):
Looks great!

lib/vector-api-client/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 mod client;
```
MOZGIII, line 1: Since this is a crate now, how about adding some crate annotations?

```rust
#![deny(
    missing_debug_implementations,
    missing_copy_implementations,
    missing_docs
)]
```

They will help find spots that can be improved and help with maintenance later on.

lib/api_client/Cargo.toml
MOZGIII: Sweet! We might also rename the dir accordingly, but that's not as important.

MOZGIII: Ah, nvm diffs didn't update properly!

leebenson: @MOZGIII - the folder was renamed to `lib/vector-api-client` -- do you mean something else?

src/api/schema/gen.rs
leebenson: 👍 Yeah, that's a good idea. I'll make the switch.

leebenson: Hmm this might be an issue:

> The build script does not have access to the dependencies listed in the dependencies or dev-dependencies section (they’re not built yet!). Also, build dependencies are not available to the package itself unless also explicitly added in the [dependencies] table.

We require the inner `api` package to build the schema, so I'm not sure if this is feasible?

lib/api_client/src/subscription.rs
```diff
@@ -216,7 +220,7 @@ impl SubscriptionClient {
 
 /// Connect to a GraphQL subscription endpoint and return an active client. Can be used for
 /// multiple subscriptions
-pub async fn make_subscription_client(
+pub async fn connect_subscription_client(
     url: &Url,
 ) -> Result<SubscriptionClient, tungstenite::error::Error> {
     let (tx, _) = connect_async(url).await?;
```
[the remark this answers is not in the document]
leebenson: Thanks, added `pub new` in ec5db3465f3769bbb8ac57c9c8439f7845271078.

`new` takes Tungstenite `WebSocketStream<TcpStream>`, so I think that's general enough now to allow the caller to set up a different kind of connection if required.

The purpose of `connect_subscription_client` is to be a helper for the common case of just supplying a `url::Url` (typically `ws://` or `wss://`), opening a connection (and implicitly validating that it's reachable), and performing the initial handshake.

src/api/schema/gen.rs
MOZGIII: We'll be able to pass it in the `build-dependencies`. They'll be two separate dependencies/builds, so it's not trivial whether we'd want to actually use it.

I wonder if there's a ready-made crate that would provide this introspection dumping for `build.rs` - seems like a pretty generic thing, so maybe we don't have to bother.

---

Might be nice to take the path as an argument rather than hard-coding it. Or, just somehow configuring it from the outside of the binary.

leebenson: I think the issue might be, the `api` code actually has to run for it to produce the resulting `schema.json`. All of the `#[derive(GraphQL)]` macros need expanding, and mant of the the result types overly internal metrics types, which necessitates Vector being compiled first.

So even if we were able to pull in `api` independently, wouldn't we also need the rest of Vector to be compiled alongside it? That's what I'm thinking.

I like the idea of this being an automatic part of the build process, but it doesn't seem like `build.rs` is designed for consuming the crate that it's pre-empting being built. Seems a bit chicken and egg.

Or am I thinking about this wrong?

lib/api_client/src/subscription.rs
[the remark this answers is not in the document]
MOZGIII: Speaking of different kinds of connections - how difficult is it to generalize over the `TcpStream`? Not saying we should do it, but it looks somewhat trivial, and may end up useful for some edge cases, like WebSocket over unix sockets via connection management agent in a GCP or similar cloud setup.

src/api/schema/gen.rs
MOZGIII: Ah, in this case we'd have to split the schema into a separate crate, both to avoid the chicken and egg problem and to reduce the compile times.

I wonder if it makes sense to move this `gen` binary to something like `vector api generate-schema` command, so anyone can use it, as it seems like all the data is in the binary either way.

lib/api_client/src/subscription.rs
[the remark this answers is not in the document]
leebenson: It seems pretty general (https://docs.rs/tungstenite/0.11.1/tungstenite/protocol/struct.WebSocket.html). It's still WebSockets, so whatever it's connected to would need to talk WS, but it works with ws/wss and has a `from_raw_socket()` method that might open up other use-cases.

I guess we just wait and see what requirements come in from the world of the esoteric use-case, and adapt as needed.

src/api/schema/gen.rs
leebenson: > Ah, in this case we'd have to split the schema into a separate crate, both to avoid the chicken and egg problem and to reduce the compile times.

I think that probably makes sense if/once topology and metrics are separate. The GraphQL API is always going to be tightly coupled to that, so if the rest of it lives in Vector, I think the API needs to, also. If we get to a point where that's pulled out, it'll unblock pulling out the API, too.

> I wonder if it makes sense to move this gen binary to something like vector api generate-schema command, so anyone can use it, as it seems like all the data is in the binary either way.

I like that idea. That could be useful for third-party tooling, where users are building their own clients around the API (of course, we could also have a hosted `schema.json` -- either internally, or in docs, or both.)

I'll kick that can down the road a bit, since it's not the immediate scope of this PR, but I think it's worth creating an issue for.

lib/api_client/src/client.rs
leebenson: I've never used a builder pattern in Rust- this looks interesting. To avoid blocking this PR, I'll go ahead and open a separate issue to learn a bit more about this, and add it later. But thanks for bringing that to my attention.

leebenson: https://github.com/timberio/vector/issues/4538

MOZGIII: `derive_builder` is a syntax sugar. The idea there is to, rather than taking `new(url: _, client: reqwest::Client)` which requires passing the `client`, or `new(url: _, client: Option<reqwest::Client>)` which can be though of "pass request, or we'll use `reqwest::Client::default()`, but in a more elegant way.

With builder (and `derive_builder` simply to type less code :smile:) it's possible to make `url` be mandatory, and `reqwest::Client` optional, and default-initialized if none passed.

leebenson: Yeah, gotcha. Makes sense. I've wanted to reach for a pattern like that a couple of times, but didn't know there was an idiomatic solution for it. I'll use that in the next round. My immediate use-case for `Url` is to unblock the `vector top` work, so I'll defer this until I've had chance to read the above link and make sure I'm doing it properly -- but it sounds ideal.

lib/api_client/src/subscription.rs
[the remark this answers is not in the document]
MOZGIII: I mean taking `WebSocketStream<T: _>` rather than `WebSocketStream<TcpStream>`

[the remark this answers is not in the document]
leebenson: Ahh gotcha, yeah possibly, although I'm not immediately sure what we'd constrain it to.

Tungstenite [seems open-ended for `Stream`](https://docs.rs/tungstenite/0.11.1/src/tungstenite/protocol/mod.rs.html#63-70), but if we change the signature of `new` to:

```rust
pub fn new<TStream>(ws: WebSocketStream<TStream>) -> Self
```

We get a mess of errors:

> error[E0277]: the trait bound `TStream: tokio::io::AsyncRead` is not satisfied
>     --> lib/vector-api-client/src/subscription.rs:141:64
>      |
> 141  |         let (mut ws_tx, mut ws_rx) = futures::StreamExt::split(ws);
>      |                                                                ^^ the trait `tokio::io::AsyncRead` is not implemented for `TStream`
>      | 
>     ::: /Users/leebenson/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.5/src/stream/stream/mod.rs:1271:14
>      |
> 1271 |     fn split<Item>(self) -> (SplitSink<Self, Item>, SplitStream<Self>)
>      |              ---- required by this bound in `futures::StreamExt::split`
>      |
>      = note: required because of the requirements on the impl of `futures::Stream` for `tokio_tungstenite::WebSocketStream<TStream>`
> help: consider restricting type parameter `TStream`
>      |
> 132  |     pub fn new<TStream: tokio::io::AsyncRead>(ws: WebSocketStream<TStream>) -> Self {
>      |                       ^^^^^^^^^^^^^^^^^^^^^^
> 
> error[E0277]: the trait bound `TStream: tokio::io::AsyncWrite` is not satisfied
>     --> lib/vector-api-client/src/subscription.rs:141:64
>      |
> 141  |         let (mut ws_tx, mut ws_rx) = futures::StreamExt::split(ws);
>      |                                                                ^^ the trait `tokio::io::AsyncWrite` is not implemented for `TStream`
>      | 
>     ::: /Users/leebenson/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.5/src/stream/stream/mod.rs:1271:14
>      |
> 1271 |     fn split<Item>(self) -> (SplitSink<Self, Item>, SplitStream<Self>)
>      |              ---- required by this bound in `futures::StreamExt::split`
>      |
>      = note: required because of the requirements on the impl of `futures::Stream` for `tokio_tungstenite::WebSocketStream<TStream>`
> help: consider restricting type parameter `TStream`
>      |
> 132  |     pub fn new<TStream: tokio::io::AsyncWrite>(ws: WebSocketStream<TStream>) -> Self {
>      |                       ^^^^^^^^^^^^^^^^^^^^^^^
> 
> error[E0277]: `TStream` cannot be unpinned
>     --> lib/vector-api-client/src/subscription.rs:141:64
>      |
> 141  |         let (mut ws_tx, mut ws_rx) = futures::StreamExt::split(ws);
>      |                                                                ^^ the trait `std::marker::Unpin` is not implemented for `TStream`
>      | 
>     ::: /Users/leebenson/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.5/src/stream/stream/mod.rs:1271:14
>      |
> 1271 |     fn split<Item>(self) -> (SplitSink<Self, Item>, SplitStream<Self>)
>      |              ---- required by this bound in `futures::StreamExt::split`
>      |
>      = note: required because of the requirements on the impl of `futures::Stream` for `tokio_tungstenite::WebSocketStream<TStream>`
> help: consider restricting type parameter `TStream`
>      |
> 132  |     pub fn new<TStream: std::marker::Unpin>(ws: WebSocketStream<TStream>) -> Self {
>      |                       ^^^^^^^^^^^^^^^^^^^^
> 
> error[E0599]: no method named `next` found for struct `futures::stream::SplitStream<tokio_tungstenite::WebSocketStream<TStream>>` in the current scope
>    --> lib/vector-api-client/src/subscription.rs:156:38
>     |
> 156 |                     res = &mut ws_rx.next() => {
>     |                                      ^^^^ method not found in `futures::stream::SplitStream<tokio_tungstenite::WebSocketStream<TStream>>`
>     | 
>    ::: /Users/leebenson/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.5/src/stream/stream/split.rs:12:1
>     |
> 12  | pub struct SplitStream<S>(BiLock<S>);
>     | -------------------------------------
>     | |
>     | doesn't satisfy `_: futures::Stream`
>     | doesn't satisfy `_: tokio::stream::StreamExt`
>     |
>     = note: the method `next` exists but the following trait bounds were not satisfied:
>             `futures::stream::SplitStream<tokio_tungstenite::WebSocketStream<TStream>>: futures::Stream`
>             which is required by `futures::stream::SplitStream<tokio_tungstenite::WebSocketStream<TStream>>: tokio::stream::StreamExt`
> 
> error[E0599]: no method named `poll` found for struct `std::pin::Pin<&mut _>` in the current scope
>    --> lib/vector-api-client/src/subscription.rs:150:17
>     |
> 150 | /                 select! {
> 151 | |                     // Break the loop if shutdown is triggered. This happens implicitly once
> 152 | |                     // the client goes out of scope
> 153 | |                     _ = &mut shutdown_rx => break,
> ...   |
> 179 | |                     },
> 180 | |                 }
>     | |_________________^ method not found in `std::pin::Pin<&mut _>`
>     |
>     = note: `fut` is a function, perhaps you wish to call it
>     = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
> 
> error[E0599]: no method named `poll` found for struct `std::pin::Pin<_>` in the current scope
>    --> lib/vector-api-client/src/subscription.rs:150:17
>     |
> 150 | /                 select! {
> 151 | |                     // Break the loop if shutdown is triggered. This happens implicitly once
> 152 | |                     // the client goes out of scope
> 153 | |                     _ = &mut shutdown_rx => break,
> ...   |
> 179 | |                     },
> 180 | |                 }
>     | |_________________^ method not found in `std::pin::Pin<_>`
>     |
>     = note: `fut` is a function, perhaps you wish to call it
>     = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
> 
> error[E0599]: no method named `send` found for struct `futures::stream::SplitSink<tokio_tungstenite::WebSocketStream<TStream>, tokio_tungstenite::tungstenite::Message>` in the current scope
>    --> lib/vector-api-client/src/subscription.rs:177:43
>     |
> 177 | ...                   let _ = ws_tx.send(Message::Text(serde_json::to_string(&p).unwrap())).await;
>     |                                     ^^^^ method not found in `futures::stream::SplitSink<tokio_tungstenite::WebSocketStream<TStream>, tokio_tungstenite::tungstenite::Message>`
>     | 
>    ::: /Users/leebenson/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.5/src/stream/stream/split.rs:46:1
>     |
> 46  | pub struct SplitSink<S, Item> {
>     | -----------------------------
>     | |
>     | doesn't satisfy `_: futures::Sink<_>`
>     | doesn't satisfy `_: futures::SinkExt<_>`
>     |
>     = note: the method `send` exists but the following trait bounds were not satisfied:
>             `futures::stream::SplitSink<tokio_tungstenite::WebSocketStream<TStream>, tokio_tungstenite::tungstenite::Message>: futures::Sink<_>`
>             which is required by `futures::stream::SplitSink<tokio_tungstenite::WebSocketStream<TStream>, tokio_tungstenite::tungstenite::Message>: futures::SinkExt<_>`

Unless there's a 5 minute version to solve this, I think it's worth waiting to see what use cases other than TCPStream we might have, before attempting to generalise it further. This already accommodates local/remote sockets so we're probably covering the 90-95% case.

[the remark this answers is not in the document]
MOZGIII: I'm pretty sure rust have just wrote the trait bounds that we're looking for right there :smile: 

Try adding `TStream: tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin`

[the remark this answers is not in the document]
MOZGIII: It also looks like this `TStream` should be a part of the struct, in an `std::marker::PhantomData<TStream>`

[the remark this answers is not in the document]
MOZGIII: If it ends up more difficult than that - let's not bother.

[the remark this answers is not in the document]
leebenson: Cool, with the addition of `'static + Send`, it seems to have worked! e85f703c51fd2bbe573478e7e40fcdb43e077fc0

lib/vector-api-client/src/lib.rs
leebenson: I think you just like causing me work 😛 

```
error: aborting due to 44 previous errors
```

Adding docs and a bucket load of `#[derive]`s now... 😄

lib/vector-api-client/graphql/schema.json
```diff
@@ -1,45 +1,106 @@
...
+            "INLINE_FRAGMENT"
+          ],
+          "name": "skip"
+        }
+      ],
+      "mutationType": null,
       "queryType": {
         "name": "Query"
       },
-      "mutationType": null,
```
MOZGIII, line 7: I wonder if the request changed, or the order for the generated fields is unstable? If we're using hashmap internally somehwere it might very well be that the generated schema fields aren't guaranteed to serialize in the predefined order. If so, we have to solve it, so regenerations produce consistent results.

lib/vector-api-client/src/lib.rs
MOZGIII: :smile: "How to get 44 new compile errors with a neat simple trick"

lib/vector-api-client/graphql/schema.json
leebenson: We bumped to async-graphql 2.0.0 [this morning](https://github.com/timberio/vector/pull/4537), so it might have tweaked how this file is being generated.

If the order winds up being unstable between `cargo run --bin graphql-schema` runs, I'll file an issue at async-graphql, but I haven't noticed that previously.

lib/vector-api-client/src/lib.rs
leebenson: Added docs, and the derived traits in 97e1add9558e270e09a62c91709ab801edf11329.

I've temporarily removed the `missing_docs` rule due to errors like this:

```
error: missing documentation for a struct field
  --> lib/vector-api-client/src/gql/metrics.rs:18:10
   |
18 | #[derive(GraphQLQuery, Debug, Copy, Clone)]
   |          ^^^^^^^^^^^^
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

This looks like something `GraphQLQuery` needs to accommodate in its macros -- is that right?

PR #4544, opened 2020-10-13
enhancement: Add vector user to adm in debian packaging

### LINKED ISSUE #4543: Add `vector` user to `adm` in deb packages
Mentioned in discord: https://discordapp.com/channels/742820443487993987/746070591097798688/765572515765944370

The user had to add the `vector` user to the `adm` group in order to consume files from `/var/log`. This is not unexpected, but we could make the experience better for people on debian/ubuntu by, optimistically, adding the `vector` user to the `adm` group during package installation. We currently do this for `systemd-journal` group so there is some precedent.

We could survey other OSes we support for similar "monitoring groups".

jamtur01:
Closes #4543 

### DIFF
--- a/distribution/debian/scripts/postinst
+++ b/distribution/debian/scripts/postinst
@@ -1,8 +1,13 @@
 #!/bin/sh
 set -e
 
-usermod -aG systemd-journal vector || true
+# Add Vector to adm group to read /var/logs
+usermod --append --gid adm vector || true
 
+# Add Vector to systemd-journal to read journald logs
+usermod --append --gid systemd-journal vector || true
+
+# Reload the daemon to reflect new group membership
 if command -v systemctl >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


distribution/debian/scripts/postinst
```diff
@@ -1,8 +1,13 @@
 #!/bin/sh
 set -e
 
-usermod -aG systemd-journal vector || true
+# Add Vector to adm group to read /var/logs
+usermod --append --gid adm vector || true
```
jszwedko, line 5: long flags :heart:

jszwedko (approved):
Thanks!

distribution/debian/scripts/postinst
jamtur01: Heh. I guess we can do that from now one... :)

PR #4546, opened 2020-10-13
chore: fix shutdown tests

lukesteensen:
It's still not clear to me how this snuck in, but it seems like we started relying on the tokio `process` feature without explicitly enabling it [here](https://github.com/timberio/vector/commit/881fbb3ef12cdc04715ea18c1a81e636ddb34c8e#diff-acc89eb37f2db3e476b2c19fc1702a4f9bd795d859c7ba1858d3caa8ac56e911R34). Presumably this worked because of some other dependency enabling the feature, but that ended up breaking [here](https://github.com/timberio/vector/runs/1244690291?check_suite_focus=true).

This change just enables the feature we rely on directly and now the tests are fine locally.


jamtur01 (approved):
Weird but LGTM. :)

PR #4561, opened 2020-10-14
chore: Fixup Cargo deny

### LINKED ISSUE #4515: Resolve RUSTSEC-2020-0052
https://github.com/crossbeam-rs/crossbeam/pull/533 fixed a security vulnerability in `crossbeam` crate and released `v0.4.4` with the fix. 

We depend on `crossbeam` through `criterion` which is only used in `dev` so it doesn't have an effect on the users.

It doesn't seem that `criterion` is yet updated with this patch.

Detected in https://github.com/timberio/vector/pull/4037/checks?check_run_id=1239361723

Hoverbear:
Fixup the cargo deny rules.

Closes #4515

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1018,12 +1018,12 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.4.3"
+version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


jszwedko (approved):
Thanks! I was curious if we would just be able to update. I'm glad that was the case. Thanks for clearing out the unused rules too.

Just noting that this closes #4515

fanatid (approved):

PR #4581, opened 2020-10-15
enhancement(observability): Topology added/removed GraphQL subscriptions

leebenson:
This PR adds `topologyAdded` and `topologyRemoved` subscriptions, for detecting changes to running configuration.

This is a precursor to having `vector top` tables auto-update when the topology changes, and for the eventual UI client to rejig topology icons on the dashboard in real-time.

Unlike subscriptions that need to sample data across a timespan, these subscriptions don't take/loop on an 'interval' to detect changes. Instead, a pub/sub pattern is used (provided by `async-graphql`'s previous `SimpleBroker` helper) to alert subscribers at the point of writing new changes, removing the need for short-lived mutexes over configuration reads.

## Demo

https://www.loom.com/share/13990dadb7544c16ac46a7aec6b8eef6


src/api/schema/broker.rs
```diff
@@ -0,0 +1,67 @@
...
+use serde::export::PhantomData;
+use slab::Slab;
+use std::any::{Any, TypeId};
+use std::collections::HashMap;
+use std::pin::Pin;
+use std::sync::Mutex;
+
+/// Current subscriptions. This is static so that 'global' changes can be published to any
+/// connecting subscriber. Subscribers fall out of scope once the subscription is terminated
+static SUBSCRIBERS: Lazy<Mutex<HashMap<TypeId, Box<dyn Any + Send>>>> = Lazy::new(Default::default);
```
Hoverbear, line 14: It looks like you're making an Anymap? https://docs.rs/anymap/0.12.1/anymap/

src/api/schema/broker.rs
```diff
@@ -0,0 +1,67 @@
...
+
+struct Senders<T>(Slab<UnboundedSender<T>>);
```
Hoverbear, line 16: Do you think we could use bounded channels here?

src/api/schema/broker.rs
leebenson: Thanks! Switched to `anymap` in c672a42dac93e2ea0d2ee7db24cf412e9d73c0ce

src/api/schema/broker.rs
leebenson: 👍 Switched to bounded chans, with a size of 20. I think this should be enough to keep back-pressure low but still feel real-time for a reasonable # of connected clients.

Updated [the client to match](https://github.com/timberio/vector/pull/4581/commits/c9316ce38590ec4c01854ab5e958f00df271696b#diff-88d0e464468fdbce20a25a363596ec78830b07a1bbe556270e60ff948ffb71f3R162); previously it was 100, which was a bit high.

leebenson:
@Hoverbear - tweaked `Broker` per your suggestions. Should be good to review again. Thanks!

Hoverbear (approved):

PR #4586, opened 2020-10-15
enhancement(kubernetes_logs source): Advanced container filtering

### LINKED ISSUE #4213: Allow filtering logs by container name in `kubernetes_logs` source via the annotation
Filtering by the pod name should already be possible with #3412 via a very efficient, resource preserving way.

However, filtering by container can't be implemented like that because this filtering has to happen over the body of the `Pod` object.

To implement this, we will allow users add an `annotation` to the `Pod`: ~`vector.dev/exclude-containers/by-name/<container_name>: "true"`~, and change the paths provider to construct globs on a container level, rather than on the pod level.

UPD: the `vector.dev/exclude-containers/by-name/<container_name>` is not allowed as a valid annotation key in the Kubernetes API, so we're using `vector.dev/exclude-containers: "<container_name_1>,<container_name_2>,<...>"` format instead.

This will allow us to effectively rely on the Kuberneres API state, and for users to pick containers to exclude with high precision.

### LINKED ISSUE #4270: Exclude files from polling at the `kubernetes_logs` source by specifying globs at the vector config level
Similar to how logs are filtered at the `file` source, we need to have the ability to filter log files via globbing over their paths.

MOZGIII:
This PR adds two new ways to filter logs to the `kubernetes_logs` source:

1. Filtering by `Pod` *annotation* in the form of `vector.dev/exclude-containers: "<container_name_1>,<container_name_2>,<...>"` - log files from the `<container_name_1>` and `<container_name_2>` are excluded.
2. Filtering by the file exclusion patterns configured at the Vector configuration at the `kubernetes_logs` source - for a pattern `/var/log/pods/pod-namepspace_pod-name_pod-uid/**` all the files matching that pattern will be excluded.

Closes #4213.
Closes #4270.

To do:

- [x] E2E tests
- [ ] docs

### DIFF
--- a/lib/k8s-e2e-tests/src/lib.rs
+++ b/lib/k8s-e2e-tests/src/lib.rs
@@ -11,41 +11,67 @@ pub fn make_framework() -> Framework {
     Framework::new(interface)
 }
 
-pub fn make_test_pod<'a>(
-    namespace: &'a str,
-    name: &'a str,
-    command: &'a str,
-    labels: impl IntoIterator<Item = (&'a str, &'a str)> + 'a,
-) -> Pod {
-    let labels: std::collections::BTreeMap<String, String> = labels
+pub fn collect_btree<'a>(
+    items: impl IntoIterator<Item = (&'a str, &'a str)> + 'a,
+) -> Option<std::collections::BTreeMap<String, String>> {
+    let collected: std::collections::BTreeMap<String, String> = items
         .into_iter()
         .map(|(key, val)| (key.to_owned(), val.to_owned()))
         .collect();
-    let labels = if labels.is_empty() {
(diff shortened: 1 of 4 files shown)


MOZGIII:
Manual E2E run: https://github.com/timberio/vector/actions/runs/309734112

MOZGIII:
I'll be merging this later today to get it into the nightly builds. E2E tests are passing, so the chances of breaking things are fairly low.
The code might be a bit odd here and there, so I'd appreciate the reviews. Any change requests not included before the merge will be addressed in a separate PR.

Hoverbear (approved):
Odd, as you mentioned, but nothing blocking. :)

PR #4591, opened 2020-10-15
chore(ci): Update localstack

jszwedko:
Updates to latest localstack tag. I need this for some of the S3+SQS
work, but figured I'd do it separately.

The biggest change is that localstack now exposes all services on the
same port, 4566, simplifying our setup.

Additional changes due to localstack changes:

I updated `aws_s3` sink tests to use "Identity" for Content-Encoding as
this appears to be the default encoding now

I added a sleep in `aws_kinesis_streams` test to allow localstack to
fully create the stream (it was getting `ResourceNotFound` errors when
the sink tried to publish via `PutRecords`). I tried a `wait_for` with
`DescribeStream`, but the stream was successfully returned by that
before it was ready, still.

I updated `aws_kinesis_firehose` to create the Elasticsearch domain as
localstack now lazily starts Elasticsearch upon domain creation.

I updated `aws_kinesis_firehose` tests to just deserialize requests
directly rather than using `elastic_responses` (which allowed me to drop
this dependency). The localstack update included an ES update which
caused me to run into elastic-rs/elastic#361
. I was going to update `elastic_responses` to include this change but
the `elastic_responses` crate was dropped in newer versions of that
repo.  I then attempted to switch to the actual client that repo
provides but ran into async issues with it depending on futures01
(elastic-rs/elastic#414). That Github issue
for that indicated to me that it seems like that repo was being
deprecated in-lieu of the official client. Given we only needed an
exceedingly small subset of functionality, I just decided to parse it
with `serde_json::Value` rather than switch to the new crate; however,
that seems like the preferred option for the future is an option for the
future if we come to need a more robust ES integration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -522,7 +522,7 @@ dependencies = [
  "futures-core",
  "futures-util",
  "hex",
- "http 0.2.1",
+ "http",
  "hyper",
  "hyper-rustls",
  "hyper-unix-connector",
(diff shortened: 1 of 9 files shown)


jamtur01 (changes requested):
Looks fine other than the two elastic_response errors?

jszwedko:
Oops, I thought I checked that we weren't using those is the `elasticsearch` sink, but I was incorrect. I'll update.

jamtur01 (approved):
LGTM

fanatid:
Can we merge it? It would be great to update localstack before https://github.com/timberio/vector/pull/2755

jszwedko:
@fanatid :+1: I was just hoping to get one more review for the rust changes.

fanatid (approved):
Thanks for removing `elastic_responses`! This was the last dependency that required an old http crate.

PR #4596, opened 2020-10-16
Ignore tests when `sinks-prometheus` not enabled

fanatid:
@juchiast what you think if we will ignore tests if feature not enabled instead of not compiling them? The good thing about this is that we will see in the output that tests exists but skipped by some cases. Also `test_generate_config` will be executed even if `sinks-prometheus` is not enabled.

To pr #4557

### DIFF
--- a/src/sources/statsd/mod.rs
+++ b/src/sources/statsd/mod.rs
@@ -182,35 +182,18 @@ impl TcpSource for StatsdTcpSource {
     }
 }
 
-#[cfg(feature = "sinks-prometheus")]
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::{
-        config,
-        sinks::prometheus::PrometheusSinkConfig,
-        test_util::{next_addr, start_topology},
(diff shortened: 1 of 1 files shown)


juchiast:
We should definitely move ` #[cfg(feature = "sinks-prometheus")]` to each test function instead of on the module, I'm not sure about using `ignore` because it look a bit obscure. 

Ideally, I think we should add `sinks-prometheus` feature to `sources-statsd`, but only on test mode. Not sure if cargo has a way to configure this. 

Btw, there are other places where tests of a component depend on another component, such as in statsd sink.

fanatid:
> Ideally, I think we should add `sinks-prometheus` feature to `sources-statsd`, but only on test mode. Not sure if cargo has a way to configure this.

Yeah, that's would be perfect, but I'm also not sure how to solve this with cargo.

Another option to panic instead of ignore, but anything is better than just skipping tests compilation :slightly_smiling_face:

PR #4597, opened 2020-10-16
  chore: rid old StreamSink

fanatid:
The successor of #3893.

Before upgrade TCP, UDP and other sinks I'd like to get feedback of this approach.

In this PR I added `AckerFramedWrite` which almost precise copy of [tokio_util::codec::FramedWrite](https://docs.rs/tokio-util/0.3.1/tokio_util/codec/struct.FramedWrite.html). `AckerFramedWrite` implements [futures::Sink](https://docs.rs/futures/0.3.6/futures/sink/trait.Sink.html) where save item size in bytes in `start_send` and call `ack` in `poll_flush` when item completely send. This better approach than #3893 because we do not wait `Poll::Ready` from flush on sending for each item.


fanatid:
@lukesteensen I had updated acking by replacing AckerFramedWrite to AckerBytesSink (logic as in old StreamAck)
https://github.com/timberio/vector/blob/068f289ffca61e5631237760c88deaf90ca0ab24/src/sinks/util/mod.rs#L127-L171

src/sinks/util/unix.rs
```diff
@@ -83,49 +95,43 @@ impl UnixConnector {
...
-}
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
 
-type UnixSocket = FramedWrite<UnixStream, BytesCodec>;
-type UnixSocket01 = CompatSink<UnixSocket, Bytes>;
+    fn call(&mut self, msg: Bytes) -> Self::Future {
+        let connect = self.connector.connect();
+        async move { connect.await?.send(msg).await.context(SendError) }.boxed()
```
lukesteensen, line 119: We should not be initiating a new connection on every event. (I think you just moved this).

src/sinks/util/unix.rs
```diff
@@ -136,131 +142,58 @@ impl UnixSink {
...
+            .map(|event| match encode_event(event, &encoding) {
+                Some(bytes) => bytes,
+                None => Bytes::new(),
+            })
+            .map(Ok)
+            .peekable();
+
+        while Pin::new(&mut input).peek().await.is_some() {
+            let mut sink = self.connect().await;
+            if let Err(error) = sink.send_all(&mut input).await {
```
lukesteensen, line 187: If I'm understanding this correctly, on every event we will initiate a new connection with a new sink stack and flush that one event all the way though? Doing things one at a time like this will likely be extremely slow and does not get us any benefit from batching writes or acks.

lukesteensen (changes requested):
The new sink adapter for acks looks good, but the `StreamSink` implementation is sending everything one at a time and that needs to be fixed.

src/sinks/util/unix.rs
fanatid: I only moved this code inside the file. Issue about this: https://github.com/timberio/vector/issues/4610

src/sinks/util/unix.rs
fanatid: No, it's not correct. It was in original PR (https://github.com/timberio/vector/pull/3893). Here we initiate connection once we have an item in the stream (`.peek()`), but not consume it. Once we have an item we connect and forward all items from the stream. If we have error on sending, we log it and wait for new item again, then start the connection, and so on.

juchiast:
Some sinks such as `vector` use a custom encoding function, I'm not sure how it can be done with StreamSink trait?

https://github.com/timberio/vector/blob/master/src/sinks/vector.rs#L56-L59

fanatid:
It was already done in https://github.com/timberio/vector/pull/3893 (so should not be hard to do same here).
`TcpSink` in `StreamSink::run` can encode item with `encode_function` which will be a member of `TcpSink`.

src/sinks/util/unix.rs
lukesteensen: Ah, that makes sense. Sorry, should have read closer! :smile:

lukesteensen (approved):
Looks good!

src/sinks/util/unix.rs
```diff
@@ -136,131 +142,58 @@ impl UnixSink {
...
-            Err(error) => {
-                emit!(UnixSocketFlushFailed {
+#[async_trait]
+impl StreamSink for UnixSink {
+    async fn run(&mut self, input: BoxStream<'_, Event>) -> Result<(), ()> {
+        let encoding = self.encoding.clone();
+        let mut input = input
+            .map(|event| match encode_event(event, &encoding) {
+                Some(bytes) => bytes,
+                None => Bytes::new(),
```
juchiast, line 180: Can we use `filter_map` here to avoid sending an empty `Bytes` to the sink?

fanatid: Yes, we can, but then we will need to add `ack(1)` and extra `emit!`. I just double-checked `FramedWrite` and `BytesCodec` -- it's completely okay send empty `Bytes`, only a few extra calls which do nothing.

juchiast: This seems subtle and can lead to incorrect acking, so I think we should add a comment for it, may be at AckerBytesSink?

fanatid: Yeah, comment would be great, I'll add.

src/sinks/util/acker_bytes_sink.rs
```diff
@@ -0,0 +1,73 @@
...
+        self.project().inner.poll_ready(cx)
+    }
+
+    fn start_send(self: Pin<&mut Self>, item: Bytes) -> Result<(), Self::Error> {
+        let pinned = self.project();
+        pinned.sizes.push(item.len());
+        pinned.inner.start_send(item)
+    }
+
+    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
```
jszwedko, line 62: This is probably just my ignorance, but can you explain how the failure case works here? What happens if, for example, the unix socket cannot send the event in `call` below?

jszwedko (commented):
Overall this looks good, I'm just wondering how the send failure case is represented and handled.

src/sinks/util/acker_bytes_sink.rs
fanatid: If `UnixSocket` cannot send the event then `FramedWrite` will pass error to us (to `result` variable), but we do not have any logic here for handling this error so we also pass it on next level. On next level [SendAll::poll](https://docs.rs/futures-util/0.3.6/src/futures_util/sink/send_all.rs.html#78-112) also pass error to our loop in `StreamSink::run` method and finally there we handler error (by emitting event and reconnecting).

fanatid:
Little more complicated than the originally proposed version because I found that [send_all](https://docs.rs/futures-util/0.3.6/src/futures_util/sink/send_all.rs.html) have one buffered item and in disconnection case, it would be probability to lose and ack one event even without an attempt to send it. Except additional wrappers and migration to StreamSink, I also changed UDP to async version and adjusted internal events.

juchiast:
`sinks::socket::test::tcp_stream_detects_disconnect` is failing.

src/sinks/util/udp.rs
```diff
@@ -104,201 +117,166 @@ impl UdpConnector {
...
+    Disconnected,
     Connecting(BoxFuture<'static, UdpSocket>),
     Connected(UdpSocket),
+    Sending(oneshot::Receiver<UdpSocket>),
 }
 
 pub struct UdpService {
-    state: UdpServiceState,
+    connector: UdpConnector,
+    state: Cell<UdpServiceState>,
```
juchiast, line 156: `Cell` is not needed

```diff
diff --git a/src/sinks/util/udp.rs b/src/sinks/util/udp.rs
index ed5217a1f..e4d3c2fc9 100644
--- a/src/sinks/util/udp.rs
+++ b/src/sinks/util/udp.rs
@@ -16,7 +16,6 @@ use futures::{future::BoxFuture, ready, stream::BoxStream, FutureExt, StreamExt}
 use serde::{Deserialize, Serialize};
 use snafu::{ResultExt, Snafu};
 use std::{
-    cell::Cell,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     pin::Pin,
     sync::Arc,
@@ -153,14 +152,14 @@ enum UdpServiceState {
 
 pub struct UdpService {
     connector: UdpConnector,
-    state: Cell<UdpServiceState>,
+    state: UdpServiceState,
 }
 
 impl UdpService {
     fn new(connector: UdpConnector) -> Self {
         Self {
             connector,
-            state: Cell::new(UdpServiceState::Disconnected),
+            state: UdpServiceState::Disconnected,
         }
     }
 }
@@ -172,7 +171,7 @@ impl tower::Service<Bytes> for UdpService {
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         loop {
-            self.state = Cell::new(match self.state.get_mut() {
+            self.state = match &mut self.state {
                 UdpServiceState::Disconnected => {
                     let connector = self.connector.clone();
                     UdpServiceState::Connecting(Box::pin(async move {
@@ -191,7 +190,7 @@ impl tower::Service<Bytes> for UdpService {
                     };
                     UdpServiceState::Connected(socket)
                 }
-            });
+            };
         }
         Poll::Ready(Ok(()))
     }
@@ -199,10 +198,11 @@ impl tower::Service<Bytes> for UdpService {
     fn call(&mut self, msg: Bytes) -> Self::Future {
         let (sender, receiver) = oneshot::channel();
 
-        let mut socket = match self.state.replace(UdpServiceState::Sending(receiver)) {
-            UdpServiceState::Connected(socket) => socket,
-            _ => panic!("UdpService::poll_ready should be called first"),
-        };
+        let mut socket =
+            match std::mem::replace(&mut self.state, UdpServiceState::Sending(receiver)) {
+                UdpServiceState::Connected(socket) => socket,
+                _ => panic!("UdpService::poll_ready should be called first"),
+            };
 
         Box::pin(async move {
             // TODO: Add reconnect support as TCP/Unix?
```

juchiast (changes requested):

src/sinks/util/tcp.rs
```diff
@@ -108,249 +94,157 @@ impl TcpConnector {
...
-    backoff: ExponentialBackoff,
-    span: tracing::Span,
-}
-
-enum TcpSinkState {
-    Disconnected,
-    Connecting(Box<dyn Future<Item = TcpOrTlsStream, Error = TcpError> + Send>),
-    Connected(TcpOrTlsStream01, OpenTokenDyn),
-    Backoff(Box<dyn Future<Item = (), Error = ()> + Send>),
+    events_counter: Arc<EventsCounter>,
```
juchiast, line 145: Everything in these sinks run in a single thread, so I don't think we need to use mutex. I think we can get away with borrow checker issue by using Option<EventsCounter>, take and move it into `BytesSink`, then take it out of BytesSink when a send error happens. This will require calling `encode_event` in `start_send` of `BytesSink`, mentioned in another comment.

I have not tried this out so not sure if it actually work.

src/sinks/util/socket_events_counter.rs
```diff
@@ -0,0 +1,169 @@
...
+    task::{Context, Poll},
+};
+use tokio::io::AsyncWrite;
+use tokio_util::codec::{BytesCodec, FramedWrite};
+
+const MAX_PENDING_ITEMS: usize = 10_000;
+
+/// Count number of encoded events and ack on request.
+pub struct EventsCounter {
+    sizes: Mutex<Vec<usize>>,
```
juchiast, line 19: If we really need to use `Mutex`, let's use `tokio::sync::Mutex` instead to avoid blocking on `lock()`.

juchiast (changes requested):

src/sinks/util/udp.rs
fanatid: Wow, thanks! I did not use `std:mem` previously.

src/sinks/util/tcp.rs
```diff
@@ -108,249 +94,157 @@ impl TcpConnector {
...
+        // We need [Peekable](https://docs.rs/futures/0.3.6/futures/stream/struct.Peekable.html) for initiating
+        // connection only when we have something to send.
+        // It would be cool encode events before `peekable()` but [SendAll](https://docs.rs/futures-util/0.3.6/src/futures_util/sink/send_all.rs.html)
+        // have buffered item, so if our connection fail we will completely lost one item.
+        // As result `EventsCounter` was added for calculating encoded events and ack only consumed items on failed send.
+        let mut input = Some(input.peekable());
+        while Pin::new(input.as_mut().unwrap()).peek().await.is_some() {
+            let events_counter = Arc::clone(&self.events_counter);
+            let encode_event = move |event| future::ready(events_counter.encode_event(event));
+            let mut stream = input.take().unwrap().filter_map(encode_event);
```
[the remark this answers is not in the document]
fanatid: The short answer is no, we need wrap a stream. Yes, we can `encode_event` from `start_send` in `BytesSink`, but.. in this case we will not count the taken item from the original stream to buffer in `send_all` (originally proposed changes did not count buffer in `send_all`).

i.e. we have stream like:
input stream -> peekable -> encode_event (counter here) -> send_all (buffer here) -> sink
if we move `encode_event` after `send_all` we will not ack enough items on disconnect

src/sinks/util/socket_events_counter.rs
fanatid: That's completely okay to use `std::sync::Mutex` in async code since our lock almost instantly. This even mentioned in `tokio::sync::Mutex` docs (https://docs.rs/tokio/0.2.21/tokio/sync/struct.Mutex.html):
> There are some situations where you should prefer the mutex from the standard library. Generally this is the case if:
> The duration of any single lock is near-instant.

I'm not completely sure but I think it's even possible to use `RefCell` here because as I understand there should not be concurrent access to `sizes` from different places. From the other side `Mutex` looks safer and I not sure about the performance difference between `Mutex` and `RefCell`.

juchiast: > That's completely okay to use std::sync::Mutex in async code since our lock almost instantly.
Seem fair.

`AtomicUsize` is also an option.

src/sinks/util/tcp.rs
ktff: I agree with the above comment.

I partly see what lead to this design, but there is no need for it. Since `send_all` of `futures` 0.3 doesn't consume the sink we can recover `EventsCounter` from it. With that there is no need for any mutex or atomic.  

And even if the above is wrong, `Mutex` is expensive to call for every event especially since there aren't other threads accessing this, so this complexity and performance cost are really not worth it.

src/sinks/util/socket_events_counter.rs
ktff: What's the need for `Vec`? Can't we count events and sum their sizes instead? If it's for internal events only, then they can be changed to accept aggregated data.

src/sinks/util/tcp.rs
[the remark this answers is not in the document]
ktff: In that case there is no need for `encode_event_inner` to be in `EventCounter`. While `on_success` seams fine to be there.

src/sinks/util/tcp.rs
fanatid: If we move `encode_event` function to `BytesSink` how we will able count events bytes in `EventsCounter`?

src/sinks/util/socket_events_counter.rs
fanatid: We can drop `Vec` in favour of `AtomicUsize` but only if can submit more events/bytes to internal metrics that actually happen, otherwise we need `Vec` for proper values =\

juchiast (changes requested):
I think we have data loss in `FramedWrite` buffer, here's a test (in `tcp.rs`):

```rust
    #[tokio::test]
    async fn events_lost_in_buffer() {
        trace_init();

        let listen_addr = next_addr();

        let server = tokio::spawn({
            let listen_addr = listen_addr.clone();
            async move {
                let mut server = tokio::net::TcpListener::bind(listen_addr).await.unwrap();

                let stream = server.accept().await.unwrap().0;
                stream.shutdown(std::net::Shutdown::Both).unwrap();

                let stream = server.accept().await.unwrap().0;
                let mut reader = tokio_util::codec::FramedRead::new(
                    stream,
                    tokio_util::codec::LinesCodec::new(),
                );
                for _ in 0..6 {
                    let x = reader.next().await.unwrap().unwrap();
                    println!("recv {}", x);
                }
                reader
                    .into_inner()
                    .shutdown(std::net::Shutdown::Both)
                    .unwrap();
            }
        });

        let events = (0..6)
            .map(|i| {
                let mut map = std::collections::HashMap::new();
                map.insert(
                    crate::config::log_schema().message_key().to_owned(),
                    crate::event::Value::from(i.to_string()),
                );
                crate::Event::Log(map.into())
            })
            .collect::<Vec<crate::Event>>();

        let (mut sender, receiver) = futures::channel::mpsc::channel(10);

        let config = TcpSinkConfig {
            address: listen_addr.to_string(),
            tls: None,
        };
        let encoding = crate::sinks::util::Encoding::Text.into();
        let encode = move |event| crate::sinks::util::encode_event(event, &encoding);
        let cx = SinkContext::new_test();
        let sink = config.build(cx, encode).unwrap().0;
        let sink = tokio::spawn(async move {
            sink.run(receiver.boxed()).await.unwrap();
        });

        for i in 0..3 {
            sender.try_send(events[i].clone()).unwrap();
        }
        tokio::time::delay_for(Duration::from_secs(1)).await;
        for i in 3..6 {
            sender.try_send(events[i].clone()).unwrap();
        }
        drop(sender);

        let (r1, r2) = future::join(sink, server).await;
        r1.unwrap();
        r2.unwrap();
    }
```

Result:
```
running 1 test
recv 3
recv 4
recv 5
thread 'sinks::util::tcp::test::events_lost_in_buffer' panicked at 'called `Option::unwrap()` on a `None` value', src/sinks/util/tcp.rs:263:49
thread 'sinks::util::tcp::test::events_lost_in_buffer' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(...)', src/sinks/util/tcp.rs:309:12
test sinks::util::tcp::test::events_lost_in_buffer ... FAILED

failures:

failures:
    sinks::util::tcp::test::events_lost_in_buffer
```

The test simulates:
- A sources that:
    - Produces 3 event.
    - Sleep for 1s.
    - Produces another 3 events.
- A server that:
    - Disconnect immediately on the first connection.
    - Read 6 events on the second connection.

Here's what happen: after 3 event are buffered on `FramedWrite` (via `start_send`), `check_shutdown` is called and `BytesSink` is dropped, so all 3 buffered events are dropped with it. Because we took those events out of the input stream, they will be lost forever.

fanatid:
I think we can remove some complexity here by implementing our own `SendAll`. But our `SendAll` will be without buffer because we will use buffer from `Peekable` as result we will not need `EventsCounter` and will able place `encode_event` before `Peekable`.
@lukesteensen are you ok with partially copy of `SendAll` here?

src/sinks/util/tcp.rs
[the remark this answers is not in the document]
ktff: This https://github.com/timberio/vector/pull/4597#discussion_r513613709 seams to be for this conversation. 

`EventsCounter` could have method like `fn add_event(&self, byte_size: usize)` which `BytesSink` would call on successful encode of an event.

src/sinks/util/socket_events_counter.rs
ktff: Regarding `Vec`
> if can submit more events/bytes to internal metrics that actually happen

true, I missed that. `Vec` is fine.

src/sinks/util/tcp.rs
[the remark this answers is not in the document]
ktff: @fanatid yep, sorry the above comment was indeed for the the other conversation, I'll correct it.

src/sinks/util/tcp.rs
ktff: `EventsCounter` could have method like `fn add_event(&self, byte_size: usize)` which `BytesSink` would call on successful encode of an event.

lukesteensen (commented):
> @lukesteensen are you ok with partially copy of SendAll here?

Yes, I think that's reasonable. It does seem like that could help simplify things a bit. We should try that, but not spend too much time before merging this. The longer it stays open the more merge conflicts are likely. We can always merge and them do some simplifications as follow ups.

fanatid:
@juchiast @ktff I added `SendAll` for `Peekable`, so `EventsCounter` removed and it's possible to use `encode_event` before `.peekable()`. Many things are simplified.

ktff (approved):
Looks good. I've just left a few doc changes. And regarding the failing test that @juchiast commented, it's fine, ideally buffers/acker should deal with it.

src/sinks/util/udp.rs
```diff
@@ -98,197 +110,168 @@ impl UdpConnector {
...
+            .map(|event| match encode_event(event) {
+                Some(bytes) => bytes,
+                None => Bytes::new(),
+            })
+            .peekable();
+
+        while Pin::new(&mut input).peek().await.is_some() {
+            let mut socket = self.connector.connect_backoff().await;
+            while let Some(bytes) = input.next().await {
+                let result = udp_send(&mut socket, &bytes).await;
```
juchiast, line 245: Should we send an empty packet here?

src/sinks/util/udp.rs
```diff
@@ -98,197 +110,168 @@ impl UdpConnector {
...
-    backoff: ExponentialBackoff,
-}
-
-enum State {
-    Initializing,
-    Connecting(Box<dyn Future<Item = UdpSocket, Error = UdpError> + Send>),
-    Connected(UdpSocket),
-    Backoff(Box<dyn Future<Item = (), Error = ()> + Send>),
+    acker: Acker,
+    encode_event: Arc<dyn Fn(Event) -> Option<Bytes> + Send + Sync>,
```
juchiast, line 213: `Box` can be used here:

```diff
diff --git a/src/sinks/util/udp.rs b/src/sinks/util/udp.rs
index 267b084d3..a777882ab 100644
--- a/src/sinks/util/udp.rs
+++ b/src/sinks/util/udp.rs
@@ -18,7 +18,6 @@ use snafu::{ResultExt, Snafu};
 use std::{
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     pin::Pin,
-    sync::Arc,
     task::{Context, Poll},
     time::Duration,
 };
@@ -210,7 +209,7 @@ impl tower::Service<Bytes> for UdpService {
 struct UdpSink {
     connector: UdpConnector,
     acker: Acker,
-    encode_event: Arc<dyn Fn(Event) -> Option<Bytes> + Send + Sync>,
+    encode_event: Box<dyn Fn(Event) -> Option<Bytes> + Send + Sync>,
 }
 
 impl UdpSink {
@@ -222,7 +221,7 @@ impl UdpSink {
         Self {
             connector,
             acker,
-            encode_event: Arc::new(encode_event),
+            encode_event: Box::new(encode_event),
         }
     }
 }
@@ -230,21 +229,15 @@ impl UdpSink {
 #[async_trait]
 impl StreamSink for UdpSink {
     async fn run(&mut self, input: BoxStream<'_, Event>) -> Result<(), ()> {
-        let encode_event = Arc::clone(&self.encode_event);
-        let mut input = input
-            // We send event empty events because we need `ack` and `emit!`.
-            .map(|event| match encode_event(event) {
-                Some(bytes) => bytes,
-                None => Bytes::new(),
-            })
-            .peekable();
+        let mut input = input.peekable();
 
         while Pin::new(&mut input).peek().await.is_some() {
             let mut socket = self.connector.connect_backoff().await;
-            while let Some(bytes) = input.next().await {
+
+            while let Some(event) = input.next().await {
+                let bytes = (self.encode_event)(event).unwrap_or_default();
                 let result = udp_send(&mut socket, &bytes).await;
                 self.acker.ack(1);
-
                 match result {
                     Ok(()) => emit!(SocketEventsSent {
                         mode: SocketMode::Udp,
@@ -255,7 +248,7 @@ impl StreamSink for UdpSink {
                         emit!(UdpSocketError { error });
                         break;
                     }
-                };
+                }
             }
         }
```

juchiast (approved):
Look good! I had some comments on UDP sink.

src/sinks/util/udp.rs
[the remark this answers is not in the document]
fanatid: Good catch! Not sure what will happen if we call it with empty `bytes`. Will fix it.

PR #4615, opened 2020-10-17
enhancement(metrics): Expose internal metrics cardinality as a internal metric counter

MOZGIII:
This PR adds an `internal_metrics_cardinality` counter, that exposes the cardinality of the internal metrics registry.

This should help with diagnosing issues like https://github.com/timberio/vector/issues/2356 and it might help with https://github.com/timberio/vector/issues/4582.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -1,5 +1,5 @@
 use crate::{event::Metric, Event};
-use metrics::{Key, Label, Recorder, Unit};
+use metrics::{Key, KeyData, Label, Recorder, Unit};
 use metrics_tracing_context::{LabelFilter, TracingContextLayer};
 use metrics_util::layers::Layer;
 use metrics_util::{CompositeKey, Handle, MetricKind, Registry};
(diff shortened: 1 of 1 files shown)


MOZGIII:
Pushed a docker image off this as `mozgiii/vector-k8s-wip:bug2`

spencergilbert:
It looks like something about this change has broken (I suspect) the `prometheus` sink. The endpoint has stopped serving metrics and requests against it just hang indefinitely.

MOZGIII:
Thanks for the feedback! I'll investigate this issue further.

It also seems like this causes our tests to hang in the CI. Very interesting, and quite surprising.

src/metrics.rs
```diff
@@ -8,11 +8,23 @@ use std::sync::Arc;
 
 static CONTROLLER: OnceCell<Controller> = OnceCell::new();
 
+// Cardinality counter parameters, expose the internal metrics registry
+// cardinality.
+// Useful for the end users to help understand the characteristics of their
+// environment and how vectors acts in it.
+const CARDINALITY_KEY_NAME: &str = "internal_metrics_cardinality";
+static CARDINALITY_KEY_DATA: KeyData = KeyData::from_static_name(CARDINALITY_KEY_NAME);
+static CARDINALITY_KEY: CompositeKey =
```
Hoverbear, line 17: Love const fn

Hoverbear (commented):
This looks good overall but I'm curious about this bug @spencergilbert is reporting!

MOZGIII:
Yes, something's wrong with the tests as well. I'll try to figure out what's wrong with this in the near future, as it might help with performance analysis.

jamtur01:
@MOZGIII What's the status here?

Hoverbear (approved):

PR #4617, opened 2020-10-17
chore(ci): Add transforms-metric_to_log transform dependency to sink-humio

jszwedko:
Given that the humio metrics sink depends on the log_to_metric
transform.

Also fix Makefile to run humio integration tests. I couldn't find a good
way to use a wildcard for the module (like
"::humio::*::integration_tests") so I just did the logs submodule for
now. It seems like there are plans to add regex support for test
filters: https://github.com/rust-lang/rust/issues/53278

Fixes https://github.com/timberio/vector/actions/runs/311846034


jamtur01 (approved):
LGTM

jamtur01:
Looks like the Cross and Mac jobs are hung. Going to re-run.

FungusHumungus (approved):
Good catch. Sorry, I should have spotted that. It looks like the check-component-features CI job picked it up too.

PR #4630, opened 2020-10-18
chore(ci): Adds benchmark results GA action

### LINKED ISSUE #2800: Add github-action-benchmark to benches suite at CI
https://github.com/rhysd/github-action-benchmark

jamtur01:
Closes #2800 

Adds https://github.com/rhysd/github-action-benchmark to our benchmark tests.

Adds commit comments like upon regression: https://github.com/rhysd/github-action-benchmark/commit/077dde1c236baba9244caad4d9e82ea8399dae20#commitcomment-36047186


Makefile
```diff
@@ -737,7 +737,7 @@ test-wasm: $(WASM_MODULE_OUTPUTS)  ### Run engine tests
 
 .PHONY: bench
 bench: ## Run benchmarks in /benches
-	${MAYBE_ENVIRONMENT_EXEC} cargo bench --no-default-features --features "${DEFAULT_FEATURES}"
+	${MAYBE_ENVIRONMENT_EXEC} cargo bench --no-default-features --features "${DEFAULT_FEATURES}" | tee output.txt
```
jszwedko, line 740: I might prefer to have this redirection in `benches.yml` given I don't think we'd want it to happen locally.

At the least, we should add it to `.gitignore`.

jszwedko (approved):
This is cool!

Makefile
jamtur01: Is there a benches.yml? I can't find any reference to this in the cargo bench docs?

jszwedko: Sorry, I meant in the GA workflow file `.github/workflows/benches.yml`. We could redirect the output there since the GA action is technically the thing that needs it and avoid creating the file locally when `make bench` is run. Not a big deal though, I'm fine with creating the file and `.gitignore`ing it as you have here.

PR #4635, opened 2020-10-19
fix: Generate valid config for all components

### LINKED ISSUE #3885: `vector generate` output does not include required fieds

juchiast:
close #3885 

Also added `generate_all` test which will generate and parse all submitted components.


juchiast:
hmm something when wrong when I use `make signoff`

PR #4636, opened 2020-10-19
fix: Generate valid config for all components

### LINKED ISSUE #3885: `vector generate` output does not include required fieds

juchiast:
close #3885 

Also added `generate_all` test which will generate and parse all submitted components.


src/transforms/add_fields.rs
```diff
@@ -51,7 +51,11 @@ inventory::submit! {
     TransformDescription::new::<AddFieldsConfig>("add_fields")
 }
 
-impl GenerateConfig for AddFieldsConfig {}
+impl GenerateConfig for AddFieldsConfig {
+    fn generate_config() -> toml::Value {
+        toml::from_str(r#"fields.name = "TODO""#).unwrap()
```
bruceg, line 56: Is this something that needs completing? If not, I suggest changing the name above.

src/transforms/swimlanes.rs
```diff
@@ -71,7 +71,14 @@ inventory::submit! {
...
 
-impl GenerateConfig for SwimlanesConfig {}
+impl GenerateConfig for SwimlanesConfig {
+    fn generate_config() -> toml::Value {
+        toml::Value::try_from(Self {
+            lanes: IndexMap::new(),
+        })
+        .unwrap()
+    }
+}
```
bruceg, line 81: Should this transform also have a `generate_config` test like the others?

src/transforms/tag_cardinality_limit.rs
```diff
@@ -69,7 +69,16 @@ inventory::submit! {
...
+impl GenerateConfig for TagCardinalityLimitConfig {
+    fn generate_config() -> toml::Value {
+        toml::Value::try_from(Self {
+            mode: Mode::Exact,
+            value_limit: default_value_limit(),
+            limit_exceeded_action: default_limit_exceeded_action(),
+        })
+        .unwrap()
+    }
+}
```
bruceg, line 81: Ditto comment about `generate_config` test

src/generate.rs
```diff
@@ -417,8 +464,12 @@ when_full = "block"
 [sinks.sink0]
 healthcheck = true
 inputs = ["TODO"]
```
bruceg, line 466: Is this still incomplete?

src/generate.rs
```diff
@@ -435,6 +486,9 @@ when_full = "block"
 inputs = []
 type = "add_fields"
 
+[transforms.transform0.fields]
+name = "TODO"
```
bruceg, line 490: I this still TODO?

bruceg (commented):
I have some questions about bits that look incomplete, but it looks ok otherwise.

src/transforms/add_fields.rs
juchiast: changed to `field_name`

src/transforms/tag_cardinality_limit.rs
juchiast: Added!

src/transforms/swimlanes.rs
juchiast: Added!

src/generate.rs
juchiast: This is generated from fields.name = "TODO" above.

src/generate.rs
juchiast: This happens when there is no input for the component, TODO here is meant for the user to fill it up. We can either output an empty list or `["TODO"]` here.

bruceg: If it is not something for a developer to fill in, I think we should avoid the "TODO" token. That implies that there is something left incomplete within Vector. This could be something like `"source-name-goes-here"` or equivalent to indicate what kind of value it should be filled with.

bruceg (approved):

jszwedko (approved):
Should we drop the default implementation of `generate_config` on `GenerateConfig`? Or is it still useful in some circumstances?

juchiast:
@jszwedko some components still use the default implementation. I removed the default implementation and added tests for those components.

src/transforms/dedupe.rs
```diff
@@ -18,8 +18,6 @@ pub enum FieldMatchConfig {
     MatchFields(Vec<String>),
     #[serde(rename = "ignore")]
     IgnoreFields(Vec<String>),
-    #[serde(skip)]
-    Default,
```
juchiast, line 22: Hit error `FieldMatchConfig::Default` cannot be serialized. I fixed this and also remove runtime panic.

PR #4653, opened 2020-10-20
chore(ci): Remove the problem matcher from the check-component-features

MOZGIII:
We have a lot of warnings coming from the `check-component-features`, and it looks like we don't intend to fix them and keep them strictly warning-less. I'd say it's thus meaningful to see those warnings in the CI, as it clutters the real issues.
The proposed fix for now implemented in this is to remove the problem matchers from the `check-component-features` CI runs.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -371,7 +371,6 @@ jobs:
     container: timberio/ci_image
     steps:
       - uses: actions/checkout@v2
-      - run: echo "::add-matcher::.github/matchers/rust.json"
       - run: make slim-builds
       - run: echo "home/runner/.local/bin" >> "$GITHUB_PATH"
       - run: make check-component-features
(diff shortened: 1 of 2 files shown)


jamtur01:
I'd prefer we fix these rather than plastering over it.

MOZGIII:
> I'd prefer we fix these rather than plastering over it.

Let's create an issue then! We should also tighten the rules at the `check-component-features` to consider critical warnings as errors (like we have it for the regular `check` job). I agree this would be a better resolution in the long run.

The fact that we have a weaker failure condition for the `check-component-features` means that up until now we didn't really consider this a big issue. We might want to discuss this is the chat with everyone to gain consensus.

jszwedko:
@MOZGIII @jamtur01 I think this can be closed since we make the check fail workflows now?

jamtur01:
Agreed.

PR #4656, opened 2020-10-20
chore(ci): Only run E2E test suite on non-PRs, or if the opt-in label is set on the PR

MOZGIII:
This PR enables an opt-in way to run the K8s E2E tests for PRs in CI.


jamtur01 (approved):
Couldn't you have the 2nd job just rely on the first? No big deal but might save some typing.

MOZGIII:
I thought about that, but this is a bit lighter on the computing resources.

PR #4661, opened 2020-10-20
fix(statsd source): Fixup statsd shutdown

Hoverbear:
Fixup a problem from https://github.com/timberio/vector/pull/4557/ where the test was not updated.


jszwedko (approved):
Thank you!

PR #4672, opened 2020-10-21
fix(file source): only sleep when backing off

lukesteensen:
This adds instrumentation around where the file source is spending its time. The ratio of time spent across various segments is reported as a debug log, so it's easily enabled with `LOG=info,file_source=debug`. The output looks roughly like the following:

```
{"checkpointing": 0.00042231457, "discovery": 0.0002874473, "other": 0.00043339972, "reading": 0.012727213, "sending": 0.98606783, "sleeping": 0.000062016334}
```

As part of adding that instrumentation, I found that it was spending a meaningful amount of time sleeping when it should not have been. It turns out this was due to calling `delay_for` with a zero duration and the fact that tokio timers only have millisecond resolution. This lead to us sleeping for a millisecond when we should not have been sleeping at all.

### DIFF
--- a/lib/file-source/src/file_server.rs
+++ b/lib/file-source/src/file_server.rs
@@ -310,10 +310,13 @@ where
             // call. Also since we are using block_on here and in the above code,
             // this should be run in it's own thread. `spawn_blocking` fulfills
             // all of these requirements.
-            match block_on(select(
-                shutdown,
-                delay_for(Duration::from_millis(backoff as u64)),
-            )) {
+            let sleep = async move {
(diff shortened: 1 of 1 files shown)


jamtur01 (approved):
LGTM

jszwedko (approved):
This is nifty! I wonder if we could publish these as internal metrics in the future.

MOZGIII:
:tada: I'm taking this for testing in my load-test setup. I could reproduce the slowdown behavior, and, hopefully, this will fix it.

PR #2755, opened 2020-06-06
feat(new sink): aws sqs

francescop:
This introduce AWS SQS as sink.
Closes #2138

PR #4674, opened 2020-10-21
chore(aws_kinesis_streams sink): remove Arc and fix event

fanatid:
Noted while worked on #2755 

- Move `AwsKinesisStreamsEventSent` to correct place. The event should be generated on success request, not on event encoding.
- Remove duplicated `Arc` from `KinesisService`. Inner value of `rusoto_core::Client` already use `Arc`, so we had `Arc` over `Arc`.

### DIFF
--- a/src/sinks/aws_kinesis_streams.rs
+++ b/src/sinks/aws_kinesis_streams.rs
@@ -27,7 +27,6 @@ use snafu::Snafu;
 use std::{
     convert::TryInto,
     fmt,
-    sync::Arc,
     task::{Context, Poll},
 };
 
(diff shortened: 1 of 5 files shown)


src/sinks/aws_kinesis_streams.rs
```diff
@@ -185,13 +182,25 @@ impl Service<Vec<PutRecordsRequestEntry>> for KinesisService {
...
         let request = PutRecordsInput {
             records,
             stream_name: self.config.stream_name.clone(),
         };
 
-        Box::pin(async move { client.put_records(request).await }.instrument(info_span!("request")))
+        Box::pin(async move {
+            client
+                .put_records(request)
+                .inspect_ok(|_| {
```
jszwedko, line 196: TIL `inspect_ok`, this is useful.

jszwedko (approved):
Thanks for these clean-ups!

PR #4681, opened 2020-10-21
enhancement: Add `ConnectionOpen` gauge

ktff:
Ref. #4505

Going with the simplicity, only one event was added for `tcp`/`unix` X `sink`/`source` connections. We can later split/enrich this event as needed. 

This event also covers other sources and sinks that use the shared code, more exactly this covers connections in: 
* `statsd` sink
* `socket` sink
* `vector` sink
* `papertrail` sink
* `syslog` source
* `vector` source
* `socket` source

There are two main problems that this PR addresses:
1. Cover all possible ways a connection can end. 
	- While the opening of a connection always has a single entry point, closing does not. 
	- This is solved by bundling, or tying in some other way, a token with the connection so that when the connection is dropped, the token is also and an event is emitted.

2. Unordered writes/emits to a gauge. 
	- In situations where we have a thread/future for every connection, just emitting current value of the gauge, noting that we are atomically modifying this value, can end with an older value of the gauge being emitted last which would then be reported as correct for an undetermined amount of time. 
	- To solve this, we have an algorithm which quickly reaches consistency, but which requires for the event to be emittable multiple times.   
	- An alternative would be to enforce sequentiality by using a lock, that would avoid the need for multiple emits, but it introduces locking for a quite a small thing.

cc. @binarylogic

### DIFF
--- a/src/internal_events/open.rs
+++ b/src/internal_events/open.rs
@@ -102,7 +102,7 @@ mod tests {
         let handles = (0..n)
             .map(|_| {
                 let gauge = gauge.clone();
-                let value = value.clone();
+                let value = Arc::clone(&value);
                 thread::spawn(move || {
                     let mut work = 0;
                     for _ in 0..m {
--- a/src/internal_events/open.rs
+++ b/src/internal_events/open.rs
@@ -93,7 +93,7 @@ mod tests {
 
     /// If this failes at any run, then the algorithm in `gauge_add` is faulty.
     #[test]
-    fn eventualy_consistent() {
+    fn eventually_consistent() {
         let n = 8;
         let m = 1000;
         let gauge = OpenGauge::new();
--- a/src/internal_events/open.rs
+++ b/src/internal_events/open.rs
@@ -5,6 +5,8 @@ use std::sync::{
     Arc,
 };
 
+pub type OpenTokenDyn = OpenToken<Box<dyn Fn(usize) + 'static + Send>>;
+
 #[derive(Debug)]
 pub struct ConnectionOpen {
     pub count: usize,
@@ -36,6 +38,12 @@ impl OpenGauge {
             gauge: self.gauge,
             emitter,
         }
+    }
+}
+
+impl Default for OpenGauge {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/internal_events/open.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/// If reporting gauges from multiple threads, they can end up in a wrong order
+/// resulting in having wrong value for a prolonged period of time.
+/// This function performs a synchronization procedure that corrects that.
+fn gauge_add(gauge: &AtomicUsize, add: isize, emitter: impl Fn(usize)) {
+    // Lock-free procedure with eventual consistency
+
+    // Initialize value and acquire older writes to gauge metric.
+    let mut value = gauge.load(Ordering::Acquire);
```
Hoverbear, line 68: It looks like this is doing https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_add ?

src/internal_events/open.rs
```diff
@@ -0,0 +1,129 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::mem::drop;
+    use std::thread;
+
+    /// If this failes at any run, then the algorithm in `gauge_add` is faulty.
+    #[test]
+    fn eventualy_consistent() {
```
Hoverbear, line 96:
```suggestion
    fn eventually_consistent() {
```

src/internal_events/open.rs
ktff: Not really, `fetch_add` would be ideal if there wasn't the issue of unordered `gauge!(...)` calls. For example, with `fetch_add`, two threads, both trying to increment gauge by 1, and `Atomic` `gauge` with start value of 0, the following situation can happen:
- thread 0: v0 = `gauge.fetch_add(1)` = 0; --------------`gauge = 1`
- thread 1: v1 = `gauge.fetch_add(1)` = 1; --------------`gauge = 2`
- thread 1: `gauge!(open_connections = v1 + 1)`; ----gauge `open_connection` = 2
- thread 0: `gauge!(open_connections = v0 + 1)`; ----gauge `open_connection` = 1

In this scenario `open_connection` gauge would be reported as 1 for as long as nothing else updates it. So to avoid that we use the commented algorithm.

Hoverbear: Ah ok, got you. :)

Hoverbear (approved):

src/sources/util/tcp.rs
```diff
@@ -13,7 +13,7 @@ use futures::{
 use futures01::Sink;
 use listenfd::ListenFd;
 use serde::{de, Deserialize, Deserializer, Serialize};
-use std::{fmt, io, net::SocketAddr, task::Poll, time::Duration};
+use std::{fmt, io, mem::drop, net::SocketAddr, task::Poll, time::Duration};
```
fanatid, line 16: How `std::mem::drop` is different from available `drop`?

src/internal_events/open.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub fn open<E: Fn(usize)>(self, emitter: E) -> OpenToken<E> {
+        gauge_add(&self.gauge, 1, &emitter);
+        OpenToken {
+            gauge: self.gauge,
+            emitter,
+        }
+    }
+}
+
+impl Default for OpenGauge {
```
fanatid, line 44: Can be defined through `#[derive(Clone, Default)]`. It's strange that clippy did not catch `new` without arguments :thinking:

src/sources/util/tcp.rs
ktff: Good catch, It isn't different. I'll change it in another PR.

PR #4684, opened 2020-10-21
chore: Split transforms into streaming and non-streaming.

Hoverbear:
Implement and close https://github.com/timberio/vector/issues/4564.

Differences to original RFC:

* @lukesteensen and I agreed that `transform_into()` was the primary for function-style transforms and have renamed it.
* `transform` has been renamed `transform_one` and is deprecated in non-test code (it can't handle 1:n outputs)
* @lukesteensen and I agreed  that `Stateless` and `Stateful` should be mooshed into `Function` since `Stateless` didn't make sense taking a `&mut self`. 

Things to pay attention to:

* All transforms are now `1:n`.
* Stream transforms no longer have oneshot/buffer-filling implementations. You need to create a `futures01::stream::iter_ok()` from the inputs then read off the stream produced in `transform`.
* Vector Unit tests now have a possible failure mode around using a TaskTransform in two different input sets, but it's unlikely to happen.
* Vector Unit tests have experienced some rework.
* Topology, for now, uses streaming mapping to do functiontransforms. (https://github.com/timberio/vector/pull/4427 has plans for this)


lukesteensen (approved):
Nice! I'm really happy with where this ended up. Thanks for sticking with it :+1: 

Pending green CI and merges conflict resolution this looks good to me!

src/sources/kubernetes_logs/mod.rs
```diff
@@ -271,30 +272,31 @@ impl Source {
...
+            .flat_map(move |event| {
+                let mut buf = Vec::with_capacity(1);
+                parser.transform(&mut buf, event);
+                futures::stream::iter(buf)
             });
 
-        let event_processing_loop = events.map(Ok).forward(out);
+        let event_processing_loop = partial_events_merger.transform(
+            Box::new(events.map(Ok).compat())
+        ).map_err(|_| unimplemented!("TODO")).compat().forward(out);
```
Hoverbear, line 299: Hey @MOZGIII , I'm wondering if you can help me with this one?

Now that the `partial_events_merger` is a `TaskTransform` it needs to be attached to the stream, but the `Sink::Error` seems to be restricted to `std::error::Error` and returning `()`? I fiddled around for a long time but I'm hoping you might have some context that could help me?

Hoverbear:
Remember how we were like "Oh, I don't think we need to worry about tasktransforms in vector self tests taking multiple inputs"? I think this is hitting that.

src/config/unit_test.rs
```diff
@@ -109,11 +109,34 @@ fn walk(
...
+        match target.transform {
+            Transform::Function(ref mut t) => {
+                for input in inputs.clone() {
+                    t.transform(&mut results, input)
+                }
+                targets = target.next.clone();
+                transforms.insert(key, target);
+            }
+            Transform::Task(t) => {
+                error!("Using a recently refactored `TaskTransform` in a unit test. You may experience limited support for multiple inputs.");
```
Hoverbear, line 123: It looks like our `reduce` behavior tests bump into the problem of consumption of the transform here. @lukesteensen would you advise disabling these behavior tests or should this ticket also refactor those?

lukesteensen: Do you have an idea for how to handle it? It'd be nice to include if it's not too big an addition, but otherwise I'm fine to open an issue and reenable in a follow up PR.

src/sources/kubernetes_logs/mod.rs
MOZGIII: Sure, I'll take a look!

src/config/unit_test.rs
Hoverbear: Let me give it a shot today.

src/sources/kubernetes_logs/mod.rs
MOZGIII:
```suggestion
        ).map_err(|_| unreachable!("TODO: remove with futures 0.3 transition")).compat().forward(out);
```

src/config/unit_test.rs
Hoverbear: Resolved!

PR #4686, opened 2020-10-21
chore(architecture): Move rusoto and http utils

jszwedko:
I need the `rusoto` module used by AWS sinks (previously in
`sinks::util:rusoto`) in the new `aws_s3` source. This pulls it up into
`src/` alongside other shared utility modules.

As part of moving it up, I also had to move up the `HttpClient` type in
the `sinks::util::http` module, which I pulled into an `http` module.

I followed the existing naming conventions for modules at the top-level
(like `kafka`, `dns`), but I admit that having an `http` module is a bit
confusing in that it conflicts with the `http` crate. I'm open to
suggestions here.


fanatid:
We also have `region` module which used only when `rusoto_core` feature is enabled. Maybe move `region` and `rusoto` to `aws` module (as folder)?

jszwedko:
@fanatid Good idea, I moved the `region` module into a submodule of `rusoto`. I was going to rename the `rusoto` module to `aws`, but all of the types and logic in it are pretty specific to `rusoto` so I feel like `rusoto` is a better name.

src/rusoto/mod.rs
```diff
@@ -30,11 +28,13 @@ use std::{
 };
 use tower::{Service, ServiceExt};
 
-pub type Client = HttpClient<util::http::HttpClient<RusotoBody>>;
+pub mod region;
```
fanatid, line 31: Maybe add `pub use region::RegionOrEndpoint`? :thinking:

jszwedko:
@fanatid another good idea :smile: I added `pub use` for `RegionOrEndpoint` and `region_from_endpoint`.

fanatid (approved):

bruceg (approved):
LGTM

Given what remains in `src/sinks/http.rs` would it be better named `http_sink.rs`? Also, would it make sense to move `struct Auth` into the top-level `src/http.rs`?

jszwedko:
@bruceg :+1: good call, I moved `Auth` up. 

> Given what remains in src/sinks/http.rs would it be better named http_sink.rs?

I think this would break convention as none of the other sinks in `src/sinks/` have this suffix.

In-case you meant `src/sinks/util/http.rs`, I feel like it is likely that we'll want to add back in some sink specific HTTP functionality so would prefer to just keep it generic for now unless you feel strongly rather than update all of the imports (unless there is an easy way to do this that I'm not aware of?).

bruceg:
> In-case you meant `src/sinks/util/http.rs`, I feel like it is likely that we'll want to add back in some sink specific HTTP functionality so would prefer to just keep it generic for now unless you feel strongly rather than update all of the imports (unless there is an easy way to do this that I'm not aware of?).

Yes, I meant `src/sinks/util/http.rs` and no I don't feel strongly at all.

jszwedko:
Issues for test failures:

https://github.com/timberio/vector/issues/4713
https://github.com/timberio/vector/issues/4714

PR #4688, opened 2020-10-22
chore: Clean up redundancies in template

juchiast:
- Remove `src_bytes`: `Template` stores a clone of `src` (`src` and `src_bytes`), but barely uses `src_bytes`.
- Flip `From<&str>` and `From<String>`: avoid cloning twice in `From<String>` impl.
- Flip `render` and `render_string`: avoid cloning and UTF-8 parsing in `render_string`.
- Change `get_ref` to return `&str`, more fitting for its usage in `aws_cloudwatch_logs`.

### DIFF
--- a/src/sinks/aws_cloudwatch_logs/mod.rs
+++ b/src/sinks/aws_cloudwatch_logs/mod.rs
@@ -13,7 +13,6 @@ use crate::{
     },
     template::Template,
 };
-use bytes::Bytes;
 use chrono::{Duration, Utc};
 use futures::{
     future::{BoxFuture, FutureExt, TryFutureExt},
@@ -286,8 +285,8 @@ impl CloudwatchLogsSvc {
         key: &CloudwatchKey,
         client: CloudWatchLogsClient,
     ) -> Self {
-        let group_name = String::from_utf8_lossy(&key.group[..]).into_owned();
-        let stream_name = String::from_utf8_lossy(&key.stream[..]).into_owned();
+        let group_name = key.group.clone();
+        let stream_name = key.stream.clone();
 
         let create_missing_group = config.create_missing_group.unwrap_or(true);
         let create_missing_stream = config.create_missing_stream.unwrap_or(true);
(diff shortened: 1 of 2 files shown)


src/template.rs
```diff
@@ -46,35 +45,35 @@ impl TryFrom<&str> for Template {
...
-            Err(TemplateError::StrftimeError)
-        } else {
-            Ok(Template {
-                src: src.into(),
-                src_bytes: Vec::from(src.as_bytes()).into(),
-                has_ts: is_dynamic,
-                has_fields: RE.is_match(src),
-            })
-        }
+        Template::try_from(src.to_owned())
```
fanatid, line 48: This will be useless `to_owned()` if we will get error from `TryFrom<String>`. I think it's a good case for [std::borrow::Cow](https://doc.rust-lang.org/std/borrow/enum.Cow.html): both `TryFrom<String>` and `TryFrom<&str>` can use `TryFrom<Cow<'_, str>>` what should help to avoid extra clones.

juchiast: Updated, thanks!

fanatid (approved):
Code looks good for me.

src/sinks/aws_cloudwatch_logs/mod.rs
```diff
@@ -502,7 +502,7 @@ async fn healthcheck(
         return Ok(());
     }
 
-    let group_name = String::from_utf8_lossy(&config.group_name.get_ref()[..]).into_owned();
+    let group_name = config.group_name.get_ref().to_owned();
```
bruceg, line 505: There may be an opportunity to do this in other places in this sink.

bruceg (approved):
LGTM

lukesteensen (approved):

src/sinks/aws_cloudwatch_logs/mod.rs
juchiast: I changed `group` and `stream` type to String.

PR #4695, opened 2020-10-22
feat(remap transform): add "remap-lang" crate

JeanMertz:
This PR updates the remap language with some necessary features (and some changes that made implementing those features easier).

These are:

* Language now lives in its own crate `remap-lang`

   This significantly improved compile times during development, and is another step towards #3784.

* Structured errors

   Helps to better understand what happens when a program fails to compile, or errors at runtime.

* Expression support (everything returns a `Value`)

   Needed to be able to use the language in contexts such as #4572.

* Trait-based function implementations

    Removes the need for complicated macro's in previous set-up, and paves the way to define remap functions outside this crate, to keep its dependencies to a minimum. The next PR will use this feature, as all existing remap functions will remain in the main Vector crate, for the time being.

* `enum_dispatch` over trait objects for improved performance

   Benchmarks will follow, but this should improve runtime performance.

* Support for non-`Value` arguments

   Similar to #4290, but without the extra layer of `QueryValue -> Value` indirection introduced in that PR. Which will make function implementations easier to write.


lib/remap-lang/src/function/split.rs
```diff
@@ -0,0 +1,90 @@
+use crate::{
+    expression::Error, Argument, ArgumentList, Expr, Expression, Function, Object, Parameter,
+    Result, State, Value,
+};
+use std::convert::{TryFrom, TryInto};
+
+#[derive(Debug)]
+pub(crate) struct Split;
```
JeanMertz, line 8: This function is more of a proof-of-concept.

For now, the idea is to leave all functions in the main Vector crate (they pull in quite a bit of external dependencies, and also use types defined in Vector itself).

In a follow-up PR this function will be moved back into Vector.

lib/remap-lang/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn insert(&mut self, path: &str, value: Value) -> std::result::Result<(), String>;
+
+    /// Find a value for a given path.
+    ///
+    /// See [`Object::insert`] for more details.
+    fn find(&self, path: &str) -> std::result::Result<Option<Value>, String>;
+}
+
+#[cfg(test)]
+mod tests {
```
JeanMertz, line 70: These are a few tests I used during development. In the next PR, this crate will be hooked up into Vector, and the existing tests there will use this crate, to prove nothing broke.

Once that's done, I'll move (most of) the unit tests over to this crate, and keep the behaviour tests in Vector.

lib/remap-lang/src/parser.rs
```diff
@@ -0,0 +1,331 @@
+#![allow(clippy::or_fun_call)]
```
JeanMertz, line 1: Note that a lot in this file is copy/pasted from the existing implementation, with the necessary changes to return expressions where relevant and some clean-up to make the parser simpler.

JeanMertz (commented):
Unfortunately, this PR isn't easy to review, as it contains both existing code copied from somewhere else and new code. It's also contained mostly in a single commit.

I left some comments to guide any reviewer.

This is the first of two or three PRs to work towards solving #4572.

The next PR will hook up this new crate within Vector itself, this will also prove that the language still works with existing tests currently living in Vector.

The PR afterwards will add the remap feature to the `reduce` transform and close #4572.

lib/remap-lang/src/error.rs
```diff
@@ -0,0 +1,96 @@
...
+    Unknown,
+}
+
+impl std::error::Error for Rule {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
+}
+
+impl fmt::Display for Rule {
```
Hoverbear, line 37: Nice job!

lib/remap-lang/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// * path indices:
+    ///
+    ///   ```txt
+    ///   .foo[2]
+    ///   ```
+    ///
+    /// When inserting into a coalesced path, the implementor is encouraged to
+    /// insert into the right-most segment if none exists, but can return an
+    /// error if needed.
+    fn insert(&mut self, path: &str, value: Value) -> std::result::Result<(), String>;
```
Hoverbear, line 61: So should this not be the result of a `target` from the grammar? We currently have this problem extensively, folks will use something like `foo.bar` where they expect it to actually be the key `"foo.bar"` not `"foo"."bar"`.

lib/remap-lang/grammar.pest
```diff
@@ -0,0 +1,80 @@
+// Root ------------------------------------------------------------------------
+
+program    = _{ SOI ~ NEWLINE* ~ (expression ~ (NEWLINE+ ~ expression)*)* ~ NEWLINE* ~ EOI }
```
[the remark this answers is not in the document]
bruceg: I haven't looked at the code yet. Does this imply that every expression clones its result `Value` to produce a return value, or is there a smarter reference happening to avoid that?

lib/remap-lang/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    /// Find a value for a given path.
+    ///
+    /// See [`Object::insert`] for more details.
+    fn find(&self, path: &str) -> std::result::Result<Option<Value>, String>;
```
Hoverbear, line 66: Similar feeling here, it seems this should use a `target` from the grammar (so a `Lookup`) type.

Hoverbear (approved):
I am surprised you can't remove some code from vector? But otherwise it looks good as a first pass.

bruceg:
> I am surprised you can't remove some code from vector? But otherwise it looks good as a first pass.

Unless I've missed it, this PR appears to have no actual connection to the rest of vector yet, which is one of my concerns. It's not clear to me how this external `struct Value` (etc) will connect with the same struct within vector.

lib/remap-lang/src/expression.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    #[error("variable error")]
+    Variable(#[from] variable::Error),
+}
+
+pub trait Expression: std::fmt::Debug {
+    fn execute(&self, state: &mut State, object: &mut dyn Object) -> Result<Option<Value>>;
+}
+
+macro_rules! enum_dispatch {
```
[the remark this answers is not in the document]
bruceg: Consider renaming this macro for clarity, since it's not really a generic enum dispatch. It is not exported, so, whatever.

lib/remap-lang/src/expression/arithmetic.rs
```diff
@@ -0,0 +1,48 @@
+use super::{Expr, Expression, Object, Result, State, Value};
+use crate::Operator;
+
+#[derive(Debug)]
+pub(crate) struct Arithmetic {
```
bruceg, line 5: Consider prefixing this to distinguish it from unary arithmetic (`-` negate, `~` binary invert, and `!` logical invert).

JeanMertz:
> Unless I've missed it, this PR appears to have no actual connection to the rest of vector yet, which is one of my concerns. It's not clear to me how this external `struct Value` (etc) will connect with the same struct within vector.

That is correct.

There will be several connection points:

1. Vector will implement `impl From<remap::Value> for Value` (and, to make it easy, we'll start with either making both the same, or similar enough to make conversions between the two seamless)
2. Vector will implement `impl remap::Object for Event`, to allow this crate to inject and fetch values from events.
3. For the `remap` transform (and afterwards, for the `reduce` transform), Vector will call `remap::Program::from_str` for the remap script defined in the `vector.toml` config, and will call `remap::Runtime::execute(&mut event, &program)` for every event that needs to be processed by the program.

These steps will happen in a follow-up PR that I'm fairly close to putting up (tomorrow).

> I am surprised you can't remove some code from vector?

I left that out of this PR, so that reviewing would be a bit easier. I tried doing that at first, but given the changes, there was nothing that Git would mark as "moved", making it as difficult to see what's new and what isn't, and adding the extra noise of seeing code being removed from Vector.

The follow-up PR will basically rip the existing implementation out of Vector, but keep the existing tests, then hook up the `remap-lang` crate, and have the same tests pass. Then, once we're happy, I'll move most unit tests over to the crate, and leave the behaviour tests in Vector itself.

lib/remap-lang/src/expression/arithmetic.rs
bruceg: The alternative would be to make this an enum using a strategy similar to enum dispatch above:
```rust
enum Arithmetic {
    Multiply(Box<Expr>, Box<Expr>),
    Negate(Box<Expr>),
}
```
That's probably more work now.

lib/remap-lang/src/expression/assignment.rs
```diff
@@ -0,0 +1,50 @@
...
+        let value = self.value.execute(state, object)?;
+
+        match value {
+            None => Ok(None),
+            Some(value) => {
+                match &self.target {
+                    Target::Variable(ident) => {
+                        state
+                            .variables_mut()
+                            .insert(ident.to_owned(), value.clone());
```
bruceg, line 39: I recognize the value of having expressions usable as a value, but these clones concern me for run-time efficiency, considering the value may be a deep structure (think `Value::Map` which may contain further maps, etc). I'm not sure off-hand how to handle it better though, and I recognize such a change would affect many return types. We'd like to return `Result<Option<&Value>>` but lifetimes almost certainly prevent that. Possibly something with `Rc<Value>`?

lib/remap-lang/src/expression/function.rs
```diff
@@ -0,0 +1,96 @@
...
+impl Function {
+    pub(crate) fn new(ident: String, arguments: Vec<(Option<String>, Argument)>) -> Result<Self> {
+        // TODO: move to `Runtime` so that we can eventually expose
+        // `register_function`.
+        let functions = vec![Split];
+
+        let function = functions
+            .iter()
+            .find(|b| b.identifier() == ident)
+            .ok_or_else(|| E::Function(ident.clone(), Error::Undefined))?;
```
bruceg, line 34: Is it safe to presume this will become a map at some point? It's only evaluated at load time, so not a big deal.

lib/remap-lang/src/expression/if_statement.rs
```diff
@@ -0,0 +1,43 @@
...
+            false_expression,
+        }
+    }
+}
+
+impl Expression for IfStatement {
+    fn execute(&self, state: &mut State, object: &mut dyn Object) -> Result<Option<Value>> {
+        match self.conditional.execute(state, object)? {
+            Some(Value::Boolean(true)) => self.true_expression.execute(state, object),
+            Some(Value::Boolean(false)) | None => self.false_expression.execute(state, object),
```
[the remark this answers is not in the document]
bruceg: I'm a little nervous about conflating `None` (undefined) with `false`, but we'll have to see how this plays out. It would probably be worth introducing a `defined` or `undef` function or equivalent later to test specifically for that case.

lib/remap-lang/src/expression/assignment.rs
FungusHumungus: I imagine most of the sizable data will ultimately be `bytes::Bytes`, which is very cheap to clone. It might not be as expensive as we think?

StephenWakely (commented):
I like the way this is going, and I love the compile times! Have run out of time for today, will have a closer look in the morning..

lib/remap-lang/src/expression/assignment.rs
bruceg: True. The most *frequently* cloned-after-assignment data would be scalars, all of which are quick to clone (thanks to `Bytes` for strings). However, we also want to move parsing functionality like `parse_json` into remap, and that will not be an inexpensive clone.

lib/remap-lang/src/lib.rs
JeanMertz: Yeah, I agree! I suspect this interface will change once we integrate `Lookup`. We don't have to worry about backward breaking changes in this crate, as for the time being we'll be the only consumer (and this won't be published either), so I'm fine using strings for now (similar to what was already there), and then move to `Lookup` in a follow-up PR once that's done.

WDYT?

lib/remap-lang/grammar.pest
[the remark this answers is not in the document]
JeanMertz: Yes, I believe `SOI` and `EOI` are required, at least from all the documentation and examples I've seen.

lib/remap-lang/src/expression/if_statement.rs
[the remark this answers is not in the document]
JeanMertz: Yeah, we definitely need to play with this to see what works best.

lib/remap-lang/grammar.pest
```diff
@@ -0,0 +1,80 @@
...
+// Statements ------------------------------------------------------------------
+
+assignment   = { target ~ "=" ~ expression }
+if_statement = { "if" ~ (boolean_expr | group) ~ block ~ ("else" ~ block)? }
+
+// Primary ---------------------------------------------------------------------
+
+primary  =  { value | variable | path | group }
+value    =  { string | float | integer | boolean | null }
+variable = ${ "$" ~ ident }
```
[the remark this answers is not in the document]
JeanMertz: It was discussed, and resulted in "do what you think is best" in #4419. But it wasn't an RFC, or a broader discussion with others in the team.

Given that, and given that we don't need variables right now to move this forward, I removed the rule from the grammar (but kept the `State` object the same, so the plumbing is still there, it just isn't used yet).

Let's continue the discussion in #4419, and I'll create a new PR at some point to re-introduce variables, if others are on board.

lib/remap-lang/grammar.pest
```diff
@@ -0,0 +1,80 @@
+// Root ------------------------------------------------------------------------
+
+program    = _{ SOI ~ NEWLINE* ~ (expression ~ (NEWLINE+ ~ expression)*)* ~ NEWLINE* ~ EOI }
+expression = _{ statement | boolean_expr | block }
+statement  = _{ assignment | if_statement }
+
+// Statements ------------------------------------------------------------------
+
+assignment   = { target ~ "=" ~ expression }
+if_statement = { "if" ~ (boolean_expr | group) ~ block ~ ("else" ~ block)? }
```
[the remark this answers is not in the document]
JeanMertz: `group` is part of `primary`, so it is an expression as well, but I opted against allowing `expression`, to make this as narrow as possible right now.

The goal of this PR (and the language in general) is not to go for maximal possibilities, but for "what's reasonable", where "reasonable" is somewhat defined in #3740 (although I agree we should be more clear on the exact goals to make decisions like these easier to reason about).

Previously only `boolean_expr` was allowed, this PR expands that with `group`. Having said that, I'll back out of `group` for now as well, to keep discussion focused on the features that we need right now, versus the features that became possible (and potentially useful), but aren't necessarily needed just yet.

lib/remap-lang/grammar.pest
```diff
@@ -0,0 +1,80 @@
...
+boolean_expr   = { equality ~ (operator_boolean_expr ~ equality)* }
+equality       = { comparison ~ (operator_equality ~ comparison)* }
+comparison     = { addition ~ (operator_comparison ~ addition)* }
+addition       = { multiplication ~ (operator_addition ~ multiplication)* }
+multiplication = { not ~ (operator_multiplication ~ not)* }
+not            = { operator_not* ~ (call | primary) }
+
+// Operators -------------------------------------------------------------------
+
+operator_boolean_expr   = { "||" | "&&" }
```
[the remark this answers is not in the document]
JeanMertz: Right. That should already be the case, as the precedence is determined by the parser (see `operation_fns`). I'm not sure what the right test to prove this would be, though, since both `(true || false) && true` and `true || (false && true)` would evaluate to the same result.

I'll update the order of this rule to make it more explicit in the grammar.

However, I just noticed there's a bug in the implementation, in that both sides of the operator are evaluated _before_ the operation is applied, which means the code doesn't short-circuit (as one would expect, I assume).

This is already the case in the current implementation, so I'll leave it as is in this PR, but I'm going to create an issue to see what the concensus is on this.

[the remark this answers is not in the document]
JeanMertz: Filed as https://github.com/timberio/vector/issues/4706.

lib/remap-lang/grammar.pest
```diff
@@ -0,0 +1,80 @@
...
+operator_addition       = { "-" | "+" }
+operator_multiplication = { "/" | "*" | "%" }
+operator_not            = { "!" }
+
+// Paths -----------------------------------------------------------------------
+
+path             = ${ ("." ~ path_segment)+ }
+path_segment     = ${ (path_field | path_coalesce) ~ path_index* }
+path_field       = ${ ident | string }
+path_coalesce    = !{ "(" ~ path_field ~ ("|" ~ path_field)+ ~ ")" }
```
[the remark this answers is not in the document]
JeanMertz: It was decided not to do that just yet, to not overcomplicate the language. But we can at some point, if we think it adds more value than it adds cognitive complexity.

lib/remap-lang/grammar.pest
```diff
@@ -0,0 +1,80 @@
...
+path_index       =  { "[" ~ path_index_inner ~ "]" }
+path_index_inner =  { "0" | ASCII_NONZERO_DIGIT ~ ASCII_DIGIT* }
```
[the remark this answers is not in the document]
JeanMertz: I assume it is. This already exists in the existing implementation. I'm hesitant to change it in this PR, but we can always circle back to this in a follow-up PR if there's a need.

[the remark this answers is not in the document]
JeanMertz: This is similar to other answers: technically we could, but it was decided not to add that yet, to reduce the cognitive complexity of the language (although I'm sure for some, _not_ having is the bigger surprise).

We can definitely change this in a follow-up PR though (and, we already allow negative indices in e.g. the `slice` function (#3755)).

lib/remap-lang/src/expression/arithmetic.rs
JeanMertz: This is mostly moved over from the existing impl. We don't support negate and binary invert arithmetic, and logical/not has its own expression implementation.

I'm going to leave this as is for now, but we can iterate on this if/when we do add all unary arithmetic options.

lib/remap-lang/src/expression/assignment.rs
JeanMertz: Yeah, I considered returning `&Value`, but decided against that change for now, as — as you mention — that has quite some impact in the implementation, and potentially needs reference counting.

I'm going to run the existing benchmarks (and potentially expand them) once this is integrated back into Vector, and there's room for improvement here, but I think we're fine for now with this clone.

_(as an aside, I removed variable support for now, so cloning doesn't happen here, but it does still happen in other places, so the point still stands)_

lib/remap-lang/src/expression/function.rs
JeanMertz: Yeah, I agree. This will change in a follow-up PR when the Vec is pulled out from here into the `Program` struct.

lib/remap-lang/grammar.pest
```diff
@@ -0,0 +1,80 @@
...
+equality       = { comparison ~ (operator_equality ~ comparison)* }
+comparison     = { addition ~ (operator_comparison ~ addition)* }
+addition       = { multiplication ~ (operator_addition ~ multiplication)* }
+multiplication = { not ~ (operator_multiplication ~ not)* }
+not            = { operator_not* ~ (call | primary) }
+
+// Operators -------------------------------------------------------------------
+
+operator_boolean_expr   = { "||" | "&&" }
+operator_equality       = { "!==" | "==" }
```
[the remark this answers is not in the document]
JeanMertz: Good catch!

lib/remap-lang/grammar.pest
```diff
@@ -0,0 +1,80 @@
...
+integer = { "-"? ~ "0" | ASCII_NONZERO_DIGIT ~ ASCII_DIGIT* }
+float   = {
+    "-"?
+    ~ "0" | ASCII_NONZERO_DIGIT ~ ASCII_DIGIT*
+    ~ ("." ~ ASCII_DIGIT* ~ (^"e" ~ ("+" | "-")? ~ ASCII_DIGIT+)?)
+        | (^"e" ~ ("+" | "-")? ~ ASCII_DIGIT+)
+}
+
+regex       = ${ "/" ~ regex_inner ~ "/" ~ regex_flags }
+regex_flags =  { ("i" | "x" | "m")* }
```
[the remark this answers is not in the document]
JeanMertz: Good catch. Will tackle this in a follow-up PR when the other functions are hooked up (which will expose the fact that this is missing right now).

lib/remap-lang/src/value.rs
```diff
@@ -0,0 +1,356 @@
+use bytes::Bytes;
+use std::collections::BTreeMap;
+use std::convert::{TryFrom, TryInto};
+use std::string::String as StdString;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum Value {
+    String(Bytes),
```
FungusHumungus, line 8: I think renaming this to `String` is confusing as it is still bytes.

I can see it does make for nicer error messages as we now get `expected type string` rather than `expected types bytes`, but I think we can still do that without changing the name here?

lib/remap-lang/src/function/split.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl Expression for SplitFn {
+    fn execute(&self, state: &mut State, object: &mut dyn Object) -> Result<Option<Value>> {
+        let string: String = self
+            .value
+            .execute(state, object)?
+            .ok_or(Error::Missing)?
+            .try_into()?;
+
```
FungusHumungus, line 62: Is there any reason why these can't be `required` and `optional` macros as before?

lib/remap-lang/src/function.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    fn try_into(self) -> std::result::Result<Expr, Self::Error> {
+        match self {
+            Argument::Expression(expr) => Ok(expr),
+            Argument::Regex(_) => Err(Error::ArgumentExprRegex),
+        }
+    }
+}
+
+pub(crate) trait Function: std::fmt::Debug {
```
FungusHumungus, line 99: It is confusing having both a trait and a struct called `Function`. Could this be renamed?

lib/remap-lang/src/function/split.rs
JeanMertz: No reason, but as mentioned in the comments, this function will be moved back to vector in a follow-up PR, where the `required` and `optional` macro's currently live.

lib/remap-lang/src/function.rs
JeanMertz: I'd rather not. This trait is publicly accessible (it isn't in this PR, but it is made public in #4709), and I think the name make sense.

Similarly, the name `expression::Function` also makes sense.

They are in different modules, and outside this crate, you only deal with one of the two (the trait), so I don't think we need to change anything here.

lib/remap-lang/src/value.rs
JeanMertz: Maybe. I like to be a bit more explicit about what type of value this is (it's similarly named `string` in the parser, not `bytes`).

It has to store `Bytes`, both for performance reasons, and because strings from an event might not be valid UTF-8.

Let's see how we like `String` for now. We can always do a refactor to rename to `Bytes` in a follow-up PR.

PR #4701, opened 2020-10-22
enhancement(data model): Add `namespace` to `Metric` 

ktff:
First of the PRs for #3896

Adds `namespace` field to `struct Metric` and wires this change through the code base. 

Does not change behavior of componenets.

### DIFF
--- a/proto/event.proto
+++ b/proto/event.proto
@@ -58,6 +58,7 @@ message Metric {
     AggregatedHistogram aggregated_histogram = 9;
     AggregatedSummary aggregated_summary = 10;
   }
+  string namespace = 11;
 }
 
 message Counter {
--- a/src/conditions/is_log.rs
+++ b/src/conditions/is_log.rs
@@ -67,6 +67,7 @@ mod test {
         assert_eq!(
             cond.check(&Event::from(Metric {
                 name: "test metric".to_string(),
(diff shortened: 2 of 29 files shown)


proto/event.proto
```diff
@@ -58,6 +58,7 @@ message Metric {
     AggregatedHistogram aggregated_histogram = 9;
     AggregatedSummary aggregated_summary = 10;
   }
+  string namespace = 11;
```
bruceg, line 61: Do we need a documentation note that new Vector can't send metrics to older Vector via the "vector" protocols due to this new field?

bruceg (approved):
Looks good, just one question.

fanatid (approved):

jszwedko (approved):
This looks good for the first step of #3896.

I agree with Bruce that we should document this as a breaking change in that the `vector` sink will only be compatible with the `vector` source if they both are running a version with this change. Is there a way we could introduce this that would retain compatibility? I lack familiarity with protobuf.

proto/event.proto
lukesteensen: Adding a field with a never-before-used tag number is backward-compatible. The older implementation will just ignore tags it doesn't know about.

lukesteensen (approved):

Issue #4649, opened 2020-10-20
Add host metrics to the GraphQL API


leebenson:
Consume the `host_metrics` source in the API, outside of source configuration.

This should be a 'hot' source - i.e. only active when requested, and relatively light; metrics should be collected only if the pertinent fields are queried.

The use-case is memory/CPU gauges in `vector top`.

Issue #4524, opened 2020-10-12
Add 'events throughput' to subscriptions


leebenson:
'Events throughput' should provide a snapshot over a given `interval: Int!`, of the number of events processed during that captured window, updated via a GraphQL subscription. It should be relevant to all topology.

Issue #4523, opened 2020-10-12
Add 'errors' query to topology


leebenson:
In #4431, an aggregated 'events' column has been added. 'Errors' should be next, showing the total errors against a given topological resource over the life of the Vector instance.

leebenson:
This is done in #4886

PR #4702, opened 2020-10-22
feat(observability): `vector top` v1 dashboard

### LINKED ISSUE #4521: Implement table redrawing for `vector top`
#4431 draws a table with initial metrics, and exits. As subscriptions are added per #4520, the table needs to remain fixed, with metrics updated in-place.

We're using [prettytable-rs](https://github.com/phsym/prettytable-rs) to draw the table to the console. There are examples of table redraw (such as tic-tac-toe) in the repo, which essentially resets the term cursor and overwrites over the old. This strategy is likely required to accommodate scenarios where column widths adjust per growing metrics, rather than moving the cursor to a specific stat position and resetting.

However, I'm curious how performant this will be with a larger table against the need to reduce flicking, when stats could be received at low ms intervals (via multiple streams). This needs testing in macOS, Linux and Windows terminals.

It's also necessary for the table to retain its original layout, so that topology rows aren't jumping around as new metrics are coming in.

### LINKED ISSUE #4522: Add bookkeeping for `vector top` topology table layouts
One of the requirements surfaced #4521, is the need for a 'fixed' position of topology as metrics are adjusting column fields.

In #4520, I described a two-step approach to retrieving stats, and subsequently subscribing to their updates. This decouples receiving metrics from the initial layout.

To further decouple GraphQL data from the layout, I think a viable approach would be to loop over received data, and store it in a separate mutable map, which becomes the canonical source of data. As new subscription responses are received, that data would be mutated in place. This would be write-heavy, with a single read when the table is subsequently redrawn.

leebenson:
This PR adds initial work toward a `vector top` dashboard to show health, topology and metrics from a local/remote Vector instance in a terminal.

Closes #4521 
Closes #4522

## Screenshot

**macOS:**

Screenshot 2020-10-22 at 19 38 13

**Windows:** (Server 2019, in this case)

Screenshot_2020-10-26_at_18 50 47

## Demo

https://www.loom.com/share/3b3c810315d84d5b81ca3fa7b2c5a20f

## Scope

The scope of this initially version is as follows:

- Implement an initial full-screen dashboard that works in *nix terminals
- Connect to a local/remote running Vector instance
- Check that the server is alive, by executing a GraphQL `{ health }` query
- Poll topology every `--refresh-interval` milliseconds (default: 500ms)
- Display sources/transforms/sinks, along with aggregate event counts, in a table
- Show the connected URL, and instructions to quit

## Next

I have started the following efforts locally, but decided to break them into follow-up PRs to cut scope:

- ~Support for Windows, via [crossterm](https://github.com/crossterm-rs/crossterm)~ **ℹ️ Windows support was added in 9ddb6c671a31a2086241cda815ce259c21c1df9c**
- Host metrics
- Improved data resolutions (full subscriptions; no polling). 

## Approach

This PR builds on earlier work to expose a GraphQL API endpoint in a running Vector instance, and a GraphQL query and subscription client to connect to it.

The dasboard is implemented with [tui-rs](https://github.com/fdehau/tui-rs), which provides common dash components such as tables, charts and styled text. This initial makes use of several components, including:

- Layout/alignment: `layout::{Alignment, Constraint, Layout, Rect}`
- Stylistic: `style::{Color, Modifier, Style}`
- Text: `text::{Span, Spans}`
- Widgets: `widgets::{Block, Borders, Paragraph, Row, Table, Wrap}`

Follow-up work will make use of `Gauge` and `Chart` types to provide visual indicators for Vector health/metric velocity.

## Notes / Known issues

Some of these issues already exist; I'll do a sweep of the missing issues and edit this PR with links as required.

These are immediate improvements that are actively being worked on. None of them are blocking, but will form the more feature-complete `vector top` v1 intended to ship to users:

### Windows support

**ℹ️ Windows support was added in 9ddb6c671a31a2086241cda815ce259c21c1df9c**

~[tui-rs](https://github.com/fdehau/tui-rs) supports multiple terminal backends. This PR uses [termion](https://github.com/redox-os/termion), which is compatible with Redox, Mac OS X, BSD, and Linux (or, in general, ANSI terminals). It lacks Windows support.~

~I started with termion, because the early documentation I found on tui tended to use it.~

~I'll be swapping this out for [crossterm](https://github.com/crossterm-rs/crossterm) next to add Windows support.~

### Host metrics

We have support for host metrics in the API in #4649, which landed today. These are ready to be consumed as gauge/text components in the dasboard next.

### Locking

The nature of the observability data that `vector top` consumes is high-contention. Some data is polled on a regular interval; other data uses GraphQL subscriptions via WebSockets and is server-pushed. There's a relatively equal weight required for read/writes. This made `RwLock` a poor default choice.

I tried several locking strategies, including [arc-swap](https://github.com/vorner/arc-swap) with immutable structs to limit 'swaps' (writes).

In the end, I settled on a fairly mundane approach, using an `Arc<Mutex<BTreeMap<String, TopologyRow>>>` to model thread-safe topology state, since local dashboard frame redraws (reads) and `--refresh-interval` polls + subscriptions (writes) tend to occur on fairly regularly schedules.

To reduce contention, expensive query operations are typically performed first, and underlying state data is cloned for re-rendering, to avoid long-running locks. Read/write entry is typically slim.

I've also broken out state into multiple structs, each locked separately. `TopologyState` governs the current topology + aggregate metrics; host metrics and streaming metrics such as i/o throughput, are likely to structured separately.

### Fluid layouts

Tui is pretty good at resizing to fit the viewport. This isn't perfect. Text can be cut off on very small console screens, or if purposefully resized to be too small.

Most of this is regular fare for terminal-based dashboards, mitigated with reasonable size allowances for columns, etc.

We may be able to detect smaller screens and adapt accordingly, but I think ultimately this is akin to our web UI approach -- we're targeting reasonably sized screens and not, say, mobile terminals or esoteric configurations.

### Polling -> Subscriptions

Topology is currently observed by polling (default: 500ms). This both fetches new topology, and updates aggregate stats. This was done for a simple starting place (that works well enough), but isn't particularly efficient. The returned data is likely to have a high degree of repetition across queries.

In #4581, I added `topologyAdded` and `topologyRemoved` subscription fields, which stream server-pushed events in response to configuration changes. These provide a more efficient mechanism for notifying `vector top` of changes, and should be utilised.

Additionally, I would like to add direct subscription support for `eventsProcessedTotal` and other metrics, which sample data over a provided `interval: Int!` and send only data that has changed since the last iteration. Instead of full topology every (default) 500ms, this would result in streaming JSON payloads similar to:

```json
{
    "data": {
        "__typename": "EventsProcessedTotal",
        "name": "source_name",
        "eventsProcessedTotal": 999
    }
}
```

Which would be more efficient for remote clients, especially as these are multiplexed over a single WebSocket connection without the overhead of a HTTP handshake per query.

### Channels vs interval loops

**ℹ️ This was added in 9ddb6c671a31a2086241cda815ce259c21c1df9c**

~Tui works on 'frames', where `.draw()` is invoked in userland and rebuilds component renders per a defined schedule.~

~Currently, that schedule is a set 250ms 'tick' controlled by:~

_~[code omitted, since it's no longer relevant]~_

~This isn't very efficient. A frame redraws every 250ms, irrespective of whether state has changed, or a key press has been registered.~

~Instead, these ought to be channels that can be `select!`ed on, to:~

~1. Redraw only when underlying observability data has changed (i.e. is server-pushed, or polled). This would also add an implicit association with the `--refresh-interval` opt. Right now, there's no correlation.~

~2. Exit on a matched key press. Stdin is captured in real-time, so this doesn't require redundant looping.~

### Additional metrics

Good candidates to pad out the metrics columns are 'errors' in #4523 and 'events throughput' in #4524. Currently, those columns default to `--` to reflect the zero values.

### TTY detection

Running this would full terminal support can yield panics such as:

```
thread '<unnamed>' panicked at 'attempt to subtract with overflow'
```

This visible, for example, when debugging in a headless terminal such as those provided by an IDE.

We should have a safe fallback such as displaying the lack of dashboard support, or (ultimately) suggesting `vector stats` for a dashboard in 'snapshot' mode, where the complexity of managing redraws, cursor positioning and layout and unlikely to surface.

For now, simple TTY detection and a user-friendly error message is sufficient. 

### Connection management

If a query fails right now, the loop terminates and the dashboard would just hang.

There should ultimately be some reconnection logic and some indication in the UI that a connection has gone down.

Both of these are relatively complex areas that will require some extra thought and coordination, and I'd consider them to be out of scope for v1.

For now, a user can just quit with 'q' and restart `vector top` for unstable connections.

### More testing

~I've only tested this on macOS locally, on a 16" Macbook Pro screen.~ Also tested on Windows Server 2019.

---

Of the above, I'd consider the priority to be:

1. TTY detection / fallback.

2. Switching `termion` -> `crossterm`, to add Windows support. I anticipate this being a relatively superficial replacement, since tui has support for multiple backends.

~3. Some local Windows testing (haven't booted it up in a few months, but I probably should.)~ **Confirmed: Works on Windows**

4. Adding additional metrics, such as errors processed, to fill out the table.

5. Switching to subscriptions to enable support for metrics such as throughput (this has the separate task of adding schema support, but that should be relatively straightforward; we already have API support to sample data over an `interval`.)

The above should be enough to release this as a v1, with the rest deferred for ongoing improvements.


leebenson:
Putting this back into draft temporarily, while I add Windows support.

leebenson:
Windows suppott has been added, and redraw/keyboard input is now based on channels and should be snappier than the previous/static 250ms tick.

leebenson:
`tui` seems to be pulling in `termion` and causing https://github.com/timberio/vector/pull/4702/checks?check_run_id=1298631594

Investigating to see if that's an optional feature we can remove

leebenson:
@sghall highlighted an issue with the CPU spiking to 100% on macOS. I could reproduce locally with both debug and release builds.

Interestingly, a 4 vCPU 2.3Ghz Intel XEON VM running Windows Server 2019 maxed out at ~18%. Comparably, my 2.3Ghz, 8 core i9 with double the ram (32GB) hit ~105%.

The previous strategy was to re-draw the terminal whenever a) new topology 'state' data was made available, or b) on a polled 50ms interval for keyboard input-- whichever came first.

I've replaced that approach with a 250ms throttle wrapper on state changes, via:

```rust
// Throttle widgets changes to 250ms, to space out re-draws
let widget_listener =
    tokio::time::throttle(tokio::time::Duration::from_millis(250), widgets.listen());

tokio::pin!(widget_listener);

loop {
    tokio::select! {
        _ = widget_listener.next() => {
            terminal.draw(|f| widgets.draw(f))?;
        },
        k = key_press_rx.recv() => {
            if let KeyCode::Esc | KeyCode::Char('q') = k.unwrap() {
                let _ = key_press_kill_tx.send(());
                break
            }
        }
    }
}
```

This effectively limits redraws to 4x per second, irrespective of updates.

The underlying stream is provided by a `tokio::sync::watch::Receiver<()>`, so we don't suffer any 'lag' / memory pressure for updates entering a buffer that may not be rendered until later. Instead, live state is pulled at the time of redraw.

cc @sghall for local confirmation.

jamtur01 (approved):
LGTM - as much as that is worth.

sghall:
@leebenson that last update definitely cut down on the CPU usage.  Getting similar numbers to yours locally (1-2%).  Thanks for taking care of that 👍

leebenson:
Merged #4761 to inherit the topology -> components naming, and `vector top` to suit it. It's inflated the size of this PR, but it'll be back down once #4761 is in master.

leebenson:
when CI is green, I'll get this merged.

It only has one reviewer (thanks @jamtur01), but it's been tested fairly extensively on macOS, Ubuntu and Windows, reports from @sghall helped fix a CPU load issue (thanks!) and it lives outside of Vector proper.

It's been in the review queue since Friday and I'm starting to base other branches off of it, so I don't want this to drift too much. If you want to review this post-merge, feel free, and I'll tackle separately.

PR #4703, opened 2020-10-22
feat(remap transform): add `replace` and `flatten` remap functions

### LINKED ISSUE #3750: New `replace` remap function

### LINKED ISSUE #3752: New `flatten` remap function

FungusHumungus:
Closes #3750 
Closes #3752 

This adds `replace` and `flatten` remap functions.

Note with the `flatten` function, it flattens both Arrays and Maps, but it does not attempt to flatten a value that is a combination of Arrays and Maps. The rules for flattening are determined according to what the top level value is. So an Array of Maps is flattened as an Array, and an Map of Arrays is flatten as a Map.

Creating this as draft for now, as it may need some additional work to merge it into #4695

### DIFF
--- a/src/mapping/query/function/flatten.rs
+++ b/src/mapping/query/function/flatten.rs
@@ -1,0 +1,362 @@
+use super::prelude::*;
+use std::collections::btree_map;
+
+/// An iterator to walk over maps allowing us to flatten nested maps to a single level.
+struct MapFlatten<'a> {
+    values: btree_map::Iter<'a, String, Value>,
+    inner: Option<Box<MapFlatten<'a>>>,
+    parent: Option<String>,
+}
+
+impl<'a> MapFlatten<'a> {
+    fn new(values: btree_map::Iter<'a, String, Value>) -> Self {
+        Self {
+            values,
+            inner: None,
+            parent: None,
+        }
+    }
+
+    fn new_from_parent(parent: String, values: btree_map::Iter<'a, String, Value>) -> Self {
+        Self {
+            values,
+            inner: None,
+            parent: Some(parent),
+        }
+    }
+
+    /// Returns the key with the parent prepended.
+    fn new_key(&self, key: &str) -> String {
+        match self.parent {
+            None => key.to_string(),
+            Some(ref parent) => format!("{}.{}", parent, key),
+        }
+    }
+}
+
+impl<'a> std::iter::Iterator for MapFlatten<'a> {
+    type Item = (String, &'a Value);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(ref mut inner) = self.inner {
+            let next = inner.next();
+            if next.is_some() {
+                return next;
+            } else {
+                self.inner = None;
+            }
+        }
+
+        let next = self.values.next();
+        match next {
+            Some((key, Value::Map(value))) => {
+                self.inner = Some(Box::new(MapFlatten::new_from_parent(
+                    self.new_key(key),
+                    value.iter(),
+                )));
+                self.next()
+            }
+            Some((key, value)) => Some((self.new_key(key), value)),
+            None => None,
+        }
+    }
+}
+
+/// Create an iterator that can walk a tree of Array values.
+/// This can be used to flatten the array.
+struct ValueFlatten<'a> {
+    values: std::slice::Iter<'a, Value>,
+    inner: Option<Box<ValueFlatten<'a>>>,
+}
+
+impl<'a> ValueFlatten<'a> {
+    fn new(values: std::slice::Iter<'a, Value>) -> Self {
+        ValueFlatten {
+            values,
+            inner: None,
+        }
+    }
+}
+
+impl<'a> std::iter::Iterator for ValueFlatten<'a> {
+    type Item = &'a Value;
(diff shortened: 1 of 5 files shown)


src/mapping/query/function/flatten.rs
```diff
@@ -0,0 +1,362 @@
...
+    type Item = (String, &'a Value);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(ref mut inner) = self.inner {
+            let next = inner.next();
+            if next.is_some() {
+                return next;
+            } else {
+                self.inner = None;
+            }
```
JeanMertz, line 47:
```suggestion
            match next {
            	Some(_) => return next,
            	None => self.inner = None,
            }
```

src/mapping/query/function/flatten.rs
```diff
@@ -0,0 +1,362 @@
...
+            Some((key, value)) => Some((self.new_key(key), value)),
+            None => None,
+        }
+    }
+}
+
+/// Create an iterator that can walk a tree of Array values.
+/// This can be used to flatten the array.
+struct ValueFlatten<'a> {
```
JeanMertz, line 67: Shouldn't this be named `ArrayFlatten`, similar to the above `MapFlatten`?

src/mapping/query/function/flatten.rs
```diff
@@ -0,0 +1,362 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        // Iterate over our inner list first.
+        if let Some(ref mut inner) = self.inner {
+            let next = inner.next();
+            if next.is_some() {
+                return next;
+            } else {
+                // The inner list has been exhausted.
+                self.inner = None;
+            }
```
JeanMertz, line 93:
```suggestion
			match next {
				Some(_) => return next,
                // The inner list has been exhausted.
				None => self.inner = None,
            }
```

src/mapping/query/function/flatten.rs
```diff
@@ -0,0 +1,362 @@
...
+
+        // Then iterate over our values.
+        let next = self.values.next();
+        if let Some(Value::Array(next)) = next {
+            // Create a new iterator for this child list.
+            self.inner = Some(Box::new(ValueFlatten::new(next.iter())));
+            self.next()
+        } else {
+            next
+        }
```
JeanMertz, line 104:
```suggestion
		match next {
			Some(Value::Array(next)) => {
        	    // Create a new iterator for this child list.
    	        self.inner = Some(Box::new(ValueFlatten::new(next.iter())));
	            self.next()			
			}
			None => next,
        }
```

src/mapping/query/function/flatten.rs
```diff
@@ -0,0 +1,362 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::mapping::query::path::Path;
+    use serde_json::json;
+
+    #[test]
+    fn check_flatten() {
+        let cases = vec![
```
JeanMertz, line 163: I believe these cases will become quite a bit more readable by applying new testing patterns, such as can be seen in this PR:

https://github.com/timberio/vector/pull/4770

Basically;

* use `map![]` to create maps
* use `into()` vs the more verbose `Value::from`
* don't test against an actual event, but use a map (`std::collections::BTreeMap` implements `remap::Object`)
* paths can be created from strings, so change `Path::from(vec![vec!["foo"]])` to `Path::from("foo")`
* test against an actual path only once, use `Literal::from(map!["foo": "bar"])` for other tests (and leave the event an empty `map![]`), to reduce the amount of unneeded boilerplate.

I realize this requires the PR to be rebased on the new remap-lang work first, so I propose we keep things as they are in this PR, and then refactor once the remap-lang PRs are merged, at which point we'll have to change this implementation anyway.

JeanMertz (approved):
Nice!

Left a couple of notes, and we'll have to rewrite some parts once the remap-lang work is merged, but the implementations themselves look good 👌

src/mapping/query/function/flatten.rs
FungusHumungus: :+1:  Gosh, how did I miss that!?

PR #4707, opened 2020-10-23
enhancement(new_relic_logs sink): Fix API max payload size

matildasmeds:
New Relic Logs API max payload size is 10^6 bytes.
While I'm not 100% sure this sets the actual JSON payload
size to where it should be, it does seem like an improvement
compared to the old value of 5MiB.

I have opted to allow the user to set a lower batch size, 
should they so prefer.

Comments & Feedback welcome!

### DIFF
--- a/src/sinks/new_relic_logs.rs
+++ b/src/sinks/new_relic_logs.rs
@@ -13,12 +13,20 @@ use indexmap::IndexMap;
 use serde::{Deserialize, Serialize};
 use snafu::Snafu;
 
+// New Relic Logs API accepts payloads up to 1MB (10^6 bytes)
+const MAX_PAYLOAD_SIZE: usize = 1000000 as usize;
+
 #[derive(Debug, Snafu)]
 enum BuildError {
     #[snafu(display(
         "Missing authentication key, must provide either 'license_key' or 'insert_key'"
     ))]
     MissingAuthParam,
+    #[snafu(display(
+        "Too high batch max size. The value must be {} bytes or less",
+        max_size
+    ))]
+    BatchMaxSize { max_size: usize },
 }
 
 #[derive(Deserialize, Serialize, Debug, Eq, PartialEq, Clone, Derivative)]
@@ -114,10 +122,14 @@ impl NewRelicLogsConfig {
         };
 
         let batch = self.batch.use_size_as_bytes()?;
+        let max_payload_size = batch.max_bytes.unwrap_or(MAX_PAYLOAD_SIZE);
+        if max_payload_size > MAX_PAYLOAD_SIZE {
+            return Err(Box::new(BuildError::BatchMaxSize {
+                max_size: MAX_PAYLOAD_SIZE,
+            }));
+        }
         let batch = BatchConfig {
-            // The max request size is 10MiB, so in order to be comfortably
-            // within this we batch up to 5MiB.
-            max_bytes: Some(batch.max_bytes.unwrap_or(bytesize::mib(5u64) as usize)),
+            max_bytes: Some(batch.max_bytes.unwrap_or(MAX_PAYLOAD_SIZE)),
             max_events: None,
             ..batch
         };
@@ -191,10 +203,7 @@ mod tests {
         );
         assert_eq!(http_config.method, Some(HttpMethod::Post));
         assert_eq!(http_config.encoding.codec(), &Encoding::Json.into());
-        assert_eq!(
-            http_config.batch.max_bytes,
-            Some(bytesize::mib(5u64) as usize)
-        );
+        assert_eq!(http_config.batch.max_bytes, Some(MAX_PAYLOAD_SIZE));
         assert_eq!(
             http_config.request.in_flight_limit,
             InFlightLimit::Fixed(100)
@@ -213,7 +222,7 @@ mod tests {
         let mut nr_config = NewRelicLogsConfig::default();
         nr_config.insert_key = Some("foo".to_owned());
         nr_config.region = Some(NewRelicLogsRegion::Eu);
-        nr_config.batch.max_size = Some(bytesize::mib(8u64) as usize);
+        nr_config.batch.max_size = Some(MAX_PAYLOAD_SIZE);
         nr_config.request.in_flight_limit = InFlightLimit::Fixed(12);
         nr_config.request.rate_limit_num = Some(24);
 
@@ -225,10 +234,7 @@ mod tests {
         );
         assert_eq!(http_config.method, Some(HttpMethod::Post));
         assert_eq!(http_config.encoding.codec(), &Encoding::Json.into());
-        assert_eq!(
-            http_config.batch.max_bytes,
-            Some(bytesize::mib(8u64) as usize)
-        );
+        assert_eq!(http_config.batch.max_bytes, Some(MAX_PAYLOAD_SIZE));
         assert_eq!(
             http_config.request.in_flight_limit,
             InFlightLimit::Fixed(12)
@@ -249,7 +255,7 @@ mod tests {
         region = "eu"
 
         [batch]
-        max_size = 8388608
+        max_size = 838860
 
         [request]
         in_flight_limit = 12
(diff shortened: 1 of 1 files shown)


src/sinks/new_relic_logs.rs
```diff
@@ -115,9 +119,8 @@ impl NewRelicLogsConfig {
 
         let batch = self.batch.use_size_as_bytes()?;
         let batch = BatchConfig {
-            // The max request size is 10MiB, so in order to be comfortably
-            // within this we batch up to 5MiB.
-            max_bytes: Some(batch.max_bytes.unwrap_or(bytesize::mib(5u64) as usize)),
+            // New Relic Logs API accepts payloads only up to 1MB (10⁶ bytes)
```
jszwedko, line 122: Did you mean to put `10MB` here?

src/sinks/new_relic_logs.rs
```diff
@@ -115,9 +119,8 @@ impl NewRelicLogsConfig {
...
+            max_bytes: Some(cmp::min(batch.max_bytes.unwrap_or(MAX_PAYLOAD_SIZE), MAX_PAYLOAD_SIZE)),
```
jszwedko, line 123: I think I'd prefer to raise an error if they pass a payload size that is too big rather than transparently cap it. I think that'll result in more easily discoverable behavior. This could be a new `BuildError`.

Hoverbear (approved):
This looks good. Thank you. 😊

jszwedko (commented):
This looks great, thanks @matildasmeds ! Just one note about raising an explicit error rather than silently capping.

I see there was some conversation around the 5 MB limit when this was written: https://github.com/timberio/vector/pull/1303/files#r353804978. I agree that we should be using the full payload size available.

I would like to test this just to verify that New Relic will actually accept a 10^6 byte body and isn't including things like HTTP headers in its calculation (though I would find this surprising).

Hoverbear:
@matildasmeds Could I ask you to [fix the DCO](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) on your commit by adding a signoff?

You should also run `make fmt` or `cargo fmt`? That'll fix the format.

You make also wish to check `make check-clippy` for any possible lints.

src/sinks/new_relic_logs.rs
Hoverbear: I see you did not get distracted by the fancy ⁶ character.

jszwedko: :smile: its presence actually inspired me to figure out how to type superscript via [compose key](http://www.kbdedit.com/manual/ex17_compose_key_chained_dead.html).

src/sinks/new_relic_logs.rs
matildasmeds: I wish I knew more about what drives the payload size. 
Other New Relic plugins implement payload splitting, in case it's too big. 
Do you think we should do the same here?

jszwedko: Do you have an example of a plugin that does that? That would increase my confidence regarding what it considers to be part of the payload in its calculation.

For payload splitting, the `batch` configuration that you are modifying here should handle that for us. It will only batch together events whose lengths sum to less than X bytes (or `n` events if `max_events` is set too).

src/sinks/new_relic_logs.rs
matildasmeds: I meant to put `1MB`, because this is the actual limit. I will also reformat the fancy character to something less fancy.

matildasmeds:
> I would like to test this just to verify that New Relic will actually accept a 10^6 byte body and isn't including things like HTTP headers in its calculation (though I would find this surprising).

The current documentation says ["Payload total size: 1MB(10^6 bytes) maximum per POST.  We highly recommend using compression."](https://docs.newrelic.com/docs/telemetry-data-platform/ingest-manage-data/ingest-apis/use-event-api-report-custom-events#instrument) I think it means that headers are not included, but I will double-check with the Ingest team. Also, hopefully the docs can be improved.

src/sinks/new_relic_logs.rs
jszwedko: Oops, my apologies, I was thinking it was supposed to be 10MiB. I do see the 1MiB documented limit. I'm confused what limit of 10 MiB that old comment was referring to.

src/sinks/new_relic_logs.rs
jszwedko: Noting that the batching seems to be before compression. 1 MiB of events is likely to compress down to a much smaller payload. I opened https://github.com/timberio/vector/issues/4716 to think about how to improve this behavior.

matildasmeds: As for payload splitting, these plugins come into mind.
There's [the fluent-bit plugin, written in go](https://github.com/newrelic/newrelic-fluent-bit-output/blob/master/record/record.go#L89-L118). And [fluentd plugin in Ruby has a similar implementation](https://github.com/newrelic/newrelic-fluentd-output/blob/master/lib/fluent/plugin/out_newrelic.rb#L127-L176).

One thing to keep in mind, is that this code is best effort, and may not be perfect. But I hope it helps.

src/sinks/new_relic_logs.rs
```diff
@@ -249,7 +255,7 @@ mod tests {
         region = "eu"
 
         [batch]
-        max_size = 8388608
+        max_size = 838860
```
matildasmeds, line 258: I'm changing the max_size to be under max limit here, so that the test should pass, and adding another test which should panic below.

src/sinks/new_relic_logs.rs
```diff
@@ -13,12 +13,20 @@ use indexmap::IndexMap;
 use serde::{Deserialize, Serialize};
 use snafu::Snafu;
 
+// New Relic Logs API accepts payloads up to 1MB (10^6 bytes)
+const MAX_PAYLOAD_SIZE: usize = 1000000 as usize;
```
Hoverbear, line 17:
```suggestion
const MAX_PAYLOAD_SIZE: usize = 1_000_000 as usize;
```

(Non blocking, nit)

Hoverbear (approved):

src/sinks/new_relic_logs.rs
jszwedko: Gotcha, thanks! Looking at the Go implementation, the cap seems to be with respect to the payload size rather than including any other parts of the request. :+1:

src/sinks/new_relic_logs.rs
jszwedko: Also nit: I might throw a link in to https://docs.newrelic.com/docs/telemetry-data-platform/ingest-manage-data/ingest-apis/use-event-api-report-custom-events#limits.

src/sinks/new_relic_logs.rs
```diff
@@ -13,12 +13,20 @@ use indexmap::IndexMap;
...
+
 #[derive(Debug, Snafu)]
 enum BuildError {
     #[snafu(display(
         "Missing authentication key, must provide either 'license_key' or 'insert_key'"
     ))]
     MissingAuthParam,
+    #[snafu(display(
+        "Too high batch max size. The value must be {} bytes or less",
+        max_size
```
jszwedko, line 27: Nit: I think you could leave out the parameter here and just use the defined constant, `MAX_PAYLOAD_SIZE`, as the parameter to the format string.

You could do maybe do something like:

```rust
     #[snafu(display(
        "Batch size of {} too high. The value must be {} bytes or less",
        batch_size,
        MAX_PAYLOAD_SIZE
     ))]
    BatchMaxSize { batch_size: usize },
```

jszwedko (approved):
Looks great, thanks for adding the error test too.

src/sinks/new_relic_logs.rs
matildasmeds: Personally, I would prefer not to include this link. The reason is that this documentation is written for the Event API, not for the Log API. Under the hood the rules are the same for both APIs, but this documentation talks explicitly of the Event API, which may be confusing to readers. Our docs team is working to provide the same information also in the Log API context.

src/sinks/new_relic_logs.rs
matildasmeds: Yes, considering this error is used only for this sink, the value is essentially a constant, rather than a parameter.

matildasmeds:
@Hoverbear & @jszwedko Nits applied! Why not make the code a bit prettier, once we're at it. As for link to docs please see the comment above. Many thanks for your prompt comments & suggestions!

src/sinks/new_relic_logs.rs
jszwedko: Ah, gotcha :+1:

jszwedko:
Looks good! Thanks again @matildasmeds

PR #4709, opened 2020-10-23
enhancement(remap transform): swap out remap impl with new remap-lang crate

JeanMertz:
This PR integrates the work done in #4695 back into Vector. It does the following:

- Allow conversions between `remap::Value` and `event::Value`.
- Implement `remap::Object` for `Event` to allow value insertions and querying.
- Swap out the existing `mapping` module in the `Remap` transform for the new `remap-lang` crate.
- Fixes some issues while running existing tests.

Some things that are **out-of-scope for this PR**, but will be done in follow-up PRs:

- Remove old remap implementation (to reduce diff noise in this PR).
- Use new `Lookup` type built by @Hoverbear for path insertions/lookup (I don't believe this is completely hooked-up yet, but will be soon!).
- Use remap in the `reduce` transform (being implemented in #4743).
- Complex performance optimizations such as reference counting values (we'll first set up a benchmark suite in the remap-lang crate before we focus on this).

I'd advice **against** reviewing the entire PR in one go, instead going commit by commit. Once you get to the _re-add "xxx" function_ commits, you can skip those, as those are mostly copy/pasting to the new trait-based implementation.


Cargo.toml
```diff
@@ -185,13 +188,13 @@ heim = { version = "0.1.0-beta.3", optional = true, features = ["full"] }
 uom = { version = "0.29.0", optional = true }
 rust_decimal = "1.8.1"
 mongodb = { version = "1.1.1", optional = true }
+anyhow = { version = "1.0.28" }
```
jszwedko, line 191: We also use `snafu`, should we standardize on one? They seem to overlap significantly.

jszwedko (approved):
This looks good to me overall. Admittedly I'm still lacking some context on the parser bits, but what I saw seemed reasonable.

I believe the `Lookup` stuff has landed. I'd be fine seeing that done as a follow-up though.

Cargo.toml
JeanMertz: Yeah, that's a good point.

I am intuitively drawn towards the thiserror+anyhow combination (and saw that anyhow was already used in Vector and so used that).

Having said that, snafu and anyhow do differ, whereas snafu and thiserror are more alike.

See also: https://old.reddit.com/r/rust/comments/dfs1zk/2019_q4_error_patterns_snafu_vs_errderive_anyhow/

Given the difference, and given we already had anyhow, and thiserror is contained to the `remap-lang` crate, I'd vote to leave things as is, and either open up a tracking issue to consolidate the libraries where needed, or leave the situation as is, and have people use the one they prefer (what matters most is how we eventually represent errors to the user, and that the errors we _do_ use implement `std::error::Error`.

bruceg: The `snafu` crate is designed for libraries, while `anyhow` is targeted at application usage, so they could potentially coexist. However, from what I am seeing, this use of `anyhow` *is* in a library, for which we would prefer the other.

JeanMertz: > However, from what I am seeing, this use of `anyhow` _is_ in a library, for which we would prefer the other.

I'm using `anyhow` it in tests, to produce a string representing the chain of errors that occurred.

bruceg: Ok, that's fair and isn't a problem.

lib/remap-lang/src/parser.rs
```diff
@@ -4,354 +4,363 @@ use crate::{
...
+            if is_escaped {
+                match c {
+                    '\\' => escaped_chars.push(c),
+                    'n' => escaped_chars.push('\n'),
+                    't' => escaped_chars.push('\t'),
+                    '"' => escaped_chars.push('"'),
+                    // This isn't reachable currently due to the explicit list of
+                    // allowed escape chars in our parser grammar. However, if that
+                    // changes then we might need to rely on this error.
+                    _ => return Err(Error::EscapeChar(c)),
```
bruceg, line 322: Could this simply default to: `_ => escaped_chars.push(c),`?

src/event/value.rs
```diff
@@ -202,6 +202,39 @@ impl TryInto<serde_json::Value> for Value {
...
+
+        match v {
+            Value::Bytes(v) => String(v),
+            Value::Integer(v) => Integer(v),
+            Value::Float(v) => Float(v),
+            Value::Boolean(v) => Boolean(v),
+            Value::Map(v) => Map(v.into_iter().map(|(k, v)| (k, v.into())).collect()),
+            Value::Array(v) => Array(v.into_iter().map(Into::into).collect()),
+            Value::Null => Null,
+            _ => todo!("Value::Timestamp"),
```
bruceg: What is the plan for the timestamp conversion?

bruceg (approved):
This looks ok to me, with the glaring exception of the separation of the two value types and the complications that leads to, all of which is out of scope of this PR.

src/mapping/query/function/mod.rs
```diff
@@ -127,37 +122,6 @@ macro_rules! build_signatures {
...
-    parse_duration => ParseDurationFn,
-    format_number => FormatNumberFn,
-    parse_url => ParseUrlFn,
-    starts_with => StartsWithFn,
-    ends_with => EndsWithFn,
-    round => RoundFn,
-    floor => FloorFn,
-    ceil => CeilFn,
-    parse_syslog => ParseSyslogFn,
     split => SplitFn,
```
FungusHumungus, line 125: Is there a reason Split has been left here?

src/event/value.rs
```diff
@@ -202,6 +202,40 @@ impl TryInto<serde_json::Value> for Value {
     }
 }
 
+impl From<remap::Value> for Value {
+    fn from(v: remap::Value) -> Self {
+        use remap::Value::*;
+
+        match v {
+            String(v) => Value::Bytes(v),
```
FungusHumungus, line 210: I'm not sure about the naming of `String` here, because really it is still `Bytes`.

StephenWakely (approved):

src/mapping/query/function/mod.rs
JeanMertz: No reason, It'll be removed in the clean-up PR which removes the old implementation.

src/event/value.rs
JeanMertz: `Timestamp` was added as a variant in a follow-up commit (when I needed it for one of the function impls).

JeanMertz:
Thanks for the feedback, @bruceg 

re:

> with the glaring exception of the separation of the two value types and the complications that leads to

To be clear (and I get why this wasn't obvious, since the `Value` type was changed in one of the function impl commits, which I mentioned not to review), `remap::Value` and `vector::event::Value` are currently identical.

There's an `impl From<event::Value> for remap::Value` (and vice versa) which basically moves the data between the container types, so the only downside is that we need an `into` call when we go from one to the other.

This might change in the future, but I don't think it's likely, and if it does, we can tackle any implications at that point.

lib/remap-lang/src/parser.rs
JeanMertz: It seems like it could, but I'll double-check (this code was copy/pasted from the original impl).

JeanMertz: One reason I could see this being confusing, is if we add an escape char such as `r` to the parser, we'd have to explicitly convert that to `\r` here, so just pushing the char as a default case won't work in that case.

I think having an explicit error makes it more clear that you need to special case the new escape character.

As the comment says, there's no way this can trigger unless the PEG is updated, so I'm inclined to leave this as a more explicit case that needs to be updated if the grammar changes.

JeanMertz: Having said that, I realized this error doesn't have to be special-cased, compared to other "invalid token" errors.

So I've updated this a bit in adcd42fe7 (keeping the original intent of returning an error if the grammar invalidates this parse function).

PR #4719, opened 2020-10-23
perf(file source): send batches of lines

lukesteensen:
This modifies the file source to send its whole `Vec` of lines at once instead of one at a time. It also attempts to unravel some of the compat layers that have accumulated in this area of the code.

Locally this shows a ~20% or more improvement to performance.

### DIFF
--- a/lib/file-source/src/file_server.rs
+++ b/lib/file-source/src/file_server.rs
@@ -65,10 +65,10 @@ where
         self,
         mut chans: C,
         mut shutdown: impl Future + Unpin,
-    ) -> Result<Shutdown, <C as Sink<(Bytes, String)>>::Error>
+    ) -> Result<Shutdown, <C as Sink<Vec<(Bytes, String)>>>::Error>
     where
-        C: Sink<(Bytes, String)> + Unpin,
-        <C as Sink<(Bytes, String)>>::Error: std::error::Error,
+        C: Sink<Vec<(Bytes, String)>> + Unpin,
+        <C as Sink<Vec<(Bytes, String)>>>::Error: std::error::Error,
     {
         let mut fingerprint_buffer = Vec::new();
 
@@ -284,7 +284,8 @@ where
             });
 
             let start = time::Instant::now();
-            let mut stream = stream::iter(lines.drain(..).map(Ok));
+            let to_send = std::mem::take(&mut lines);
+            let mut stream = stream::once(futures::future::ok(to_send));
             let result = block_on(chans.send_all(&mut stream));
             match result {
                 Ok(()) => {}
--- a/src/line_agg.rs
+++ b/src/line_agg.rs
@@ -135,7 +135,11 @@ impl<K, C> Logic<K, C> {
     }
 }
 
-impl<T, K, C> LineAgg<T, K, C> {
+impl<T, K, C> LineAgg<T, K, C>
+where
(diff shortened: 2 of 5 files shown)


src/line_agg.rs
```diff
@@ -135,7 +135,11 @@ impl<K, C> Logic<K, C> {
     }
 }
 
-impl<T, K, C> LineAgg<T, K, C> {
+impl<T, K, C> LineAgg<T, K, C>
+where
+    T: Stream<Item = (K, Bytes, C)> + Unpin,
+    K: Hash + Eq + Clone,
+{
```
lukesteensen, line 142: I added this to improve the error messages we get when `LineAgg` was created with something that didn't `impl Stream` properly. This catches it a bit earlier and makes things less confusing.

src/sources/file.rs
```diff
@@ -225,25 +225,15 @@ pub fn file_source(
...
-            });
-            let logic = line_agg::Logic::new(config);
-            Box::new(Compat::new(
-                LineAgg::new(rx.map(|(line, src)| (src, line, ())), logic)
-                    .map(|(src, line, _context)| (line, src))
-                    .map(Ok),
-            ))
-        };
-        let messages: Box<dyn Stream<Item = (Bytes, String), Error = ()> + Send> =
+        let (tx, rx) = futures::channel::mpsc::channel::<Vec<(Bytes, String)>>(2);
```
lukesteensen, line 233: I reduced the size of the channel significantly because there's now the possibility that file source reads will far outrun what we've been able to process through the rest of the topology. This limits it to roughly `4 * max_read_bytes * num_files` (2 batches in the channel and one on either end).

lib/file-source/src/file_server.rs
```diff
@@ -284,7 +284,8 @@ where
             });
 
             let start = time::Instant::now();
-            let mut stream = stream::iter(lines.drain(..).map(Ok));
+            let to_send = std::mem::take(&mut lines);
```
MOZGIII, line 287: Very elegant! I thought this would be more difficult!

Hoverbear (approved):

src/line_agg.rs
MOZGIII, line 142: This might not be necessary, or if really needed it's better to move the `new` fn to the lower block where generic contraints are explicit already.

src/sources/file.rs
```diff
@@ -292,6 +284,17 @@ pub fn file_source(
     }))
 }
 
+fn wrap_with_line_agg(
```
MOZGIII, line 287: Good thing that this is now extracted!

MOZGIII (approved):

src/sources/file.rs
MOZGIII: I was going to ask about it, makes a lot of sense!

MOZGIII:
Let me just test this with a k8s E2E flow

src/line_agg.rs
MOZGIII: Github doesn't merge the comments :( See https://github.com/timberio/vector/pull/4719#discussion_r511015595
Let's just move the `new` to the block below. This is just a nit though.

lukesteensen: We can't move `new` because it's not part of the `Stream` trait. Again, this isn't required at all, it just helps give a more useful error message for users of `LineAgg`.

PR #4730, opened 2020-10-24
fix(architecture): Enable API (+ client) on musl targets

leebenson:


jamtur01 (approved):
LGTM

PR #4733, opened 2020-10-25
enhancement(logplex source): annotate logs with query parameters

### LINKED ISSUE #4732: Support annotating logs with URL parameters for the logplex source
Discussion in discord: https://discordapp.com/channels/742820443487993987/746070591097798688/769706197547876362

In its current form, it seems like the `logplex` source is most useful when used as the logdrain for a single Heroku app because it isn't possible to differentiate the logs by Heroku app.

Example set of Heroku logs for a single app:

```
2020-10-24T08:40:12.187823+00:00 app[web.1]: * Listening on tcp://0.0.0.0:38071
2020-10-24T08:40:12.188274+00:00 app[web.1]: Use Ctrl-C to stop
2020-10-24T08:40:12.621499+00:00 heroku[web.1]: State changed from starting to up
2020-10-24T08:40:13.137687+00:00 heroku[router]: at=info method=GET path="/robots.txt" host=adoptadrainsf-staging.herokuapp.com request_id=a05a4bad-43b5-4bbd-9138-4e7a32d4d0c2 fwd="157.55.39.39" dyno=web.1 connect=0ms service=8ms status=200 bytes=333 protocol=http
2020-10-24T08:40:31.705732+00:00 heroku[router]: at=info method=GET path="/sitemap.xml" host=adoptadrainsf-staging.herokuapp.com request_id=db5e6e9b-729a-4cf7-9976-857953979ddf fwd="52.247.221.252" dyno=web.1 connect=3ms service=101ms status=200 bytes=686 protocol=http
2020-10-24T09:15:37.291335+00:00 heroku[web.1]: Idling
```

Here we can see that Heroku uses the app name field to differentiate between runtime logs, router logs, and app logs.

The proposed idea, which I also like, is to allow parsing out URL parameters for that source and adding them as log fields. This would let users set their logdrain to things like:

```
https://url.for.vector:443?app=myapplicationname
```

The `logplex` source could then use this to annotate all of the logs with `app`.

We could also consider this for other HTTP-based sources. We have some precedent of doing so with headers for the generic `http` source.

CGA1123:
Allow annotation of logs via query parameters for http + logplex source,
in a similar fashion as is currently supported for HTTP headers.

This is particularly useful with the logplex source as it allows for a
single vector instance to be used as a drain for multiple Heroku
applications.

Closes #4732

Signed-off-by: Christian Gregg <christian@bissy.io>

### DIFF
--- a/src/sources/http.rs
+++ b/src/sources/http.rs
@@ -119,16 +119,13 @@ fn add_headers(
     headers: HeaderMap,
 ) -> Vec<Event> {
     for header_name in headers_config {
-        let value = headers
-            .get(header_name)
-            .map(HeaderValue::as_bytes)
-            .map(<&[u8]>::to_owned)
-            .map(Bytes::from);
+        let value = headers.get(header_name).map(HeaderValue::as_bytes);
 
         for event in events.iter_mut() {
-            event
-                .as_mut_log()
-                .insert(header_name as &str, Value::from(value.to_owned()));
+            event.as_mut_log().insert(
+                header_name as &str,
+                Value::from(value.map(Bytes::copy_from_slice)),
+            );
         }
     }
 
--- a/src/sources/http.rs
+++ b/src/sources/http.rs
@@ -4,7 +4,7 @@ use crate::{
     },
     event::{Event, Value},
     shutdown::ShutdownSignal,
-    sources::util::{ErrorMessage, HttpSource, HttpSourceAuthConfig},
+    sources::util::{ErrorMessage, HttpSource, HttpSourceAuthConfig, add_query_parameters},
     tls::TlsConfig,
     Pipeline,
 };
@@ -126,27 +126,6 @@ fn add_headers(
         for event in events.iter_mut() {
             event.as_mut_log().insert(
                 header_name as &str,
-                Value::from(Bytes::from(value.to_owned())),
-            );
-        }
-    }
-
-    events
-}
-
-fn add_query_parameters(
-    mut events: Vec<Event>,
-    query_parameters_config: &[String],
-    query_parameters: HashMap<String, String>,
-) -> Vec<Event> {
-    for query_parameter_name in query_parameters_config {
-        let value = query_parameters
-            .get(query_parameter_name)
-            .map(String::as_bytes)
-            .unwrap_or_default();
-        for event in events.iter_mut() {
-            event.as_mut_log().insert(
-                query_parameter_name as &str,
                 Value::from(Bytes::from(value.to_owned())),
             );
         }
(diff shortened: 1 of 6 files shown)


src/sources/http.rs
```diff
@@ -124,6 +134,27 @@ fn add_headers(
     events
 }
 
+fn add_query_params(
```
CGA1123, line 137: This is duplicated exactly in `logplex`, maybe could be pulled into `util/http.rs`?

src/sources/logplex.rs
```diff
@@ -23,6 +24,7 @@ use warp::http::{HeaderMap, StatusCode};
 #[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct LogplexConfig {
     address: SocketAddr,
+    query_params: Vec<String>,
```
jszwedko, line 27: Since this will be pulled out as the public facing user configuration, could we call it `query_parameters`?

src/sources/logplex.rs
```diff
@@ -108,6 +121,27 @@ fn decode_message(body: Bytes, header_map: HeaderMap) -> Result<Vec<Event>, Erro
...
+fn add_query_params(
+    mut events: Vec<Event>,
+    query_params_config: &[String],
+    query_params: HashMap<String, String>,
+) -> Vec<Event> {
+    for query_param_name in query_params_config {
+        let value = query_params
+            .get(query_param_name)
+            .map(String::as_bytes)
+            .unwrap_or_default();
```
jszwedko, line 133: I'm seeing that this matches `add_headers`, but I'm wondering if we shouldn't map `None` to `event::Value::Null` rather than the empty string. That would help distinguish parameters that are missing from those that exist, but are empty. I think `add_headers` may have predated `event::Value::Null`; I wouldn't mind seeing that updated too.

src/sources/http.rs
jszwedko: :+1: I think that's a good idea. Maybe `src/sources/util/http.rs` since I think only sources would make use of it.

jszwedko (commented):
Thanks for this @CGA1123 !

Just one note about mapping non-existent parameters, but I think the rest looks good.

We'll want to add this to the docs too. If you feel like taking a shot, you can find them in `docs/`, otherwise we can add them once this is merged.

CGA1123:
Hi @jszwedko thanks for the review, I think I have addressed your comments!
I took a shot at updating the docs as well 😄 

CI seems to have hanged on the mac build though, not sure why...

jszwedko (approved):
Looks great! Thanks @CGA1123 . I kicked CI to run it again and will merge once it is green.

CGA1123:
Yay! Thanks @jszwedko 😄 

Will be trying this out in the coming days on our apps at work I think 🚀

PR #4735, opened 2020-10-25
enhancement(reload): Resolve port conflict in sinks

### LINKED ISSUE #4031: Wait for sinks to finish on reload
Currently during reload we are only waiting for sources to finish, but we should also wait for sinks to finish for two reasons:
* Some sinks hold user defined ports so we need to wait for the sinks to release them.
   * Currently the only sink that does this is `prometheus` sink which will panic, and cause vector to shutdown, if we try to change it without changing the address port.
* If the old and the new sink with the same destination are allowed to run at the same time, their event streams can be intermingled from the point of view of the destination which can cause out of order events.

It's probably the cause of #3727.

ktff:
Closes #4031 

Adds `Resource` which configs of sinks can report as having. This information is used to:
* Validate that there aren't port conflicts among the sinks in the config.
* Detect conflicting sinks, among a new and an old config, during reload. If found, we wait for them to finish before starting new ones. Crucially, sinks with resources should not use them during build phase, otherwise conflict can still occur. 

### Todo
After merging this, open issues for:
- [x] Extending resources to sources
	- Will enable detection of more resource conflicts in config.
	- Reload will be more robust.
	- Resource conflicts between sinks and sources would be detected. 
- [x] Enhance error reporting of conflicting usages of resources. 
	- Useful after resources have been extended to sources.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -190,6 +190,9 @@ pub trait SinkConfig: core::fmt::Debug + Send + Sync {
 
     fn sink_type(&self) -> &'static str;
 
+    /// Resources that the sink is using.
+    /// These resources shouldn't be used in the build method as that can result
+    /// in a build error. Only the sinks are contrained by this.
     fn resources(&self) -> Vec<Resource> {
         Vec::new()
     }
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -7,7 +7,9 @@ use component::ComponentDescription;
 use indexmap::IndexMap; // IndexMap preserves insertion order, allowing us to output errors in the same order they are present in the file
 use serde::{Deserialize, Serialize};
 use snafu::{ResultExt, Snafu};
+use std::collections::{HashMap, HashSet};
 use std::fs::DirBuilder;
+use std::hash::Hash;
 use std::path::PathBuf;
 
 pub mod api;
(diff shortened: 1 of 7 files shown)


src/config/mod.rs
```diff
@@ -187,6 +189,13 @@ pub trait SinkConfig: core::fmt::Debug + Send + Sync {
     fn input_type(&self) -> DataType;
 
     fn sink_type(&self) -> &'static str;
+
+    /// Resources that the sink is using.
+    /// These resources shouldn't be used in the build method as that can result
+    /// in a build error. Only the sinks are contrained by this.
+    fn resources(&self) -> Vec<Resource> {
```
lukesteensen, line 196: If we make this return a `HashSet` instead of `Vec` we could maybe simplify the `conflicts` method and just verify that the `intersection` with every other set is empty.

src/topology/mod.rs
```diff
@@ -400,6 +430,17 @@ impl RunningTopology {
         for name in &diff.sinks.to_add {
             self.setup_inputs(&name, new_pieces);
         }
+
+        // Since inputs of diff.sinks.to_change need to be replaced to avoid loosing events
```
lukesteensen, line 434:
```suggestion
        // Since inputs of diff.sinks.to_change need to be replaced to avoid losing events
```

lukesteensen (approved):
This turned out really well! I left a couple of small comments, but nothing that needs to block.

src/config/mod.rs
ktff: It would make it a bit simpler, but it would also make impl of `resources` function a bit more verbose. There is also one more reason, I see error reporting of conflicts (Todo: add issue link here) to be better grouped around resources than components, so current version of `conflicts` will be more natural to extend.

PR #4736, opened 2020-10-26
fix(ci): Run /lib/* tests in  on pull requests

### LINKED ISSUE #2588: Run tests in `lib/*` on pull requests
Are tests run in the `lib` subdirectories? I just checked submitted a [pull request](https://github.com/timberio/vector/pull/2587) that touched only files in `lib/codec`, and no tests were run. Looking at other pull requests, it doesn't appear tests are being run in `lib` at all.

Since changes to the sources in `lib` can fix or cause bugs in the rest of the code, we should at least be running standard tests when those change. Additionally, running tests in those directories would be a good idea.

jamtur01:

This adds running the `lib/*` tests for every PR - these were not previously run as far as I can see.

Closes #2588

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -260,7 +260,7 @@ cross-image-%:
 
 .PHONY: test
 test: ## Run the unit test suite
-	${MAYBE_ENVIRONMENT_EXEC} cargo test --no-fail-fast --no-default-features --features ${DEFAULT_FEATURES} ${SCOPE} -- --nocapture
+	${MAYBE_ENVIRONMENT_EXEC} cargo test --no-fail-fast --no-default-features --workspace --features ${DEFAULT_FEATURES} ${SCOPE} -- --nocapture
 
 .PHONY: test-all
 test-all: test test-behavior test-integration ## Runs all tests, unit, behaviorial, and integration.


Makefile
```diff
@@ -260,7 +260,7 @@ cross-image-%:
 
 .PHONY: test
 test: ## Run the unit test suite
-	${MAYBE_ENVIRONMENT_EXEC} cargo test --no-fail-fast --no-default-features --features ${DEFAULT_FEATURES} ${SCOPE} -- --nocapture
+	${MAYBE_ENVIRONMENT_EXEC} cargo test --no-fail-fast --no-default-features --workspace --features ${DEFAULT_FEATURES} ${SCOPE} -- --nocapture
```
Hoverbear, line 263: Please note the file source tests on Windows may take quite some minutes due to the poor perf.

Hoverbear (approved):
While this is fine I'm not sure this will have the impact you want on Windows, the file tests take quite some time.

jamtur01:
@Hoverbear I mean the Check Components check is longer at 46 minutes? So it's moot perhaps?

PR #4750, opened 2020-10-26
chore(ci): Re-introducing grub build fix

jamtur01:

https://github.com/timberio/vector/pull/3962


Hoverbear (approved):

jamtur01:
This just needs the bootstrap to work so skipping actually completing all the tests.

MOZGIII:
We can actually just uninstall grub instead, it's not like the instance we use is to be ever booted again after us.

jamtur01:
It gets sucked with `apt upgrade` I believe.

PR #4764, opened 2020-10-26
fix(ci): Fix unset $PUSH for build-docker.sh

jszwedko:
Resolves
https://github.com/timberio/vector/commit/26be119ac6c506cd50001b5fdc94adafdecb0f2d#r43596614

I opted to use the eval syntax mentioned in
http://mywiki.wooledge.org/BashFAQ/050#I_only_want_to_pass_options_if_the_runtime_data_needs_them
to still allow for quoting here to satsify shellcheck. This also seems
preferable to the field splitting example there (which isn't relevant in
our case since it is a single argument, but could be in other uses of
`${FOO+'--bar' 'baz'}` expansions).

Thanks for catching that @MOZGIII !

### DIFF
--- a/scripts/build-docker.sh
+++ b/scripts/build-docker.sh
@@ -35,11 +35,11 @@ build() {
     docker buildx create --use --name vector-builder
     docker buildx install
 
-    docker buildx build \
+    eval docker buildx build \
       --platform="$PLATFORM" \
       --tag "$TAG" \
       target/artifacts \
-      -f "$DOCKERFILE" "${PUSH:+--push}"
+      -f "$DOCKERFILE" "${PUSH:+'--push'}"
   else
     docker build \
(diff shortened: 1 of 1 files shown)


MOZGIII:
I'd rather avoid using `eval`, it can be very surprising. Why don't we use just bash arrays instead? It's going to be much easier to understand. I take a stab at it!

jszwedko:
@MOZGIII that'd be great if you already have an idea of what it'd look like. I feel like I always struggle with bash arrays and building commands like this.

MOZGIII (changes requested):
I created the https://github.com/timberio/vector/pull/4900 with the bash arrays

jszwedko:
:+1: Closing this in-lieu of #4900

PR #4767, opened 2020-10-26
chore: Added support for an /etc/default/vector environment file.

### LINKED ISSUE #4748: Add environment support to systemd unit file
Adding support for environment variable management to the system unit file.

c.f. #4741

jamtur01:
Closes #4748

### DIFF
--- a/distribution/systemd/vector.service
+++ b/distribution/systemd/vector.service
@@ -11,7 +11,7 @@ ExecStart=/usr/bin/vector
 ExecReload=/bin/kill -HUP $MAINPID
 Restart=no
 AmbientCapabilities=CAP_NET_BIND_SERVICE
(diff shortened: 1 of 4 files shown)


distribution/systemd/vector.service
```diff
@@ -11,6 +11,7 @@ ExecStart=/usr/bin/vector
 ExecReload=/bin/kill -HUP $MAINPID
 Restart=no
 AmbientCapabilities=CAP_NET_BIND_SERVICE
+EnvironmentFile=/etc/default/vector
```
jszwedko, line 14: Not a big deal since we are packaging one, but I've usually seen these as `EnvironmentFile=-` to avoid erroring if the file doesn't exist.

jszwedko (approved):
I like it!

PR #4773, opened 2020-10-27
chore(ci): Run behavior tests for PRs

jszwedko:
They seem to run fairly quickly (2 minutes locally).

This in response to a commit sneaking through to master that broke one
of them:
https://github.com/timberio/vector/runs/1314766689?check_suite_focus=true

Also renames the shutdown test action in the master workflow to match
the fact that it runs miscellaneous tests.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -131,7 +131,7 @@ jobs:
         run: cargo test --no-fail-fast --no-default-features --features default-msvc
 
   test-misc:
-    name: Shutdown - Linux
+    name: Miscellaneous - Linux
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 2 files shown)


jszwedko:
I think it'd be nice to break up these actions more so that we could, for example: build, cache that build, and then run `make test` and `make behavior-test` as separate actions using the cached build, but hesitated to take on that refactoring as part of this. Just leaving it here as food for thought as we make more changes to our GA workflows.

jamtur01 (approved):
Good call. +1.

jszwedko:
Failing test fixed by https://github.com/timberio/vector/pull/4703/commits/83ff1d2f62862c00b64905fa2ac8595834584f5f

PR #4779, opened 2020-10-27
feat(new source): `aws_s3` source

### LINKED ISSUE #1017: New `aws_s3` source
Needs Source: AWS S3

jszwedko:
See RFC: https://github.com/timberio/vector/blob/master/rfcs/2020-09-29-4155-aws-s3-source.md
Fixes #1017 

This is the initial implementation of an `aws_s3` source that relies an AWS SQS for bucket notifications to inform of new S3 objects to consume. See RFC for discussion of other approaches and why this one was chosen. The source does have an enum `strategy` configuration to allow for additional approaches (like SNS or long polling) to be supported.

The basic flow is:

User setup:
* bucket is created
* queue is created
* bucket is configured to notify the queue for ObjectCreated events
* vector is configured with the `aws_s3` source using the queue configuration whereupon it will process the ObjectCreated events to read each S3 object.

Example configuration:

```toml
[sources.s3]
  type = "aws_s3"

  region = "us-east-1"
  strategy = "sqs"
  compression = "gzip"

  sqs.queue_name = "jesse-test-flow-logs"
  sqs.delete_message = false

[sinks.console]
  type = "console"
  inputs = ["s3"]
  encoding.codec = "json"
```

The commits can be viewed in-order, but the resulting diff probably isn't to bad either (it's mostly just `sources/aws_s3`). It may be worth looking at the added cue documentation first.

The source also behaves very much like the `file` source where it emits one event per-line, but also supports the same multiline configuration that the `file` source supports. **Note** there is a rough edge here where the `multiline` config supports a `timeout_ms` option that isn't really applicable here but is applied just the same.

Future work:
1. Additional codec support (like `application/ndjson`). For now, this acts very much like the `file` source. This could be looped into the general work around codecs  https://github.com/timberio/vector/issues/4278
2. Additional compression formats (Snappy, LZ4, Zip). This was requested in the original issue. I started with just supporting the formats that were supported out-of-the-box by the `async_compression` crate we are using.
3. (potential) multi-region support. Currently we only support reading from a queue and a bucket in the same region. I expect this will cover most cases since AWS requires the bucket to publish notifications to a queue in the same region. One could forward messages from a queue in one region to another, but this seems unlikely. I'd prefer to wait and see if anyone asks for multi-region support; especially given that fluentbit and filebeat have the same restriction.
4. Concurrent processing. Right now one message is processed at a time which leads to predictable behavior, but we may observe some performance improvements by processing multiple objects at once. The benefit should be vetted though, the process may be limited by incoming network bandwidth anyway.
5. Refresh the message visibility timeout. Right now, the visibility timeout is set once, when the message is retrieved, but we could refresh this timeout if we are still processing a message when it gets close to the end of the timeout to avoid another vector instance picking it up. This would let users have the best of both worlds: short visibility timeouts to quickly reprocess messages when a `vector` instance falls over, but also avoiding concurrent processing of messages for large objects where the processing time exceeds the visibility timeout.

I'll create issues for 2 and 5. I think the others can be left until we observe their necessity.


docs/reference/components/aws.cue
```diff
@@ -0,0 +1,153 @@
+package metadata
```
jszwedko, line 1: Moved from `docs/reference/components/sinks/aws.cue` and updated to cover both sources and sinks.

src/sources/aws_s3/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+use rusoto_sqs::SqsClient;
+use serde::{Deserialize, Serialize};
+use snafu::{ResultExt, Snafu};
+use std::convert::TryInto;
+
+pub mod sqs;
+
+#[derive(Copy, Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum Compression {
```
jszwedko, line 22: It might be nice to pull this up and share it with the `file` source in the future.

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,733 @@
...
+            })?;
+
+        let queue_url = queue_url_result
+            .queue_url
+            .ok_or(IngestorNewError::MissingQueueUrl {
+                name: config.queue_name.clone(),
+                owner: config.queue_owner.clone(),
+            })?;
+
+        // This is a bit odd as AWS wants an i64 for this value, but also doesn't want negative
```
jszwedko, line 154: Ideas here welcome :smile:

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,733 @@
...
+            Some(body) => {
+                let r = super::s3_object_decoder(
+                    self.compression,
+                    &s3_event.s3.object.key,
+                    object.content_encoding.as_deref(),
+                    object.content_type.as_deref(),
+                    body,
+                );
+
+                // Record the read error saw to propagate up later so we avoid ack'ing the SQS
```
jszwedko, line 332: Ideas here also welcome :smile:

src/sources/aws_s3/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+                expected
+            );
+        }
+    }
+}
+
+#[cfg(feature = "aws-s3-integration-tests")]
+#[cfg(test)]
+mod integration_tests {
+    use super::{sqs, AwsS3Config, Compression, Strategy};
```
jszwedko, line 256: I abstained from using `use super::*` here to see how it felt. I think I like the explicitness better.

jszwedko (commented):
I think this could probably be better refactored by having the SQS bits implement an interface to drive a top-level object reader by having it simply return fields to read, but I figured that was better left until we actually add another "strategy" which could then refactor this a bit to share the logic doing the object reads and event creation.

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,517 @@
...
+                //
+                // FramedRead likely stops when it gets an i/o error but I found it more clear to
+                // show that we `take_while` there hasn't been an error
+                //
+                // This can result in objects being partially processed before an error, but we
+                // prefer duplicate lines over message loss. Future work could include recording
+                // the offset of the object that has been read, but this would only be relevant in
+                // the case that the same vector instance processes the same message.
+                let mut read_error: Option<String> = None;
+                let mut lines: Box<dyn Stream<Item = Bytes> + Send + Unpin> = Box::new(
```
bruceg, line 348: What about having `Item = Result<Bytes, io::Error>` and handle the errors at a later step?

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,517 @@
...
+                                Err(err) => {
+                                    read_error = Some(err.to_string());
+                                    false
+                                }
+                            })
+                        })
+                        .map(|r| r.unwrap()), // validated by take_while
+                );
+                if let Some(config) = &self.multiline {
+                    lines = Box::new(
```
bruceg, line 362: Suggest:
```rust
let lines = match &self.multiline {
    None => lines,
    Some => Box::new(…),
};
```

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,517 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "PascalCase")]
+struct S3Event {
+    records: Vec<S3EventRecord>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct S3EventRecord {
+    event_version: String, // TODO compare >=
```
bruceg, line 446: Is this still TODO?

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,517 @@
...
+    {
+        use serde::de::Error;
+
+        let s = String::deserialize(deserializer)?;
+
+        let mut parts = s.splitn(2, ':');
+
+        let kind = parts
+            .next()
+            .ok_or_else(|| D::Error::custom("Missing event type"))?
```
bruceg, line 477: Event "type" or "kind"?

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,517 @@
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_str(&format!("{}:{}", self.name, self.kind))
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
```
bruceg, line 501: These three `rename_all` are unnecessary.

src/sources/aws_s3/mod.rs
```diff
@@ -0,0 +1,477 @@
...
+    None,
+    Gzip,
+    Zstd,
+}
+
+impl Default for Compression {
+    fn default() -> Self {
+        Compression::Auto
+    }
+}
```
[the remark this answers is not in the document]
jszwedko: TIL about the `derivative` crate. This should let me clean up a few things here :smile:

src/sources/aws_s3/mod.rs
```diff
@@ -0,0 +1,477 @@
...
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(default, deny_unknown_fields)]
+struct AwsS3Config {
+    #[serde(flatten)]
+    region: RegionOrEndpoint,
+
+    compression: Compression,
+
+    strategy: Strategy,
```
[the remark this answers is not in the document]
jszwedko: I was thinking that the `serde(default)` on the struct definition would do this, but maybe I'm wrong?

src/sources/aws_s3/sqs.rs
jszwedko: Good call, I wanted to use `type` but given it is a reserved keyword, I shifted to `kind`. I'll update this.

src/sources/aws_s3/sqs.rs
jszwedko: They are, strictly, redundant, but I find them useful as a sort of documentation about the serialized form.

I'm just pulling out `bucket` and `object` here but there are additional keys, like `s3SchemaVersion`, that would need to be translated to camelCase.

src/sources/aws_s3/mod.rs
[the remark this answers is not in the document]
bruceg: Ah, you are right, of course. I think I have used that attribute on too many things where I only wanted some of the attributes to default, that I forgot to look at the container attribute.

src/internal_events/mod.rs
```diff
@@ -14,6 +14,8 @@ mod aws_cloudwatch_logs_subscription_parser;
 mod aws_kinesis_firehose;
 #[cfg(feature = "sinks-aws_kinesis_streams")]
 mod aws_kinesis_streams;
+#[cfg(any(feature = "sources-aws_s3", feature = "sinks-aws_s3"))]
+pub mod aws_s3;
```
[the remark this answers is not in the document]
jszwedko: :+1: I meant to call this out too as it diverges from our existing pattern of re-exporting the types in each module. I liked this as it allowed me to refer to them as: `use crate::internal_events::aws_s3::source::{....}` thus avoiding the need to include the "namespace" of each event as part of its name.

If we like this pattern, I can document it to prefer for new internal events (and to slowly migrate existing ones as they are touched).

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,517 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    async fn handle_sqs_message(
+        &self,
+        message: Message,
+        out: Pipeline,
```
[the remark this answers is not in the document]
jszwedko: `send_all` needs `self` so I don't think I can avoid cloning. I might be missing something though.

src/sources/aws_s3/sqs.rs
jszwedko: Ah, yes, thanks. I missed that when combing for lingering TODOs.

src/sources/aws_s3/mod.rs
```diff
@@ -0,0 +1,477 @@
...
+    #[snafu(display("Unable to initialize: {}", source))]
+    Initialize { source: sqs::IngestorNewError },
+    #[snafu(display("Unable to create AWS client: {}", source))]
+    Client { source: crate::Error },
+    #[snafu(display("Unable to create AWS credentials provider: {}", source))]
+    Credentials { source: crate::Error },
+    #[snafu(display("sqs configuration required when strategy=sqs"))]
+    ConfigMissing,
+    #[snafu(display("could not parse region configuration: {}", source))]
+    RegionParse { source: rusoto::region::ParseError },
```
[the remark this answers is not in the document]
jszwedko: I updated the `snafu(display())` on `rusoto::region::ParseError`. I feel like this display message is still useful for tying it back to the actual configuration value that failed to parse. In this case, there is only one `Region` in the config, but we could add more.

src/sources/aws_s3/sqs.rs
jszwedko: I actually tried something like this, but `LineAgg` needs `Bytes` so I'd just need to catch it there, which isn't much later. I'm definitely interested in improving this though as it seems like the pattern probably won't be uncommon.

jszwedko:
For the S3 event record version validation, I ended up pulling in the `semver` crate to handle the deserialization and comparison. It's a small crate and we were already indirectly depending on it:

```
$ cargo tree -i  semver      
semver v0.9.0
├── cargo-lock v4.0.1
│   └── built v0.4.3
│       [build-dependencies]
│       └── vector v0.11.0 (/home/jesse/workspace/vector)
└── rustc_version v0.2.3
    [build-dependencies]
    ├── cast v0.2.3
    │   ├── criterion v0.3.3
    │   │   [dev-dependencies]
    │   │   └── vector v0.11.0 (/home/jesse/workspace/vector)
    │   └── criterion-plot v0.4.3
    │       └── criterion v0.3.3 (*)
    ├── raw-cpuid v8.1.1
    │   └── heim-virt v0.1.0-beta.1
    │       └── heim v0.1.0-beta.3
    │           └── vector v0.11.0 (/home/jesse/workspace/vector)
    ├── rusoto_core v0.45.0
    │   ├── rusoto_cloudwatch v0.45.0
    │   │   └── vector v0.11.0 (/home/jesse/workspace/vector)
    │   ├── rusoto_firehose v0.45.0
    │   │   └── vector v0.11.0 (/home/jesse/workspace/vector)
    │   ├── rusoto_kinesis v0.45.0
    │   │   └── vector v0.11.0 (/home/jesse/workspace/vector)
    │   ├── rusoto_logs v0.45.0
    │   │   └── vector v0.11.0 (/home/jesse/workspace/vector)
    │   ├── rusoto_s3 v0.45.0
    │   │   └── vector v0.11.0 (/home/jesse/workspace/vector)
    │   ├── rusoto_sts v0.45.0
    │   │   └── vector v0.11.0 (/home/jesse/workspace/vector)
    │   └── vector v0.11.0 (/home/jesse/workspace/vector)
    └── rusoto_signature v0.45.0
        ├── rusoto_core v0.45.0 (*)
        └── vector v0.11.0 (/home/jesse/workspace/vector)
```

jszwedko:
Noting for myself that I need to update this to not allow `timeout_sec`s to apply for the multiline aggregation.

src/internal_events/mod.rs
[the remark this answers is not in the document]
JeanMertz: I'm a fan of this, as long as we apply it consistently 👍

docs/reference/components/sources/aws_s3.cue
```diff
@@ -0,0 +1,225 @@
...
+		}
+
+		requirements: []
+		warnings: []
+		notices: []
+	}
+
+	configuration: {
+		strategy: {
+			common:      true
```
JeanMertz, line 62: Should we set this to `false`, given that there isn't any strategy other than the default to pick at the moment?

docs/reference/components/sources/aws_s3.cue
```diff
@@ -0,0 +1,225 @@
...
+						required:    false
+						warnings: []
+						type: uint: {
+							default: 15
+							unit:    "seconds"
+						}
+					}
+					visibility_timeout_secs: {
+						common:      false
+						description: "The visibility timeout to use for messages in secords. This controls how long a message is left unavailable when a Vector receives it. If Vector does not delete the message before the timeout expires, it will be made reavailable for another consumer; this can happen if, for example, the `vector` process crashes."
```
JeanMertz, line 106: s/a Vector/Vector

docs/reference/components/sources/aws_s3.cue
```diff
@@ -0,0 +1,225 @@
...
+						}
+					}
+					delete_message: {
+						common:      true
+						description: "Whether to delete the message once Vector processes it. It can be useful to set this to `false` to debug or during initial Vector setup."
+						required:    false
+						warnings: []
+						type: bool: default: true
+					}
+					queue_name: {
```
JeanMertz, line 121: This one is missing the `common` flag

docs/reference/components/sources/aws_s3.cue
```diff
@@ -0,0 +1,225 @@
...
+					types.end = "timestamp|%s"
+				```
+
+				To parse AWS load balancer logs, the `regex_parser` transform can be used:
+
+				```toml
+				[transforms.elasticloadbalancing_fields_parsed]
+					type = "regex_parser"
+					inputs = ["s3"]
+					regex = '^(?P<type>[\\w]+) (?P<timestamp>[\\w:.-]+) (?P<elb>[^\\s]+) (?P<client_host>[\\d.:-]+) (?P<target_host>[\\d.:-]+) (?P<request_processing_time>[\\d.-]+) (?P<target_processing_time>[\\d.-]+) (?P<response_processing_time>[\\d.-]+) (?P<elb_status_code>[\\d-]+) (?P<target_status_code>[\\d-]+) (?P<received_bytes>[\\d-]+) (?P<sent_bytes>[\\d-]+) "(?P<request_method>[\\w-]+) (?P<request_url>[^\\s]+) (?P<request_protocol>[^"\\s]+)" "(?P<user_agent>[^"]+)" (?P<ssl_cipher>[^\\s]+) (?P<ssl_protocol>[^\\s]+) (?P<target_group_arn>[\\w.:/-]+) "(?P<trace_id>[^\\s"]+)" "(?P<domain_name>[^\\s"]+)" "(?P<chosen_cert_arn>[\\w:./-]+)" (?P<matched_rule_priority>[\\d-]+) (?P<request_creation_time>[\\w.:-]+) "(?P<actions_executed>[\\w,-]+)" "(?P<redirect_url>[^"]+)" "(?P<error_reason>[^"]+)"'
```
JeanMertz, line 205: nit: might be worthwhile to split this regex over multiple lines for readability (and enable the flag that makes the regex ignore whitespace and comments).

src/internal_events/aws_s3.rs
```diff
@@ -0,0 +1,119 @@
...
+            warn!(message = "Deletion of SQS message failed", %self.message_id, %self.error);
+        }
+
+        fn emit_metrics(&self) {
+            counter!("sqs_message_delete_failed_total", 1);
+        }
+    }
+
+    #[derive(Debug)]
+    pub(crate) struct SqsS3EventRecordIgnoredInvalidEvent<'a> {
```
JeanMertz, line 103: s/IngoredInvalidEvent/InvalidEventIgnored

?

src/internal_events/aws_s3.rs
```diff
@@ -0,0 +1,119 @@
...
+        pub bucket: &'a str,
+        pub key: &'a str,
+        pub kind: &'a str,
+        pub name: &'a str,
+    }
+
+    impl<'a> InternalEvent for SqsS3EventRecordIgnoredInvalidEvent<'a> {
+        fn emit_logs(&self) {
+            warn!(message = "Ignored S3 record as event kind that was not ObjectCreated.", %self.bucket, %self.key, %self.kind, %self.name);
```
JeanMertz, line 112: This warning seems fairly technical. I can't quite decipher what this is trying to tell. What does it mean for an "event kind" to not be "ObjectCreated"?

(I'm not only asking for myself, but also as a user, to perhaps guide us towards a better error message)

src/sources/aws_s3/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+    key: &str,
+) -> Option<Compression> {
+    content_encoding
+        .and_then(content_encoding_to_compression)
+        .or_else(|| content_type.and_then(content_type_to_compression))
+        .or_else(|| object_key_to_compression(key))
+}
+
+fn content_encoding_to_compression(content_encoding: &str) -> Option<Compression> {
+    use Compression::*;
```
JeanMertz, line 202: Just a minor nit, but for just two cases, and the fact that it requires the non-standard way of writing `Option::None`, I would forgo this line, and just write `Compression::Gzip` and `Compression::Zstd`.

src/sources/aws_s3/mod.rs
JeanMertz: I would personally use `use super::*` in tests, especially since basically all the use statements below are redundant to what's already imported in the parent module.

But, it's a minor nit, so I'm fine leaving this experiment in.

src/sources/aws_s3/sqs.rs
JeanMertz: Alternatively, limit the timeout secs config to a `u32` and cast it to `i64` safely.

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+                bucket: &s3_event.s3.bucket.name,
+                key: &s3_event.s3.object.key,
+                kind: &s3_event.event_name.kind,
+                name: &s3_event.event_name.name,
+            });
+            return Ok(());
+        }
+
+        // S3 has to send notifications to a queue in the same region so I don't think this will
+        // actually ever be it unless messages are being forwarded from one queue to another
```
JeanMertz, line 305:
```suggestion
        // actually ever be hit unless messages are being forwarded from one queue to another.
```

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+            Some(body) => {
+                let object_reader = super::s3_object_decoder(
+                    self.compression,
+                    &s3_event.s3.object.key,
+                    object.content_encoding.as_deref(),
+                    object.content_type.as_deref(),
+                    body,
+                );
+
+                // Record the read error saw to propagate up later so we avoid ack'ing the SQS
```
JeanMertz, line 347:
```suggestion
                // Record the read error seen to propagate up later so we avoid ack'ing the SQS
```

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+                // message
+                //
+                // String is used as we cannot take clone std::io::Error to take ownership in
```
JeanMertz, line 350:
```suggestion
                // String is used as we cannot clone std::io::Error to take ownership in
```

src/sources/aws_s3/sqs.rs
JeanMertz: I'm not really sure what the issue is here?

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+                // prefer duplicate lines over message loss. Future work could include recording
+                // the offset of the object that has been read, but this would only be relevant in
+                // the case that the same vector instance processes the same message.
+                let mut read_error: Option<std::io::Error> = None;
+                let lines: Box<dyn Stream<Item = Bytes> + Send + Unpin> = Box::new(
+                    FramedRead::new(object_reader, BytesDelimitedCodec::new(b'\n'))
+                        .map(|res| {
+                            res.map_err(|err| {
+                                read_error = Some(err);
+                                ()
```
JeanMertz, line 366:
```suggestion
```

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+                // FramedRead likely stops when it gets an i/o error but I found it more clear to
+                // show that we `take_while` there hasn't been an error
+                //
+                // This can result in objects being partially processed before an error, but we
+                // prefer duplicate lines over message loss. Future work could include recording
+                // the offset of the object that has been read, but this would only be relevant in
+                // the case that the same vector instance processes the same message.
+                let mut read_error: Option<std::io::Error> = None;
+                let lines: Box<dyn Stream<Item = Bytes> + Send + Unpin> = Box::new(
+                    FramedRead::new(object_reader, BytesDelimitedCodec::new(b'\n'))
```
JeanMertz, line 362: Should we be worried about files stored with CRLF newlines?

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+                        .map(|res| {
+                            res.map_err(|err| {
+                                read_error = Some(err);
+                                ()
+                            })
+                            .ok()
+                        })
+                        .take_while(|res| futures::future::ready(res.is_some()))
+                        .map(|r| r.unwrap()), // validated by take_while
```
JeanMertz, line 371: nit: instead of `unwrap` and a code comment, use `expect("validated by take_while")`

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+                );
+
+                let lines = match &self.multiline {
+                    Some(config) => Box::new(
+                        LineAgg::new(
+                            lines.map(|line| ((), line, ())),
+                            line_agg::Logic::new(config.clone()),
+                        )
+                        .map(|(_src, line, _context)| line),
```
JeanMertz, line 380:
```suggestion
                        .map(|(_, line, _)| line),
```

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+                    futures::future::ready(Some(Ok(event)))
+                });
+
+                let mut send_error: Option<futures01::sync::mpsc::SendError<Event>> = None;
+                out.send_all(Compat::new(Box::pin(stream)))
+                    .compat()
+                    .await
+                    .map_err(|err| {
+                        send_error = Some(err);
+                        ()
```
JeanMertz, line 409:
```suggestion
```

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+
+#[derive(Clone, Debug)]
+struct S3EventName {
+    kind: String,
+    name: String,
+}
+
+// https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html#supported-notification-event-types
+//
+// we could enums here, but that seems overly brittle as deserialization would  break if they add
```
JeanMertz, line 488:
```suggestion
// we could use enums here, but that seems overly brittle as deserialization would  break if they add
```

src/sources/aws_s3/sqs.rs
```diff
@@ -0,0 +1,543 @@
...
+// new event types or names
```
JeanMertz, line 489: Can't we still use enums, but have a catch all that at least logs if something is received that's unexpected (or just ignore "other")?

Still, this seems fine by me as well 👍

JeanMertz (approved):
Sweet!

Left some comments, nothing major, except for maybe Windows newlines.

docs/reference/components/sources/aws_s3.cue
jszwedko: I actually discovered that `required` cannot be used with `common` in our cue schema. I guess the thinking is that `required` implies `common`.

src/internal_events/aws_s3.rs
jszwedko: :+1: I'll clarify. This is basically intending to convey that it won't processing other object events (like object deletions). The bucket notifications should only be configured to forward `ObjectCreated` events, but this exists in case they accidentally configure it to send other types too.

docs/reference/components/sources/aws_s3.cue
JeanMertz: Ah I see. I believe I saw a few others that had `required` and `common` defined though?

src/sources/aws_s3/sqs.rs
jszwedko: Probably, but it seems like the other sources just deal with `\n`s (`file`, `udp`, `tcp`, `statsd`, `http`). I'll open a follow up issue to address this across the board if that's ok with you.

src/sources/aws_s3/sqs.rs
jszwedko: I actually like the `_src` and `_context` as it tells me what these variables would be.

src/sources/aws_s3/sqs.rs
jszwedko: :+1: I'll investigate this.

src/sources/aws_s3/sqs.rs
jszwedko: It just feels a little messy to me to record intermediate state rather than just streaming. We ended up having a conversation about it in discord though: https://discordapp.com/channels/742820443487993987/751107681217019944/771486734110162964

My conclusion from that conversation is that it'd be nice to use more `TryStream`s and have `send_all` stop at the first error and return it so that I could flow all errors and then handle them in one place.

src/sources/aws_s3/mod.rs
jszwedko: Yeah, I'm not 100% on it. Some of the imports are duplicated, but there is also a much smaller set of imports in the integration test and I found it easier to see where certain types were coming from rather than having to check first the `integration_tests` imports and then `super`'s imports.

docs/reference/components/sources/aws_s3.cue
jszwedko: Sorry, I mean if `required` is `true`. If it is `false`, `common` is required to be defined (if I remember correctly).

src/sources/aws_s3/sqs.rs
JeanMertz: I guess so. I'm used to having my editor show me that information when I want to know, and since we're dropping the values here, we "don't care what they are", so the extra naming seems like noise to me.

But it's a minor nit, and it's fine to leave them in 👍

docs/reference/components/sources/aws_s3.cue
JeanMertz: Ah yes, sorry, I had checked that, but I somehow thought I had seen a required `true` and `common` used together, but I was wrong!

src/sources/aws_s3/sqs.rs
jszwedko: Opened https://github.com/timberio/vector/issues/4857

src/sources/aws_s3/sqs.rs
jszwedko: True, I probably wouldn't lean on them as much if I configured `vim` better.

Though, it is still nice when browsing the code in Github or other "dumb" editors.

src/sources/aws_s3/sqs.rs
jszwedko: I came back around the other way on this one :smile: I'd like to, at least, recognize any new `ObjectCreated` events without needing to add new enum kinds. I'll just leave this as-is until we find a need.

src/sources/aws_s3/sqs.rs
jszwedko: :+1: good idea. It'll artificially limit the value, but I can't imagine anyone wanting to set a value higher than u32 anyway.

jszwedko:
@bruceg do you mind giving this one another look?

bruceg (approved):
Just one suggestion about a long function, nothing blocking though.

jszwedko:
Note for myself that I broke the deseralization of SQS messages in a recent commit due to relying on the `semvar`'s serde support which requires a fully formed version number (e.g. `2.1.0` rather than `2.1`). I opened https://github.com/steveklabnik/semver/issues/219 but I'll probably back this out to do custom parsing for now until upstream support is added (if it is at all).

jszwedko:
Windows failure is the memory issue that also exists on master.

PR #4781, opened 2020-10-27
chore(journald source): remove extra shutdown

fanatid:
Resolve comment https://github.com/timberio/vector/pull/4243/#discussion_r511314467
@juchiast I only have a question is it correct to stop source and return from function? Or we should block a function forever? (with something like `futures::future::pending().await`)

### DIFF
--- a/src/sources/journald.rs
+++ b/src/sources/journald.rs
@@ -182,7 +182,6 @@ impl JournaldConfig {
             include_units,
             exclude_units,
             channel: out,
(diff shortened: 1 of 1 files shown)


lukesteensen (approved):
If a source knows for sure that it cannot receive any more events then it should go ahead and shut down, so this seems correct to me.

juchiast (approved):
Look good!

PR #4783, opened 2020-10-27
chore(ci): Run test-shutdown and test-cli on PRs

jszwedko:
These seem to run quickly:

* ~50s for test-shutdown, most of which is the 30s sleep for kafka in
  the Makefile
* ~2s for test-cli

Given this, I think it is worth running them on PRs.

I just copied the miscellaneous test action from the master workflow
(moving test-behavior down).

Again, I think it'd be nice to have an action that ran the build setup
and cached the result to split up these tests into separate actions, but
this'll do for now. For one, the current behavior will have it fail as
soon as it hits an error where it'd be nice to run them all regardless.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -96,6 +96,25 @@ jobs:
       - run: echo "::add-matcher::.github/matchers/rust.json"
       - run: make slim-builds
       - run: make test
+
+  test-misc:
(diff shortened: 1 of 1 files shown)


jamtur01 (approved):
LGTM

jszwedko:
Failure fixed by: https://github.com/timberio/vector/pull/4782

PR #4784, opened 2020-10-27
enhancement(observability): Component counter metrics

leebenson:
This PR introduces a root `componentEventsProcessedTotal` subscription field, to mimic a pub/sub pattern for returning streaming `EventsProcessedTotal` metrics against a component name, whenever the current value is greater than any previously returned payload.

The practical use-case for this is to eliminate the need for polling via `query { components { ... } }`, and reduce the payload size / bandwidth by scoping return data to a singular metric with relevant changes.

This is done via a generalised `component_counter_metrics()` helper, returning an `impl Stream<Item = Metric>`, which returns a stream of changes (filtered by metric name) as a 'raw' `event::Metric` type, to be transformed upstream into the relevant GraphQL.

This reduces the surface of a resolver to something like this:

```rust
/// Component events processed metrics. Streams new data as the metric increases
async fn component_events_processed_total(
    &self,
    #[arg(default = 1000, validator(IntRange(min = "100", max = "60_000")))] interval: i32,
) -> impl Stream<Item = ComponentEventsProcessedTotal> {
    component_counter_metrics("events_processed_total", interval)
        .map(ComponentEventsProcessedTotal::new)
}
```

### `vector.toml` config

The query below uses the following config:

```toml
[api]
  enabled = true
  
[sources.gen]
  type = "generator"
  batch_interval = 5

[sinks.out]
  # General
  type = "blackhole"
  inputs = ["gen"]
  print_amount = 100
```

Note that this raises 2x new JSON events every 5 seconds, which will be relevant below.

### Query

```gql
subscription {
  componentEventsProcessedTotal(interval: 1000) {
    name
    metric {
      timestamp
      eventsProcessedTotal
    }
  }
}
```

### Results

Based on the above config + query, the following payloads are returned over a multiplexed WebSocket:

**1:**

```json
{
  "data": {
    "componentEventsProcessedTotal": {
      "metric": {
        "eventsProcessedTotal": 14,
        "timestamp": "2020-10-27T19:37:39.908953+00:00"
      },
      "name": "gen"
    }
  }
}
```

**2:**

```json
{
  "data": {
    "componentEventsProcessedTotal": {
      "metric": {
        "eventsProcessedTotal": 14,
        "timestamp": "2020-10-27T19:37:39.908984+00:00"
      },
      "name": "out"
    }
  }
}
```

**3:**

```json
{
  "data": {
    "componentEventsProcessedTotal": {
      "metric": {
        "eventsProcessedTotal": 16,
        "timestamp": "2020-10-27T19:37:44.909184+00:00"
      },
      "name": "gen"
    }
  }
}
```

**4:**

```json
{
  "data": {
    "componentEventsProcessedTotal": {
      "metric": {
        "eventsProcessedTotal": 16,
        "timestamp": "2020-10-27T19:37:44.909246+00:00"
      },
      "name": "out"
    }
  }
}
```

Note that despite passing `interval: 1000`, the timestamps are 5 seconds apart -- reflecting the `batch_interval` set in `vector.toml`.

Therefore, `interval` represents the minimum lower bound between payloads of the same component name. The upper bound is determined by event velocity.


leebenson:
Expanded this PR a little to include errors and bytes processed.

leebenson:
One thing I noticed switching to subscriptions locally using this work, is that `get_metrics()` can return metrics in any order.

I assumed this would have been ordered implicitly, by virtue of `emit!`ing as source (-> transform) -> sink.

This makes for slightly unexpected UX, in that sinks metrics can be sent back to the client first, making it look like a sink has processed an event before the source has sent it. Even though the delay is often just a few milliseconds between payloads, its enough to look unusual.

Since metrics are captured over an `interval`, it should be trivial to sort the results in-memory before emitting from the stream. This only needs to be done at the GraphQL level and should be low-cost. I'll make that switch now before this is ready to merge.

leebenson:
This is ready to review. I added a `get_metrics_sorted()` function, that returns a stream of results over an `interval`, sorted by source -> transform -> sink.

Since clients are connecting over TCP vs WS, these should always be received in order on the client.

It's not a huge deal -- the client can handle asynchronous packets in an arbitrary order - but I think this makes for more expected UX.

leebenson:
Discovered a bug this morning with `SubscriptionClient`, with this offending block:

```rust
impl Subscription {
    /// Returns a new Subscription, that is associated with a particular GraphQL query
    pub fn new(id: Uuid, client_tx: broadcast::Sender<Payload>) -> Self {
        let (tx, _) = broadcast::channel(1);
        Self { id, tx, client_tx }
    }
}
```

The buffer on the broadcast channel should have been set to `10`, not `1`. This had the effect of starving the sender whenever there was already a `Payload` en route to the client.

This wasn't obvious when metrics were sent in an arbitrary order (or when there were more than two components configured) but when there's exactly one source and one sink, this had the effect of skipping every 2nd stream, because the bounded channel would have been exhausted until it was fully emitted. The client would receive metrics from the source or sink, but never the other!

I fixed that, and added a test to assert the order is correct.

leebenson:
Closing this as a standalone PR and folding work over into the main `vector top` branch.

PR #4789, opened 2020-10-28
chore(ci): don't expand build failure link previews

JeanMertz:
To prevent links posted through the hook from expanding, since they never contain any useful information:

image


jszwedko (approved):
Very welcomed :smile:

PR #4795, opened 2020-10-28
feat(kubernetes_logs source): Add ingestion timestamp

MOZGIII:
This PR adds (an opt-in) filling of the ingestion timestamp to the `kubernetes_logs` source.

### DIFF
--- a/src/sources/kubernetes_logs/mod.rs
+++ b/src/sources/kubernetes_logs/mod.rs
@@ -94,6 +94,9 @@ pub struct Config {
     glob_minimum_cooldown_ms: usize,
 
     /// A field to use to set the timestamp when Vector ingested the event.
+    /// This is useful to compute the latency between important event processing
+    /// stages, i.e. the time delta between log line was written and when it was
+    /// processed by the `kubernetes_logs` source.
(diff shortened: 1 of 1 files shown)


lukesteensen (changes requested):
The `file` source should be adding a `timestamp` itself, but it looks like it does not do that currently. I'd rather we add this functionality there rather than for k8s specifically. I'm also not sure we need to allow overriding the key.

MOZGIII:
Kubernetes adds it's own `timestamp`, and adding the time when we picked up the message is not meaningful for most of the users (they want to see the time of the event as in the log file).

In the `file` source, this already happens, but it's hidden in this line: https://github.com/timberio/vector/blob/b036f347049844f500d2bcb3ebb460c3d6d41d2c/src/sources/file.rs#L310

jamtur01:
@MOZGIII What's the status here?

MOZGIII:
> @MOZGIII What's the status here?

I want to talk with @lukesteensen about this, seems like there was a bit of confusion here, but it should be easy to resolve.

lukesteensen (approved):
Ah, I see. That makes sense.

PR #4799, opened 2020-10-29
chore(file source): differentiate between types of fingerprints

lukesteensen:
This sets the stage for a transition from the old strategy of checksum fingerprinting to a new unified one. By differentiating between the different types of checksums on disk, we are able to intelligently upgrade existing checksums without breaking compatibility with existing stored checkpoints.

In this PR we change the in-memory representation of fingerprints and introduce a new format for writing them to disk. It also preps `FirstLineChecksum` to be the unified strategy by adding `ignored_header_bytes`. Next, we will add a path to transparently upgrade the `checksum` strategy to use `FirstLineChecksum` internally.

I did not give a huge amount of thought to the new format, since we will also be migrating away from the current filename-based storage. It should only be used as part of this transition period and can be removed in a future release.

I would recommend reading the commits one by one. Two simply move code around and are isolated to make the diff easier to follow.

### DIFF
--- a/lib/file-source/src/checkpointer.rs
+++ b/lib/file-source/src/checkpointer.rs
@@ -1,0 +1,105 @@
+use super::{FileFingerprint, FilePosition};
+use glob::glob;
+use std::{
+    collections::HashMap,
+    fs, io,
+    path::{Path, PathBuf},
+    time,
+};
+
+pub struct Checkpointer {
+    directory: PathBuf,
+    glob_string: String,
+    checkpoints: HashMap<FileFingerprint, FilePosition>,
+}
+
+impl Checkpointer {
+    pub fn new(data_dir: &Path) -> Checkpointer {
+        let directory = data_dir.join("checkpoints");
+        let glob_string = directory.join("*").to_string_lossy().into_owned();
+        Checkpointer {
+            directory,
+            glob_string,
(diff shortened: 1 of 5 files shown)


lib/file-source/src/checkpointer.rs
```diff
@@ -24,12 +24,53 @@ impl Checkpointer {
...
+    /// only allows [0-9a-f], we can use any character outside of that range as a magic byte
+    /// identifier for the newer formats.
     fn decode(&self, path: &Path) -> (FileFingerprint, FilePosition) {
+        use FileFingerprint::*;
+
         let file_name = &path.file_name().unwrap().to_string_lossy();
-        scan_fmt!(file_name, "{x}.{}", [hex FileFingerprint], FilePosition).unwrap()
+        match file_name.chars().next().expect("empty file name") {
+            'g' => {
+                let (c, pos) = scan_fmt!(file_name, "g{x}.{}", [hex u64], FilePosition).unwrap();
```
Hoverbear, line 56: I think `scan_fmt_some` would avoid the unwraps here, but it's just a nit

Hoverbear (approved):

MOZGIII (approved):
We are definitely moving in the right direction!

PR #4800, opened 2020-10-29
fix(config): support numbers in paths

### LINKED ISSUE #4794: `rename_fields` transform does not parse field names properly
This is branched of from `master` @ `2fdbd762ef80ebd325bfb4fc92a5bcb00adcdd2a`.

```toml
[transforms.rename]
  inputs = [".."]
  fields.k8s_logs_ingestion_timestamp = "..."
```

`k8s_logs_ingestion_timestamp` is not recognized properly.

JeanMertz:
This adds support for numbers in path segments.

Number support in remap was never added — and never caught in tests as we always use path names such as `foo` and `a` — but since remap hasn't officially shipped yet, this wasn't an issue.

However, now that the remap parser is used for path lookup (#4066) in the Vector config file itself, it _is_ exposed publicly, and so this needs to be addressed.

Note that this _only_ fixes numbers in paths for the Lookup code. The new remap-lang crate (#4695) has its own grammar file and already supports numbers in paths. I'm working on unifying the two (and fixing quoted paths in lookup, re https://github.com/timberio/vector/pull/4277#issuecomment-707771117) in a separate PR, but this should at least unblock releasing v0.11.

fixes #4794.

---

It would be good for @Hoverbear to weigh in on the current test-suite for the new lookup code, and how certain we are the tests cover all previous ways one could write paths to access event fields within the Vector config file.

It might be that we only accidentally missed the lacking support for numbers, or it might be that we need to add more tests to make sure the current path parsing rules are backward compatible with v0.10.

The intent in remap was always to be conservative in which characters we allow in regular path notation (`.foo`), so that we keep our options open towards using the restricted characters for other use-cases in the language, while simultaneously allowing an escape-hatch using quoted paths (`."@foo"`)

After this PR, the allowed set of characters is:

```
("a".."z" | "A".."Z" | 0..9 | "_" | "-" )+
```

Which strikes a balance between being able to use regular paths as often as possible while not restricting ourselves too much in the parser.

Anything else requires the quoted path notation to work.

Given that the same parser is used both for remap-lang and the regular path lookup in Vector configs, the above rule applies to both situations, even though we can be more lenient in what we allow in regular paths in Vector configs (given that we're not actually parsing the remap language itself, but only the path rule).

Are we certain the above covers all previous use-cases of path-based lookup using the old `PathComponent` system? @Hoverbear did a great job adding a lot of test-cases, but it's unclear to me if these cases are based on all the path syntax we currently support in v0.10.

**edit** I found a couple of backward-breaking changes in `Lookup`. See: https://github.com/timberio/vector/issues/4794#issuecomment-718693677

### DIFF
--- a/src/event/lookup/test.rs
+++ b/src/event/lookup/test.rs
@@ -159,7 +159,8 @@ fn parse_artifact(path: impl AsRef<Path>) -> std::io::Result<String> {
     let mut buf = Vec::new();
     test_file.read_to_end(&mut buf)?;
     let string = String::from_utf8(buf).unwrap();
-    Ok(string.trim_end_matches('\n').to_owned())
+    // remove trailing newline introduced by editors
+    Ok(string.trim_end().to_owned())
 }
 
 // This test iterates over the `tests/data/fixtures/lookup` folder and ensures the lookup parsed,
--- a/src/event/lookup/test.rs
+++ b/src/event/lookup/test.rs
@@ -159,7 +159,7 @@ fn parse_artifact(path: impl AsRef<Path>) -> std::io::Result<String> {
     let mut buf = Vec::new();
     test_file.read_to_end(&mut buf)?;
     let string = String::from_utf8(buf).unwrap();
-    Ok(string)
+    Ok(string.trim_end_matches('\n').to_owned())
 }
 
 // This test iterates over the `tests/data/fixtures/lookup` folder and ensures the lookup parsed,
--- a/src/mapping/parser/mod.rs
+++ b/src/mapping/parser/mod.rs
@@ -571,6 +571,13 @@ mod tests {
                 ))]),
             ),
             (
+                ".123 = 38",
+                Mapping::new(vec![Box::new(Assignment::new(
+                    "123".to_string(),
+                    Box::new(Literal::from(Value::from(38))),
+                ))]),
+            ),
+            (
                 ".foo = \"bar\"",
                 Mapping::new(vec![Box::new(Assignment::new(
                     "foo".to_string(),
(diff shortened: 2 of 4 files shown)


src/event/lookup/test.rs
```diff
@@ -159,7 +159,7 @@ fn parse_artifact(path: impl AsRef<Path>) -> std::io::Result<String> {
     let mut buf = Vec::new();
     test_file.read_to_end(&mut buf)?;
     let string = String::from_utf8(buf).unwrap();
-    Ok(string)
+    Ok(string.trim_end_matches('\n').to_owned())
```
JeanMertz, line 162: This is failing the CI on windows (`\r` 😅), will fix soon

src/mapping/parser/mod.rs
```diff
@@ -563,6 +563,13 @@ mod tests {
     #[test]
     fn check_parser() {
         let cases = vec![
+            (
+                ".v3ctor = \"bar\"",
+                Mapping::new(vec![Box::new(Assignment::new(
+                    "v3ctor".to_string(),
+                    Box::new(Literal::from(Value::from("bar"))),
+                ))]),
+            ),
```
FungusHumungus, line 572: It would be worth adding a test to see what would happen if the path field name consisted soley of numbers to check if the path gets parsed as a path or a number first.

What would the following parse as:

```
.123 = 38
```

JeanMertz: It would parse as a path, since we don't support that float notation, you'd need to write `0.123` for it to be parsed as a float.

I'll add a test to validate that behaviour 👍

jszwedko (approved):
:tada:

StephenWakely (approved):

Hoverbear (approved):

Hoverbear:
@JeanMertz Indeed I think we should add some more lookup tests! I agree. :)

PR #4801, opened 2020-10-29
enhancement(config): Remove error message for compression when using AWS ElasticSearch

smlgbl:
GZip compression works with the latest AWS ElasticSearch versions - see https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/release-notes.html


smlgbl:
For enabling gzip compression on older or already existing AWS ElasticSearch domains, see
https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/gzip.html#gzip-enable

steveteuber (approved):

jszwedko (approved):
Thanks @smlgbl !

I note that the default is no compression, so people will be able to opt-in.

PR #4802, opened 2020-10-29
feat(remap): add variable support

### LINKED ISSUE #4419: Local variables in the Remap language

JeanMertz:
This adds variable support to the remap language:

```rust
// variable assignment
$foo = "bar"

// path assignment and variable reference
.bar = $foo
```

This was removed in https://github.com/timberio/vector/pull/4695 (340842453a65875e3898f145bc332d2c4e104e76) to reduce the complexity of that PR and have it move forward without too much unrelated discussions.

Now that all the recent remap-lang changes have been merged, I'm re-introducing the concept of variables in this PR.

## Motivation

I'm happy to _not_ do this if there is opposition to this, but I'll posit my motivation for why I think it makes sense to add variables to the language.

**only manipulate the target object if you need to**

Remap language already supports assigning to paths, but if you want to hold a temporary value, right now you must assign and then delete that value to a field within the target object:

```rust
.url_parts = parse_url(.my_url) // note: we don't support "parse_url(.my_url).scheme"
.scheme = .url_parts.scheme
del(".url_parts")
```

Not only is this inconvenient, it also introduces two failure modes:

* You forget to delete the temporary field, polluting your events
* You assign a temporary value to an existing field, overwriting the previous value

**make conditions more readable**

Now that we support the Remap language in conditions (#4743), variables can help keep those conditions more readable:

```rust
$starts_with_date = match(trim(.timestamp), /^\d{4}-\d{2}-\d{2}.+/)
$other_check = to_bool(/* ... */)
$final_check = /* ... */ || /* ... */

$starts_with_date && $other_check && $final_check
```

**allow variable assignment in immutable contexts**

Not only that, but we plan on introducing an "immutable mode" for Remap (#4744), which would make the target object immutable. This does not apply to variables, since their mutability is "internal", meaning they don't mutate the outside world, and so there is no harm in being able to assign to variables in an immutable context.

Meaning, without variables, you wouldn't even be able to do the things we did in the first and second example because you can't assign values to target paths in an immutable context.

## Syntax

For a discussion and arguments for the `$foo = "bar"` syntax, see the related issue (https://github.com/timberio/vector/issues/4419). The TL;DR is that this syntax is similar to jq, and it prevents accidental variable assignments if you'd write `foo = "bar"` but intended to write `.foo = "bar"` (which would result in an error-free, but unexpected outcome of the script).

## Future Work

With this change, you can **store** values in variables, and **retrieve** those values, but you can't **traverse** containers:

```rust
.foo.bar = "baz" // works
$foo = .foo      // works

$foo             // returns { "bar": "baz" }
$foo.bar         // fails to compile
```

I think the latter should be valid syntax.

closes #4419

### DIFF
--- a/lib/remap-lang/src/lib.rs
+++ b/lib/remap-lang/src/lib.rs
@@ -134,6 +134,7 @@ mod tests {
     #[test]
     fn it_works() {
         let cases: Vec<(&str, Result<Option<Value>>)> = vec![
+            (r#".foo = null || "bar""#, Ok(Some("bar".into()))),
             (r#"$foo = null || "bar""#, Ok(Some("bar".into()))),
             // (r#".foo == .bar"#, Ok(Some(Value::Boolean(false)))),
             (
--- a/lib/remap-lang/src/lib.rs
+++ b/lib/remap-lang/src/lib.rs
@@ -134,7 +134,7 @@ mod tests {
     #[test]
     fn it_works() {
         let cases: Vec<(&str, Result<Option<Value>>)> = vec![
-            (r#".foo = null || "bar""#, Ok(Some("bar".into()))),
+            (r#"$foo = null || "bar""#, Ok(Some("bar".into()))),
             // (r#".foo == .bar"#, Ok(Some(Value::Boolean(false)))),
             (
                 r#".foo == .bar"#,
@@ -152,11 +152,11 @@ mod tests {
             //     r#".a.b.(c | d) == .e."f.g"[2].(h | i)"#,
             //     Ok(Some(Value::Boolean(false))),
             // ),
-            (".bar = true\n.foo = .bar", Ok(Some(Value::Boolean(true)))),
+            ("$bar = true\n.foo = $bar", Ok(Some(Value::Boolean(true)))),
             (
                 r#"{
-                    .foo = "foo"
-                    .foo = .foo + "bar"
+                    $foo = "foo"
+                    .foo = $foo + "bar"
                     .foo
                 }"#,
                 Ok(Some("foobar".into())),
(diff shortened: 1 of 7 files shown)


lib/remap-lang/src/parser.rs
```diff
@@ -381,18 +398,18 @@ mod tests {
             ),
             (
                 ".foo = to_string",
-                vec![" 1:8\n", "= expected if_statement, not, path, or block"],
+                vec![" 1:8\n", "= expected assignment, if_statement, not, or block"],
```
JeanMertz, line 401: More readable error reporting is something I'll be working on separately, so you can ignore the changes here and below.

lib/remap-lang/src/lib.rs
```diff
@@ -134,7 +134,7 @@ mod tests {
     #[test]
     fn it_works() {
         let cases: Vec<(&str, Result<Option<Value>>)> = vec![
-            (r#".foo = null || "bar""#, Ok(Some("bar".into()))),
+            (r#"$foo = null || "bar""#, Ok(Some("bar".into()))),
```
FungusHumungus, line 137: Can we keep the path tests and add the variable tests to ensure it works for both cases?

FungusHumungus (approved):
Nice. For the record, I'm all for variables in the language. 

It definitely does need to work with containers to be complete, though.

lib/remap-lang/src/lib.rs
JeanMertz: I'll add it, but note that these tests aren't that relevant right now. They mostly served me during development, the actual tests will be moved over from the Vector crate in the next couple of days.

jszwedko (approved):
:+1: I think this'll be very useful.

cc/ @lukesteensen since I think he's been pretty interested in features expanding the language.

When are we gonna allow:

```
$foo = "bar"
$bar = "hello"
$$foo // hello
```

:trollface:

bruceg (approved):

PR #4804, opened 2020-10-29
chore(transforms): Do not use lookups in transforms right now.

Hoverbear:
Triage #4273 before 0.11.


JeanMertz (approved):
Issue https://github.com/timberio/vector/issues/4794 contains the relevant tests that need to pass before `Lookup` can be used for path field lookups without breaking existing configurations.

Issue #4284, opened 2020-10-03
Benchmarking tests are broken


jamtur01:
The benchmarking tests are broken:

https://github.com/timberio/vector/runs/1201731003?check_suite_focus=true

```text
     Running target/release/deps/bench-8e5b7ca31779d8cb
Gnuplot not found, using plotters backend
Benchmarking pipe
Benchmarking pipe: Warming up for 3.0000 s
Benchmarking pipe: Collecting 10 samples in estimated 5.9976 s (20 iterations)
Benchmarking pipe: Analyzing
pipe                    time:   [287.14 ms 290.26 ms 293.82 ms]
                        thrpt:  [32.458 MiB/s 32.856 MiB/s 33.213 MiB/s]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild

Benchmarking pipe_with_tiny_lines
thread 'main' panicked at 'assertion failed: `(left == right)`
Benchmarking pipe_with_tiny_lines: Warming up for 3.0000 s
  left: `100000`,
 right: `0`', benches/bench.rs:151:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: bench failed
make: *** [Makefile:726: bench] Error 101
Error: Process completed with exit code 2.
```

PR #4809, opened 2020-10-29
chore(tests): Get benchmarks running again

jszwedko:
The commits in this PR can be viewed in order.

The high-level changes are:

* Fixes WASM and regular benchmarks so they run again
* Adds `--all-targets` to `make test` and `make test-wasm` so that the benchmarks are also tested. Currently this is kinda slow as we have many benchmarks where a single sample takes many seconds, but I intend to improve this by removing the internal batching from the samples where it isn't necessary. I was originally going to add `make` target specifically to run the benchmarks as tests, but I couldn't find way to do this (`cargo test --benches` runs the unit tests too; `cargo test --benches -- --bench` skips the unit tests but runs the benchmarks as benchmarks and not tests).
* Adds the WASM tests and benchmark tests to CI; these currently don't seem to be have run in CI anywhere

Closes: #4284


benches/lua.rs
```diff
@@ -59,109 +57,118 @@ fn add_fields(c: &mut Criterion) {
...
         })
-        .sample_size(10),
+        .throughput(Throughput::Elements(1)),
     );
 }
 
 fn field_filter(c: &mut Criterion) {
-    let num_events: usize = 100_000;
-
+    let num_events = 10;
```
Hoverbear, line 87: How would you feel about making this per-event?

benches/isolated_buffering.rs
```diff
@@ -30,16 +28,11 @@ impl Sink for NullSink {
 }
 
 fn benchmark_buffers(c: &mut Criterion) {
-    let num_lines: usize = 100_000;
+    let num_lines: usize = 10_000;
```
Hoverbear, line 31: How would you feel about reducing this batch size? It's a lot of events since criterion will run this in a loop with warmup.

benches/files.rs
```diff
@@ -12,7 +12,7 @@ use vector::{
 };
 
 fn benchmark_files_without_partitions(c: &mut Criterion) {
-    let num_lines: usize = 100_000;
+    let num_lines: usize = 10_000;
```
Hoverbear, line 15: Similarly, how would you feel about reducing this down? Criterion is already running in a fairly hot loop and this is a very coarse grained measurement.

benches/buffering.rs
```diff
@@ -1,25 +1,19 @@
...
 use tempfile::tempdir;
 use vector::test_util::{
-    next_addr, runtime, send_lines, start_topology, wait_for_tcp, CountReceiver,
+    next_addr, random_lines, runtime, send_lines, start_topology, wait_for_tcp, CountReceiver,
 };
 use vector::{buffers::BufferConfig, config, sinks, sources};
 
 fn benchmark_buffers(c: &mut Criterion) {
-    let num_lines: usize = 100_000;
+    let num_lines: usize = 10_000;
```
Hoverbear, line 11: I feel like I'm repeating myself, but this is a very coarse grained measurement and criterion is already doing several thousand iterations for us.

benches/bench.rs
```diff
@@ -460,6 +462,7 @@ fn benchmark_regex(c: &mut Criterion) {
 
 fn benchmark_complex(c: &mut Criterion) {
     let num_lines: usize = 100_000;
```
Hoverbear, line 464: "Ana repeats herself about coarse grained measurements"

benches/bench.rs
```diff
@@ -626,11 +630,17 @@ fn benchmark_complex(c: &mut Criterion) {
                         let output_lines_200 = output_lines_200.await.len();
                         let output_lines_404 = output_lines_404.await.len();
 
+                        // binomial distribution
+                        let sample_stdev = (output_lines_all as f64
```
Hoverbear, line 634: Could you use `debug_assert` for the test runs, so that they don't affect measurement in release? https://doc.rust-lang.org/std/macro.debug_assert.html

benches/batch.rs
```diff
@@ -89,8 +87,6 @@ fn partitioned_batching(
             },
         )
     })
-    .sample_size(10)
-    .noise_threshold(0.05)
     .throughput(Throughput::Bytes((num_events * event_len) as u64))
 }
```
Hoverbear, line 92: Below, in this file, theyre are a bench of benches calling this code that use high `num_events` numbers, and I repeat myself about coarse grained samples and criterion already doing this.

Hoverbear (commented):
Really great work so far. :) I'm fine merging this, but I would advise we either reduce our self-set batches and rely on Criterion's facilities a bit more. Maybe i'ts best to plan some future work for that though.

benches/lua.rs
```diff
@@ -9,167 +8,121 @@ use vector::{
...
-    let value_bytes_native = Bytes::from(value).into();
-    let value_bytes_v1 = Bytes::from(value).into();
-    let value_bytes_v2 = Bytes::from(value).into();
+    let mut group = c.benchmark_group("lua_add_fields");
+    group.throughput(Throughput::Elements(1));
 
-    c.bench(
-        "lua_add_fields",
-        Benchmark::new("native", move |b| {
-            b.iter_with_setup(
```
jszwedko, line 25: Switched away from `b.iter_with_setup` as:

* It's soft deprecated
* It was running this once per iteration which seemed unnecessary and possibly working against the purpose of the benchmark to measure the throughput

benches/lua.rs
```diff
@@ -9,167 +8,121 @@ use vector::{
...
     let key = "the key";
     let value = "this is the value";
 
-    let value_bytes_native = Bytes::from(value).into();
-    let value_bytes_v1 = Bytes::from(value).into();
-    let value_bytes_v2 = Bytes::from(value).into();
+    let mut group = c.benchmark_group("lua_add_fields");
+    group.throughput(Throughput::Elements(1));
 
-    c.bench(
```
jszwedko, line 22: Switched away from `c.bench` to creating a benchmark group given `c.bench` is deprecated: https://docs.rs/criterion/0.3.3/src/criterion/lib.rs.html#1634

benches/lua.rs
jszwedko: I did switch this to `iter_batched`, but I think this one still fits better running a small batch given we want to, sometimes, filter an event (10% of the time). I could augment it to be an individual event that is randomly filtered or not filtered, but I figured there could be some difference in execution time for a filtered event vs. an unfiltered event.

I could change this to be two benchmarks for the filter though, one where it always filters and one where it never filters.

benches/lua.rs
```diff
@@ -9,167 +8,121 @@ use vector::{
...
-        })
-        .with_function("v2", move |b| {
-            b.iter_with_setup(
-                || {
-                    let config = format!(
-                        r#"
-                        hooks.process = """
-                            function (event, emit)
-                                event.log['{}'] = '{}'
+    let mut benchmarks = [
```
jszwedko, line 20: The body of the iter loop for each benchmark was the same so I refactored this to be an array of (name, transform) tuples. I do get the feeling that this could be a little overboard though :smile: .

benches/lua.rs
```diff
@@ -9,167 +8,121 @@ use vector::{
...
+                transforms::lua::v2::Lua::new(&toml::from_str::<LuaConfig>(&config).unwrap())
+                    .unwrap(),
+            ) as Box<dyn Transform>
+        }),
+    ];
+
+    for (name, transform) in benchmarks.iter_mut() {
+        group.bench_function(name.to_owned(), |b| {
+            b.iter(|| {
+                let event = black_box(transform.transform(event.clone()).unwrap());
```
jszwedko, line 55: I added these `black_box` to ensure that the compiler doesn't optimize away the benchmark due to it discovering that the value is never actually used (except in debug mode where there is an assertion).

criterion automatically wraps inputs when `bench_with_input` is used and appears to wrap outputs of the iter closure. Instead of `black_box` I could write this as:

```
                 let event = transform.transform(event.clone()).unwrap();
                 debug_assert_eq!(event.as_log()[key], value.to_owned().into());
                 event
```

To return the event and have criterion wrap it in `black_box`.

benches/lua.rs
```diff
@@ -9,167 +8,121 @@ use vector::{
     Event,
 };
 
-fn add_fields(c: &mut Criterion) {
+fn bench_add_fields(c: &mut Criterion) {
```
jszwedko, line 11: It might be easiest to view this diff side-by-side.

benches/lua.rs
Hoverbear: I like it! You might prefer to use `AddFields::from(toml::from_str("my config block"))` for consistency's sake.

benches/lua.rs
```diff
@@ -9,167 +8,121 @@ use vector::{
...
-                    let transform = rt.block_on(async move {
-                        transforms::field_filter::FieldFilterConfig {
-                            field: "the_field".to_string(),
-                            value: "0".to_string(),
-                        }
-                        .build()
-                        .await
-                        .unwrap()
-                    });
+            rt.block_on(async move {
```
Hoverbear, line 78: Oh no rustfmt :(

jszwedko:
@Hoverbear I have some merge conflicts to resolve tomorrow, but I think this is ready for review in the meanwhile. I recommend checking out the last 4 commits. https://github.com/timberio/vector/pull/4809/commits/92aed118508201bb5e8317ab7478584f577805a8 is the big one. It was mostly mechanical, but resulted in a rather large diff. I can try to split it up if needed.

I don't think the result is perfect, but I do think it is an improvement. I'd still like to see the large end-to-end benchmarks improved (or possibly removed if they are not valuable).

I was hoping to get your thoughts on testing the WASM benchmarks separately too (https://github.com/timberio/vector/pull/4809/commits/9ee9aee51d6216ad46a8332694c9a05ac082cc0e). It's a bit awkward as it requires a few features aside from the `transforms-wasm` itself. I'm almost wondering if it'd be better to do one build that has WASM _and_ all the default features for testing rather than separating them out (maybe a `test-with-wasm` rather than `test-wasm`).

Hoverbear:
> It's a bit awkward as it requires a few features aside from the transforms-wasm itself. I'm almost wondering if it'd be better to do one build that has WASM and all the default features for testing rather than separating them out (maybe a test-with-wasm rather than test-wasm).

That totally works for me!

Hoverbear:
Noting that this does not appear to compile yet:

```bash
ana@autonoma:~/git/timberio/vector$ git log -p
commit 9ee9aee51d6216ad46a8332694c9a05ac082cc0e (HEAD -> fix-benchmarks, origin/fix-benchmarks)
Author: Jesse Szwedko <jesse@szwedko.me>
Date:   Wed Nov 4 22:21:25 2020 +0000
# ...
ana@autonoma:~/git/timberio/vector$ git pull
Already up to date.
ana@autonoma:~/git/timberio/vector$ make bench
cargo bench --no-default-features --features "default" | tee bench_output.txt
   Compiling vector v0.11.0 (/home/ana/git/timberio/vector)
error[E0601]: `main` function not found in crate `http`
   --> benches/http.rs:1:1
    |
1   | / use criterion::{criterion_group, BatchSize, BenchmarkId, Criterion, Throughput};
2   | | use futures::{compat::Future01CompatExt, TryFutureExt};
3   | | use hyper::{
4   | |     service::{make_service_fn, service_fn},
...   |
102 | |
103 | | criterion_group!(http, benchmark_http);
    | |_______________________________________^ consider adding a `main` function to `benches/http.rs`

error[E0601]: `main` function not found in crate `event`
  --> benches/event.rs:1:1
   |
1  | / use bytes::Bytes;
2  | | use criterion::{criterion_group, BatchSize, Criterion};
3  | | use serde_json::{json, Value};
4  | | use vector::{
...  |
98 | |
99 | | criterion_group!(event, benchmark_event);
   | |_________________________________________^ consider adding a `main` function to `benches/event.rs`

error[E0601]: `main` function not found in crate `files`
  --> benches/files.rs:1:1
   |
1  | / use bytes::Bytes;
2  | | use criterion::{criterion_group, BatchSize, Criterion, Throughput};
3  | | use futures::{compat::Future01CompatExt, stream, SinkExt, StreamExt};
4  | | use std::convert::TryInto;
...  |
93 | |
94 | | criterion_group!(files, benchmark_files_without_partitions);
   | |____________________________________________________________^ consider adding a `main` function to `benches/files.rs`

warning: unused import: `approx::assert_relative_eq`
 --> benches/bench.rs:3:5
  |
3 | use approx::assert_relative_eq;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  |

error[E0601]: `main` function not found in crate `batch`
   --> benches/batch.rs:1:1
    |
1   | / use bytes::Bytes;
2   | | use criterion::{criterion_group, Criterion, Throughput};
3   | | use futures::{compat::Future01CompatExt, future};
4   | | use futures01::{Sink, Stream};
...   |
174 | |     }
175 | | }
    | |_^ consider adding a `main` function to `benches/batch.rs`

For more information about this error, try `rustc --explain E0601`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...

For more information about this error, try `rustc --explain E0601`.

For more information about this error, try `rustc --explain E0601`.

For more information about this error, try `rustc --explain E0601`.
```

jszwedko:
Doh, I must have broken it. I'll resolve it, the merge conflicts, and add the `test-with-wasm` `make` target.

jszwedko:
Noting that this should be working now, but I commented out:

* The WASM benchmarks given the transform needs to be updated
* The Lua benchmarks pending https://github.com/timberio/vector/pull/4875 as Ana and I could not satisfy the borrow checker. We believe the changes in that PR should make it easier (specifically the addition of `Sync` to `FunctionTransform`).

Hoverbear:
Confirming #4875 will unblock the lua benches.

Makefile
```diff
@@ -785,18 +792,25 @@ $(WASM_MODULE_OUTPUTS): ### Build a specific WASM module
...
 .PHONY: bench
 bench: ## Run benchmarks in /benches
-	${MAYBE_ENVIRONMENT_EXEC} cargo bench --no-default-features --features "${DEFAULT_FEATURES}" | tee bench_output.txt
+	${MAYBE_ENVIRONMENT_EXEC} cargo bench --no-default-features --features "${DEFAULT_FEATURES}"
 	${MAYBE_ENVIRONMENT_COPY_ARTIFACTS}
 
+.PHONY: bench-all
+bench-all: ### Run default and WASM benches
+	# bench-all: $(WASM_MODULE_OUTPUTS) # TODO(jesse): renable after wasm transform builds again
+	# ${MAYBE_ENVIRONMENT_EXEC} cargo bench --no-default-features --features "${DEFAULT_FEATURES} wasm-benches"
```
Hoverbear, line 807: Probably better to put this after the normal benches.

Makefile
```diff
@@ -785,18 +792,25 @@ $(WASM_MODULE_OUTPUTS): ### Build a specific WASM module
...
+	${MAYBE_ENVIRONMENT_EXEC} cargo bench --no-default-features --features "${DEFAULT_FEATURES}"
```
Hoverbear, line 808: If you make these two commands use the same feature set you can probably skip a long compile time.

Hoverbear:
@jszwedko I spotted this error while testing locally: 
image

Hoverbear (approved):
Oh yes I'm so excited to be able to use these again. :) I think other than the issue I raised it looks good.

This is great:

image

Makefile
jszwedko: :+1: I mean to replace 808 with 807 rather than having both in there

jszwedko:
Just noting that https://github.com/timberio/vector/pull/4875 is required for renabling the wasm benchmarks too for similar reasons to the lua one where we are benchmarking both `FunctionTransform` and `TaskTransform` together.

Hoverbear:
@jszwedko I think that PR will merge soon but if you want we could merge this first and I can repair them in that PR.

Makefile
```diff
@@ -272,7 +272,14 @@ cross-image-%:
 
 .PHONY: test
 test: ## Run the unit test suite
-	${MAYBE_ENVIRONMENT_EXEC} cargo test --no-fail-fast --no-default-features --workspace --features ${DEFAULT_FEATURES} ${SCOPE} -- --nocapture
+	${MAYBE_ENVIRONMENT_EXEC} cargo test --no-fail-fast --no-default-features --workspace --features ${DEFAULT_FEATURES} ${SCOPE} --all-targets -- --nocapture
+
+.PHONY: test-all-components
+test-all-components: ## Test with all components enabled
```
Hoverbear, line 278: It seems our naming convention suggests `test-components` here.

Hoverbear:
@jszwedko I think my final review comment before approval is to ask if we must have `benches/default_benchmarks` and `benches/wasm_benchmarks`... Could we have `benches/` and `benches/wasm/`?

jszwedko:
Merged in #4875 and updated lua and wasm transforms. I'll wait for #4875 to merge before merging this one.

> @jszwedko I think my final review comment before approval is to ask if we must have `benches/default_benchmarks` and `benches/wasm_benchmarks`... Could we have `benches/` and `benches/wasm/`?

:+1: I'm fine with that.

jszwedko:
Moved the default benchmarks back up.

I'll double check this on Monday, but I think it should be good to go.

benches/wasm/mod.rs
```diff
@@ -57,51 +57,6 @@ pub fn protobuf(c: &mut Criterion) {
     });
 }
 
-pub fn drop(criterion: &mut Criterion) {
```
jszwedko: @Hoverbear I'm curious if you have any thoughts here. This benchmark currently hangs forever as `bench_group_transforms_over_parameterized_event_sizes` determines when an event has been processed by polling the output of the transform (which in this case, no event is emitted). I could not determine a good way to otherwise determine when an event has been fully processed by a transform.

Makefile
```diff
@@ -849,18 +856,23 @@ $(WASM_MODULE_OUTPUTS): ### Build a specific WASM module
...
 test-wasm: export TEST_LOG=vector=trace
 test-wasm: $(WASM_MODULE_OUTPUTS)  ### Run engine tests
-	${MAYBE_ENVIRONMENT_EXEC} cargo test wasm --no-fail-fast --no-default-features --features "wasm" --lib -- --nocapture
+	${MAYBE_ENVIRONMENT_EXEC} cargo test wasm --no-fail-fast --no-default-features --features "transforms-field_filter transforms-wasm transforms-lua transforms-add_fields" --lib --all-targets -- --nocapture
 
 ##@ Benching (Supports `ENVIRONMENT=true`)
 
 .PHONY: bench
 bench: ## Run benchmarks in /benches
-	${MAYBE_ENVIRONMENT_EXEC} cargo bench --no-default-features --features "${DEFAULT_FEATURES}" | tee bench_output.txt
```
Hoverbear, line 858: Why this? Criterion generates rich HTML reports in `target/criterion` no?

jszwedko: Do you mean the `tee`? That was already there as part of https://github.com/timberio/vector/pull/4630

I agree though, I'll revisit this as part of #4758

Hoverbear: Oh! Ok! I missed this!

benches/wasm/mod.rs
Hoverbear: Ohhhh, I think we need to just drop this for now then. :( With the task transform changes I think the only assertion we can make is that "nothing ever arrives" which seems to be passing right now.

jszwedko:
@Hoverbear I think this is _finally_ ready to be merged.

Hoverbear (approved):
Rest looks good :)

.github/workflows/master.yml
```diff
@@ -421,6 +421,7 @@ jobs:
       - cancel-previous
       - cross-linux
       - test-linux
+      - test-wasm
```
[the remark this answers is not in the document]
jszwedko: Oops, yep, thank you :smile:

Makefile
```diff
@@ -272,7 +272,14 @@ cross-image-%:
 
 .PHONY: test
 test: ## Run the unit test suite
-	${MAYBE_ENVIRONMENT_EXEC} cargo test --no-fail-fast --no-default-features --workspace --features ${DEFAULT_FEATURES} ${SCOPE} -- --nocapture
+	${MAYBE_ENVIRONMENT_EXEC} cargo test --no-fail-fast --no-default-features --workspace --features ${DEFAULT_FEATURES} ${SCOPE} --all-targets -- --nocapture
```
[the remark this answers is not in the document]
jszwedko: Hmm, would I not want to run those tests too though?

PR #4814, opened 2020-10-30
feat(remap): support else-if conditional expression

JeanMertz:
This updates the remap-lang grammar to extend if statements with support for else-if conditionals:

```rust
if false {
	0
} else if false { 
	1
} else if true {
	2
} else {
	3
}
// => 2
```

I had this lying around from my previous work on refactoring the remap-lang implementation, I pulled it out at the last minute to not overcomplicate the PR (similar to #4802), but after reading @FungusHumungus RFC in #4808 I figured it'd be worth to re-submit the PR.

This doesn't do anything that's described in @FungusHumungus's RFC _other than_ extending if statements to support `else if` cases, which is a trivial thing to add and something people probably expect to exist regardless.

As with variables (#4802), I'm happy to drop this proposal, but I think its table stakes for any language, even one that focusses on simplicity (or perhaps even more so, as it would be unintuitive to many not to have this feature). It could be said that the same can be achieved through nested if-else statements, but that can become tedious rather quickly.

In terms of implementation, I opted not to make the blocks optional for single expressions (which would have made the explicit support for `else if` in the grammar unnecessary), to not extend other parts of the language in this PR. We could still do that in a follow-up PR, but I'm hesitant to do that right now (or at all).

### DIFF
--- a/lib/remap-lang/src/lib.rs
+++ b/lib/remap-lang/src/lib.rs
@@ -169,6 +169,35 @@ mod tests {
                 "#,
                 Ok(Some(true.into())),
             ),
+            (r#"if false { 1 }"#, Ok(None)),
+            (r#"if true { 1 }"#, Ok(Some(1.into()))),
+            (r#"if false { 1 } else { 2 }"#, Ok(Some(2.into()))),
+            (r#"if false { 1 } else if false { 2 }"#, Ok(None)),
+            (r#"if false { 1 } else if true { 2 }"#, Ok(Some(2.into()))),
+            (
+                r#"if false { 1 } else if false { 2 } else { 3 }"#,
+                Ok(Some(3.into())),
+            ),
+            (
+                r#"if false { 1 } else if true { 2 } else { 3 }"#,
+                Ok(Some(2.into())),
+            ),
+            (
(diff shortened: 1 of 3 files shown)


jszwedko (approved):
:+1: I'd definitely expect this to be present given we have conditionals at all.

lib/remap-lang/src/lib.rs
```diff
@@ -169,6 +169,35 @@ mod tests {
...
+                Ok(Some(3.into())),
+            ),
+            (
+                r#"if false { 1 } else if true { 2 } else if false { 3 } else { 4 }"#,
+                Ok(Some(2.into())),
+            ),
+            (
+                r#"if false { 1 } else if false { 2 } else if false { 3 } else { 4 }"#,
+                Ok(Some(4.into())),
+            ),
```
FungusHumungus, line 200: Tiny point. I think it would be worth adding a test for:

```
if false { 1 } else { if false { 2 } else { 3 } } else { 5 }
```

If we ever decide to make surrounding blocks with '{}' optional, this test might help remind us how bad an idea that might be!

FungusHumungus (approved):
Nice! I really like how you've managed to keep this change contained within the parser.

lukesteensen (approved):

PR #4815, opened 2020-10-30
chore: Implement a workaround for symlinks on Windows at helm-dependencies script

### LINKED ISSUE #4592: Helm dependency update breaks symlinks on Windows
I checked our scripts on Windows, and under git-bash the Helm dependency update script breaks the symlink structure.

We should be able to fix this by special-casing our script to use Windows-specific `mklink` rather than `ln` if we detect we're running in Windows, which is also easy to do.

MOZGIII:
Closes #4592.

To do:

- [x] test on Windows
- [x] test on Linux

---

Note, my local Git version on Windows was worried that the symlinks were different at first, but after updating Git to the latest version I see no issues.


Hoverbear (approved):
I wish people would simply not use git bash. Sigh.

PR #4818, opened 2020-10-30
chore(deps): upgrade to 1.47 toolchain

jszwedko:
Includes clippy changes.


src/sinks/util/auto_concurrency/tests.rs
```diff
@@ -630,7 +630,7 @@ async fn all_tests() {
     // The first delay takes just slightly longer than all the rest,
     // which causes the first test to run differently than all the
     // others. Throw in a dummy delay to take up this delay "slack".
-    let _ = tokio::spawn(async move { delay_for(Duration::from_millis(1)) }).await;
```
jszwedko, line 633: @bruceg clippy caught the fact that we were not awaiting on the `delay_for` here. I'm wondering if that had something to do with the weird timing issue you saw?

I believe I fixed this correctly by simply returning the future from `delay_for` rather than including an `async {}` block; I felt this better than just adding an `.await` in the block.

src/kubernetes/stream.rs
```diff
@@ -35,6 +35,8 @@ where
             }
         }
         decoder.finish().map_err(|data| Error::UnparsedDataUponCompletion { data })?;
+
+        ()
```
jszwedko, line 39: @MOZGIII I had some trouble figuring out what this looked like expanded, but I think it is a clippy false-positive resulting from it thinking that we could simply do:

```
         decoder.finish().map_err(|data| Error::UnparsedDataUponCompletion { data })
```

instead, but that doesn't compile in this context. I resolved by simply returning `()` at the end instead, but I'm curious what you think.

jszwedko: The lint error was:

```
warning: unneeded `return` statement    
   --> src/kubernetes/stream.rs:21:5 
    |                                                
21  | /     try_stream! {                            
22  | |         let mut decoder: MultiResponseDecoder<T> = MultiResponseDecoder::new();
23  | |                                              
24  | |         debug!(message = "Streaming the HTTP body.");
...   |                                              
37  | |         decoder.finish().map_err(|data| Error::UnparsedDataUponCompletion { data })?;
38  | |     }                                        
    | |     ^ in this macro invocation              
    | |_____|                                        
    |                                                
    |                                                
   ::: /home/jesse/.cargo/registry/src/github.com-1ecc6299db9ec823/async-stream-impl-0.3.0/src/lib.rs:210:1
    |                                                
210 | / pub fn try_stream(input: TokenStream) -> TokenStream {
211 | |     let mut stmts = match parse_input(input) {
212 | |         Ok(x) => x,                                                                               
213 | |         Err(e) => return e.to_compile_error().into(),
...   |                                                                                                   
242 | |     .into()                                                                                       
243 | | }                                            
    | |_- in this expansion of `try_stream!`
    |                                                                                                     
    = note: `#[warn(clippy::needless_return)]` on by default                     
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
```

src/sinks/util/auto_concurrency/tests.rs
jszwedko: Rather, hmm, I see this line was added specifically for that issue. In this case I don't think it would have waited at all though so I'm curious what side-effects it has.

src/sinks/prometheus.rs
```diff
@@ -214,10 +214,7 @@ fn encode_metric_datum(
                 statistic: StatisticKind::Histogram,
             } => {
                 // convert distributions into aggregated histograms
-                let mut counts = Vec::new();
-                for _ in buckets {
-                    counts.push(0);
-                }
+                let mut counts = vec![0; buckets.len()];
```
bruceg, line 217: Fun, another merge conflict for me :)

src/sources/host_metrics.rs
```diff
@@ -867,10 +867,7 @@ mod tests {
 
     #[tokio::test]
     async fn filters_on_collectors() {
-        let all_metrics = HostMetricsConfig::default()
-            .capture_metrics()
-            .await
-            .collect::<Vec<_>>();
+        let all_metrics_count = HostMetricsConfig::default().capture_metrics().await.count();
```
bruceg, line 870: Nice, totally makes more sense than my copy.

src/sinks/util/auto_concurrency/tests.rs
bruceg: I'm getting a bit of a Twilight Zone feeling. I have no idea why both these lines are in there. I thought the delay required a separate task since the clock was paused from this task, but removing the `tokio::spawn` line completely produced the same result… or at least it does now.

bruceg (approved):
Looks good, and you can go one step further and remove that extraneous `tokio::spawn` line as well.

src/sinks/util/auto_concurrency/tests.rs
jszwedko: I know that feeling. Do you want me to remove the spawn and the comment above it?

bruceg: Correct, just drop the spawn, leave the real delay.

src/kubernetes/stream.rs
fanatid: @jszwedko I removed `()` which you added and do not see any clippy errors, am I doing something wrong?

jszwedko: @fanatid ah, you are right! This only shows up with `+nightly` which I was using to make use of `cargo clppy --fix`. I'll drop it.

PR #4831, opened 2020-11-01
enhancement(log_to_metric transform): Add `namespace` option

ktff:
Ref. #3896 

Updates transforms to expose `namespace` field. Only `log_to_metric` transform needed to be changed. 

Adds `namespace` option to `log_to_metric` transform which is templateable and sets `namespace` field of created metric.


jszwedko (approved):
Looks good!

PR #4835, opened 2020-11-02
enhancement(kubernetes platform): Add the ability to set conatiner ports at vector-agent Helm chart

### LINKED ISSUE #3808: Add configurable ports for k8s pods
### Current Vector Version

```
vector 0.11.0 (g50a9c3a x86_64-unknown-linux-musl 2020-09-10)
```

Chart version: vector-0.11.0-nightly-2020-09-10

### Use-cases

When utilizing the `prometheus` sink (my current use case of exposing `internal_metrics`) you'd need to expose a port to allow for the scraping of metrics. 

### Attempted Solutions

Hand editing of the helm templates

### Proposal

Allow users to pass a port in with the `values.yaml` file (either just the port number or entire spec) to expose the metrics endpoint

### References

N/A

_Note: I'll likely have a PR for this soon as I'll need to implement it for my current use_

MOZGIII:
This PR adds the ability to configure container ports. It is mostly useful to bind names to the exposed ports, which is required, in particular, if you're using `prometheus-operator`.

Closes #3808.

### DIFF
--- a/distribution/helm/vector-agent/templates/daemonset.yaml
+++ b/distribution/helm/vector-agent/templates/daemonset.yaml
@@ -58,6 +58,10 @@ spec:
             {{- with .Values.env }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
+          ports:
+            {{- with .Values.extraContainersPorts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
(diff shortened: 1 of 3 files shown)


spencergilbert:
This looks like it should satisfy the needs of the PodMonitor CRD - I'm not sure if we want/need to support just one or both of PodMonitors and ServiceMonitors.

MOZGIII:
We're focusing on just exposing the ports here, but it's a requirement for both `PodMonitor` and `ServiceMonitor`, as well as the `annotations` discovery mechanism. The root cause is that the `kubernets_sd_configs` in Prometheus doesn't work unless container `ports` are present.

jamtur01 (approved):
LGTM

PR #4836, opened 2020-11-02
enhancement(journald source): Restart journalctl on errors, save checkpoint on shutdown

### LINKED ISSUE #1626: Handle potential crash of journalctl
If the `journalctl` program crashes or otherwise exits, the `journald` source makes no attempt to restart it. While this has not been observed to happen in practice, this is still a potential failure case that should be caught and dealt with.

This will require some backoff however, as errors during start up (missing program or libraries, permissions issues, etc) will look the same as a crash and will end up rapidly restarting a program that can't run.

### LINKED ISSUE #4738: Save checkpoint before shutdown
Would be nice to have.

juchiast:
close #1626
close #4738 


src/sources/journald.rs
```diff
@@ -372,32 +293,134 @@ where
...
+    if unit.contains('.') {
+        unit.into()
+    } else {
+        format!("{}.service", unit)
+    }
 }
 
-fn decode_record(text: &str, remap: bool) -> Result<Record, JsonError> {
-    let mut record = serde_json::from_str::<JsonValue>(&text)?;
+fn decode_record(line: &[u8], remap: bool) -> Result<Record, JsonError> {
```
juchiast, line 422: Changed to bytes to avoid unnecessary clone (https://github.com/timberio/vector/pull/4243#discussion_r498610519).

src/sources/journald.rs
```diff
@@ -372,32 +297,134 @@ where
...
+///
+/// Code uses `start_journalctl` below,
+/// but we need this type to implement fake journald source in testing.
+type StartJournalctlFn = Box<
+    dyn Fn(
+            &Option<String>, // cursor
+        ) -> crate::Result<(BoxStream<'static, io::Result<Bytes>>, StopJournalctlFn)>
+        + Send
+        + Sync,
+>;
```
bruceg, line 343: What are the advantages of using this boxed function over the trait type it appears to be replacing? Could this be accomplished using a trait, which is (arguably) more idiomatic for Rust and Vector?

juchiast: Previously, it returns `Self` and `Self` needs to implement `Stream`, I found that hard to follow and would change it to return a BoxStream if I keep the trait.

Personally, I assume more than one type are being used when I see a trait or a generic type, but there was only one type/function here (except in testing). And to me, this looks more like a function than an object/interface, after all it was a trait with only one function.

jamtur01:
@juchiast What's the status here - can this be merged? Or further updates needed?

juchiast:
Hi @jamtur01, this can be merged, I'm waiting for a review.

fanatid (approved):
LGTM, but would like to wait for Bruce's approval.

bruceg (approved):

Issue #4247, opened 2020-10-01
Convert Service sink adapters to StreamSink


lukesteensen:
For many of our sinks that are based on `tower::Service` we use shared components like `PartitionBatchSink` and `BatchSink` to convert them to `futures01::Sink`. As part of our effort to remove `futures01`, we should convert these components to implement our new `StreamSink` abstraction.

This should hopefully let us simplify the code a bit, but the focus at first should be on converting them in the most straightforward way possible. As a follow-up, once they're converted and working, we can make a pass to simplify as much as possible and move away from the `poll`-style towards an `async`/`await` style.

Issue #2942, opened 2020-07-01
Migrate relevant sinks to Service sink variant


lukesteensen:
Blocked on #2938 

See [Architecture RFC](https://github.com/timberio/vector/blob/master/rfcs/2020-06-18-2625-architecture-revisit.md).

PR #4846, opened 2020-11-02
chore: add VectorSink::Sink

fanatid:
Part of #2942 #4247 

Only introduce variant and upgrade pulsar and kafka. Integration tests for kafka fail for me locally, will mark as draft for some time.

### DIFF
--- a/src/sinks/kafka.rs
+++ b/src/sinks/kafka.rs
@@ -189,19 +189,16 @@ impl Sink<Event> for KafkaSink {
     type Error = ();
 
     fn poll_ready(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        if matches!(self.poll_delivery_fut(cx), Poll::Pending)
-            && self.delivery_fut.len() >= SEND_RESULT_LIMIT
-        {
-            Poll::Pending
-        } else {
-            Poll::Ready(Ok(()))
+        match self.poll_delivery_fut(cx) {
+            Poll::Pending if self.delivery_fut.len() >= SEND_RESULT_LIMIT => Poll::Pending,
+            _ => Poll::Ready(Ok(())),
         }
     }
 
     fn start_send(mut self: Pin<&mut Self>, item: Event) -> Result<(), Self::Error> {
         assert!(
             self.delivery_fut.len() < SEND_RESULT_LIMIT,
-            "Expected to `poll_ready` called first."
+            "Expected `poll_ready` to be called first."
         );
 
         let topic = self.topic.render_string(&item).map_err(|missing_keys| {
--- a/src/sinks/kafka.rs
+++ b/src/sinks/kafka.rs
@@ -7,27 +7,32 @@ use crate::{
     sinks::util::encoding::{EncodingConfig, EncodingConfigWithDefault, EncodingConfiguration},
     template::{Template, TemplateError},
 };
-use futures::{compat::Compat, FutureExt};
-use futures01::{
-    future as future01, stream::FuturesUnordered, Async, AsyncSink, Future, Poll, Sink, StartSend,
-    Stream,
+use futures::{
+    channel::oneshot::Canceled, future::BoxFuture, ready, stream::FuturesOrdered, FutureExt, Sink,
+    Stream, TryFutureExt,
 };
 use rdkafka::{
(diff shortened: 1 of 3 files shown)


src/sinks/pulsar.rs
```diff
@@ -51,17 +49,13 @@ pub enum Encoding {
...
     encoding: EncodingConfig<Encoding>,
-    producer: Arc<Mutex<Producer<TokioExecutor>>>,
-    in_flight: FuturesUnordered<MetadataFuture<SendFuture, usize>>,
-    // ack
-    seq_head: usize,
-    seq_tail: usize,
-    pending_acks: HashSet<usize>,
+    producer: Option<Producer<TokioExecutor>>,
+    in_flight_prepare:
+        Option<BoxFuture<'static, (Producer<TokioExecutor>, Result<SendFuture, PulsarError>)>>,
```
fanatid, line 54: Will replace to `enum` tomorrow and write few explanation comments for a review. Today is too late :(

src/sinks/kafka.rs
```diff
@@ -141,95 +151,136 @@ impl KafkaSink {
...
 
-        let mut record = FutureRecord::to(&topic).key(&key).payload(&body[..]);
+        let seqno = self.seq_head;
+        self.seq_head += 1;
 
-        if let Some(Value::Timestamp(timestamp)) = item.as_log().get(log_schema().timestamp_key()) {
-            record = record.timestamp(timestamp.timestamp_millis());
-        }
+        let producer = Arc::clone(&self.producer);
+        self.delivery_fut.push(Box::pin(async move {
```
fanatid, line 216: `send_result` is not async and previously we returned `NotReady` if the queue was full. In the new `Sink` trait we can not return not ready, so for the case when sending blocked (queue full) I added `FuturesUnordered` where `send_result` called with 10ms delay. A number of futures limited by `SEND_RESULT_LIMIT` (set to 5).
I'm not completely sure that this is the best solution, maybe have 1 future with 1ms (or something like this) delay will be better.

src/sinks/pulsar.rs
```diff
@@ -123,75 +119,104 @@ impl PulsarSinkConfig {
...
+    fn start_send(mut self: Pin<&mut Self>, item: Event) -> Result<(), Self::Error> {
+        assert!(
+            matches!(self.state, PulsarSinkState::Ready(_)),
+            "Expected to `poll_ready` called first."
+        );
 
-    fn start_send(&mut self, item: Self::SinkItem) -> StartSend<Self::SinkItem, Self::SinkError> {
         let message = encode_event(item, &self.encoding).map_err(|_| ())?;
 
-        let producer = Arc::clone(&self.producer);
```
fanatid, line 155: I added an internal state instead `Arc<Mutex<T>>`.

src/sinks/kafka.rs
```diff
@@ -141,95 +151,136 @@ impl KafkaSink {
...
+            Poll::Pending
+        } else {
+            Poll::Ready(Ok(()))
+        }
+    }
+
+    fn start_send(mut self: Pin<&mut Self>, item: Event) -> Result<(), Self::Error> {
+        assert!(
+            self.delivery_fut.len() < SEND_RESULT_LIMIT,
+            "Expected to `poll_ready` called first."
```
bruceg, line 204:
```suggestion
            "Expected `poll_ready` to be called first."
```

bruceg (approved):
Looks ok on my end, but I don't know what the best approach for the semantics of this should be, particularly your questions about the librdkafka interface.

src/sinks/kafka.rs
lukesteensen: Just to follow up on this, I think what you have here is reasonable.

If we wanted to be more precise, we could try to tie the retries to the completion of the `in_flight` futures instead of a delay. So when one of those is complete we try again to send data, assuming that the completed request could have opened up space in the buffer. But that could be a bit tricky.

fanatid: Oh, right, that sounds more correct way. I agree that this can be tricky, but should I try to add it?

lukesteensen: It's probably worth spending a little bit of time to see how it would work, but not a lot if it turns out to be complicated.

fanatid: Added in https://github.com/timberio/vector/pull/5007

PR #4850, opened 2020-11-03
chore(tests): Fix starting influxdb under Podman

bruceg:
The networking is shared within a pod in Podman, and both the v1 influxd
agents attempt to use TCP port 8088 (RPC service). This configures the
v1 TLS agent's RPC port to TCP port 8089 to avoid the conflict.


jamtur01 (approved):
LGTM

PR #4858, opened 2020-11-03
chore: Make FunctionTransform take an immutable self.

Hoverbear:
As a result of #4684 we can now make our functiontransforms take an immutable self.


lukesteensen (commented):
@Hoverbear is this still something you think is desirable? From my perspective, I'd rather leave it as-is and not have to deal with the `RefCell`s.

Hoverbear:
Let's pan it.

PR #4859, opened 2020-11-04
enhancement(remap): display full error chain

JeanMertz:
This change makes sure we get more detailed error messages in Vector's log output whenever a remap-lang error occurs.

It changes this:

```
error for function "foo_func"
```

Into this:

```
remap error: error for function "foo_func": missing required argument "arg1" (position 0)
```

### DIFF
--- a/lib/remap-lang/src/error.rs
+++ b/lib/remap-lang/src/error.rs
@@ -1,4 +1,5 @@
 use crate::{expression, function, parser::Rule, value};
+use std::error::Error as StdError;
 use std::fmt;
 
 #[derive(thiserror::Error, Debug, PartialEq)]
@@ -40,8 +41,8 @@ impl From<&str> for Error {
     }
 }
 
-impl std::error::Error for Rule {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+impl StdError for Rule {
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
         None
     }
 }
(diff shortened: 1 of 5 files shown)


FungusHumungus (approved):
Nice.

lib/remap-lang/src/error.rs
```diff
@@ -105,3 +106,44 @@ impl fmt::Display for Rule {
...
+    }
+}
+
+impl fmt::Display for RemapError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut details = "remap error".to_owned();
+        let mut error: Option<&(dyn StdError + 'static)> = Some(&self.0);
+
+        while let Some(current) = error {
+            details = format!("{}: {}", details, current);
```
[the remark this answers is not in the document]
JeanMertz: Agreed. I'll make the change.

lib/remap-lang/src/error.rs
```diff
@@ -105,3 +106,44 @@ impl fmt::Display for Rule {
...
+pub struct RemapError(pub(crate) Error);
+
+impl StdError for RemapError {
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        Some(&self.0)
+    }
+}
+
+impl fmt::Display for RemapError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
jszwedko, line 120: It seems like this could have some unfortunate effects if the source also does this recursive printing where you'd get the same error messages duplicated (consider if `source` is itself a `RemapError`).

I'm wondering if the `fmt::Display` implementation should be decoupled from the "error stack printing".

Not a big deal though, I don't think it is likely to cause issues. It just stood out to me.

jszwedko (approved):
This is a great UX improvement :tada:

lib/remap-lang/src/error.rs
JeanMertz: The source of `RemapError` is the concrete type `remap::Error`, so there should be no way for that to happen.

What could happen, is that the parent error that stores this error iterates over the sources, which indeed would produce duplicate error messages.

One way to prevent that is to "erase" the nested errors here, by returning `None` for `source()`, effectively making this the only error object to be returned by remap.

I'm not sure what the right approach is here, but I agree, it's unlikely to happen, and if it does, we can change it then.

I almost feel like our tracing macro's should do this by default, when we use `%error`.

jszwedko: Gotcha, yeah, I see. This would only be an issue at a higher level. Agreed with punting for now.

PR #4860, opened 2020-11-04
enhancement(remap): support "g" regex flag

JeanMertz:
This re-adds support for the `g` regex flag, which was temporarily removed as part of the remap refactoring PRs.

### DIFF
--- a/lib/remap-lang/grammar.pest
+++ b/lib/remap-lang/grammar.pest
@@ -62,7 +62,7 @@ float   = ${
 }
 
 regex       = ${ "/" ~ regex_inner ~ "/" ~ regex_flags }
-regex_flags =  { ("i" | "x" | "m")* }
+regex_flags =  { ("g" | "i" | "x" | "m")* }
 
 // Other ----------------------------------------------------------------------
 
--- a/lib/remap-lang/src/function.rs
+++ b/lib/remap-lang/src/function.rs
@@ -1,6 +1,7 @@
 use crate::{Expression, Result, Value};
 use core::convert::TryInto;
 use std::collections::HashMap;
+use std::ops::Deref;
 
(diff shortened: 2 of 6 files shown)


jszwedko (commented):
I feel like I'm missing something :smile: Where is this flag used?

jszwedko:
Doh, never mind, I see #4861

jszwedko (approved):

jszwedko (changes requested):
Actually, I may have approved this too soon :smile: 

Thinking about it more, I'm wondering, instead of a `g` flag, if it might be better to just have `replace` take an optional parameter for `all` This would allow either regex or plaintext to be used to replace either just the first, or all matches.

As it stands, I think it's a little odd that `replace` (in #4861) replaces all, by default, when plaintext is used, but only the first match when a regex is used.

JeanMertz:
> Thinking about it more, I'm wondering, instead of a `g` flag, if it might be better to just have `replace` take an optional parameter for `all` This would allow either regex or plaintext to be used to replace either just the first, or all matches.

Yeah, I've been thinking the same.

I might actually prefer that solution as well. I'll close this PR, and update #4861. I might go ahead and add a `count` variable, with `-1` meaning all, and anything else meaning 0 or more. It would default to `-1`.

FungusHumungus:
> As it stands, I think it's a little odd that `replace` (in #4861) replaces all, by default, when plaintext is used, but only the first match when a regex is used.

I wonder if the replace should be changed to only replace the first match when plaintext is used. This is how it works in Javascript. To do a global match you need to fall back to regexes with the global flag set. So that would keep it consistent with JS.

An all parameter is also a nice idea.

JeanMertz:
Updated https://github.com/timberio/vector/pull/4861, closing this one. Thank you for the feedback @jszwedko 🎉

PR #4861, opened 2020-11-04
feat(remap transform): add `replace` function

JeanMertz:
This re-adds the `replace` function, which was initially introduced in #4703, but wasn't ported to the new set-up yet.

This is dependent on #4860.


jszwedko (approved):
Looks good!

It'd be nice to support using the capture groups in the replacement text, but we can do that as a follow-up. I'll open an issue.

jszwedko (changes requested):
Apologies for prematurely approving. I have the same question on this as I have over on #4860 about handling global vs. only one find/replace.

> Thinking about it more, I'm wondering, instead of a g flag, if it might be better to just have replace take an optional parameter for all This would allow either regex or plaintext to be used to replace either just the first, or all matches.
> 
> As it stands, I think it's a little odd that replace (in #4861) replaces all, by default, when plaintext is used, but only the first match when a regex is used.

https://github.com/timberio/vector/pull/4860#pullrequestreview-523423209

bruceg (approved):

jszwedko:
:+1:

PR #4864, opened 2020-11-04
fix(remap): unescape forward slash in regex

JeanMertz:
Rust's `regex` crate takes its regex as a string, such as:

```rust
regex::Regex::new("hello / world")
```

In Remap, a regex is surrounded by forward slashes, which requires forward slashes inside the regex to be escaped:

```javascript
match(.foo, /hello \/ world/)
```

Before, we'd pass the entire value `hello \/ world` to `Regex::new` as a string, which would result in an error, since it doesn't accept escaped forward slashes.

This PR unescapes those slashes, so that the regex can be properly parsed by the `regex` crate.

### DIFF
--- a/lib/remap-lang/src/lib.rs
+++ b/lib/remap-lang/src/lib.rs
@@ -131,6 +131,34 @@ mod tests {
     use super::*;
     use std::collections::HashMap;
 
+    #[derive(Debug, Clone)]
+    struct RegexPrinter;
+    impl Function for RegexPrinter {
+        fn identifier(&self) -> &'static str {
+            "regex_printer"
+        }
+
+        fn compile(&self, mut arguments: ArgumentList) -> Result<Box<dyn Expression>> {
+            Ok(Box::new(RegexPrinterFn(arguments.required_regex("value")?)))
+        }
+
+        fn parameters(&self) -> &'static [Parameter] {
+            &[Parameter {
+                keyword: "value",
+                accepts: |_| true,
+                required: true,
+            }]
+        }
+    }
+
+    #[derive(Debug, Clone)]
+    struct RegexPrinterFn(regex::Regex);
+    impl Expression for RegexPrinterFn {
+        fn execute(&self, _: &mut State, _: &mut dyn Object) -> Result<Option<Value>> {
+            Ok(Some(format!("regex: {:?}", self.0).into()))
+        }
+    }
+
     #[test]
     fn it_works() {
         let cases: Vec<(&str, Result<Option<Value>>)> = vec![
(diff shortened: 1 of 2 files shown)


lib/remap-lang/src/lib.rs
```diff
@@ -131,6 +131,34 @@ mod tests {
     use super::*;
     use std::collections::HashMap;
 
+    #[derive(Debug, Clone)]
+    struct RegexPrinter;
```
FungusHumungus, line 135: This is clever. Nice idea!

StephenWakely (approved):

lib/remap-lang/src/lib.rs
jszwedko, line 135: Is the program:

```
/foo/
```

Valid? It'd be unusual, but I'd expect that to simply return the regex as the result of the script.

Asking as, if it is, I'm wondering why we need this printer.

jszwedko (approved):
:+1: good catch.

In the future, it'd be nice to support other regex delimiters too to ease using regexs where the patterns have a lot of `/`s in them. Perhaps even allowing users to choose their own via an `r` prefix (like `r#foo/bar#`).

lib/remap-lang/src/lib.rs
JeanMertz: No, that's unsupported, regexes are only valid as function arguments (which is indeed why we need this printer function).

See the relevant grammar:

https://github.com/timberio/vector/blob/a87b132e6c00c56467a261a42a52be564962dc54/lib/remap-lang/grammar.pest#L22

jszwedko: Aha :+1: . I'm wondering if treating regexes as values would simplify things (or make them more complicated :smile: ).

For example, it would allow things like:

```
$my_regex = /foo/
replace(.message, $my_regex, "something")
```

I suppose an issue is that `event::Value` has no concept of regexes.

JeanMertz: To expand on that:

> but I'd expect that to simply return the regex as the result of the script

What `Value` type would you expect this to return? Introducing a special `Regex` value type seems unwanted, as we can't serialize that to anything meaningful in JSON for example, unless we wrap the regex in a string (but then we can't distinguish it from a regular string when deserializing that same JSON payload.

Alternatively, we could have it serialize to some kind of map with a `kind` field and a `regex` field, and in fact that's what we initially pursued, but we considered the use-case to be so exotic, meanwhile it could introduce unknown performance bottlenecks since you'd have to parse the regex every time a new event is captures that contains a "regex field".

So in the end, we decided it was best to just limit regexes to function arguments, and hard-code them into the language grammar, instead of representing them as a `Value`.

Hoverbear (approved):

PR #4875, opened 2020-11-04
chore(transforms): Add inline functions to pipeline

### LINKED ISSUE #4807: Add `Vec<dyn TransformFn>` to `Pipeline`
As part of [RFC 3791](https://github.com/timberio/vector/blob/master/rfcs/2020-10-06-3791-composing-components-pt-1.md) implement the steps:

* Add `Vec<dyn TransformFn>` to `Pipeline`.
* Implement composed sources as facades that prepend the relevant `TransformFn` to the `Pipeline` passed to `SourceConfig::build`.

Hoverbear:
Closes #4807 .


Hoverbear:
@jszwedko I mentioned this might fix your benching issue since it makes `FunctionTransform: Sync`

src/transforms/aws_ec2_metadata.rs
```diff
@@ -170,8 +172,21 @@ impl TransformConfig for Ec2Metadata {
     }
 }
 
-impl FunctionTransform for Ec2MetadataTransform {
-    fn transform(&mut self, output: &mut Vec<Event>, mut event: Event) {
+impl TaskTransform for Ec2MetadataTransform {
```
jszwedko, line 175: I actually think this would be fine left as a `FunctionTransform`. Cloning it would just clone the `ReadHandle` to the evmap. The state refresh happens in a separate tokio task spawned once at config build time.

src/pipeline.rs
```diff
@@ -1,12 +1,30 @@
...
 pub struct Pipeline {
     inner: Sender<Event>,
+    // We really just keep this around in case we need to rebuild.
+    #[derivative(Debug = "ignore")]
+    inlines: Vec<Box<dyn FunctionTransform>>,
+    enqueued: VecDeque<Event>,
+}
+
+impl Pipeline {
+    fn try_empty_queue(&mut self) -> Poll<(), <Self as Sink>::SinkError> {
```
jszwedko, line 19: Would this also make sense as `try_flush`? `try_empty_queue` kinda makes me think it might just drop things rather than forwarding them (though the return signature would make it clearer to me that it is sending).

jszwedko (commented):
This looks good! Just one note about the `Ec2MetadataTransform`.

Will there be later work to actually populate `inlines`?

src/transforms/aws_ec2_metadata.rs
Hoverbear: In this case, it's because evmap is not sync, so we'd have to wrap it in something sync. We'd have to wrap it in something like an `Arc<Mutex<T>>` to do this which would not be great for perf. 😭 

```bash
ana@autonoma:~/git/timberio/vector$ make check-clippy
cargo clippy --workspace --all-targets --features all-integration-tests -- -D warnings
    Checking vector v0.11.0 (/home/ana/git/timberio/vector)
error[E0277]: `std::cell::Cell<()>` cannot be shared between threads safely
   --> src/transforms/aws_ec2_metadata.rs:174:6
    |
174 | impl FunctionTransform for Ec2MetadataTransform {
    |      ^^^^^^^^^^^^^^^^^ `std::cell::Cell<()>` cannot be shared between threads safely
    | 
   ::: src/transforms/mod.rs:152:59
    |
152 | pub trait FunctionTransform: Send + dyn_clone::DynClone + Sync {
    |                                                           ---- required by this bound in `transforms::FunctionTransform`
    |
    = help: within `transforms::aws_ec2_metadata::Ec2MetadataTransform`, the trait `std::marker::Sync` is not implemented for `std::cell::Cell<()>`
    = note: required because it appears within the type `std::marker::PhantomData<std::cell::Cell<()>>`
    = note: required because it appears within the type `evmap::ReadHandle<std::string::String, bytes::Bytes>`
    = note: required because it appears within the type `transforms::aws_ec2_metadata::Ec2MetadataTransform`

For more information about this error, try `rustc --explain E0277`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error[E0277]: `std::cell::Cell<()>` cannot be shared between threads safely
   --> src/transforms/aws_ec2_metadata.rs:174:6
    |
174 | impl FunctionTransform for Ec2MetadataTransform {
    |      ^^^^^^^^^^^^^^^^^ `std::cell::Cell<()>` cannot be shared between threads safely
    | 
   ::: src/transforms/mod.rs:152:59
    |
152 | pub trait FunctionTransform: Send + dyn_clone::DynClone + Sync {
    |                                                           ---- required by this bound in `transforms::FunctionTransform`
    |
    = help: within `transforms::aws_ec2_metadata::Ec2MetadataTransform`, the trait `std::marker::Sync` is not implemented for `std::cell::Cell<()>`
    = note: required because it appears within the type `std::marker::PhantomData<std::cell::Cell<()>>`
    = note: required because it appears within the type `evmap::ReadHandle<std::string::String, bytes::Bytes>`
    = note: required because it appears within the type `transforms::aws_ec2_metadata::Ec2MetadataTransform`

For more information about this error, try `rustc --explain E0277`.
error: build failed
make: *** [Makefile:812: check-clippy] Error 101
```

jszwedko: Aha, I see, hmm....

It looks like they provide a `ReadHandleryFactory` that is `Sync`: https://docs.rs/evmap/10.0.2/evmap/struct.ReadHandleFactory.html. Generating new `ReadHandle`s from that does require a lock though.

I guess given those limitations of evmap, this seems :+1:

jszwedko (approved):

src/pipeline.rs
```diff
@@ -17,30 +35,127 @@ impl Sink for Pipeline {
...
+            let mut new_working_set = Vec::with_capacity(working_set.len());
+            for event in working_set.drain(..) {
+                inline.transform(&mut new_working_set, event);
+            }
+            core::mem::swap(&mut new_working_set, &mut working_set);
+        }
+        self.enqueued.extend(working_set);
+
+        self.try_flush()?;
+        Ok(AsyncSink::Ready)
```
lukesteensen, line 51: If I'm reading this right it seems like we'll never return `NotReady` from `start_send`. I think we need to bound the size of `enqueued` and try to flush and `poll_complete` existing events before accepting and transforming more.

lukesteensen (changes requested):

src/pipeline.rs
Hoverbear: Agree!

Hoverbear: I am not sure it is correct to cap the size of `enqueued` given that we know it will only be holding 1 event's work of processed messages at any given time and we are not permitted to block or discard messages here. Could you review https://github.com/timberio/vector/pull/4875/commits/4966d1fd5104b099b36e69654853042e2c1bbe07 see if this resolves your concerns?

src/pipeline.rs
```diff
@@ -17,30 +35,131 @@ impl Sink for Pipeline {
...
+        debug_assert!(self.enqueued.is_empty());
         self.inner.poll_complete()
     }
 }
 
 impl Pipeline {
     #[cfg(test)]
-    pub fn new_test() -> (Self, Receiver<Event>) {
-        Self::new_with_buffer(100)
+    pub fn new_test(inlines: Vec<Box<dyn FunctionTransform>>) -> (Self, Receiver<Event>) {
```
Hoverbear, line 67: https://github.com/timberio/vector/pull/4941#discussion_r520944681

> I might prefer we leave this method as-is and add another helper if/when we need it to include inlines. That way we don't need to change everywhere we use this without setting any.

I think this is fine, I'll add a builder.

lukesteensen (approved):
Looks like a few red CI checks still, but other than those this looks good to me!

Hoverbear:
Oy let me fix that!

PR #4881, opened 2020-11-05
enhancement(prometheus source): Add TLS and authentication options

### LINKED ISSUE #3797: Prometheus source to support TLS configuration and authorization

bruceg:
Closes #3797 
Supercedes #4878 

The `auth` config metadata is shared with the Clickhouse sink. Some better way of sharing that stanza would be good, but I am unclear what is the best option for that.

### DIFF
--- a/docs/reference/components/sources/prometheus.cue
+++ b/docs/reference/components/sources/prometheus.cue
@@ -31,6 +31,13 @@ components: sources: prometheus: {
 					ssl: "optional"
 				}
 			}
+			tls: {
+				enabled:                true
+				can_enable:             false
+				can_verify_certificate: true
+				can_verify_hostname:    true
+				enabled_default:        false
+			}
 		}
 		multiline: enabled: false
 	}
--- a/docs/reference/components/sources/prometheus.cue
+++ b/docs/reference/components/sources/prometheus.cue
@@ -76,6 +76,52 @@ components: sources: prometheus: {
 				unit:    "seconds"
 			}
 		}
+		auth: {
+			common:      false
+			description: "Options for the authentication strategy."
+			required:    false
+			warnings: []
+			type: object: {
+				examples: []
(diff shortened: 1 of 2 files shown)


ktff (approved):
Looks good.

> The auth config metadata is shared with the Clickhouse sink. Some better way of sharing that stanza would be good

Maybe how `_tls_connect` from `componenets.cue` is used.

PR #4883, opened 2020-11-05
fix(blackhole sink): Make print_amount an optional option with a default of 1000

### LINKED ISSUE #3886: make `blackhole` sink's `print_amount` field optional
While testing out a local Vector configuration in https://github.com/timberio/vector/issues/230#issuecomment-692727213, I ran into https://github.com/timberio/vector/issues/3885.

That got me thinking, I wonder if we should consider making `print_amount` an optional field. Its purpose is orthogonal to what the actual sink is supposed to do, and given that this sink seems like a good fit for quickly testing something, the less configuration it requires, the better.

jamtur01:
Closes #3886 

### DIFF
--- a/src/sinks/blackhole.rs
+++ b/src/sinks/blackhole.rs
@@ -18,7 +18,7 @@ pub struct BlackholeSink {
 }
 
 #[derivative(Default)]
-#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[derive(Clone, Debug, Default, Derivative, Deserialize, Serialize)]
 pub struct BlackholeConfig {
     #[derivative(Default(value = "1000"))]
     #[serde(default)]
--- a/src/sinks/blackhole.rs
+++ b/src/sinks/blackhole.rs
@@ -18,7 +18,7 @@ pub struct BlackholeSink {
 }
 
 #[derivative(Default)]
-#[derive(Clone, Debug, Default, Derivative, Deserialize, Serialize)]
+#[derive(Clone, Debug, Derivative, Deserialize, Serialize)]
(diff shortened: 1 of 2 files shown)


src/sinks/blackhole.rs
```diff
@@ -17,11 +17,16 @@ pub struct BlackholeSink {
     acker: Acker,
 }
 
-#[derive(Deserialize, Serialize, Debug, Clone)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
 pub struct BlackholeConfig {
+    #[serde(default = "default_print_amount")]
```
jszwedko, line 22: @bruceg recently pointed me to our use of [`Derivative`](https://mcarton.github.io/rust-derivative/) elsewhere which can be used to write this something like:

```
#[derivative(Default)]
#[derive(Clone, Debug, Default, Deserialize, Serialize)]
pub struct BlackholeConfig {
    #[derivative(Default(value="1000"))]
    #[serde(default)]
    pub print_amount: usize,
}
```

To save the need to define a `const fn`.

jszwedko (approved):
:+1: this has always bothered me too.

src/sinks/blackhole.rs
jamtur01: Oh neat. Thanks.

jamtur01: So in order to use the value of `print_amount` in internal metrics, I needed to appease serde anyway so it's now a mix of the above.

jszwedko: Aha, I thought `#[serde(default)]` would take care of that.

PR #4886, opened 2020-11-05
feat(observability): `vector top`, release candidate

leebenson:
This is a release candidate for `vector top`, v1.

It introduces several improvements over earlier work, including:

- GraphQL queries after first paint, have been switched to GraphQL subscriptions. This enables the server to push streaming payloads for events and topology changes, instead of re-polling over the (default) 500ms interval.

- A `.graphqlconfig` has been added to the root, to improve `.graphql` file editing experience in IDEs that support it.

- The pub/sub broker implementation has been simplified to use channels, reducing the footprint of external libs such as slab and anymap.

- Fixed a bug with `SubscriptionClient`, that would silently drop payloads on streams receiving data in quick succession.

- Reduced lock contention by switching the dashboard to channels. I suspect this is why the CPU has dropped from 50%+ in a release build on a 100ms stream, to ~2-3% locally. Based on this, I was able to remove the static 250ms redraw and move back to redrawing on state changes, which doesn't seem to have had a negative impact on CPU.

- Added some new tests around notifications for adding/deleting components.

It's currently in draft due to one remaining issue with an intermittant misfire of the initial draw that I spotted on a new release build. Investigating...


src/top/cmd.rs
```diff
@@ -80,20 +36,40 @@ pub async fn cmd(opts: &super::Opts) -> exitcode::ExitCode {
         }
     }
 
-    // Create initial topology, to be shared by the API client and dashboard renderer
-    let state = Arc::new(WidgetsState::new(url, ComponentsState::new()));
+    // Create a metrics state updater
+    let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
```
Hoverbear, line 40: Do you think we could pick a reasonable bound size for this?

Hoverbear:
Confirming this works on Windows (I think the lack of value in bytes processed is expected here):
image
image


I did note during testing that if I have `vector top` connected, terminate the vector it's connected to, then restart it, the `top` dashboard won't refresh.

leebenson:
> I did note during testing that if I have vector top connected, terminate the vector it's connected to, then restart it, the top dashboard won't refresh.

Yeah, that's an unfortunate side-effect of the lack of any connection management, at present. There needs to be an attempt at reconnection, potentially a back-off strategy for retries, and some way to reflect the connection status in the dashboard. 

All of those things are on my radar, but probably for the 1.5 release.

We'll be more resilient out-the-box with the web UI because those things have at least some level of library support baked in, whereas my home-brew `SubscriptionClient` is pretty bare-bones atm.

leebenson:
Ok, I've taken this out of review now.

I've spent almost the entirety of the last 3 days trying to hunt down a heisenbug that prevented a certain arrangement of subscriptions from firing at all. Depending on the arrangement of those subscriptions, it happened between ~10-50% of the time.

I was finally able to replicate it properly 100% of the time, although the exact cause of it is still a mystery. @sunli829, I could use your eyes on this, if you have a little time to look -- thanks in advance!

Here's the issue...

We have multiple subscription fields which either:

1. Collect metrics.
2. Subscribe to a pub/sub pattern.

Collecting metrics works perfectly. We can get events/errors/bytes processed, and multiplex those requests over a single client. Any interval works -- from as low as 50ms, to no upper limit. Metrics never skip a beat, irrespective of how many of them we request. All good there.

However, pub/sub is a different story. We have two subscriptions -- `componentAdded` and `componentRemoved` -- which subscribe to changes made to topology. I've tried multiple strategies... a 'simple broker' pattern, and (more recently, and the approach taken in this PR) to just use a Tokio broadcast channel.

When a _dedicated_ client connection is made to the GraphQL API server and **one** of these is requested, everything works perfectly well. We receive metrics and component changes. However, when we attempt to multiplex one of these pub/sub requests with metrics, it only works 10-50% of the time. When we attempt to use both componentAdded and componentRemoved together... it fails 100% of the time.

By "fail", I mean that after sending the request -- we get nothing back from the API server. It prevents any further payloads against that same connection.

At first, I suspected it was an issue with my home-brew `SubscriptionClient`, or possibly the tungstenite WebSocket lib. So I broke that out into a separate, smaller app (for faster compilation) and stripped back everything to just observe raw socket messages. The same problem occurred. With a certain combination of subscription requests, _any_ payload ceases to be returned. With other combinations, payloads would _occasionally_ be returned (and work reliably ongoing, when they were). In other invocations, no payload was returned at all -- hinting at some kind of race condition.

I then suspected it could be my pub-sub approach. I tried multiple approaches/libs, including `bus`, as well as sending a `tokio::sync::broadcast::Receiver` through context. They each encountered the same issue.

After much experimentation with `websocat` directly (taking my `SubscriptionClient` out of the equation), I was able to replicate the issue 100% of the time with the following payload:

```bash
➜  vector git:(leebenson/top-subscriptions) websocat ws://127.0.0.1:8686/graphql
{"id":"b8869530-853f-4feb-bb1d-fd2ede9582f1","type":"start","payload":{"operationName":"ComponentAddedSubscription","query":"subscription ComponentAddedSubscription {\n  componentAdded {\n    __typename\n    name\n  }\n}\n","variables":null}}
{"id":"83a9e880-15e6-4c82-890a-fdac34be3101","type":"start","payload":{"operationName":"ComponentRemovedSubscription","query":"subscription ComponentRemovedSubscription {\n    componentRemoved {\n        __typename\n        name\n    }\n}\n","variables":null}}
{"id":"c7186e5d-d333-4e9b-8153-56d0e8f628ed","type":"start","payload":{"operationName":"ComponentEventsProcessedTotalSubscription","query":"subscription ComponentEventsProcessedTotalSubscription($interval: Int!) {\n  componentEventsProcessedTotal(interval: $interval) {\n    name\n    metric {\n      eventsProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
{"id":"2d7ce3e3-c336-473e-8c4e-8b51dd42948a","type":"start","payload":{"operationName":"ComponentBytesProcessedTotalSubscription","query":"subscription ComponentBytesProcessedTotalSubscription($interval: Int!) {\n  componentBytesProcessedTotal(interval: $interval) {\n    name\n    metric {\n      bytesProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
```

**Sending those four subscriptions in a single client causes the GraphQL server to prevent all payloads, 100% of the time.**

The workaround is to break them out into separate clients:

```bash
➜  vector git:(leebenson/top-subscriptions) websocat ws://127.0.0.1:8686/graphql
{"id":"b8869530-853f-4feb-bb1d-fd2ede9582f1","type":"start","payload":{"operationName":"ComponentAddedSubscription","query":"subscription ComponentAddedSubscription {\n  componentAdded {\n    __typename\n    name\n  }\n}\n","variables":null}}
```

```bash
{"id":"83a9e880-15e6-4c82-890a-fdac34be3101","type":"start","payload":{"operationName":"ComponentRemovedSubscription","query":"subscription ComponentRemovedSubscription {\n    componentRemoved {\n        __typename\n        name\n    }\n}\n","variables":null}}
{"id":"c7186e5d-d333-4e9b-8153-56d0e8f628ed","type":"start","payload":{"operationName":"ComponentEventsProcessedTotalSubscription","query":"subscription ComponentEventsProcessedTotalSubscription($interval: Int!) {\n  componentEventsProcessedTotal(interval: $interval) {\n    name\n    metric {\n      eventsProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
```

etc... works fine.

@sunli829 - any ideas why that might be?

Again, we can multiplex _some_ combination of subscriptions. Any number of metrics together is fine. One of the pub/sub subscriptions + metrics is _sometimes_ fine (10-50% success.) Two pub/subs + metrics always fails.

As a workaround for now, I have updated this PR to create a separate client for each subscription. That's not ideal-- it's 4 long-lived connections where only should be required, and it'll complicate the web client (urql) where, by default, a single client is assumed for all subscriptions.

Ideally we can find the root cause of this and switch back to using just a single connection, but I could use your help @sunli829 to ascertain whether this is something in async-graphql, or whether (more likely) it's something I've introduced.

sunli829:
@leebenson Would you please provide a minimal use case that can reproduce the problem?

leebenson:
Here's a demo of what's happening: https://www.loom.com/share/9e969a322bab4719a882d85da9c799b5

@sunli829 -- to replicate, I'd recommend cloning this branch, and using the following `vector.toml`:

```toml
[api]
  enabled = true
  
[sources.gen1]
  type = "generator"
  batch_interval = 0.01

[transforms.gen1_add_field]
  # General
  type = "add_fields"
  inputs = ["gen1"]

  # Fields
  fields.message = "hello!"

[sinks.out]
  # General
  type = "blackhole"
  inputs = ["gen1_add_field"]
  print_amount = 100000
```

Fire Vector up, and then in another terminal window, use [websocat](https://github.com/vi/websocat):

```bash
➜ websocat ws://127.0.0.1:8686/graphql
{"id":"c7186e5d-d333-4e9b-8153-56d0e8f628ed","type":"start","payload":{"operationName":"ComponentEventsProcessedTotalSubscription","query":"subscription ComponentEventsProcessedTotalSubscription($interval: Int!) {\n  componentEventsProcessedTotal(interval: $interval) {\n    name\n    metric {\n      eventsProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
{"id":"2d7ce3e3-c336-473e-8c4e-8b51dd42948a","type":"start","payload":{"operationName":"ComponentBytesProcessedTotalSubscription","query":"subscription ComponentBytesProcessedTotalSubscription($interval: Int!) {\n  componentBytesProcessedTotal(interval: $interval) {\n    name\n    metric {\n      bytesProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
```

You should see metrics coming back every second.

Now, try this -- this should **fail**:

```bash
➜ websocat ws://127.0.0.1:8686/graphql
{"id":"b8869530-853f-4feb-bb1d-fd2ede9582f1","type":"start","payload":{"operationName":"ComponentAddedSubscription","query":"subscription ComponentAddedSubscription {\n  componentAdded {\n    __typename\n    name\n  }\n}\n","variables":null}}
{"id":"83a9e880-15e6-4c82-890a-fdac34be3101","type":"start","payload":{"operationName":"ComponentRemovedSubscription","query":"subscription ComponentRemovedSubscription {\n    componentRemoved {\n        __typename\n        name\n    }\n}\n","variables":null}}
{"id":"c7186e5d-d333-4e9b-8153-56d0e8f628ed","type":"start","payload":{"operationName":"ComponentEventsProcessedTotalSubscription","query":"subscription ComponentEventsProcessedTotalSubscription($interval: Int!) {\n  componentEventsProcessedTotal(interval: $interval) {\n    name\n    metric {\n      eventsProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
{"id":"2d7ce3e3-c336-473e-8c4e-8b51dd42948a","type":"start","payload":{"operationName":"ComponentBytesProcessedTotalSubscription","query":"subscription ComponentBytesProcessedTotalSubscription($interval: Int!) {\n  componentBytesProcessedTotal(interval: $interval) {\n    name\n    metric {\n      bytesProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
```

The pub/sub approach I'm using is in https://github.com/timberio/vector/blob/leebenson/top-subscriptions/src/api/schema/components.rs

Thanks!

sunli829:
Are there any differences between the two subscription implementations?

```
{"id":"c7186e5d-d333-4e9b-8153-56d0e8f628ed","type":"start","payload":{"operationName":"ComponentEventsProcessedTotalSubscription","query":"subscription ComponentEventsProcessedTotalSubscription($interval: Int!) {\n  componentEventsProcessedTotal(interval: $interval) {\n    name\n    metric {\n      eventsProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
{"id":"2d7ce3e3-c336-473e-8c4e-8b51dd42948a","type":"start","payload":{"operationName":"ComponentBytesProcessedTotalSubscription","query":"subscription ComponentBytesProcessedTotalSubscription($interval: Int!) {\n  componentBytesProcessedTotal(interval: $interval) {\n    name\n    metric {\n      bytesProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
```

```
{"id":"b8869530-853f-4feb-bb1d-fd2ede9582f1","type":"start","payload":{"operationName":"ComponentAddedSubscription","query":"subscription ComponentAddedSubscription {\n  componentAdded {\n    __typename\n    name\n  }\n}\n","variables":null}}
{"id":"83a9e880-15e6-4c82-890a-fdac34be3101","type":"start","payload":{"operationName":"ComponentRemovedSubscription","query":"subscription ComponentRemovedSubscription {\n    componentRemoved {\n        __typename\n        name\n    }\n}\n","variables":null}}
{"id":"c7186e5d-d333-4e9b-8153-56d0e8f628ed","type":"start","payload":{"operationName":"ComponentEventsProcessedTotalSubscription","query":"subscription ComponentEventsProcessedTotalSubscription($interval: Int!) {\n  componentEventsProcessedTotal(interval: $interval) {\n    name\n    metric {\n      eventsProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
{"id":"2d7ce3e3-c336-473e-8c4e-8b51dd42948a","type":"start","payload":{"operationName":"ComponentBytesProcessedTotalSubscription","query":"subscription ComponentBytesProcessedTotalSubscription($interval: Int!) {\n  componentBytesProcessedTotal(interval: $interval) {\n    name\n    metric {\n      bytesProcessedTotal\n    }\n  }\n}\n","variables":{"interval":1000}}}
```

sunli829:
It seems to be related to this?

https://github.com/timberio/vector/blob/c85c4351c3b4a738e2d61b979f6f74c1fdf17ea9/src/api/schema/components.rs#L242

leebenson:
> It seems to be related to this?

That was my suspicion, but I'm not sure why. I tried:

- Having a single static channel to subscribe to changes
- Having multiple static channels, for each change type, so different fields wouldn't 'collide'.
- Your previous `SimpleBroker` (from an older version of async-graphql)
- Lock-free pub/sub libs such as [bus](https://crates.io/crates/bus)
- Subscribing to a non-static channel through context

Each produced the same result.

It also doesn't explain why it works when using separate clients vs. all multiplexed over a single client.

Any thoughts on that?

sunli829:
Modify it to see if it will report an error.

```rust
            COMPONENT_CHANGED.send(ComponentChanged::Removed(
                COMPONENTS
                    .read()
                    .expect(INVARIANT)
                    .get(name)
                    .expect(INVARIANT)
                    .clone(),  
            )).unwrap();  // <<<<<<<<<-------------
```

https://github.com/timberio/vector/blob/c85c4351c3b4a738e2d61b979f6f74c1fdf17ea9/src/api/schema/components.rs#L368

leebenson:
Adding `.unwrap()` adds a requirement to derive `Debug` in related components, but it doesn't result in any errors.

sunli829:
I successfully compiled `vector` and I am working on this strange problem.

leebenson:
Thanks @sunli829, I appreciate your help with this!

sunli829:
I can reproduce this problem, but now I am going to bed, and I may be able to solve it tomorrow. This should not be a big problem.

sunli829:
I want to sleep well at night, so I solved this problem before going to bed, and now I immediately release a new version of `async-graphql`.

sunli829:
Please upgrade async-graphql to `v2.0.13`, I'm sorry for wasting your three days. 😂

lib/vector-api-client/graphql/queries/components.graphql
```diff
@@ -5,5 +5,8 @@ query ComponentsQuery {
     eventsProcessedTotal {
       eventsProcessedTotal
     }
+    bytesProcessedTotal {
```
lucperkins, line 8: Heads up: there's been some significant renaming work in VectorMetricsVille. These metrics are now prefixed with `vector_` on the Prometheus side of things. I'm not sure if that entails that these should be renamed, but wanted to let you know.

leebenson:
@sunli829, thanks so much for fixing this - works great!

I've gotta fix some tests locally, and then this PR should be good to go.

sunli829:
Please upgrade `async-graphql` to `v2.0.15`. Because of a bug, the old parameter attribute `#[arg(..)]` did not report an error, and this attribute did not actually take effect. @leebenson 

https://github.com/timberio/vector/blob/ba0f987ccd00a69a97007624612f61eed7c62b01/src/api/schema/metrics/mod.rs#L72

src/api/schema/metrics/bytes_processed.rs
```diff
@@ -26,8 +26,38 @@ impl ProcessedBytesTotal {
...
+        );
+
+        Self { name, metric }
+    }
+}
+
+#[Object]
+impl ComponentBytesProcessedTotal {
+    /// Component name
+    async fn name(&self) -> String {
```
sunli829, line 55: A borrow should be returned to avoid cloning.

```rust
async fn name(&self) -> &String { ... }
```

src/api/schema/metrics/errors.rs
```diff
@@ -0,0 +1,63 @@
+use crate::event::{Metric, MetricValue};
+use async_graphql::Object;
+use chrono::{DateTime, Utc};
+
+pub struct ErrorsTotal(Metric);
```
sunli829, line 5: I think this can be done to avoid cloning.

```rust
pub struct ErrorsTotal<'a>(&'a Metric);
```

src/api/schema/metrics/mod.rs
```diff
@@ -69,17 +77,54 @@ impl MetricsSubscription {
         })
     }
 
+    /// Component events processed metrics. Streams new data as the metric increases
+    async fn component_events_processed_total(
+        &self,
+        #[arg(default = 1000, validator(IntRange(min = "100", max = "60_000")))] interval: i32,
```
sunli829, line 83: After upgrading to async-graphql `v2.0.15`, errors will be reported correctly here.

lucperkins (approved):
I made a quick pass from a usability perspective but not much surface area on that end of things, so my approval carries little weight 😄 Exciting stuff, though.

src/api/schema/components.rs
```diff
@@ -8,7 +8,7 @@ use std::{
 };
 use tokio::stream::{Stream, StreamExt};
 
-const INVARIANT: &str =
+pub const INVARIANT: &str =
     "It is an invariant for the API to be active but not have COMPONENTS. Please report this.";
```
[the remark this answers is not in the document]
leebenson: Agreed, I'll make that clearer.

FWIW, the use of 'invariant' here is due to this being an internal error, rather than something that the user has done wrong. Technically, this error is only invoked if there's an issue acquiring a lock on the components, so the message is even vaguer. I'll clear that up.

src/api/schema/metrics/mod.rs
leebenson: That's great, thanks! This actually caught a silent error, where a `--refresh-interval` below 100 (ms) was previously allowed, but should have failed. Great work on input validation too, btw... really nice to be able to define custom errors so easily!

src/api/schema/metrics/bytes_processed.rs
leebenson: Ah, cool. I wasn't sure if lifetimes were affected in Object wrappers, so I defaulted to cloning everything. This is great. I've done a sweep for all `.clone()`s, and updated to `&str` for string types 👍

leebenson:
I updated this PR to ignore the tests that tend to trip up with global metrics/tracing enabled. Tackling separately in #4925.

lib/vector-api-client/src/gql/components.rs
```diff
@@ -36,3 +78,43 @@ impl std::fmt::Display for components_query::ComponentsQueryComponentsOn {
         write!(f, "{}", res)
     }
 }
+
+impl std::fmt::Display
+    for component_added_subscription::ComponentAddedSubscriptionComponentAddedOn
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let res = match self {
+            component_added_subscription::ComponentAddedSubscriptionComponentAddedOn::Source => {
```
Hoverbear, line 87: Minor nit that you can make this quite a bit less noisy with `use component_added_subscription::ComponentAddedSubscriptionComponentAddedOn` in this file.

Hoverbear (approved):

lib/vector-api-client/src/gql/components.rs
leebenson: Unfortunately I don't think that will work, due to `component_added_subscription` and ilk being auto-generated in this same file.

PR #4887, opened 2020-11-05
enhancement(kubernetes_logs source): Add PodIPs into Pod Metadata event annotations

### LINKED ISSUE #4740: Add Pod IP option to kubernetes_logs
### Current Vector Version
```
vector 0.11.0 (gda89b0b x86_64-unknown-linux-gnu 2020-10-06)
```

### Use-cases

In order to tie in logs with external monitoring systems that might only record IP addresses (e.g. AWS ELB logs), it's sometimes useful to know the Kubernetes Pod IPs.

### Attempted Solutions

Host IP is not good enough for K8s AWS VPC CNI networking, or I expect most other forms of CNIs you will find in the K8s eco-system.

The previous version of your K8s solution, the metadata transform, looks like it had this ability though I admit I never used it at the time.  See https://github.com/timberio/vector/blob/34d392a1cbf30b0fd94a6988db237d1550758451/src/transforms/kubernetes/mod.rs#L399

### Proposal

Add this as an option to the kubernetes_logs source, along with the the existing labels/pod_uid data - I'm pretty sure it's readily available in the K8s data you can fetch when getting pod labels.

Not sure if any of the other missing bits of info compared to the v0.9.2 functionality are that necessary (none yet jumps out to me) but you could consider them also if easy.  I only have a definite use case for Pod IP.

### References

None other than the previous code ref above, and it's presumed list of available information at https://github.com/timberio/vector/blob/34d392a1cbf30b0fd94a6988db237d1550758451/src/transforms/kubernetes/mod.rs#L357-L370

eeyun:
Closes #4740 

Will probably need a thorough review @MOZGIII but ultimately pretty trivial. We tap `k8s_openapi` to get the `PodStatus` data that holds `PodIPs`. This also bumps the k8s feature level to `1_16` to support returning both the IPv4 _and_ IPv6 addresses attached to a pod in the case that both things do in fact exist.

### DIFF
--- a/lib/k8s-test-framework/src/framework.rs
+++ b/lib/k8s-test-framework/src/framework.rs
@@ -1,8 +1,8 @@
 //! The test framework main entry point.
 
 use super::{
-    exec_tail, log_lookup, namespace, test_pod, up_down, vector, wait_for_resource,
-    wait_for_rollout, Interface, Reader, Result,
+    exec_tail, kubernetes_version, log_lookup, namespace, test_pod, up_down, vector,
+    wait_for_resource, wait_for_rollout, Interface, Reader, Result,
 };
 
 /// Framework wraps the interface to the system with an easy-to-use rust API
@@ -66,6 +66,12 @@ impl Framework {
     /// `namespace`.
     pub fn exec_tail(&self, namespace: &str, resource: &str, file: &str) -> Result<Reader> {
         exec_tail(&self.interface.kubectl_command, namespace, resource, file)
+    }
+
+    /// Exect a `kubectl --version`command returning a K8sVersion Struct
+    /// containing all version information  of the running Kubernetes test cluster.
+    pub async fn kubernetes_version(&self) -> Result<kubernetes_version::K8sVersion> {
+        kubernetes_version::get(&self.interface.kubectl_command).await
     }
 
     /// Wait for a set of `resources` in a specified `namespace` to achieve
--- a/lib/k8s-test-framework/src/lib.rs
+++ b/lib/k8s-test-framework/src/lib.rs
@@ -19,6 +19,7 @@ mod exec_tail;
 pub mod framework;
 mod helm_values_file;
 pub mod interface;
+pub mod kubernetes_version;
 mod lock;
 mod log_lookup;
 pub mod namespace;
--- a/src/sources/kubernetes_logs/pod_metadata_annotator.rs
+++ b/src/sources/kubernetes_logs/pod_metadata_annotator.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use evmap::ReadHandle;
 use k8s_openapi::{
-    api::core::v1::{Container, Pod, PodSpec},
+    api::core::v1::{Container, Pod, PodSpec, PodStatus},
     apimachinery::pkg::apis::meta::v1::ObjectMeta,
 };
 use serde::{Deserialize, Serialize};
@@ -20,6 +20,8 @@ pub struct FieldsSpec {
     pub pod_name: String,
     pub pod_namespace: String,
     pub pod_uid: String,
+    pub pod_ip: String,
+    pub pod_ips: String,
     pub pod_labels: String,
     pub pod_node_name: String,
     pub container_name: String,
@@ -32,6 +34,8 @@ impl Default for FieldsSpec {
             pod_name: "kubernetes.pod_name".to_owned(),
             pod_namespace: "kubernetes.pod_namespace".to_owned(),
             pod_uid: "kubernetes.pod_uid".to_owned(),
+            pod_ip: "kubernetes.pod_ip".to_owned(),
+            pod_ips: "kubernetes.pod_ips".to_owned(),
             pod_labels: "kubernetes.pod_labels".to_owned(),
             pod_node_name: "kubernetes.pod_node_name".to_owned(),
             container_name: "kubernetes.container_name".to_owned(),
@@ -72,6 +76,7 @@ impl PodMetadataAnnotator {
 
         annotate_from_file_info(log, &self.fields_spec, &file_info);
         annotate_from_metadata(log, &self.fields_spec, &pod.metadata);
+
         if let Some(ref pod_spec) = pod.spec {
             annotate_from_pod_spec(log, &self.fields_spec, pod_spec);
 
@@ -82,6 +87,10 @@ impl PodMetadataAnnotator {
             if let Some(container) = container {
                 annotate_from_container(log, &self.fields_spec, container);
             }
+        }
+
+        if let Some(ref pod_status) = pod.status {
+            annotate_from_pod_status(log, &self.fields_spec, pod_status);
         }
         Some(())
     }
@@ -130,6 +139,24 @@ fn annotate_from_pod_spec(log: &mut LogEvent, fields_spec: &FieldsSpec, pod_spec
     }
 }
 
+fn annotate_from_pod_status(log: &mut LogEvent, fields_spec: &FieldsSpec, pod_status: &PodStatus) {
+    for (ref key, ref val) in [(&fields_spec.pod_ip, &pod_status.pod_ip)].iter() {
+        if let Some(val) = val {
+            log.insert(key, val.to_owned());
+        }
+    }
+
+    for (ref key, ref val) in [(&fields_spec.pod_ips, &pod_status.pod_ips)].iter() {
+        if let Some(val) = val {
+            let inner: Vec<String> = val
+                .iter()
+                .filter_map(|v| v.ip.clone())
+                .collect::<Vec<String>>();
+            log.insert(key, inner);
+        }
+    }
+}
+
 fn annotate_from_container(log: &mut LogEvent, fields_spec: &FieldsSpec, container: &Container) {
     for (ref key, ref val) in [(&fields_spec.container_image, &container.image)].iter() {
         if let Some(val) = val {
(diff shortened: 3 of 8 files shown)


src/sources/kubernetes_logs/pod_metadata_annotator.rs
```diff
@@ -130,6 +136,24 @@ fn annotate_from_pod_spec(log: &mut LogEvent, fields_spec: &FieldsSpec, pod_spec
...
 
+fn annotate_from_pod_status(log: &mut LogEvent, fields_spec: &FieldsSpec, pod_status: &PodStatus) {
+    k8s_if_le_1_16! {
+        for (ref key, ref val) in [(&fields_spec.pod_ips, &pod_status.pod_ip)].iter() {
+            if let Some(val) = val {
+                log.insert(key, val.to_owned());
+            }
+        }
+    }
+    k8s_if_ge_1_15! {
```
MOZGIII, line 147: This is a compile-type check, and we don't allow specializing the `kube-openapi` features. We also don't need it.

I'd suggest that we simply set *both* `pod_ips` and `pod_ip` in the event. If both the `pod_status.pod_ip` and `pod_status.pod_ips` are present - we'll have some duplication in the event, but it's fine since it's up to the Kubernetes project decide how to deal with this duplication, and I'd say we should avoid intervening with custom logic here - and just pass values as is. It's a wise choice to keep the implementation direct, and the behaviour easy to understand for the users.

So, I suggest we just have something like this:

```rust
        for (ref key, ref val) in [(&fields_spec.pod_ip, &pod_status.pod_ip)].iter() {
            if let Some(val) = val {
                log.insert(key, val.to_owned());
            }
        }
        for (ref key, ref val) in [(&fields_spec.pod_ips, &pod_status.pod_ips)].iter() {
            if let Some(val) = val {
                let inner: Vec<String> = val.iter().filter_map(|v| v.ip.clone()).collect::<Vec<String>>();
                // assign an array here
            }
        }
```

In both cases, we just pass through the values without any kind of custom logic to derive a single value. KISS way.

src/sources/kubernetes_logs/pod_metadata_annotator.rs
```diff
@@ -130,6 +136,24 @@ fn annotate_from_pod_spec(log: &mut LogEvent, fields_spec: &FieldsSpec, pod_spec
...
+        for (ref key, ref val) in [(&fields_spec.pod_ips, &pod_status.pod_ips)].iter() {
+            if let Some(val) = val {
+                let inner: Vec<String> = val.iter().filter_map(|v| v.ip.clone()).collect::<Vec<String>>();
+                log.insert(key, format!("{:?}", inner));
```
MOZGIII, line 151: Can we assign an array here, instead of a string?

src/sources/kubernetes_logs/pod_metadata_annotator.rs
```diff
@@ -323,6 +348,58 @@ mod tests {
...
+                    let mut log = LogEvent::default();
+                    log.insert("kubernetes.pod_ips", "192.168.1.2");
+                    log
+                },
+            ),
+            (
+                FieldsSpec {
+                    pod_node_name: "node_name".to_owned(),
+                    ..Default::default()
+                },
```
MOZGIII, line 375: This test should probably use the `FieldsSpec::default()`, and there should be another one that asserts that custom `pod_ips` (and `pod_ip`) is properly utilized.

src/sources/kubernetes_logs/pod_metadata_annotator.rs
```diff
@@ -323,6 +348,58 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_annotate_from_pod_status() {
+        let cases = vec![
```
MOZGIII, line 353: Tests are missing some important test cases:
- non-default `FieldsSpec` with custom field names for `pod_ips` (and `pod_ip`)
- a test case where `PodStatus::pod_ips` is set, but the `PodStatus::pod_ip` is absent

src/sources/kubernetes_logs/pod_metadata_annotator.rs
```diff
@@ -72,6 +75,9 @@ impl PodMetadataAnnotator {
 
         annotate_from_file_info(log, &self.fields_spec, &file_info);
         annotate_from_metadata(log, &self.fields_spec, &pod.metadata);
+        if let Some(ref pod_status) = pod.status {
```
MOZGIII, line 78: Nit: to match the `Pod` fields definition, let's move this block to the very end of this fn, below the `pod.spec` processing.
Fields to in the order of `metadata` -> `spec` -> `status`.

src/sources/kubernetes_logs/pod_metadata_annotator.rs
```diff
@@ -20,6 +21,7 @@ pub struct FieldsSpec {
     pub pod_name: String,
     pub pod_namespace: String,
     pub pod_uid: String,
+    pub pod_ips: String,
```
MOZGIII, line 24: I think we should allow assigning both `pod_ip` and `pod_ips` fields in the event. See my other comments for rationale.

MOZGIII (changes requested):
Great PR!

I left some comments in the code, the theme is mostly on eliminating some bits of custom "magic" and gonig with a bit more KISS approach.

There are some generic things that should be addressed:
- when working on the Kubernetes-related code, add a `ci-condition: k8s e2e tests enable` to opt-in into CI running end-top-end tests on your PR; for this PR, also enable `ci-condition: k8s e2e all targets` to run the test suite on more Kubernetes versions (without it tests are only ran at over a small subset of test targets - typically the latest versions);
- please add the assertions to the E2E tests, see:
https://github.com/timberio/vector/blob/aade9e6e5c236f454c010e5e7366722ffa9b93b8/lib/k8s-e2e-tests/tests/vector-agent.rs#L438-L452

Keep in mind that E2E tests ensure our code works across all of our supported Kubernetes versions, and since we're not bumping the MSKV, the test code should be adequate for versions 1.14-1.19 (current latest). Ping me if you need help!

src/sources/kubernetes_logs/pod_metadata_annotator.rs
eeyun: I'm not 100% certain on this second one. I don't think there should ever be a case where `PodStatus` includes`PodIPs` with entries but no `PodIP`. Perhaps I'm misunderstanding that though?

MOZGIII: This is correct - my idea is why don't we test this anyway. It might be the case that it'll be deprecated in the future or sth, so it is useful to have it covered. Plus, it's sort of a matrix of possibilities we test, and without this case, the coverage would be incomplete 😄

eeyun: cool sounds good then :+1:

lib/k8s-test-framework/src/framework.rs
```diff
@@ -68,6 +68,12 @@ impl Framework {
         exec_tail(&self.interface.kubectl_command, namespace, resource, file)
     }
 
+    /// Exect a `kubectl --version`command returning a json object
+    /// with all version information available for parsing
```
MOZGIII, line 72:
```suggestion
    /// Exec a `kubectl --version` command and obtain version information of
    /// the Kubernetes cluster the test framework is running against.
```

lib/k8s-test-framework/src/framework.rs
```diff
@@ -68,6 +68,12 @@ impl Framework {
...
+    pub async fn version(&self) -> Result<version::K8sVersion> {
```
MOZGIII, line 73:
```suggestion
    pub async fn kubernetes_version(&self) -> Result<version::K8sVersion> {
```

I think it is more meaningful to explicitly mention that this is about the Kubernetes version in the signature.

lib/k8s-test-framework/src/lib.rs
```diff
@@ -29,6 +29,7 @@ pub mod test_pod;
 mod up_down;
 mod util;
 pub mod vector;
+pub mod version;
```
MOZGIII, line 32:
```suggestion
mod kubernetes_version;
```

Given the structure of the test framework, it is unlikely we'll be adding more version-related concerns to this mod, so let's specialize the name with `kubernetes_`.

lib/k8s-test-framework/src/version.rs
```diff
@@ -0,0 +1,70 @@
+//! Perform a version lookup.
+use super::Result;
+
+use std::process::Stdio;
+use tokio::process::Command;
+
+/// Exec a `kubectl` command to pull down the kubernetes version
+/// metadata for a running cluster for use in the test framework
+pub async fn get_version(kubectl_command: &str) -> Result<K8sVersion> {
```
MOZGIII, line 9:
```suggestion
pub async fn kubernetes_version(kubectl_command: &str) -> Result<K8sVersion> {
```

eeyun: So I think if we change the module name to `kubernetes_version` we might want to keep the function as `get_version` so we end up with a function signature of `kubernetes_version::get_version()` which i think is more descriptive than `kubernetes_version::kubernetes_version`. Do you feel ok with that?

MOZGIII: What I was thinking is this looks a lot like

https://github.com/timberio/vector/blob/b401de37aacb16d189fe0aa93ad5d202c80ddcbc/lib/k8s-test-framework/src/lib.rs#L24

which we then reexport as

https://github.com/timberio/vector/blob/b401de37aacb16d189fe0aa93ad5d202c80ddcbc/lib/k8s-test-framework/src/lib.rs#L42

for better usability.

In this case, we probably can't get away wich such reexport, as we're also exporting the `K8sVersion` - sp yeah, I guess it is fine.

We can reduce it even further to `get`, so the usage will be `kubernetes_version::get()`, and the type would be addressed `kubernetes_version::K8sVersion`.

eeyun: Done!

MOZGIII (approved):

PR #4888, opened 2020-11-05
chore: rid VectorSink::Futures01Sink

fanatid:
Part of #2942 #4247. Based and blocked by #4846.


fanatid:
The next step I planning to remove `BatchSink` and use `PartitionBatchSink` everywhere (partition key where `BatchSink` currently used will be `()`).

src/sinks/util/sink.rs
```diff
@@ -105,141 +109,132 @@ where
...
+            if self.buffer.is_some() && self.batch.is_empty() {
+                let item = self.buffer.take().unwrap();
+                self.as_mut().start_send(item)?;
+
+                if self.buffer.is_some() {
+                    unreachable!("Empty buffer overflowed.");
                 }
+
+                continue;
             }
```
bruceg, line 219: Does this fail the borrow checker (because `self` stays borrowed by `item`)?
```suggestion
            if self.batch.is_empty {
                if let Some(item) = self.buffer.take() {
                    self.as_mut().start_send(item)?;

                    if self.buffer.is_some() {
                        unreachable!("Empty buffer overflowed.");
                    }

                    continue;
                }
            }
```

src/sinks/util/sink.rs
```diff
@@ -424,163 +310,137 @@ where
...
+                    if (*this.closing && !batch.is_empty()) || batch.was_full() {
+                        true
+                    } else {
+                        let linger = this
+                            .lingers
+                            .get_mut(&partition)
+                            .expect("linger should exists for poll_flush");
+                        matches!(linger.poll_unpin(cx), Poll::Ready(()))
+                    }
+                };
```
bruceg, line 375:
```suggestion
                if (*this.closing && !batch.is_empty())
                    || batch.was_full()
                    || matches!(
                        this
                            .lingers
                            .get_mut(&partition)
                            .expect("linger should exists for poll_flush")
                            .poll_unpin(cx),
                        Poll::Ready(())
                       ) {
```
(and drop the `if should_send {` below)

You might even be able to turn this into a pure iterator:
```rust
    let partitions_ready: Vec<_> = this.partitions.iter()
      .filter(|(partition, batch)| …above condition…)
      .map(|(partition, _)| partition.clone())
      .collect();
```

bruceg (approved):
Looks ok, just a few non-blocking suggestions. I'm so ready for there to be just one futures.

src/sinks/util/sink.rs
fanatid: No, there is no borrow checker issues, I just do not like inner if-conditions :slightly_smiling_face: It would be nice to have something like `if let Some(item) = self.buffer.take() && self.batch.is_empty()`.
`unwrap` looks worse than inner if-conditions, I'll change.

src/sinks/util/sink.rs
fanatid: iter-filter-map-collect will not work due to borrow checker errors :disappointed: but I'll change `should_send`

fanatid:
Thanks for a review @bruceg

PR #4896, opened 2020-11-06
chore(ci): Ensure misc job only runs with changed sources

jamtur01:


jszwedko (approved):
Thanks! Do we need this in the master workflow too?

jamtur01:
@jszwedko No - I think we run everything there always?

jszwedko:
Ah, I didn't realize we weren't doing the same filtering there. :+1:

PR #4900, opened 2020-11-06
fix(ci): Fix passing --push arg at build-docker.sh

MOZGIII:
Supersedes #4764.
Fixes https://github.com/timberio/vector/commit/26be119ac6c506cd50001b5fdc94adafdecb0f2d#r43596614


jszwedko (approved):
That was simpler than I expected :smile:

Issue #3910, opened 2020-09-16
Log schemas


vector-vic:
A common need for Vector users is the ability to map data according to different schemas. This is a key requirement for Vector since it aims to be schema, standard, and vendor-neutral. In order to deliver on this claim, Vector must not only support a variety of schemas independently, but it must also assist in the interchange between them.

# Use Cases

## Transitioning to Vector

Schemas create very heavy lock-in. This is because most downstream systems depend on this schema. To name a few:

1. Alerts.

2. Graphs and dashboards.

3. Storages.

4. Humans.

Changing a schema can break all of these things which usually is not acceptable. To prevent this, Vector must adopt their current schema in a way that downstream dependencies do not notice.

## Transitioning Vendors

The use case above illustrates the need for Vector to support a single schema at a time, but there are cases where a user would need to support multiple. For example, when transitioning vendors. Vector must not only support the "read" schema but also transform the data to an entirely new "write" schema.

For example, if a user is transitioning from Splunk to Elasticsearch, Vector must ingest the data under the [Splunk Common Information Model](https://docs.splunk.com/Documentation/CIM/4.17.0/User/Overview) and transform it to the [Elastic Common Schema](https://www.elastic.co/guide/en/ecs/current/index.html).

## Automatic Dashboards, Alerts, & Insights

A benefit of using a vendor's agent is that it'll unlock automatic dashboards, alerts, and other features. This not only saves a considerable amount of time and effort, but you can effectively delegate the management of these things to your chosen vendor. For example, I assume that DataDog, and their community, continually improve their dashboards. In this case, it's very important that Vector can transparently adopt the DataDog schema so that DataDog Vector users can receive the same benefit. It also alleviates us from having to maintain these entities as well. 

# Schemas

1. [Elastic Common Schema](https://www.elastic.co/guide/en/ecs/current/index.html)

2. [Splunk Common Information Model (CIM)](https://docs.splunk.com/Documentation/CIM/4.17.0/User/Overview)

3. [OpenTelemetry Log Data Model](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/data-model.md)

4. [GELF](https://docs.graylog.org/en/3.2/pages/gelf.html#gelf-payload-specification)

5. [DataDog's reserved log attributes](https://docs.datadoghq.com/logs/log_collection/?tab=http#reserved-attributes)

6. ...and more

# Proposal

[...]

PR #4902, opened 2020-11-06
feat(remap): compile-time program result type checking

JeanMertz:
This PR implements RFC #4862.

_This PR is best reviewed \*per commit\*. While the full diff is rather large, most of that comes from numerous tests._

## User-Level Changes

1. **does not** alter any of the grammar of the Remap language.
2. **does not** add any new functions.
3. **does not** change anything for the Remap _transform_
3. **does** invalidate certain scripts in certain component conditions.

As for the last point, the one place this currently invalidates certain Remap scripts is if you use remap as a condition.

Take for example:

```toml
[transforms.reduce]
    type = "reduce"
    starts_when.type = "remap"
    starts_when.source = "1337"
```

The `Condition` trait we use only allows returning boolean values (for performance reasons), so the only thing we could do was to accept the above `starts_when` condition, but always return `false` because `1337` is not a boolean. 

After this PR, we can check the return value at boot-time, which allows us to return an error and prevent the above example config from being valid, pointing out that `starts_when.source` needs to resolve to a boolean. A user can use any of the available functions or operations to make sure the return value is a boolean:

* `to_bool(1337)` — true
* `1337 >= 0` — true
* `contains(to_string(1337), substring = "133")` — true
* etc

This PR has **no impact** on the Remap transform, as the return value of the script isn't relevant to the outcome of the transform (the script is used to manipulate events, not to resolve to a final value).

## Program Type Check

When compiling a Remap script, the callee informs the program which results it expects. The following constraints can be defined:

* fallible or not
* optional value or not _(support for optional values will probably be removed from Remap soon, to make things easier to grok)_
* one or more value type constraints

For example:

```rust
TypeDef {
	fallible: false,
	optional: false,
	constraint: Constraint::Any,
}
```

The above constraint requires the program to be infallible, and return a concrete value type of any kind.

Exact value constraints can be defined as `Any`, `Exact(Boolean)`. Multiple value kinds are defined as `OneOf(vec![Integer, Float])`.

The program returns an error at compile-time if the result does not match the expectation, e.g.:

> remap error: program error: expected to be infallible, but is not

or

> remap error: program error: expected to resolve to string or float values, but instead resolves to an error, or integer or boolean values

The `fallible` check applies to _all_ (root) expressions in a program (since any failing expression will fail the entire program), whereas the `optional` and `constraint` checks only apply to the last expression (the one that resolves to the final value of the program).

## Expression Type Definitions

To make the above work, every expression now implements:

```rust
fn type_def(&self, state: &state::Compiler) -> TypeDef;
```

There are several rules that apply:

* Path queries (`.foo.bar`) resolve to `Any` (because we don't use any schemas currently, but that could change with #3910), _unless_ a path is assigned a value, in which case the path gets whatever constraint the value has (e.g. `.foo = true` will type check `.foo` as a `Boolean`).
* Literal values type check to their concrete type.
* Variable references type-check to whatever value is assigned to them last (e.g. `$foo = true; $foo = "bar"; $foo` type-checks to a string)
* Some wrapper-expressions type check to their inner expressions (e.g. a `block` expression delegates type checking to the final expression in the block, and an if-else statement merges the if and else type checks together)
* Functions implement their own `type_def`, and have to make sure the contract between implementation and `TypeDef` holds.

## Future Work

While we're currently only using this information on a program-level, we _can_ (and probably will) use this on an expression-level, to reject certain programs.

For example:

```rust
$foo = "bar"
if $foo {
	// ...
}
```

The above will fail at runtime because `"foo"` is of kind `String`, whereas the condition in an if-statement has to resolve to a boolean.

In a future PR, we can (at compile-time) ask the constraints of `$foo`, and know that it'll never resolve to a boolean, and thus return a compile-time error. To resolve this, one could instead write `if to_bool($foo)`. The same applies to certain operators (e.g. `"foo" > 2` would fail to compile).

### DIFF
--- a/lib/remap-lang/src/expression.rs
+++ b/lib/remap-lang/src/expression.rs
@@ -1,4 +1,4 @@
-use crate::{Object, Result, State, Value, ValueKind};
+use crate::{CompilerState, Object, Result, State, Value, ValueKind};
 
 pub(super) mod arithmetic;
 pub(super) mod assignment;
@@ -73,10 +73,30 @@ impl ResolveKind {
     pub fn is_exact(&self) -> bool {
         matches!(self, ResolveKind::Exact(_))
     }
+
+    pub fn is_any(&self) -> bool {
+        matches!(self, ResolveKind::Any)
+    }
+
+    pub fn is_maybe(&self) -> bool {
+        matches!(self, ResolveKind::Maybe(_))
+    }
+
+    pub fn value_kinds(&self) -> Vec<ValueKind> {
+        use ResolveKind::*;
+
+        match self {
+            Any => ValueKind::all(),
+            OneOf(v) => v.clone(),
+            Exact(v) => vec![*v],
+            Maybe(v) => v.value_kinds(),
+        }
+    }
 }
 
 pub trait Expression: Send + Sync + std::fmt::Debug + dyn_clone::DynClone {
     fn execute(&self, state: &mut State, object: &mut dyn Object) -> Result<Option<Value>>;
+    fn resolves_to(&self, state: &CompilerState) -> ResolveKind;
 }
 
 dyn_clone::clone_trait_object!(Expression);
@@ -102,6 +122,12 @@ macro_rules! expression_dispatch {
             fn execute(&self, state: &mut State, object: &mut dyn Object) -> Result<Option<Value>> {
                 match self {
                     $(Expr::$expr(expression) => expression.execute(state, object)),+
+                }
+            }
+
+            fn resolves_to(&self, state: &CompilerState) -> ResolveKind {
+                match self {
+                    $(Expr::$expr(expression) => expression.resolves_to(state)),+
                 }
             }
         }
--- a/lib/remap-lang/src/expression.rs
+++ b/lib/remap-lang/src/expression.rs
@@ -1,4 +1,5 @@
 use crate::{CompilerState, Object, Result, State, Value, ValueKind};
+use std::fmt;
 
 pub(super) mod arithmetic;
 pub(super) mod assignment;
@@ -70,18 +71,22 @@ pub enum ResolveKind {
 }
 
 impl ResolveKind {
+    /// Returns `true` if this is a [`ResolveKind::Exact`].
     pub fn is_exact(&self) -> bool {
         matches!(self, ResolveKind::Exact(_))
     }
 
+    /// Returns `true` if this is a [`ResolveKind::Any`].
     pub fn is_any(&self) -> bool {
         matches!(self, ResolveKind::Any)
     }
 
+    /// Returns `true` if this is a [`ResolveKind::Maybe`].
     pub fn is_maybe(&self) -> bool {
         matches!(self, ResolveKind::Maybe(_))
     }
 
+    /// Get a collection of [`ValueKind`]s accepted by this [`ResolveKind`].
     pub fn value_kinds(&self) -> Vec<ValueKind> {
         use ResolveKind::*;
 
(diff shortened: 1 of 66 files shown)


lib/remap-lang/src/expression.rs
```diff
@@ -92,6 +97,88 @@ impl ResolveKind {
...
+        }
+
+        if let Maybe(kind) = other {
+            return Maybe(Box::new(self.merge(kind)));
+        }
+
+        if self.is_any() || other.is_any() {
+            return Any;
+        }
+
```
FungusHumungus:
```rust
match (self, other) {
   (Maybe(kind), Maybe(other)) => return Maybe(Box::new(kind.merge(other)),
   (Maybe(kind), _) => return Maybe(Box::new(kind.merge(other)),
   (_, Maybe(other)) => return Maybe(Box::new(self.merge(other))),
   (Any, _) => return Any,
   (_, Any) => return Any,
   _ => ()
}
```

Maybe a bit of duplication, but I think this makes it clearer what it is trying to do..

lib/remap-lang/src/type_def.rs
```diff
@@ -0,0 +1,127 @@
...
+    ///
+    /// Some expressions are infallible (e.g. the [`Literal`] expression, or any
+    /// custom function designed to be infallible).
+    pub fallible: bool,
+
+    /// True, if an expression can resolve to "nothing".
+    ///
+    /// For example, and if-statement without an else-condition can resolve to
+    /// nothing if the if-condition does not match.
+    pub optional: bool,
```
FungusHumungus, line 20: I don't think `optional` is the best name. It implies that you don't have to use it, which isn't exactly what it means. Perhaps `nullable` would be better?

lib/remap-lang/src/expression.rs
FungusHumungus: Ignore this, I see it's all changed..

lib/remap-lang/src/type_def.rs
JeanMertz: > It implies that you don't have to use it, which isn't exactly what it means.

Could you expand on this a bit more? That is how it is intended to work (and should work) right now.

Having said that, I'm not too worried about this for now, I'm already working on a follow-up PR to rip "optional expressions" out completely, and just have those situations resolve to `null`.

E.g.:

```rust
.foo = if false { "bar" } // .foo == null
```

lib/remap-lang/src/expression/arithmetic.rs
```diff
@@ -45,4 +52,223 @@ impl Expression for Arithmetic {
...
+        use value::{Constraint::*, Kind::*};
+        use Operator::*;
+
+        let constraint = match self.op {
+            Or => self
+                .lhs
+                .type_def(state)
+                .constraint
+                .merge(&self.rhs.type_def(state).constraint),
+            Multiply | Add => OneOf(vec![String, Integer, Float]),
```
FungusHumungus, line 66: `Multiply` should check the typedef of `self.lhs` to see if we are working with `String` or `Integer` or `Float`. 

`Add` should probably check both sides, but I'm not exactly sure what should happen if both are different?

lib/remap-lang/src/type_def.rs
FungusHumungus: `optional` means you don't have to do something. I think it this case, I guess the implication is that the type doesn't have to resolve to something? However, it is resolving to something - that something being `null`. So I think `nullable` is a more accurate term.

FungusHumungus:
I do really like this. It is a great start, and gives us a lot of potential to take things much further.

Thinking ahead to template fields, it would be really useful if we could limit template fields to just return Strings (or at least reject `Map` and `Array`). The issue is that since paths resolve to `Any` many template fields that use simple paths would be rejected for example `{{ .foo }}`, `{{ .foo + .bar }}`.

Obviously, path schemas would alleviate this issue. But in the absense of those, I think it would be useful to specify a type constraint of `String` if the type is known, but allow it if it isn't. So, it could allow `{{ .foo }}` - that would have to be a runtime error if `.foo` was the wrong type, but it would reject `{{ contains("..", "..") }}` as that is definitely the wrong type.

lib/remap-lang/src/type_def.rs
JeanMertz: That’s incorrect. It _does_ resolve to “nothing”. Because the return value of an expression is `Result<Option<Value>>`.

So `optional` means it can return `Ok(None)`, which is different from `Ok(Some(Null))`.

Having said that. I’ve already pushed a branch that has expressions return `Result<Value>`, and that does simplify things. So we can ignore this for now.

JeanMertz: Pushed the relevant PR at #5053. That PR does not yet remove the `optional` flag from `TypeDef`, but that will happen in a follow-up PR.

lib/remap-lang/src/expression/arithmetic.rs
JeanMertz: Ah yes, I forgot to finish this impl. I'll do so in a follow-up PR 👍

JeanMertz:
> I think it would be useful to specify a type constraint of `String` if the type is known, but allow it if it isn't.

Yeah, I like that. I'll make a note and tackle that in a follow-up PR, so that we can use it in the templating work we're doing 👍

FungusHumungus (approved):

JeanMertz:
Merging this, as @FungusHumungus has approved it, and there are quite a few PRs dependent on this one being merged.

I messaged @jszwedko that I'd be happy to receive a post-merge review. I'll tackle any review comments in a follow-up PR, if needed.

PR #4928, opened 2020-11-09
enhancement(observability): Kind/type for `vector top`

### LINKED ISSUE #4927: Vector top: update `type` - `kind`; add type
The use of `type` currently to express the component kind is a misnomer. This should be replaced with `kind`, and `type` should be used to display the underlying component type -- e.g. Prometheus, console, etc.

leebenson:
Updates component `Type` -> `Kind`, and swaps type out for the _destination_ type -- console, blackhole, etc:

image

Closes #4927 

Rather than make this an interface `type` field (which also has the unfortunate side-effect of clashing with a Rust keyword), each component type gets its own `<component>Type` field, a la:

```gql
query ComponentsQuery {
  components {
    __typename
    name
    ... on Source {
      sourceType
    }
    ... on Transform {
      transformType
    }
    ... on Sink {
      sinkType
    }
    eventsProcessedTotal {
      eventsProcessedTotal
    }
    bytesProcessedTotal {
      bytesProcessedTotal
    }
  }
}
```

These are currently implemented as strings (since the component config itself is returning a string), but could be moved to enums later if warranted. Since this is currently internal schema and not documented anywhere, I haven't made any provisions to anticipate a later switch of types, so the field name is simply of type `String!` for now.

@jszwedko - tagging you on this, since you brought the distinction of kind vs. type to my attention-- thanks!


jszwedko (commented):
:+1: Happy to see this added!

Do we anticipate any component kinds not having a `type`? I might add it to the `Component` interface type as: `componentType` rather than having a specific type field for each component.

leebenson:
> Do we anticipate any component kinds not having a type? I might add it to the Component interface type as: componentType rather than having a specific type field for each component.

That's definitely an option. I didn't in case we wound up splitting this into `enum SourceType` etc, vs. having a catch-all `ComponentType` which would list sources side-by-side with, say, sinks. I thought this might be useful if we later have config mutation that re-use these enums for input for a little more implicit validation.

jszwedko:
> > Do we anticipate any component kinds not having a type? I might add it to the Component interface type as: componentType rather than having a specific type field for each component.
> 
> That's definitely an option. I didn't in case we wound up splitting this into `enum SourceType` etc, vs. having a catch-all `ComponentType` which would list sources side-by-side with, say, sinks. I thought this might be useful if we later have config mutation that re-use these enums for input for a little more implicit validation.

Yeah, that's fair, if we were using enums for this. We could just add `componentType` be a `String` for now for use by GraphQL clients that are just generically handling components (which I imagine to be the vast majority of cases) and then always add `sourceType`, `sinkType`, etc. later if we find it useful to have enum results.

leebenson:
Thanks @jszwedko, that approach makes sense. Updated in 032ad7c023800004be990299930873e52724e287

jszwedko (approved):
:tada:

PR #4937, opened 2020-11-09
chore: Correct the inverted condition at docker push

MOZGIII:
This PR restores the push capability of our docker build script.


jszwedko (approved):
I missed that too.

PR #4941, opened 2020-11-10
chore(transforms): Migrate EventProcessed Events outside of Transforms

Hoverbear:
Builds off #4875 to wrap the event processed event emits out in the topology builder so they don't need to be written per transform. Closes https://github.com/timberio/vector/issues/4931

TODO: Ensure spans are captured right.


src/internal_events/events_processed.rs
```diff
@@ -0,0 +1,14 @@
+use crate::internal_events::InternalEvent;
+use metrics::counter;
+
+/// A metric to denote the number of events processed by a topology component.
+/// This is wired up already for transforms in the `topology::builder::build_pieces` function,
+/// so you don't need to do that yourself.
+#[derive(Debug)]
+pub struct EventProcessed;
+
+impl InternalEvent for EventProcessed {
```
lukesteensen, line 10: _nit:_ The file name is plural but the struct name is singular :slightly_smiling_face:

src/pipeline.rs
```diff
@@ -17,30 +35,127 @@ impl Sink for Pipeline {
...
+        debug_assert!(self.enqueued.is_empty());
         self.inner.poll_complete()
     }
 }
 
 impl Pipeline {
     #[cfg(test)]
-    pub fn new_test() -> (Self, Receiver<Event>) {
-        Self::new_with_buffer(100)
+    pub fn new_test(inlines: Vec<Box<dyn FunctionTransform>>) -> (Self, Receiver<Event>) {
```
lukesteensen, line 63: I might prefer we leave this method as-is and add another helper if/when we need it to include inlines. That way we don't need to change everywhere we use this without setting any.

lukesteensen (commented):
This is looking pretty good!

src/internal_events/events_processed.rs
Hoverbear: Oh thank you.

src/pipeline.rs
Hoverbear: This is part of #4875. I'll open a comment linking to this there.

Hoverbear:
image
Seems like the per-transform metrics are working

Hoverbear:
@leebenson I think later the `BytesProcessed` can slot into https://github.com/timberio/vector/blob/a3084ae1359f672b7cae85563a018dcbeb9322d1/src/topology/builder.rs#L125 and https://github.com/timberio/vector/blob/a3084ae1359f672b7cae85563a018dcbeb9322d1/src/topology/builder.rs#L136 ?

leebenson:
👍 Yeah, seems like an ideal place for it.

Hoverbear:
@leebenson Do you want me to throw it in right now?

leebenson:
Please do. Right now, we only get bytes with sinks. Do we need to juice up sources too?

Hoverbear:
Outside the scope of this PR. :) I'll inject a `BytesProcessed` event here.

Hoverbear:
@leebenson How do you want to measure `BytesProcessed` given there is no defined way to measure bytes for an internal structure which may be using referential bytes types or have differing sizes based on encoding?

leebenson:
> @leebenson How do you want to measure BytesProcessed given there is no defined way to measure bytes for an internal structure which may be using referential bytes types or have differing sizes based on encoding?

I assumed just the byte count of the raw, pre-processed incoming payload (like here, measured at the point of `emit!`ing the `EventProcessed`). Is this feasible?

Hoverbear:
Hmmm if we kept our data serialzied in byte buffer or something it would be. But since the values held by `LogEvent` are possibly things like `Bytes` or their own maps, I'm not sure we have an efficient or correct way to say "This event was X number of bytes at the source and is still X number of bytes", or even to say "This even processed Y number of bytes" without having to check the length of every field.

While I think it's possible, I think it's a feature we should consider carefully, as it may require some more architectural work.

jszwedko:
Just dropping a reference to #4705 which discussed similar shortcomings of measuring "bytes processed" when the sink is not doing any serialization. There are some options discussed there.

Hoverbear:
It is, but I'm not really sure how to measure that without actually going and scanning the size of each thing and that being quite expensive?

src/topology/builder.rs
```diff
@@ -130,8 +132,12 @@ pub async fn build_pieces(
             }
             Transform::Task(t) => {
                 let filtered = filter_event_type(input_rx, input_type);
+                let decorated = Box::new(filtered.map(move |event| {
```
lukesteensen, line 135: Could be nicer to use `inspect` here rather than `map`.

lukesteensen (approved):
Looks good to me!

src/topology/builder.rs
Hoverbear: Heck, yes!

Hoverbear:
I think this is obsolete.

PR #4945, opened 2020-11-10
enhancement(observability): Humanized formatting for `vector top` metrics

### LINKED ISSUE #4595: Do a sweep for `--human` stats in `vector top`
`vector top -h` produces 'human-readable' stats, to summarise large numbers into groups such as `k`, `Mb`, etc.

Currently, it displays some odd numbers -- such as `1.00` for `events_processed`. Grouping should only happen > 1k for this particular stat.

Do a general sweep to ensure the numbers match expectations. This should be done after the main `vector top` stats have been added, so it can be done in a single pass.

leebenson:
This PR adds a `-h` option to `vector top`, to optionally 'humanize' formatting of metrics. Closes #4595.

By default, metrics are separated by thousands:

image

With `-h` provided, this becomes:

image

Numbers >= 1000 are formatted using [human-format](https://github.com/BobGneu/human-format-rs)'s `Scales::SI` type, to 2 decimal places, using the following suffixes:

```rust
Scales {
  base: 1000,
  suffixes: vec![
      "".to_owned(),
      "k".to_owned(),
      "M".to_owned(),
      "B".to_owned(),
      "T".to_owned(),
      "P".to_owned(),
      "E".to_owned(),
      "Z".to_owned(),
      "Y".to_owned(),
  ],
}
```

I chose this scale over `Scales::Binary`, since it applies more generally to stats such as `events` vs. just byte counts.


leebenson:
Noting `--help` with this, now displays:

```
Display topology and metrics in the console, for a local or remote Vector instance

USAGE:
    vector top [FLAGS] [OPTIONS]

FLAGS:
    -h, --human-metrics    Humanize metrics, using numeric suffixes - e.g. 1,100 = 1.10 k, 1,000,000 = 1 M
        --help             Prints help information
    -V, --version          Prints version information

OPTIONS:
    -i, --refresh-interval <refresh-interval>    How often the screen refreshes (in milliseconds) [default: 500]
    -u, --url <url>                              Vector GraphQL API server endpoint
```

src/top/dashboard.rs
```diff
@@ -16,14 +18,46 @@ use tui::{
...
+    fn thousands_format(&self) -> String {
+        match self {
+            0 => "--".into(),
+            _ => self.to_formatted_string(&Locale::en),
+        }
+    }
+}
+
+/// Format metrics, using the 'humanized' format, abbreviating with suffixes
+trait HumanFormatter {
```
jszwedko, line 36: Not a big deal given the simplicity of this, but might be nice to throw a couple tests in.

jszwedko (approved):
Looks great! I think this makes it much more legible.

src/top/dashboard.rs
leebenson: Thanks for prompting for tests, @jszwedko 👍

This is a good reminder why even seemingly obvious assumptions ought to be tested. Parsing a 13 digit `f64` panicked, which would have caught anyone with a `processed_bytes` value of > 1TB.

I swapped human-format out for [number-prefix](https://crates.io/crates/number_prefix), which also appears far more active (both downloads + updates.), and also added a `human_format_bytes` after realising that I'd neglected to accommodate binary stats with a 1024 multiple.

The new panel looks like this:

image

jszwedko (approved):
Looks good! I'm surprised that other library wouldn't handle large values.

Also :+1: for formatting binary units as powers of 2.

leebenson:
> Looks good! I'm surprised that other library wouldn't handle large values.

I was surprised too, given the explicit support for suffixes that are only relevant with higher values.

This:

```rust
human_format::Formatter::new().with_decimals(2).format(1_100_000_000_000.00);
```

Panics with:

```
thread 'main' panicked at 'attempt to multiply with overflow'
```

PR #4953, opened 2020-11-11
fix(observability): Fix bytes processed total query

sghall:
ping @leebenson 'Ello Guv'na! 😄

Attempting to fix an issue where the bytes processed always comes back as null on queries (not subscriptions).  Believe it's a typo in the API that's causing it.  This is **not** noticeable on the `vector top` because the initial query is immediately updated by the subscriptions.  The typo is understandable given that events goes the other way with `events_processed_total`.

If you run a query against master like this you will always get null even if there are results:

```gql
query {
  sinks {
    name
    bytesProcessedTotal {
      bytesProcessedTotal
    }
  }
}
```


leebenson:
Nice catch @sghall 🎉

PR #4958, opened 2020-11-11
enhancement(observability): Batch events processed total

leebenson:
Adds a new `componentEventsProcessedTotalBatch` subscription, to introduce the option of receiving an array of payloads over `interval` alongside individual streams. 

@sghall brought this to my attention as potentially useful for the web UI, where it might be less expensive to update state atomically against a single `interval` vs. every stream, which can be numerous.

It also has the side-effect of shaving off a few bytes from the total payload, not repeating the typical ` { data { ... } }` setup.

## Example

Previously, we had this:

```gql
subscription {
  componentEventsProcessedTotal(interval:1000) {
    name
    metric {
      eventsProcessedTotal
    }
  }
}
```

Which returned individual payloads, like:

```json
{
  "data": {
    "componentEventsProcessedTotal": {
      "metric": {
        "eventsProcessedTotal": 52488
      },
      "name": "gen7"
    }
  }
}
```

This still exists, but now we also have:

```gql
subscription {
  componentEventsProcessedTotalBatch(interval:1000) {
    name
    metric {
      eventsProcessedTotal
    }
  }
}
```

Which returns _all_ metrics within the `interval` period as an array:

```json
{
  "data": {
    "componentEventsProcessedTotalBatch": [
      {
        "metric": {
          "eventsProcessedTotal": 55078
        },
        "name": "gen4"
      },
      {
        "metric": {
          "eventsProcessedTotal": 55078
        },
        "name": "gen7"
      },
      {
        "metric": {
          "eventsProcessedTotal": 55078
        },
        "name": "gen1"
      },
      {
        "metric": {
          "eventsProcessedTotal": 55078
        },
        "name": "gen5"
      },
      {
        "metric": {
          "eventsProcessedTotal": 55078
        },
        "name": "gen1_add_field"
      },
      {
        "metric": {
          "eventsProcessedTotal": 110156
        },
        "name": "out"
      }
    ]
  }
}
```

---

@sghall, if you like this approach, I'll roll it out to the other subscriptions separately.

### DIFF
--- a/src/api/schema/metrics/mod.rs
+++ b/src/api/schema/metrics/mod.rs
@@ -86,6 +86,20 @@ impl MetricsSubscription {
             .map(ComponentEventsProcessedTotal::new)
     }
 
+    /// Component events processed metrics, received in batches containing metrics over `interval`
+    async fn component_events_processed_total_batch(
+        &self,
+        #[graphql(default = 1000, validator(IntRange(min = "10", max = "60_000")))] interval: i32,
+    ) -> impl Stream<Item = Vec<ComponentEventsProcessedTotal>> {
+        component_counter_metrics_batch(interval, &|m| m.name == "events_processed_total").map(
+            |m| {
+                m.into_iter()
+                    .map(ComponentEventsProcessedTotal::new)
+                    .collect()
+            },
+        )
+    }
+
     /// Bytes processed metrics
     async fn bytes_processed_total(
         &self,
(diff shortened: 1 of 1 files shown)


jszwedko:
:+1: I might have actually expected `componentEventsProcessedTotal` to behave like that already (perhaps even more so if it was named slightly differently: `componentEventsProcessedTotals`).

This seems like `vector top` could take advantage of this as well to do one refresh.

leebenson:
> This seems like vector top could take advantage of this as well to do one refresh.

We _could_, but there isn't any known/visible pressure refreshing per payload currently, so we do get the advantage of being able to update on payloads being streamed in that are individually 5-10x smaller than batching and waiting for the fuller update.

It depends on the size of the config, but this might add some extra network pressure pushing it all to batching.

sghall:
Nice @leebenson! I'll pull this branch and try it out today  👍

sghall (approved):
Had a chance to try this out today.  Works great! 👍    I'm putting up a PR with the bytes and errors subscriptions using the same pattern.

PR #4996, opened 2020-11-11
enhancement: Extend `Resource` to sources 

### LINKED ISSUE #4827: Extend `Resource` to sources
With #4031 we've introduced concept of `Resource` to sinks which enables us to detect port conflicts and only wait for sink to finish if necessary. `Resource` can also be extended to sources for benefits of:
- Detection and reporting of conflicts on config level for sources.
- We will be able to handle conflicts between sinks and sources.
- Reload will be more robust.

ktff:
Closes #4827

Also extends `Fanout` with ability to wait. With that we can shutdown sinks before building new pieces so the resources can be freely used in build methods. cc @lukesteensen 

Also notable extension is the introduction of `SystemFdOffset` and `Stdin` resources. Where with `Stdin`, we will enforce usage of only one `stdin` source, and while that technically is a breaking change, the usage of multiple `stdin` sources is similar to using same port for multiple sockets so it's more of a bug fix than that we are breaking something.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -197,8 +197,6 @@ pub trait SinkConfig: core::fmt::Debug + Send + Sync {
     fn sink_type(&self) -> &'static str;
 
     /// Resources that the sink is using.
-    /// These resources shouldn't be used in the build method as that can result
-    /// in a build error. Only sinks are constrained by this.
     fn resources(&self) -> Vec<Resource> {
         Vec::new()
     }
@@ -545,9 +543,9 @@ mod test {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "sources-stdin", feature = "sinks-console"))]
 mod resource_tests {
-    use super::Resource;
+    use super::{load_from_str, Resource};
     use std::collections::HashSet;
 
     #[test]
@@ -584,5 +582,24 @@ mod resource_tests {
             conflicting,
             vec!["sink_0", "sink_1", "sink_2"].into_iter().collect()
         );
+    }
+
+    #[test]
+    fn config_conflict_detected() {
+        assert!(load_from_str(
+            r#"
+        [sources.in0]
+        type = "stdin"
+  
+        [sources.in1]
+        type = "stdin"
+  
+        [sinks.out]
+        type = "console"
+        inputs = ["in0","in1"]
+        encoding = "json"
+        "#
+        )
+        .is_err());
     }
 }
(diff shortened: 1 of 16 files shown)


src/topology/fanout.rs
```diff
@@ -122,18 +127,32 @@ impl Sink for Fanout {
...
+                Ok(AsyncSink::NotReady(item)) => return Ok(AsyncSink::NotReady(item)),
+                Ok(AsyncSink::Ready) => self.i += 1,
+                Err(()) => self.handle_sink_error(self.i)?,
+            }
+        } else {
+            // process_control_messages ended because control channel returned
+            // NotReady so it's fine to return NotReady here since the control
+            // channel will notify current task when it receives a message.
+            return Ok(AsyncSink::NotReady(item));
         }
```
bruceg, line 156:
```suggestion
        match sink.as_mut() {
            Some(sink) => match sink.start_send(item) {
                Ok(AsyncSink::NotReady(item)) => return Ok(AsyncSink::NotReady(item)),
                Ok(AsyncSink::Ready) => self.i += 1,
                Err(()) => self.handle_sink_error(self.i)?,
            }
            None => {
                // process_control_messages ended because control channel returned
                // NotReady so it's fine to return NotReady here since the control
                // channel will notify current task when it receives a message.
                return Ok(AsyncSink::NotReady(item));
            }
        }
```

bruceg (approved):
Looks good. I added a couple of non-blocking suggestions.

src/config/mod.rs
```diff
@@ -279,6 +285,12 @@ impl Resource {
     }
 }
 
+impl From<SocketAddr> for Resource {
+    fn from(addr: SocketAddr) -> Self {
+        Self::Port(addr.port())
```
lukesteensen, line 290: I'm not sure if this really matters or not, but couldn't a system with two different network interfaces have listeners on the same port number on each without conflicting? To handle that I think we'd need to do some translation of `0.0.0.0` as "all interfaces" and probably some other things as well. Likely not worthwhile.

lukesteensen (approved):
Nice!

src/config/mod.rs
jszwedko: :+1: it does seem like we should be considering the address with the port.

I think the only special consideration will be `0.0.0.0` and `::` for all addresses.

ktff: True, good catch. It should be just a matter of implementing custom comparison traits.

PR #4998, opened 2020-11-11
enhancement(sources): make scrape interval configurable

### LINKED ISSUE #4997: Make the scrape interval of internal metrics configurable
The `internal_metrics` has a hardcoded scrape interval of every 2 seconds. Would like to make that configurable so it can be controlled in a push based metrics model.

drunkirishcoder:
closes #4997

### DIFF
--- a/src/sources/internal_metrics.rs
+++ b/src/sources/internal_metrics.rs
@@ -15,6 +15,7 @@ use serde::{Deserialize, Serialize};
 use std::time::Duration;
 use tokio::select;
 
+#[serde(deny_unknown_fields)]
 #[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct InternalMetricsConfig {
     #[serde(default = "default_scrape_interval_secs")]
--- a/src/sources/internal_metrics.rs
+++ b/src/sources/internal_metrics.rs
@@ -1,5 +1,5 @@
 use crate::{
-    config::{DataType, GenerateConfig, GlobalOptions, SourceConfig, SourceDescription},
+    config::{DataType, GlobalOptions, SourceConfig, SourceDescription},
     metrics::Controller,
     metrics::{capture_metrics, get_controller},
     shutdown::ShutdownSignal,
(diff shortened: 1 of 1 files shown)


jamtur01:
Nice. +1.

src/sources/internal_metrics.rs
```diff
@@ -13,16 +13,30 @@ use futures::{
...
+pub fn default_scrape_interval_secs() -> u64 {
+    2
+}
 
 inventory::submit! {
     SourceDescription::new::<InternalMetricsConfig>("internal_metrics")
 }
 
-impl_generate_config_from_default!(InternalMetricsConfig);
+impl GenerateConfig for InternalMetricsConfig {
```
jszwedko, line 32: I think we could implement `Default` here instead. That would have the advantage of also being able to use `InternalMetricsConfig::default()` in tests and other places.

src/sources/internal_metrics.rs
```diff
@@ -13,16 +13,30 @@ use futures::{
...
+use tokio::select;
 
 #[derive(Deserialize, Serialize, Debug, Clone, Default)]
-pub struct InternalMetricsConfig {}
+pub struct InternalMetricsConfig {
+    #[serde(default = "default_scrape_interval_secs")]
+    scrape_interval_secs: u64,
+}
+
+pub fn default_scrape_interval_secs() -> u64 {
```
jszwedko, line 24: I think this can be a `const fn`.

jszwedko (approved):
Thanks @drunkirishcoder ! Two minor comments, but otherwise this looks good.

jszwedko:
Looks good!

I added a ` #[serde(deny_unknown_fields)]` when I realized a typo in my config was causing this setting to not be applied when I was testing it out.

I'll merge once CI is green.

jszwedko:
Test failures seem unrelated:

* Windows OOM issue we've been seeing recently

OSX:

```
 test config::watcher::tests::file_update ... Nov 11 22:12:12.133  INFO vector::config::watcher: Creating configuration file watcher.
Nov 11 22:12:12.136  INFO vector::config::watcher: Watching configuration files.
thread 'main' panicked at 'Test timed out', src/config/watcher.rs:150:13
```

PR #5001, opened 2020-11-12
fix: Prometheus uses cumulative histogram

### LINKED ISSUE #4628: `prometheus` sink isn't sending cumulative histogram
Prometheus uses [cumulative histograms](https://prometheus.io/docs/concepts/metric_types/#histogram), and while in the `prometheus` sink we are computing cumulative histogram from `distribution` metric, we aren't doing anything with already aggregated `histogram` which isn't cumulative. 

A simple fix would be to adjust already aggregated `histogram` in `prometheus` sink

juchiast:
close #4628 

### DIFF
--- a/src/sinks/prometheus/collector.rs
+++ b/src/sinks/prometheus/collector.rs
@@ -136,12 +136,16 @@ pub(super) trait MetricCollector {
                     count,
                     sum,
                 } => {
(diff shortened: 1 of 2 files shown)


bruceg (approved):

ktff (approved):
Well, that was nice and short.

PR #5002, opened 2020-11-12
enhancement(observability): Added batch subscriptions for component bytes and errors

sghall:
This PR is targeting https://github.com/timberio/vector/pull/4958.  I branched off of there and added the batch subscriptions for component bytes total and component errors total.  I'll wait for that to clear and then get this one in after review.

cc @leebenson 
I added the other subscriptions based off the pattern for the events.  I tested these out today locally in the app and they work great.  Comparing `vector top` (using the non-batched versions) and the batched data in the app looks to line up perfectly. Performance in the app with these subscriptions is much better.

**Note:** I regenerated the schema here.  Wasn't sure if I should do that.

### DIFF
--- a/lib/vector-api-client/graphql/schema.json
+++ b/lib/vector-api-client/graphql/schema.json
@@ -1706,6 +1706,45 @@
                 }
               ],
               "deprecationReason": null,
+              "description": "Component events processed metrics, received in batches containing metrics over `interval`",
+              "isDeprecated": false,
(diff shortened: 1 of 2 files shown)


leebenson (approved):
Looks great @sghall ! 🎉

leebenson:
> Note: I regenerated the schema here. Wasn't sure if I should do that.

👍 Yup, `make build-graphql-schema` after changes to the API is the right approach.

Issue #4925, opened 2020-11-09
Update problematic `api` tests to run in different processes


leebenson:
API tests are currently failing in #4886 due to shared global state. This isn't simple to isolate currently, due to the global nature of tracing and metrics.

A workaround could be using [rusty-fork](https://crates.io/crates/rusty-fork), but known disadvantages are:

- It doesn't currently [support async tests](https://github.com/AltSysrq/rusty-fork/issues/9).
- Wrapping tests in a `rusty_fork_test!` macro disables code analysis in IDEs, making it harder to write/maintain tests.

Need to investigate whether there's an option to combine with `#[tokio::test]` and/or create a thin wrapping function which executes the test in a separate runtime, so tests themselves can be written as regular functions.

jszwedko:
I have no idea how much effort this would be, but my initial thought would be to get rid of the global state and have metrics be associated with `Topology`.

leebenson:
👍 I think that's a better approach; this is the kick-the-can-down-the-road option. I don't know exactly how much ground this would shift, but in the API for example, there'd be a fair amount of reworking to not invoke `capture_metrics()` directly.

PR #5004, opened 2020-11-12
enhancement(observability): API batch support + tests

leebenson:
Follow-up to @sghall's #5002 to add support for `*Batch` subscriptions in the API client, and some tests to assert batching is happening as expected. I also realised I neglected to add a test for the previous `bytesProcessedTotal` and component variants, which have been added here.

No tests for error metrics yet, since those aren't typically thrown. Will probably tackle after #4925 is solved.

### DIFF
--- a/lib/vector-api-client/src/gql/metrics.rs
+++ b/lib/vector-api-client/src/gql/metrics.rs
@@ -143,7 +143,7 @@ impl MetricsSubscriptionExt for crate::SubscriptionClient {
         self.start::<ComponentEventsProcessedTotalBatchSubscription>(&request_body)
     }
 
-    /// Executes a components events processed total metrics subscription
+    /// Executes a components bytes processed total metrics subscription
     fn component_bytes_processed_total_subscription(
         &self,
         interval: i64,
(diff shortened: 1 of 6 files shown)


sghall:
Nice @leebenson.  I'll look at this today.  Want to spend some time to really understand the test setup and patterns here. 👍

lib/vector-api-client/src/gql/metrics.rs
```diff
@@ -110,4 +154,16 @@ impl MetricsSubscriptionExt for crate::SubscriptionClient {
 
         self.start::<ComponentBytesProcessedTotalSubscription>(&request_body)
     }
+
+    /// Executes a components events processed total metrics subscription
```
sghall, line 158: nit:  

/// Executes a components **bytes** processed total batch metrics subscription

tests/api.rs
```diff
@@ -388,8 +388,169 @@ mod tests {
...
+                    .unwrap()
+                    .data
+                    .unwrap()
+                    .component_bytes_processed_total,
+            );
+        }
+
+        // Should only show byte totals on sinks
+        assert_eq!(map[&0].name, "bytes_processed_total_sink");
+        assert_eq!(map[&0].name, "bytes_processed_total_sink");
```
sghall, line 501: nit:  duplicate assertion here

sghall (approved):
LGTM 👍

tests/api.rs
leebenson: Thanks! That second one should pulling from `map[&1].name`.

PR #5006, opened 2020-11-12
chore(kubernetes platform): Only add checksum if not using external configmap

### LINKED ISSUE #4812: Only add config checksum annotation to the DaemonSet/StatefulSet in Helm charts if managed config is used
If external ConfigMap is used, adding the annotation with a checksum to the DaemonSet/StatefulSet is pointless.
We should condition adding the annotation by whether the external ConfigMap is used or not.

eeyun:
Closes #4812 

### DIFF
--- a/distribution/helm/vector-agent/templates/daemonset.yaml
+++ b/distribution/helm/vector-agent/templates/daemonset.yaml
@@ -14,7 +14,9 @@ spec:
   template:
     metadata:
       annotations:
+        {{- if not  .Values.externalConfigMap }}
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- end }}
         {{- with .Values.podAnnotations }}
           {{- toYaml . | nindent 8 }}
         {{- end }}


distribution/helm/vector-agent/templates/daemonset.yaml
```diff
@@ -14,10 +14,12 @@ spec:
   template:
     metadata:
       annotations:
+        {{- if not  .Values.externalConfigMap }}
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
         {{- with .Values.podAnnotations }}
           {{- toYaml . | nindent 8 }}
         {{- end }}
+        {{- end }}
```
MOZGIII, line 22:
```suggestion
        {{- if not .Values.externalConfigMap }}
        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
        {{- end }}
        {{- with .Values.podAnnotations }}
          {{- toYaml . | nindent 8 }}
        {{- end }}
```

eeyun: Oh shoot, yeah good catch!

MOZGIII (approved):

PR #5008, opened 2020-11-12
enhancement(remap): support enum variants for function arguments

JeanMertz:
This is another improvement to the Remap language that moves one type of runtime errors to compile-time.

Specifically, this PR adds support for "enums" as function arguments.

As an example, this was possible before:

```rust
.variant = "SHA-512/INVALID"
sha2("my value", variant = .variant)
// ^ Runtime error: unknown SHA-2 algorithm variant
```

With this change, the variant has to be a literal string, and has to match the accepted enum variants. If it doesn't, the program won't compile:

```rust
.variant = "SHA-512/INVALID"
sha2("my value", variant = .variant)
// ^ Compiler error: unexpected expression, expected Literal, got Path

sha2("my value", variant = "SHA-512/INVALID")
// ^ Compiler error: unknown enum variant: SHA-512/INVALID, must be one of: SHA-224, SHA-256, <snip>
```

For now, the enum check logic requires enums to be strings, we can expand this to allow other values as well, but there's no need right now, so I wanted to start as simple as possible.

Aside from this change, the plumbing introduced to make this possible will also allow us to add other compile-time checks in the future.

For example, the `del` function currently works by providing a string `del(".foo")`, we want to use `del(.foo)`. With the changes in this PR, we can make it a compile-time error to provide anything other than a path as the function argument, without having to hard-code this function into the language grammar (as we did previously).

### DIFF
--- a/lib/remap-lang/src/expression/argument.rs
+++ b/lib/remap-lang/src/expression/argument.rs
@@ -1,0 +1,88 @@
+use crate::{
+    expression, expression::Error as E, state, value, Expr, Expression, Object, Result, TypeDef,
+    Value,
+};
+
+#[derive(thiserror::Error, Clone, Debug, PartialEq)]
+pub enum Error {
+    #[error(r#"invalid value type (got "{0}")"#)]
+    Invalid(value::Kind),
+}
+
+#[derive(Clone)]
+pub struct Argument {
+    expression: Box<Expr>,
+    ident: &'static str,
+    keyword: &'static str,
+    validator: fn(&Value) -> bool,
+
+    // used for error messages
+    function_ident: &'static str,
+}
+
+impl std::fmt::Debug for Argument {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Argument")
+            .field("expression", &self.expression)
+            .field("ident", &self.ident)
+            .field("keyword", &self.keyword)
+            .field("validator", &"fn(&Value) -> bool".to_owned())
+            .finish()
+    }
+}
+
+impl Argument {
+    pub fn new(
+        expression: Box<Expr>,
+        ident: &'static str,
+        keyword: &'static str,
+        validator: fn(&Value) -> bool,
+        function_ident: &'static str,
+    ) -> Self {
+        Self {
+            expression,
+            ident,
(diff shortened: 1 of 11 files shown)


lib/remap-lang/src/expression/argument.rs
```diff
@@ -0,0 +1,88 @@
...
+};
+
+#[derive(thiserror::Error, Clone, Debug, PartialEq)]
+pub enum Error {
+    #[error(r#"invalid value type (got "{0}")"#)]
+    Invalid(value::Kind),
+}
+
+#[derive(Clone)]
+pub struct Argument {
```
FungusHumungus, line 13: I find it confusing that we have multiple objects called `Argument` now.  We have `expression::Argument` and `function::Argument`.

Perhaps the `function::Argument` could be renamed `function::ArgumentType`?

Hoverbear (approved):

jszwedko:
I think, in this particular case, restricting the input to `sha2` and friends seems OK, but, more generally, it does seem like it'd be nice to allow variables (or paths) to be used as arguments like your example:

```
$variant = "SHA-512"
sha2("my value", variant = $variant)
```

I'm imagining setting a variable to one of the algorithms based on some aspect of the input.

JeanMertz:
> it does seem like it'd be nice to allow variables (or paths) to be used as arguments like your example:

While it's technically possible, I think the use-case is limited. Perhaps at some point we can tackle this, but I considered it out of scope for the first iteration.

> I'm imagining setting a variable to one of the algorithms based on some aspect of the input.

Just to be clear, are you saying you'd set the variable to a literal based on input, and then use that variable as value for the argument in the function? Or would you set the variable to whatever value the path has?

If the latter, then that was how it worked before this PR, but that means you run the risk of using an invalid variant, and having a runtime error.

If the former, I'd say for now, if you really want to dynamically set the value of an enum argument, you can use if statements:

```rust
if .foo == "SHA-224" {
	sha2("my value", variant = "SHA-224")
} else {
		sha2("my value", variant = "SHA-256")
}
```

jszwedko:
> If the latter, then that was how it worked before this PR, but that means you run the risk of using an invalid variant, and having a runtime error.

I was thinking this.

We can always loosen it later if people ask for it though (allowing for the runtime error there again).

PR #5015, opened 2020-11-13
enhancement(remap): use path arguments for `del` and `only_field` functions

JeanMertz:
Previously, when a function took some action on the path (as opposed to the value the path evaluated to), you'd write:

```rust
del(".foo")
only_fields(".foo.bar", ".baz")
```

If you provided anything other than a string expression, or if the string contained an invalid path (e.g. `".foo hello world"`), you'd get a runtime error.

With this change, you now write:

```rust
del(.foo)
del(.foo.bar, .baz)
```

These functions validate at compile-time that their arguments are "path expressions". Any other kind of expression is a compile-time error.

_PR is best reviewed per commit_

ref #4947

### DIFF
--- a/lib/remap-lang/src/expression/assignment.rs
+++ b/lib/remap-lang/src/expression/assignment.rs
@@ -75,7 +75,7 @@ impl Expression for Assignment {
             Target::Path(path) => state
                 .path_query_type(&path.as_string())
                 .cloned()
-                .expect("variable must be assigned via Assignment::new"),
+                .expect("path must be assigned via Assignment::new"),
         }
     }
 }
--- a/lib/remap-lang/src/expression/assignment.rs
+++ b/lib/remap-lang/src/expression/assignment.rs
@@ -1,5 +1,8 @@
 use super::Error as E;
-use crate::{state, Expr, Expression, Object, Result, TypeDef, Value};
+use crate::{
+    expression::{Path, Variable},
+    state, Expr, Expression, Object, Result, TypeDef, Value,
+};
 
 #[derive(thiserror::Error, Clone, Debug, PartialEq)]
 pub enum Error {
@@ -9,8 +12,8 @@ pub enum Error {
 
 #[derive(Debug, Clone)]
 pub enum Target {
-    Path(Vec<Vec<String>>),
-    Variable(String),
+    Path(Path),
+    Variable(Variable),
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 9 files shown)


lib/remap-lang/src/expression/path.rs
```diff
@@ -28,14 +28,31 @@ impl Path {
...
+            .iter()
+            .map(|c| {
+                c.iter()
+                    .map(|p| p.replace(".", "\\."))
+                    .collect::<Vec<_>>()
+                    .join(".")
+            })
+            .collect::<Vec<_>>()
+            .join(".")
+    }
```
JeanMertz, line 47: I'll note that path-based lookup is still a bit awkward right now. I plan on overhauling this before 0.12 ships with Remap. This will likely also involve re-introducing the `Lookup` work that Ana worked on.

So until then, please ignore the awkwardness here (and in other places) when it comes to converting between Remap-based paths, and Vector event paths.

lib/remap-lang/src/function.rs
```diff
@@ -1,7 +1,27 @@
...
+    expression::{self, Path},
+    Expr, Expression, Result, Value,
+};
 use core::convert::{TryFrom, TryInto};
 use std::collections::HashMap;
 
+// workaround for missing variable argument length.
+//
+// We'll come up with a nicer solution at some point. It took Rust five
+// years to support [0; 34].
```
bruceg, line 11: I assume this is a metric we are not seeking to emulate… :grin:

bruceg (approved):
I'm not excited about the variadic argument support, but otherwise looks ok.

lib/remap-lang/src/expression/assignment.rs
```diff
@@ -24,11 +27,12 @@ impl Assignment {
...
 
         match &target {
-            Target::Variable(ident) => state.variable_types_mut().insert(ident.clone(), type_def),
-            Target::Path(segments) => {
-                let path = crate::expression::path::segments_to_path(segments);
-                state.path_query_types_mut().insert(path, type_def)
-            }
+            Target::Variable(variable) => state
+                .variable_types_mut()
+                .insert(variable.ident().to_owned(), type_def),
```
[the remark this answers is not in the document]
JeanMertz: Yeah, I see no reason why we can't do that. I'm somewhat used to always using some kind of primitive as the key, but I guess in this case it results in an extra unnecessary allocation.

JeanMertz:
> I'm not excited about the variadic argument support, but otherwise looks ok.

You're not excited about my clever workaround @bruceg? 😛 You know what they say... fake it till you make it.

PR #5016, opened 2020-11-13
enhancement(observability): Add I/O (throughput) columns to `vector top`

### LINKED ISSUE #4524: Add 'events throughput' to subscriptions

leebenson:
Adds `I/O` columns to `vector top`, representing throughput:

image

Closes #4524

Notes:

## New throughput subscriptions

There are 4 new subscriptions for events/bytes throughput, offering streaming individual or 'batched' updates:

**Bytes throughput, singular:**

```gql
subscription ComponentBytesProcessedThroughputSubscription($interval: Int!) {
    componentBytesProcessedThroughput(interval: $interval) {
        name
        throughput
    }
}
```

**Bytes throughput, batched:**

```gql
subscription ComponentBytesProcessedThroughputBatchSubscription($interval: Int!) {
    componentBytesProcessedThroughputBatch(interval: $interval) {
        name
        throughput
    }
}
```

**Events throughput, singular:**

```gql
subscription ComponentEventsProcessedThroughputSubscription($interval: Int!) {
    componentEventsProcessedThroughput(interval: $interval) {
        name
        throughput
    }
}
```

**Events throughput, batched:**

```gql
subscription ComponentEventsProcessedThroughputBatchSubscription($interval: Int!) {
    componentEventsProcessedThroughputBatch(interval: $interval) {
        name
        throughput
    }
}
```

## Moving to batch subscriptions

In previous versions, the Vector API client used the 'singular' streaming subscriptions. This had the effect of updating individual stats and redrawing the screen on each new payload. Care was taken for metrics to be 'ordered' in source -> transform -> sink, so as to avoid the UX of seeing a sink update before a source, for example.

While I didn't observe any particular CPU or network pressure with this approach (a release build with as low as a 10ms sample rate topped out at 2-3%), there's a trade-off between receiving the streamed data as quickly as the API makes it available, and subsequently allocating a new `Vec<Metric>` to update state that may make this approach less feasible for lower powered hardware, such as IoT scenarios.

To mimic the web UI in this regard, I decided to swap out these singular streamed for the 'batched' variants. These are still streamed, and sampled over the interval as before. The difference is that instead of yielding each result to the connected WebSocket, each payload is collected as a Vec and emitted once per `interval`. The network traffic is slightly reduced, since each payload only has one `{"data": { ... }}` JSON wrapper, and the rows are iterated over in `vector top` before updating state atomically.

The singular version is still offered in our API schema, for flexibility to other clients that may have a use-case for individual payloads.

## Sampling I/O

Throughput subscriptions differ slightly from other metrics, in that the first payload is emitted on the trailing edge of `interval`, and not immediately. This is to ignore the first emission, since the interval between the subscription and the last capture would otherwise be an arbitrary sample and therefore quite meaningless.

The effect in `vector top` is that other stats are aggregated first, with I/O columns showing as `--` until `interval` has elapsed the payload has been processed.

## Added a sample rate blurb

Notice in the title of `vector top`, there's now a "Sampling @ Xms" blurb in grey, to make it clearer what the rate is.

I've also updated options to take an `--interval` instead of the previous `--refresh-interval`, as 'refresh' here is misleading-- the terminal will typically refresh at at least 4x the specified rate, since there are now 4 active metrics subscriptions per session, as well as subscriptions for observing topology changes.

`vector top --help` now displays:

```
Display topology and metrics in the console, for a local or remote Vector instance

USAGE:
    vector top [FLAGS] [OPTIONS]

FLAGS:
    -h, --human-metrics    Humanize metrics, using numeric suffixes - e.g. 1,100 = 1.10 k, 1,000,000 = 1.00 M
        --help             Prints help information
    -V, --version          Prints version information

OPTIONS:
    -i, --interval <interval>    Interval to sample metrics at, in milliseconds [default: 500]
    -u, --url <url>              Vector GraphQL API server endpoint
```

CC @lucperkins - this retains the same `-i` shorthand as before, but I think explains the intent a little better. What do you think? Will need to update the Cue config accordingly.

## Reviewers

I've tagged the following reviewers, for these reasons:

- @binarylogic - Interested in your thoughts on the layout of `vector top`, per the above screenshot. Are the column labels of "I/O" clear that they represent throughput? Should we space the columns differently to make it clear they're 'grouped' with events/bytes? Should the humanized stats be enabled explicitly per `-h` like it is now -- or be the default? (without this setting, all stats - including I/O - are separated by thousands, without a suffix.)

- @sghall - Thoughts on adding these subscriptions to the web UI and general feedback to the current state of `vector top` welcomed. Any extra requirements that spring to mind?

- @Hoverbear - general code review, please.

Thanks all!


binarylogic:
@leebenson in the screenshot what is I/O? Is it per second?

leebenson:
> @leebenson in the screenshot what is I/O? Is it per second?

All metrics are sampled at the rate provided by `--interval`, which defaults to 500ms.

binarylogic:
Got it, do you think it would be clearer to include that in the output? It might be too verbose, but it wasn't immediately obvious to me.

leebenson:
👍 I put a "Sampling rate @ <X>ms` in the title for this reason, but I don't think it's obvious enough. Could put it in the column heading, but it might expand it a bit wide. I'll try different formats.

binarylogic:
Yeah, if it's too many characters don't worry about it.

Hoverbear:
@leebenson Can we control batchsize with the subscription? I don't see it in the example.

leebenson:
> @leebenson Can we control batchsize with the subscription? I don't see it in the example.

Not directly, but the batch size will always be equal to the number of components sampled over `interval`, so this should never get too crazy.

The main benefit of the batch approach is that it gives the client a single array to update state atomically. If this was to be broken out into pages, it wouldn't serve that intended use-case very well. The 'singular' versions are likely a better fit for that, which are the same queries without the `batch` suffix. Those stream payloads individually, emitted as quickly as possible after `interval` has elapsed between samples.

jszwedko:
I mentioned this over here: https://github.com/timberio/vector/pull/4958#issuecomment-725475098

I might actually drop the word "batch" from those subscriptions and instead call it `componentEventsProcessedTotals` to make it clearer that all component totals are returned. "batch" makes me think that there should be a configurable batch size and that a given batch may not actually contain all of the components.

leebenson:
The motivation for adding the `batch` suffix was mostly to distinguish between it and the singular version. I thought the two versions separated only by an `s` might be harder to grok on quick glance, although if 'batch' is also likely to conjure connotations, I'm not sure that assumption is necessarily valid. I think having the option for the singular versions might still be useful in cases where receiving slimmer, streaming payloads is preferred, so I'd rather keep both of them around in any case.

I'm fine with making switching `batch` -> `s` for clarity. CC @sghall for awareness that this will introduce a breaking change to the current UI once added, but it should be a trivial switch.

jszwedko:
> The motivation for adding the `batch` suffix was mostly to distinguish between it and the singular version. I thought the two versions separated only by an `s` might be harder to grok on quick glance, although if 'batch' is also likely to conjure connotations, I'm not sure that assumption is necessarily valid. I think having the option for the singular versions might still be useful in cases where receiving slimmer, streaming payloads is preferred, so I'd rather keep both of them around in any case.
> 
> I'm fine with making switching `batch` -> `s` for clarity. CC @sghall for awareness that this will introduce a breaking change to the current UI once added, but it should be a trivial switch.

:thinking: the `s` is definitely a subtle difference. I also see `allComponentEventsProcessedTotals` maybe.

sghall:
> CC @sghall for awareness that this will introduce a breaking change to the current UI once added, but it should be a trivial switch.

Right on.  Yeah,  I'm fine with whatever you land on.  I may have inserted "batch" into the lexicon here and agree it could be confusing.

Hoverbear (approved):
I feel the code overall looks good! I share the confusion/concerns around the name "batch" though, since it implies something different in other parts of vector.

sghall (approved):
Looks good 👍  Had a chance to try out the batch subscriptions today and they worked well.

leebenson:
@binarylogic - I actually don't think column intervals look too bad... what d'ya think?

image

binarylogic:
I like it 👍

leebenson:
@jszwedko, I like `allComponentEventsProcessedTotals` -- switching to that format now...

leebenson:
Some changes since the last review round:

- 'Batch' subscriptions have been updated to use the `all*` pattern everywhere.
- I/O columns now show as `I/O @ <X>ms`, where `X` is the provided/default `--interval`
- Shrinking a terminal to < 80 chars will display as:

image

https://github.com/fdehau/tui-rs/issues/418 indicates improvements coming to TUI which may address column inconsistencies; this is just an interim workaround.

jszwedko (approved):
In general this looks good to me.

> The singular version is still offered in our API schema, for flexibility to other clients that may have a use-case for individual payloads.

I might drop the single subscription versions until someone asks for them. I'm having a hard time imagining a use-case where you'd want that over the set.

For throughput, I could also see us just having the client calculate that based on the differences between the totals (timestamps could be added to facilitate this) to keep the API general, but it seems fine to cater the API specifically to the client's needs (something GraphQL extols).

leebenson:
> I might drop the single subscription versions until someone asks for them. I'm having a hard time imagining a use-case where you'd want that over the set.

👍 I'll zap them. I'm not sure there's a good use for them now, either.

> For throughput, I could also see us just having the client calculate that based on the differences between the totals (timestamps could be added to facilitate this) to keep the API general, but it seems fine to cater the API specifically to the client's needs (something GraphQL extols).

I'd err toward keeping these calculations in Vector. I'm thinking of scenarios where prior state may not be retained in the client for calculating diffs, such as immutable stream -> render pipelines where functions just receive the 'new' state. It's also pretty inexpensive, and broadens the base of observability we're offering the end-client, IMO.

jszwedko:
> I'd err toward keeping these calculations in Vector. I'm thinking of scenarios where prior state may not be retained in the client for calculating diffs, such as immutable stream -> render pipelines where functions just receive the 'new' state. It's also pretty inexpensive, and broadens the base of observability we're offering the end-client, IMO.

That's fair, it does reduce the need for client state.

leebenson:
@jszwedko, I zapped the singular queries in 616643df937841e22f6317e7b6d69a3d2810f0ec.

My initial objection of distinguishing the two query fields by plurals no longer stands, so I think your original suggestion of the `componentEventsProcessedTotals`-style queries actually makes more sense than the `all*` prefix, given this.

I also think it'll be less confusing if we wind up adding filtering.

jszwedko:
> @jszwedko, I zapped the singular queries in [616643d](https://github.com/timberio/vector/commit/616643df937841e22f6317e7b6d69a3d2810f0ec).
> 
> My initial objection of distinguishing the two query fields by plurals no longer stands, so I think your original suggestion of the `componentEventsProcessedTotals`-style queries actually makes more sense than the `all*` prefix, given this.
> 
> I also think it'll be less confusing if we wind up adding filtering.

Nice! I like it. :tada:

binarylogic:
Me too, nice work.

PR #5018, opened 2020-11-13
enhancement(observability): API version + hostname queries

### LINKED ISSUE #5014: Add a `version` query to the GraphQL API
To mimic `vector --version`.

leebenson:
Adds the following query fields to the GraphQL API:

```gql
{
  version
  hostname
}
```

Returning (on my machine, anyway):

```json
{
  "data": {
    "hostname": "Lees-MacBook-Pro-2.local",
    "version": "0.11.0 (gd0b414e x86_64-apple-darwin 2020-11-12)"
  }
}
```

Closes #5014


jszwedko (approved):
Not a big deal, but I might group these under a `query { meta { version hostname } }`.

I might also make hostname nullable in-case it cannot be fetched so that it doesn't fail the entire query (i.e. `version` would still be returned).

leebenson:
Thanks @jszwedko, added both of those suggestions in de297dcee173e93320c99a66c9a61797168208c8.

The new query is:

```gql
{
  meta {
    version
    hostname
  }
}
```

Returning:

```json
{
  "data": {
    "meta": {
      "hostname": "Lees-MacBook-Pro-2.local",
      "version": "0.11.0 (gd0b414e x86_64-apple-darwin 2020-11-12)"
    }
  }
}
```

jamtur01 (approved):
LGTM!

jszwedko:
@leebenson Actually now I'm putting my GraphQL schema hat back on :smile:  and think that we might even want to do something like:

```graphql
type Version {
  major String!
  minor String!
  patch String!
}

type BuildVersion {
  version: Version!
  sha: String!
  platform: Platform!
}

type Platform {
 machine: String!
 vendor: String!
 operatingSystem: String!
}

type Meta {
  version: Version
}
```

Admittedly we probably don't need those just yet, but you could leave room for them by calling what we currently have `versionString` instead of `version` to allow them to be added later with backwards compatibility.

I realize this is all in-flux right now so compatibility is less of a concern, but maybe something to keep in mind as we release this, people start using it, and we need to start worrying about API compatibility.

leebenson:
Ha, I had similar thoughts too, but thought having parity with `vector --version` would be a good enough v1.

FWIW, this is where it's pulling from in `src/lib.rs`:

```rust
pub fn vector_version() -> impl std::fmt::Display {
    #[cfg(feature = "nightly")]
    let pkg_version = format!("{}-nightly", built_info::PKG_VERSION);

    #[cfg(not(feature = "nightly"))]
    let pkg_version = built_info::PKG_VERSION;

    pkg_version
}

pub fn get_version() -> String {
    let pkg_version = vector_version();
    let commit_hash = built_info::GIT_VERSION.and_then(|v| v.split('-').last());
    let built_date = chrono::DateTime::parse_from_rfc2822(built_info::BUILT_TIME_UTC)
        .unwrap()
        .format("%Y-%m-%d");
    let built_string = if let Some(commit_hash) = commit_hash {
        format!("{} {} {}", commit_hash, built_info::TARGET, built_date)
    } else {
        built_info::TARGET.into()
    };
    format!("{} ({})", pkg_version, built_string)
}

#[allow(unused)]
mod built_info {
    include!(concat!(env!("OUT_DIR"), "/built.rs"));
}
```

Happy to tackle breaking this out further, if there's a use-case for it. @jamtur01, do we need that resolution?

leebenson:
> I realize this is all in-flux right now so compatibility is less of a concern, but maybe something to keep in mind as we release this, people start using it, and we need to start worrying about API compatibility.

Agreed, there will come a point soon where we need to consider stability. In this pre-release mode, I think we're in a safer place to be led by requirements (mostly from the web UI, presumably.) I don't mind breaking things so much if it's only me having to clean it up!

jszwedko:
:+1: definitely agreed. I think we could just name the field differently, `versionString` rather than `version` which would both:

* Not deal with breaking up the version string until we need it
* Allow us to later add the `version` field, if we find the need, with structured version information without breaking API compatibility

This is largely based on what I'm remembering of GraphQL best practices.

leebenson:
Agreed, I think that's an easy win. Will make that switch. The old balancing act between not backing ourselves into a corner to remain sufficiently flexible, vs. being so explicit that names wind up embedding type info for perceived flexibility we may never need. It's not always clear which side of the line that should always fall.

I appreciate you prompting these graph design discussions and helping to think this through. Team Alloy FTW 😄

jszwedko:
It's always a balance :smile:

PR #5021, opened 2020-11-13
enhancement(pulsar sink): introduce encoding schema and pulsar avro schema

KannarFr:
WIP

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -194,8 +194,6 @@ anyhow = { version = "1.0.28" }
 snap = { version = "1.0.2", optional = true }
 dyn-clone = "1.0.3"
 indoc = "1.0.3"
-
-# For AVRO
 avro-rs = "0.11.0"
 
 # For WASM
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,6 +195,9 @@ snap = { version = "1.0.2", optional = true }
 dyn-clone = "1.0.3"
 indoc = "1.0.3"
 
+# For AVRO
+avro-rs = "0.11.0"
+
 # For WASM
 vector-wasm = { path = "lib/vector-wasm", optional = true }
 lucetc = { git = "https://github.com/bytecodealliance/lucet.git", rev = "b1863dacc8c92c11e5434fc8815d9b9a26cfe3db", optional = true }
--- a/src/sinks/pulsar.rs
+++ b/src/sinks/pulsar.rs
@@ -60,6 +60,7 @@ enum PulsarSinkState {
 
 struct PulsarSink {
     encoding: EncodingConfig<Encoding>,
+    avro_schema: Option<avro_rs::Schema>,
     state: PulsarSinkState,
     in_flight:
         FuturesUnordered<BoxFuture<'static, (usize, Result<CommandSendReceipt, PulsarError>)>>,
@@ -155,8 +156,22 @@ async fn healthcheck(producer: PulsarProducer) -> crate::Result<()> {
 
 impl PulsarSink {
     fn new(producer: PulsarProducer, encoding: EncodingConfig<Encoding>, acker: Acker) -> Self {
+        let schema = match &encoding.codec() {
+            Encoding::Avro => {
+                if let Some(schema) = &encoding.schema() {
+                    avro_rs::Schema::parse_str(schema).ok()
+                } else {
+                    None
+                }
+            }
+            _ => {
+                None
+            }
+        };
+
         Self {
             encoding,
+            avro_schema: schema,
             state: PulsarSinkState::Ready(Box::new(producer)),
             in_flight: FuturesUnordered::new(),
             acker,
@@ -201,7 +216,7 @@ impl Sink<Event> for PulsarSink {
             "Expected `poll_ready` to be called first."
         );
 
-        let message = encode_event(item, &self.encoding).map_err(|_| ())?;
+        let message = encode_event(item, &self.encoding, &self.avro_schema).map_err(|_| ())?;
 
         let mut producer = match std::mem::replace(&mut self.state, PulsarSinkState::None) {
             PulsarSinkState::Ready(producer) => producer,
@@ -258,7 +273,7 @@ impl Sink<Event> for PulsarSink {
     }
 }
 
-fn encode_event(mut item: Event, encoding: &EncodingConfig<Encoding>) -> crate::Result<Vec<u8>> {
+fn encode_event(mut item: Event, encoding: &EncodingConfig<Encoding>, avro_schema: &Option<avro_rs::Schema>) -> crate::Result<Vec<u8>> {
     encoding.apply_rules(&mut item);
     let log = item.into_log();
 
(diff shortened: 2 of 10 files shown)


KannarFr:
I'd like to make schema type depending on codec type so make encoding initialization parse schema if avro and use it in encode_event.

src/sinks/pulsar.rs
```diff
@@ -249,6 +268,12 @@ fn encode_event(mut item: Event, encoding: &EncodingConfig<Encoding>) -> crate::
             .get(log_schema().message_key())
             .map(|v| v.as_bytes().to_vec())
             .unwrap_or_default(),
+        Encoding::Avro => {
+            let schema = avro_rs::Schema::parse_str(encoding.schema().as_ref().unwrap()).unwrap();
```
Hoverbear, line 272: Generally we try to avoid `unwrap()` calls, since this function returns a `crate::Result` you should be able to use like this?

```suggestion
            let schema = avro_rs::Schema::parse_str(encoding.schema().as_ref()?)?;
```

Hoverbear, line 272: I'm curious how long parsing the schema takes. Perhaps we should be storing the built schema at the sink level so it can be used repeatedly?

KannarFr: Yup, this was my point at https://github.com/timberio/vector/pull/5021#issuecomment-726785043, but put it to EncodingCondig is quite a lot of work. As Schema depends on transforms applied, I agree with you to put the built schema at the sink level.

KannarFr: Can't apply it as you suggest.

KannarFr: Done 160ffcb at the sink level, not sure how to do it using log_schema.

Hoverbear: Nicee!!!! It looks much better. :)

src/sinks/pulsar.rs
```diff
@@ -136,8 +156,20 @@ async fn healthcheck(producer: PulsarProducer) -> crate::Result<()> {
 
 impl PulsarSink {
     fn new(producer: PulsarProducer, encoding: EncodingConfig<Encoding>, acker: Acker) -> Self {
+        let schema = match &encoding.codec() {
+            Encoding::Avro => {
+                if let Some(schema) = &encoding.schema() {
+                    avro_rs::Schema::parse_str(schema).ok()
+                } else {
+                    None
```
Hoverbear, line 164: Hmmmm...  I think this `new` function should return a `crate::Result<Self>` now, this way we can return an error here if the user asks for avro without a schema.

```rust
    fn new(producer: PulsarProducer, encoding: EncodingConfig<Encoding>, acker: Acker) -> crate::Result<Self> {
        let schema = match &encoding.codec() {
            Encoding::Avro => {
                if let Some(schema) = &encoding.schema() {
                    avro_rs::Schema::parse_str(schema).ok()
                } else {
                    return Err("Avro requires a schema, specify a schema file with `encoding.schema`.".into())
                }
            }
            _ => None,
        };

        Ok(Self {
            encoding,
            avro_schema: schema,
            state: PulsarSinkState::Ready(Box::new(producer)),
            in_flight: FuturesUnordered::new(),
            acker,
            seq_head: 0,
            seq_tail: 0,
            pending_acks: HashSet::new(),
        })
    }
```

KannarFr: Correct!

src/sinks/pulsar.rs
```diff
@@ -249,6 +285,12 @@ fn encode_event(mut item: Event, encoding: &EncodingConfig<Encoding>) -> crate::
             .get(log_schema().message_key())
             .map(|v| v.as_bytes().to_vec())
             .unwrap_or_default(),
+        Encoding::Avro => {
+            let value = avro_rs::to_value(log).unwrap();
+            let resolved_value =
+                avro_rs::types::Value::resolve(value, avro_schema.as_ref().unwrap())?;
```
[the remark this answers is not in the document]
KannarFr: Applied.

KannarFr:
Invesgatigating


```
Nov 19 15:46:53.922 ERROR sink{component_kind="sink" component_name=pulsar component_type=pulsar}: vector::topology: An error occurred that vector couldn't handle.
```

KannarFr:
@Hoverbear fixed!

Hoverbear:
Great! @KannarFr can you help me by fixing the DCO?

```bash
git rebase HEAD~9 --signoff
git push --force-with-lease origin pulsar-avro
```

I also noted `cargo fmt` must be run.

Cargo.toml
```diff
@@ -196,6 +196,9 @@ anyhow = { version = "1.0.28" }
 snap = { version = "1.0.1", optional = true }
 dyn-clone = "1.0.3"
 
+# For AVRO
```
Hoverbear, line 199:
```suggestion
```

Cargo.toml
```diff
@@ -196,6 +196,9 @@ anyhow = { version = "1.0.28" }
 snap = { version = "1.0.1", optional = true }
 dyn-clone = "1.0.3"
```
Hoverbear, line 198:
```suggestion
```

src/sinks/pulsar.rs
```diff
@@ -182,7 +221,7 @@ impl Sink<Event> for PulsarSink {
             "Expected `poll_ready` to be called first."
         );
 
-        let message = encode_event(item, &self.encoding).map_err(|_| ())?;
+        let message = encode_event(item, &self.encoding, &self.avro_schema).map_err(|e| error!("Error during encoding: {}", e))?;
```
Hoverbear, line 224: Ah, sorry, this should be an internal event. Can you create a `pulsar.rs` in https://github.com/timberio/vector/tree/master/src/internal_events and create an error like: https://github.com/timberio/vector/blob/e769bccd78bbfa3098282b85567dad0e4f72763e/src/internal_events/remove_fields.rs#L13-L22 ?

Then, you can call it like so:

https://github.com/timberio/vector/blob/e769bccd78bbfa3098282b85567dad0e4f72763e/src/transforms/remove_fields.rs#L72-L74

Hoverbear (commented):
The rest of this PR looks good. :)

KannarFr:
Misscliked for the review request sorry JeanMertz (don't want to ping you more with @).

KannarFr:
@Hoverbear is ok? :)

KannarFr:
@Hoverbear rebased!

Cargo.toml
Hoverbear: Still an issue. :)

src/sinks/pulsar.rs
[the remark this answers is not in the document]
Hoverbear: I see a remaining unwrap! Were you unable to remove it?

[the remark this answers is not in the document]
KannarFr:
```
error[E0277]: the trait bound `std::option::NoneError: StdError` is not satisfied
   --> src/sinks/pulsar.rs:303:75
    |
303 |                 avro_rs::types::Value::resolve(value, avro_schema.as_ref()?)?;
    |                                                                           ^ the trait `StdError` is not implemented for `std::option::NoneError`
    |
    = note: required because of the requirements on the impl of `From<std::option::NoneError>` for `Box<dyn StdError + std::marker::Send + std::marker::Sync>`
    = note: required by `std::convert::From::from`
```

Not sure I can apply it smoothly, @Hoverbear WDYT?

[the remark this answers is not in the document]
Hoverbear: You're right! I'm sorry! I forgot this nuance!

Hoverbear (approved):
I note clippy is mad. If you could fix that I think this is good to merge!

Hoverbear (approved):

PR #5026, opened 2020-11-13
enhancement(file source): add internal option to ignore missing files

### LINKED ISSUE #3662: Spam `ERROR` logs from Vector when kubernetes_logs sees standard container restarts
### Current Vector Version
```
vector 0.11.0 (g75844bc x86_64-unknown-linux-gnu 2020-08-28)
```

### Use-cases
Viewing vector's own logs to see if the installation is working

### Attempted Solutions
Install vector slightly unusually (running in a systemd service on the host - but this isn't the cause of my issue), view its logs, get worried by lots of "noise" **ERRORs** in the logs like:
```
Sep 01 11:47:15.483 ERROR source{name=kubernetes_logs type=kubernetes_logs}:file_server: vector::internal_events::file: failed reading file for fingerprinting. path="/var/log/pods/platform_cluster-autoscaler-aws-cluster-autoscaler-78c69b78-jkpdl_cb3de7e0-67e1-416f-9c3e-e0ad2e640b67/aws-cluster-autoscaler/0.log" error=Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

### Proposal
K8s as part of it's standard behaviour (at least in v1.16+ using the common Docker container driver), creates symlinks in `/var/log/pods/NAMESPACE_PODNAME_HASH/CONTAINERNAME/`, called `0.log`, `1.log`, etc., directed to the actual log file owned by the docker daemon `/var/lib/docker/containers/HASH/HASH-json.log`.   It makes `0.log` for the first time it runs the container, and `1.log`, `2.log`, after any restarts - which can be a standard occurrence in the case of eventually consistent liveness/readiness checks.  Docker is then free to cleanup old containers to reclaim disk space, which will mean in time `0.log` becomes a dangling symlink - hence the error vector issues.  Logs were being correctly read from the `1.log` symlink as hoped in the given example.

I realise this is **very** nice-to-have, but it would be a better K8s UX if these expected situations did not yield ERROR level logs from Vector (drowning out other real ERRORs), if I accept the actual dangling symlink is an error for the low-level Vector file-logging component you make up the K8s feature from.

I would accept adding a new option "ignore_dangling_symlinks" that would get passed though from the kubernetes_logs code to vector's file scanning routines, but would prefer this default to true and not be a needed option at all.

lukesteensen:
Closes #3662

This doesn't add any user-facing config option yet, but it does turn off these errors for the kubernetes source.

### DIFF
--- a/lib/file-source/src/fingerprinter.rs
+++ b/lib/file-source/src/fingerprinter.rs
@@ -8,7 +8,13 @@ use std::{
 };
 
 #[derive(Clone)]
-pub enum Fingerprinter {
+pub struct Fingerprinter {
+    pub strategy: FingerprintStrategy,
+    pub ignore_not_found: bool,
+}
+
+#[derive(Clone)]
+pub enum FingerprintStrategy {
     Checksum {
         bytes: usize,
         ignored_header_bytes: usize,
@@ -60,14 +66,14 @@ impl Fingerprinter {
     ) -> Result<FileFingerprint, io::Error> {
         use FileFingerprint::*;
 
-        match *self {
-            Fingerprinter::DevInode => {
+        match self.strategy {
+            FingerprintStrategy::DevInode => {
                 let file_handle = File::open(path)?;
                 let dev = file_handle.portable_dev()?;
                 let ino = file_handle.portable_ino()?;
                 Ok(DevInode(dev, ino))
             }
-            Fingerprinter::Checksum {
+            FingerprintStrategy::Checksum {
                 ignored_header_bytes,
                 bytes,
             } => {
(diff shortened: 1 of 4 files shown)


lib/file-source/src/fingerprinter.rs
```diff
@@ -100,13 +106,19 @@ impl Fingerprinter {
...
+            .map_err(|error| match error.kind() {
+                io::ErrorKind::UnexpectedEof => {
                     if !known_small_files.contains(path) {
                         emitter.emit_file_checksum_failed(path);
                         known_small_files.insert(path.clone());
                     }
-                } else {
+                }
+                io::ErrorKind::NotFound => {
+                    if self.ignore_not_found {
```
MOZGIII, line 117:
```suggestion
                    if !self.ignore_not_found {
```

lukesteensen: I double negatived myself when I needed a triple negative :facepalm:

MOZGIII (approved):
LGTM, but let's wait for E2E tests to complete

PR #5027, opened 2020-11-13
fix(topology): Fix fanout remove-while-iterating bug

### LINKED ISSUE #4457: Add test coverage of fanout failure behavior
c.f. #4455 

When multiple sinks are enabled, Vector sends events to all downstream sinks. If a single sink fails then Vector's fanout behavior is to cease sending to the failed sink and continue sending to any additional functioning sinks.

See https://github.com/timberio/vector/blob/master/src/topology/fanout.rs#L74 from https://github.com/timberio/vector/pull/214.

We should ensure test coverage of this behavior.

bruceg:
Closes #4457

Are there other cases of fanout error handling that should be covered?


lukesteensen (approved):
Great catch! You seem to have the cases covered as far as I can tell.

bruceg:
> Great catch! You seem to have the cases covered as far as I can tell.

Not quite, I still got it wrong when more than one sink has errors (fixed now).

ktff (approved):

PR #5050, opened 2020-11-16
fix(http sink): Fix JSON compression

### LINKED ISSUE #4280: HTTP sink generates corrupted gzip data with json encoding
If you configure the http sink to use gzip compression with json encoding, the data the sink generates is not valid gzip compressed data.

You can reproduce this by changing the `http_happy_path_post` test to configure it with `json` rather than `ndjson`.

The issue is caused [here](https://github.com/timberio/vector/blob/master/src/sinks/http.rs#L208). The `body` that `build_request` gets is gzipped compressed data, and the sink is just wrapping this in `[...]`, the brackets are not being passed through the compressor and thus corrupts the data.  

### Vector Version

master

### Vector Configuration File

```
[sinks.http]
  inputs = ["socket"]
  type = "http"
  encoding.codec = "json"
  compression = "gzip"
  uri = "http://localhost:8888"
  batch.max_events = 5
  batch.timeout_secs = 10
```

### Expected Behavior

The sink should either error saying you can't combine json and gzip, or it should gzip the json correctly.

### Actual Behavior

Corrupted data is produced.

### Example Data

Compressed data surrounded by `[...]`

```
[^_<8b>^H^@^@^@^@^@^Dÿ<9d><90>A
<83>0^PE¯R²Ö0<99>$Nâ9ºê¦<88>^D-VGLº(¥wïx<82><8a>^L³y|þ<83>ÿQ#ç¢Ze<90>4È^YU
©9åÜ^MIèÂÛ$ ókëÓ½¼×^]fî§T^D<97><87>$K7¯<92>D@¨^MÔ<80>W4­¥ÖDí  ¢ñò7õ­þ<9a>x<90>Ê3&^DíÀº<88>Î<9b>C¦mz¦K^^^W>¯´:<92>#²<80>ö¸rááì<9a>^X4ù^P<9a>àC³^K^?<9d>·ãä¸^A^@]
```

### References

Related to issues discussed in this pull request #4164 regarding adding gzip compression to the Datadog log sink.

juchiast:
close #4280 

Same approach as datadog_logs sink.

### DIFF
--- a/src/sinks/http.rs
+++ b/src/sinks/http.rs
@@ -3,6 +3,7 @@ use crate::{
     event::Event,
     http::{Auth, HttpClient},
     sinks::util::{
+        buffer::compression::GZIP_DEFAULT,
         encoding::{EncodingConfig, EncodingConfiguration},
         http::{BatchedHttpSink, HttpSink},
         BatchConfig, BatchSettings, Buffer, Compression, InFlightLimit, TowerRequestConfig,
@@ -10,6 +11,7 @@ use crate::{
     },
     tls::{TlsOptions, TlsSettings},
 };
+use flate2::write::GzEncoder;
 use futures::{future, FutureExt, SinkExt};
 use http::{
     header::{self, HeaderName, HeaderValue},
@@ -20,6 +22,7 @@ use indexmap::IndexMap;
 use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
 use snafu::{ResultExt, Snafu};
+use std::io::Write;
 
 #[derive(Debug, Snafu)]
 enum BuildError {
@@ -123,7 +126,6 @@ impl SinkConfig for HttpSinkConfig {
         let mut config = self.clone();
         config.uri = build_uri(config.uri.clone()).into();
 
-        let compression = config.compression;
         let batch = BatchSettings::default()
             .bytes(bytesize::mib(10u64))
             .timeout(1)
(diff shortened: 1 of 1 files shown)


src/sinks/http.rs
```diff
@@ -227,8 +229,13 @@ impl HttpSink for HttpSinkConfig {
             .uri(uri)
             .header("Content-Type", ct);
 
-        if let Some(ce) = self.compression.content_encoding() {
-            builder = builder.header("Content-Encoding", ce);
+        if let Compression::Gzip(level) = self.compression {
```
jszwedko, line 232: Could we make this a `match` to ensure we handle any additional compression formats we add in the future?

jszwedko (approved):
Looks good! Just one note about guarding against future issues if we add more compression formats.

src/sinks/http.rs
fanatid: I think `clippy` will throw errors for match with 1 arm..

jszwedko: It seems like it could have at least two arms though, no? `Compression::Gzip(level)` and `None`.

fanatid:
```rust
        match self.compression {
            Compression::Gzip(level) => {
                builder = builder.header("Content-Encoding", "gzip");

                let level = level.unwrap_or(GZIP_DEFAULT) as u32;
                let mut w = GzEncoder::new(Vec::new(), flate2::Compression::new(level));
                w.write_all(&body).expect("Writing to Vec can't fail");
                body = w.finish().expect("Writing to Vec can't fail");
            }
            Compression::None => {}
        }
```
works for clippy :+1:

fanatid (approved):

src/sinks/http.rs
jszwedko: :+1: that would assuage my concern that we forget to update this if we add another supported compression format.

PR #5052, opened 2020-11-16
chore(remap): simplify value type checking

JeanMertz:
This PR is an enhancement to #4902.

It changes `value::Kind` to support bitwise operations. This in turn simplifies the logic around value constraints and allows us to remove the `value::Constraint` type entirely. It also means `Kind` and `TypeDef` are now `Copy`.


FungusHumungus (approved):

FungusHumungus:
There's a few clippy errors. I'm not sure at what point they crept in!

PR #5053, opened 2020-11-16
enhancement(remap): expressions no longer return an option

JeanMertz:
This PR changes the expression trait `execute` function to return `Result<Value>` instead of the old `Result<Option<Value>>`.

This came up in multiple discussions, and with the infallible expressions RFC (#4840) stalled for now, removing the extra indirection makes things simpler to work with.

The user-facing implication is that an if-statement without an else block resolves to `null` if the conditional does not match. This was also discussed in several places and deemed to be an acceptable (or even good) trade-off.

E.g.:

```rust
// before
.foo = if false { "foo" } // .foo assignment is skipped

// after
.foo if false { "foo" } // .foo == null

// after, same result as before
if false { .foo = "foo" } //.foo assignment is skipped
```

_note: I'll be updating the `TypeDef` implementations in a follow-up PR (basically removing all references to `optional`)_

### DIFF
--- a/lib/remap-lang/src/expression/block.rs
+++ b/lib/remap-lang/src/expression/block.rs
@@ -12,18 +12,12 @@ impl Block {
 }
 
 impl Expression for Block {
-    fn execute(
-        &self,
-        state: &mut state::Program,
-        object: &mut dyn Object,
-    ) -> Result<Option<Value>> {
-        let mut value = None;
-
-        for expr in &self.expressions {
-            value = expr.execute(state, object)?;
-        }
-
-        Ok(value)
+    fn execute(&self, state: &mut state::Program, object: &mut dyn Object) -> Result<Value> {
+        self.expressions
+            .iter()
+            .map(|expr| expr.execute(state, object))
+            .last()
+            .unwrap_or(Ok(Value::Null))
     }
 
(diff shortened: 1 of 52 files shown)


lib/remap-lang/src/expression/block.rs
```diff
@@ -12,18 +12,12 @@ impl Block {
...
-            value = expr.execute(state, object)?;
-        }
-
-        Ok(value)
+    fn execute(&self, state: &mut state::Program, object: &mut dyn Object) -> Result<Value> {
+        self.expressions
+            .iter()
+            .map(|expr| expr.execute(state, object))
+            .last()
+            .unwrap_or(Ok(Value::Null))
```
FungusHumungus, line 20: This is going to ignore any errors that occur in all but the last expression, which is not what we want.

StephenWakely (changes requested):
This definitely simplifies things. Just that one thing which looks like a bug..

lib/remap-lang/src/expression/block.rs
JeanMertz: Good catch! 👀 will fix.

bruceg (commented):
Looks good except for the check failures and the issue @FungusHumungus noted.

PR #5054, opened 2020-11-16
chore(remap): remove required! and optional! macros

JeanMertz:
With the Expression trait updated to remove `None` results in #5053, the usage of the `required!` and `optional!` macro's made functions more difficult to parse as a reader than not using them.

They hid so little implementation detail, and code within them wasn't being formatted by rustfmt, which made me decide to remove them entirely.

I think this improves readability of expressions, at the cost of sometimes one or two more lines of code within the expressions themselves.

Additionally, a few PRs back, helper functions were added to the `Value` type, which makes it easier to get a specific value kind through `{as,as_mut,try,unwrap}_{string,integer,...}`, that the resulting code is often easier to write and understand than with the macros.

Another added benefit is that we were kind of fast-and-loose with resolving `Cow<str>` to an owned string, as it made using the macros easier (due to lifetime issues). Without the macros, I felt more pushed toward only cloning the string if needed, which removed extra allocations in quite a few function implementations.

_note that in a few places I use `let bytes = ...try_string()?;`, which is obviously awkward. @FungusHumungus has requested this on multiple occasions, so in a follow-up PR I'll be renaming `Value::String` to `Value::Bytes`._

### DIFF
--- a/src/remap/function/ceil.rs
+++ b/src/remap/function/ceil.rs
@@ -47,16 +47,16 @@ impl CeilFn {
 
 impl Expression for CeilFn {
     fn execute(&self, state: &mut state::Program, object: &mut dyn Object) -> Result<Value> {
-        let precision =
-            optional!(state, object, self.precision, Value::Integer(v) => v).unwrap_or(0);
-        let res = required!(state, object, self.value,
-                            Value::Float(f) => {
-                                Value::Float(round_to_precision(f, precision, f64::ceil))
-                            },
-                            v@Value::Integer(_) => v
-        );
+        let precision = match &self.precision {
+            Some(expr) => expr.execute(state, object)?.try_integer()?,
+            None => 0,
+        };
 
-        Ok(res)
+        match self.value.execute(state, object)? {
+            Value::Float(f) => Ok(round_to_precision(f, precision, f64::ceil).into()),
+            v @ Value::Integer(_) => Ok(v),
+            _ => unreachable!(),
+        }
     }
 
     fn type_def(&self, state: &state::Compiler) -> TypeDef {
--- a/lib/remap-lang/src/expression/if_statement.rs
+++ b/lib/remap-lang/src/expression/if_statement.rs
@@ -1,4 +1,3 @@
-use super::Error as E;
 use crate::{state, value, Expr, Expression, Object, Result, TypeDef, Value};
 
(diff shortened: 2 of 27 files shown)


FungusHumungus (approved):
This is well worth it to avoid the extra string allocations!

src/remap/function/ceil.rs
```diff
@@ -47,16 +47,16 @@ impl CeilFn {
...
-        Ok(res.into())
+        let precision = match &self.precision {
+            Some(expr) => expr.execute(state, object)?.try_integer()?,
+            None => 0,
+        };
+
+        match self.value.execute(state, object)? {
+            Value::Float(f) => Ok(round_to_precision(f, precision, f64::ceil).into()),
+            v @ Value::Integer(_) => Ok(v),
+            _ => unreachable!(),
```
jszwedko, line 58: Does the compile-time checking validate that it is only ever called with expressions resolving to floats or integers? I still need to review that PR :sweat_smile:

jszwedko (approved):
Looks good! Definitely more straight-forward.

src/remap/function/ceil.rs
JeanMertz: No, but we have a separate runtime check to do that (although I'm looking into merging the two). For `ceil`, check the [`parameters()` function here](https://github.com/timberio/vector/blob/37719591828ec599bd3d93a8e1baf2defc2932ec/src/remap/function/ceil.rs#L12-L25).

The `accepts` function runs before passing an argument to the function, and it [fails if the function returns `false`](https://github.com/timberio/vector/blob/37719591828ec599bd3d93a8e1baf2defc2932ec/lib/remap-lang/src/expression/function.rs#L185-L191).

JeanMertz: The reason this happens at runtime right now, is that it would be fairly annoying to have to wrap all fields with `to_*` functions to coerce them.

We'll start to introduce more strict checks where it makes sense, and with https://github.com/timberio/vector/pull/5060 we might potentially do compile-time checks that accept only a specific type _or_ "any" (so that paths don't have to be coerced).

FungusHumungus: I had to double check that this worked when looking at this PR, because it isn't obvious that this is what is happening. I can confirm that it worked! The `Box<dyn Expression>` that `self.value` is here will be the `Argument` struct and that does the type check on execution.

PR #5056, opened 2020-11-16
enhancement(remap): undefined path or variable return null

JeanMertz:
This changes path resolution to resolve to `null` if the path cannot be found, instead of returning a runtime error. The same applies to variables.

I also removed a lot of useless function tests that were really testing the `Path` expression, instead of the function itself.

ref #4947.


StephenWakely (approved):
With undefined paths returning null, I imagine a common use case will be to coalesce paths with the `|` operator. I think we should ensure this area is well tested..

It seems all the arithmetic are largely in as behaviour tests. It would be nice to see some path coalescing tests there as well.

PR #5057, opened 2020-11-16
chore(remap): remove "optional" from TypeDef

JeanMertz:
This is a follow-up to #5053. In this PR, the `optional` field is removed from `TypeDef`, as expressions can no longer return `None`.


FungusHumungus (approved):
Some CI fixes are needed.

This definitely saves us from having to worry about what this field should actually be called!

PR #5059, opened 2020-11-16
enhancement(remap): improve arithmetic type checking

JeanMertz:
As mentioned by @FungusHumungus in https://github.com/timberio/vector/pull/4902#discussion_r523466906, this PR updates our arithmetic type checking to be more precise.

It also fixes a few other inconsistencies with arithmetic operations.

### DIFF
--- a/lib/remap-lang/src/expression/arithmetic.rs
+++ b/lib/remap-lang/src/expression/arithmetic.rs
@@ -49,7 +49,7 @@ impl Expression for Arithmetic {
 
         match self.op {
             Or if lhs_def.kind.is_null() => rhs_def,
-            Or if !lhs_def.kind.is_null() && !lhs_def.kind.is_boolean() => lhs_def,
+            Or if !lhs_def.kind.is_boolean() => lhs_def,
             Or => type_def,
             And if lhs_def.kind.is_null() => lhs_def.with_constraint(Kind::Boolean),
             And => type_def
--- a/lib/remap-lang/src/expression/arithmetic.rs
+++ b/lib/remap-lang/src/expression/arithmetic.rs
@@ -26,10 +26,7 @@ impl Expression for Arithmetic {
             Divide => lhs.try_div(rhs),
             Add => lhs.try_add(rhs),
             Subtract => lhs.try_sub(rhs),
-
-            // TODO: make `Or` infallible, `Null`, `false` and `None` resolve to
-            // rhs, everything else resolves to lhs
-            Or => lhs.try_or(rhs),
+            Or => Ok(lhs.or(rhs)),
             And => lhs.try_and(rhs),
             Remainder => lhs.try_rem(rhs),
             Equal => Ok(lhs.eq_lossy(&rhs).into()),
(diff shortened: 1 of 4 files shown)


lib/remap-lang/src/expression/arithmetic.rs
```diff
@@ -46,16 +43,28 @@ impl Expression for Arithmetic {
...
+        let lhs_def = self.lhs.type_def(state);
+        let rhs_def = self.rhs.type_def(state);
+        let type_def = lhs_def | rhs_def;
 
-        TypeDef {
-            fallible: true,
-            kind,
+        match self.op {
+            Or if lhs_def.kind.is_null() => rhs_def,
+            Or if !lhs_def.kind.is_null() && !lhs_def.kind.is_boolean() => lhs_def,
```
FungusHumungus, line 52: The ` !lhs_def.kind.is_null() ` is redundant here.

lib/remap-lang/src/expression/arithmetic.rs
```diff
@@ -46,16 +43,28 @@ impl Expression for Arithmetic {
...
+            Equal | NotEqual => type_def.with_constraint(Kind::Boolean),
+            Greater | GreaterOrEqual | Less | LessOrEqual => type_def
+                .fallible_unless(Kind::Integer | Kind::Float)
+                .with_constraint(Kind::Boolean),
+            Subtract | Divide | Remainder => type_def
+                .fallible_unless(Kind::Integer | Kind::Float)
+                .with_constraint(Kind::Integer | Kind::Float),
+            Multiply | Add => type_def
+                .fallible_unless(Kind::Bytes | Kind::Integer | Kind::Float)
+                .with_constraint(Kind::Bytes | Kind::Integer | Kind::Float),
```
FungusHumungus, line 67: I don't think this is quite what I was suggesting. `with_constraint` will override all the constraints of `type_def`. But, if we know that `type_def` is `Kind::Bytes`, then it should stay as `Kind::Bytes` and not have `Kind::Integer | Kind::Float` added on.

So 

`"some text" + "more text"` should constrain to just `Kind::Bytes`.
`.foo + "some text"` should be `Kind::Bytes`.
`3 + .foo` should be `Kind::Integer`
`.foo + .bar` should be `Kind::Bytes | Kind::Integer | Kind::Float`  because we have no further info on what each side should be.

I don't think it is a disaster with the current way, but I don't think there is much of a cost to being more precise here.

jszwedko: I agree. It would be nice to restrict the type when we can.

lib/remap-lang/src/value.rs
```diff
@@ -332,33 +332,31 @@ impl Value {
...
             Value::Null => rhs,
-            Value::Boolean(lhv) => match rhs {
-                Value::Boolean(rhv) => (lhv || rhv).into(),
-                _ => return Err(err()),
-            },
-            _ => return Err(err()),
-        };
-
-        Ok(value)
+            Value::Boolean(lhv) if lhv == false => rhs,
```
jszwedko, line 342: Just calling attention to the failing check.

jszwedko (approved):
Looks good. I agree with @FungusHumungus though, that it'd be nice to restrict the types when we know the types of both operands.

lib/remap-lang/src/expression/arithmetic.rs
JeanMertz: Good point, I'll make the change.

JeanMertz: I decided against changing this in this PR — given that the changes in this PR already move us in the right direction, and this PR has been blocking other pending work — but I agree we should make further improvements in a follow-up PR.

I filed #5246 as a tracking issue.

PR #5064, opened 2020-11-16
feat(remap): support query/assignment of object root

### LINKED ISSUE #5009: Allow assignment to `.` for `remap`
### Current Vector Version

```
vector 0.11.0 (gbc2533d x86_64-unknown-linux-gnu 2020-11-12)
```

### Use-cases

Flattening an entire payload using `flatten`.

This was requested in discord: https://discord.com/channels/742820443487993987/746070591097798688/776522817952284692

I thought I might be able to do:

``toml
[sources.stdin]
  type = "stdin"

[transforms.json]
  type = "json_parser"
  inputs = ["stdin"]

[transforms.remap]
  # General
  type = "remap" # required
  inputs = ["json"] # required

  source = """
    . = flatten(.)
  """

[sinks.console]
  type = "console"
  inputs = ["json"]
  encoding.codec = "json"
```

But it doesn't like that:

```
./target/debug/vector --config /tmp/tmp.toml
Nov 12 19:48:32.248  INFO vector::app: Log level is enabled. level="info"
Nov 12 19:48:32.249  INFO vector::app: Loading configs. path=["/tmp/tmp.toml"]
Nov 12 19:48:32.263  WARN vector::config::compiler: Transform "remap" has no consumers
Nov 12 19:48:32.264  INFO vector::sources::stdin: Capturing STDIN.
Nov 12 19:48:32.264 ERROR vector::topology: Configuration error. error=Transform "remap": remap error: parser error:  --> 1:6
  |
1 |     . = flatten(.)␊
  |      ^---
  |
  = expected path_segment
```

Notably we might want to disallow assignment of anything other than an object (like `. = 5`).

<!---
In order to properly evaluate a feature request, it is necessary to understand the use-cases for it.

Please describe below the _end goal_ you are trying to achieve that has led you to request this feature.

Please keep this section focused on the problem and not on the suggested solution. We'll get to that in a moment, below!
-->

### PR DESCRIPTION
**Pull request by JeanMertz**
This change enables assigning to, or fetching the root of an event.

This is now possible:

```rust
. = .foo   // will error if .foo isn't a map
$foo = .   // stores a map value containing the root fields and their values
```

I had to hack the parser rules a bit to make this work, and have useable error messages. That is to say, technically, you can write:

```rust
.foo= "bar" // equal to:
.foo = "bar"
```

Compared to:

```rust
.= .foo  // parser expects a space after a root `.`
. = .foo // ok
```

I'll probably circle back to this in the future to see if we can make it properly parse `.` and `.foo` without requiring the space, and without creating confusing error messages, but I deemed the `.= .foo` syntax enough of an outlier that I'm happy to commit this as-is.

If others feel differently, I'm happy to keep this PR open until that issue is resolved, though.

closes #5009.

### DIFF
--- a/lib/remap-lang/grammar.pest
+++ b/lib/remap-lang/grammar.pest
@@ -41,7 +41,8 @@ operator_not            = { "!" }
 
 // Paths -----------------------------------------------------------------------
 
-path             = ${ ("." ~ path_segment)+ }
+path             = ${ path_root | ("." ~ path_segment)+ }
+path_root        = ${ ". " }
 path_segment     = ${ (path_field | path_coalesce) ~ path_index* }
 path_field       = ${ ident | string }
 path_coalesce    = !{ "(" ~ path_field ~ ("|" ~ path_field)+ ~ ")" }
--- a/lib/remap-lang/src/error.rs
+++ b/lib/remap-lang/src/error.rs
@@ -93,6 +93,7 @@ impl fmt::Display for Rule {
             path_field,
             path_index,
             path_index_inner,
+            path_root,
             path_segment,
             primary,
             program,
--- a/lib/remap-lang/src/lib.rs
+++ b/lib/remap-lang/src/lib.rs
@@ -237,6 +237,7 @@ mod tests {
             (r#"null || false"#, Ok(()), Ok(false.into())),
             (r#"false || null"#, Ok(()), Ok(().into())),
             (r#"null || "foo""#, Ok(()), Ok("foo".into())),
+            (r#". = "bar""#, Ok(()), Ok("bar".into())),
         ];
 
         for (script, compile_expected, runtime_expected) in cases {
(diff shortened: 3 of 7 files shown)


lib/remap-lang/grammar.pest
```diff
@@ -41,7 +41,8 @@ operator_not            = { "!" }
 
 // Paths -----------------------------------------------------------------------
 
-path             = ${ ("." ~ path_segment)+ }
+path             = ${ path_root | ("." ~ path_segment)+ }
+path_root        = ${ ". " }
```
FungusHumungus, line 45: I think `path_root = ${"." ~ (!"=")?}` might do it.

StephenWakely (approved):
All good apart from the CI errors!

Hoverbear (approved):
Other than @FungusHumungus 's comment this looks good. :)

lib/remap-lang/grammar.pest
JeanMertz: This unfortunately isn't sufficient. However, I am about to put up a PR that makes this better (although still not 100% correct).

I'll get this one in as-is, and have the other PR iterate on it further.

PR #5066, opened 2020-11-16
chore(ci): Fix hanging splunk_hec source partial test

### LINKED ISSUE #5058: splunk_hec source `partial` test is hanging
Causing CI to hang on master.

```
TEST_LOG=info cargo test --lib --no-default-features --features sources-splunk_hec,sinks-splunk_hec sources::splunk_hec::tests::partial -- --nocapture
```

```
Nov 16 16:16:46.296 ERROR vector::internal_events::splunk_hec::source: Invalid request body. error=Custom { kind: UnexpectedEof, error: Error("EOF while parsing an object", line: 1, column: 17) } rate_limit_secs=10
Nov 16 16:16:46.296 ERROR vector::internal_events::splunk_hec::source: Error processing request. error=InvalidDataFormat { event: 1 } rate_limit_secs=10
```

jszwedko:
Fixes #5058 

So this fixes the hanging test, but I feel like it isn't the right fix. Seemingly, if only one event is published, `collect_n()` fails to ever poll an event. Pushing this up to get @Hoverbear's thoughts.

The hanging behavior was introduced by #4875

### DIFF
--- a/src/sources/splunk_hec.rs
+++ b/src/sources/splunk_hec.rs
@@ -1056,7 +1056,7 @@ mod tests {
     async fn partial() {
         trace_init();
 
-        let message = r#"{"event":"first"}{"event":"second""#;
+        let message = r#"{"event":"first"}{"event":"second"}{"event": ""#;
         let (source, address) = source().await;
 
         assert_eq!(


src/sources/splunk_hec.rs
```diff
@@ -1056,7 +1056,7 @@ mod tests {
     async fn partial() {
         trace_init();
 
-        let message = r#"{"event":"first"}{"event":"second""#;
+        let message = r#"{"event":"first"}{"event":"second"}{"event": ""#;
```
Hoverbear, line 1059: I think this test breaking here is a sign we have done something wrong. This change seems to invalidate the assertion of the test?

jszwedko: Yeah, exactly :smile: I'm pretty sure there was an issue introduced in #4875 where it won't forward a single event until another is published. I was hoping you might be able to take a look.

Apologies, I could have made it clearer in the PR description that I don't think this is the right fix. I pushed it just for demonstration purposes.

Hoverbear: Ah ok, let me investigate.

jszwedko:
Superceded by #5069

PR #5074, opened 2020-11-17
enhancement(observability): Rename `version` -> `versionString` in GraphQL schema

leebenson:
Small amendment to replace `{ meta { version } }` with `{ meta { versionString } }`, to accommodate an upgrade path where an eventual `version` returns semantic parts.

Also API client support via a new `MetaQueryExt` trait + extension methods, and a test.


jszwedko (approved):
Looks great!

PR #5076, opened 2020-11-17
chore: use PartitionBatchSink in BatchSink

fanatid:
This addition to #4888. `BatchSink` and `PartitionBatchSink` have the same logic in `Sink` implementation. In the same time `BatchSink` can use `PartitionBatchSink` if all keys will have, for example `()`.


lukesteensen (commented):
Code looks good! Do you think the shutdown test failure is related or independent?

fanatid:
Failed test is independent, I think this fail from previous week :disappointed: 
https://discord.com/channels/742820443487993987/751109034827972709/778647217275797575

lukesteensen (approved):

PR #5079, opened 2020-11-17
chore(file source): move checkpoint persistence to separate task

### LINKED ISSUE #5078: Move file source checkpoint persistence to its own task
From the [File Source RFC](https://github.com/timberio/vector/blob/master/rfcs/2020-10-15-3480-file-source-rework.md)

The writing of file checkpoints to disk should be removed from the main read loop of the file source and moved into an independent task. This will isolate the read loop from potential performance problems in checkpointing, as well as allow us to specify the persistence interval more precisely.

lukesteensen:
Closes #5078

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1726,6 +1726,7 @@ dependencies = [
  "bytes 0.5.6",
  "chrono",
  "crc",
+ "dashmap",
  "flate2",
  "futures 0.3.5",
  "glob 0.3.0",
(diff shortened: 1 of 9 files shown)


MOZGIII (approved):
The code looks good. I'm a bit concerned about the possible race conditions, need some more time to find if it's the case or not - but I'll approve, as it shouldn't be a blocker looking at the high-level algorithm.

Would be great to add a test case with frozen time and code advancing step by step and us asserting that the state is valid and matches the expectation after each step.

PR #5082, opened 2020-11-17
fix(topology): Flush sinks in a couple of more places

### LINKED ISSUE #5081: Survey stream frowarding for missing flushes
Broken off from #5069 

We should do a survey for other places where errorable streams are being forwarded (via `Stream.forward` or `Sink.send_all`) for missing flushes.

jszwedko:
Fixes #5081

I surveyed the other usages of `Stream.forward` and `Sink.send_all`
looking for other places that should be flushing. I just found these
two, the rest follow two patterns:

* `Sink.send_all` / `Stream.forward` used with streams only containing
  `Ok`s
* Consuming error in stream and returning `None` to end the stream

### DIFF
--- a/src/sinks/statsd.rs
+++ b/src/sinks/statsd.rs
@@ -244,7 +244,7 @@ mod test {
     use super::*;
     use crate::{event::Metric, test_util::*};
     use bytes::Bytes;
-    use futures::{compat::Sink01CompatExt, TryStreamExt};
+    use futures::{compat::Sink01CompatExt, SinkExt, TryStreamExt};
     use futures01::sync::mpsc;
     use tokio::net::UdpSocket;
     use tokio_util::{codec::BytesCodec, udp::UdpFramed};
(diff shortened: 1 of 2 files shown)


bruceg (approved):

lukesteensen (approved):

fanatid:
Maybe I miss something but [forward](https://docs.rs/futures-util/0.3.6/src/futures_util/stream/stream/forward.rs.html#68) and [send_all](https://docs.rs/futures-util/0.3.6/src/futures_util/sink/send_all.rs.html#102) do flushing. Or issue related with `compat` layer?

jszwedko:
@fanatid they do, but only if the stream being sent completes without error; that is, if doesn't error when polled. This was the original error we saw with the `splunk_hec` source (#5058) where, when the stream had an error, the previous event on that stream was never flushed (`poll_complete()` was never called on `Pipeline`).

Issue #5042, opened 2020-11-15
Restructure generated Kubernetes YAMLs to match Helm charts


MOZGIII:
We have a number of Helm charts, and we need to provide YAML files and `Kustomization` bases for each of the Helm charts (based on the deployment modes):

Before 0.11 release (done):

- prepare the structure and only ship `vector-agent`-based YAMLs

After 0.11:

- `vector`
- `vector-aggregator`

MOZGIII:
Closed via #4404.

PR #5087, opened 2020-11-17
chore(kubernetes platform): Rework Kubernetes YAMLs to present multiple deployment modes

MOZGIII:
Implements the first part of the #5042, does not close it yet.

After this is complete, the https://github.com/timberio/vector-k8s-examples needs to be updated too.


distribution/kubernetes/vector-all/resources.yaml
```diff
@@ -117,7 +117,7 @@ spec:
         - name: vector
           securityContext:
             {}
-          image: "timberio/vector:latest-alpine"
```
eeyun, line 120: I'm curious why the switch from alpine to debian?

scripts/kubernetes-yaml/vector-all/values.yaml
```diff
@@ -2,8 +2,6 @@ vector-agent:
   image:
     # Use the latest release image.
     version: latest
-    # Use the alpine flavor of vector image.
-    base: alpine
```
eeyun, line 6: Is this fine to just remove or should it be replaced with debian as well?

eeyun (approved):
Changes look fine other than a couple simple questions.

distribution/kubernetes/vector-all/resources.yaml
MOZGIII: `alpine` uses `musl`, and musl has some perf issues. `debian` uses `glibc`.
`debian` shown about 20%  higher throughput than `alpine` in our load tests.

scripts/kubernetes-yaml/vector-all/values.yaml
MOZGIII: Just removing it is sufficient. The semanics here is we want to use whatever the default is in the Helm chart, and avoid special-casing for the geneated YAMLs

PR #5095, opened 2020-11-18
enhancement(elasticsearch sink): Support basic-auth credentials in endpoint configuation

### LINKED ISSUE #5019: Elasticsearch: support basic-auth credentials in "host" URL
### Current Vector Version
```
vector 0.10.0 (g0f0311a x86_64-unknown-linux-musl 2020-07-22)
```

### Use-cases

I'm trying to configure the elasticsearch sink's hostname and authentication using environment variables.

### Attempted Solutions

I can configure the sink like this:
```toml
[sinks.my_elasticsearch_sink]
type = "elasticsearch"
[...]
host = "${ES_HOST}"
auth.user = "${ES_USER}"
auth.password = "${ES_PASSWORD}"
```

### Proposal

Allow to include basic-auth credentials in the URL passed to the "host" field.
```
https://elastic:password@example.com:9200
```

So the config would be simplified to:
```toml
[sinks.my_elasticsearch_sink]
type = "elasticsearch"
[...]
host = "${ES_HOST}"
```

### References

Passing basic-auth in the "host" is supported by Logstash.  
It makes configuration a bit easier, because a single field contains all necessary information to communicate to the Elasticsearch server, so only a single field needs to be changed when changing the server.

juchiast:
close #5019 


jszwedko (approved):
:+1:

I was a bit surprised that `hyper` doesn't handle that for you, but this does appear to be the case.

It would be nice to generalize this HTTP configuration at some point so that we have consistent handling across our sinks and sources rather than needing to continue adding this sort of functionality piece-meal. For example, this adds support to including the username/password in the URI for the `elasticsearch` sink, but not to the `http` sink.

Maybe we could do that as part of https://github.com/timberio/vector/issues/3676

src/sinks/elasticsearch.rs
```diff
@@ -483,6 +478,24 @@ fn maybe_set_id(key: Option<impl AsRef<str>>, doc: &mut serde_json::Value, event
     }
 }
 
+fn get_and_strip_basic_auth(endpoint: &str) -> Option<(String, String)> {
```
bruceg, line 481: I'm not sure this is an improvement over returning `Option<(String, ElasticSearchAuth)>` and restore the apply functionality above.

juchiast: I agree, but `ElasticSearchAuth::apply` was not being used, so I went with a minimal change instead.

juchiast: I moved `get_and_strip_basic_auth` to `src/http.rs`, it now returns `Auth`.

src/http.rs
```diff
@@ -181,18 +185,92 @@ pub enum Auth {
 }
 
 impl Auth {
+    pub fn get_and_strip_basic_auth(url: &str) -> Option<(String, Self)> {
```
bruceg, line 188: Given the usage of this, a return type of `(String, Option<Self>)` might work better. You already make the relevant decisions here, so it would avoid needing to do the `match` in the caller. Of course, then you couldn't use `?` directly, at least not without breaking this into two parts.

bruceg (commented):
This looks much better. I have just one thought about the return from `get_and_strip_basic_auth`.

PR #5096, opened 2020-11-18
chore: handle TOO_MANY_REQUESTS in AWS sinks

fanatid:
Discussin in sqs sink: https://github.com/timberio/vector/pull/4675#discussion_r525282085

### DIFF
--- a/src/rusoto/mod.rs
+++ b/src/rusoto/mod.rs
@@ -11,7 +11,7 @@ use rusoto_core::{
     request::{
         DispatchSignedRequest, DispatchSignedRequestFuture, HttpDispatchError, HttpResponse,
     },
-    ByteStream, Region,
+    ByteStream, Region, RusotoError,
 };
 use rusoto_credential::{
     AutoRefreshingProvider, AwsCredentials, ChainProvider, CredentialsError, ProvideAwsCredentials,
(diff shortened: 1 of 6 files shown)


jszwedko (approved):
:+1: nice refactoring!

src/rusoto/mod.rs
```diff
@@ -334,3 +334,16 @@ impl From<Option<SignedRequestPayload>> for RusotoBody {
         RusotoBody { inner }
     }
 }
+
+pub fn is_retriable_error<T>(error: &RusotoError<T>) -> bool {
```
bruceg, line 338: This seems like it would be better as:
```rust
pub impl RusotoError<T> {
    pub fn is_retriable_error(&self) -> bool;
}
```

bruceg (commented):
Agreed with @jszwedko nice refactor, with one suggestion.

src/rusoto/mod.rs
fanatid: Is this okay to extend crate? (in case of `vector` can be crate itself)

bruceg: No, you're right, that's an external type so we can't impl for it. I misread this as for the `AwsRusotoError`.

PR #5101, opened 2020-11-18
fix(tests): Fix flaky `timely_shutdown_docker` test

### LINKED ISSUE #5046: Flaky `timely_shutdown_docker`
The test is failing because started vector instance exited with non 0 exit code. It isn't config since then it wouldn't be flaky. Also one clue is the lack of shutdown logs. That mostly points to either crash, or more probably we never actually started which narrows the problem space.

Another possibility is that `docker` is misbehaving on CI, but in that case we should make source resistant to it.

https://github.com/timberio/vector/runs/1403149743

bruceg:
Give vector an extra second to start up.

Closes #5046


jszwedko (approved):
It looks like this worked! Nice sleuthing.

PR #5111, opened 2020-11-19
chore(http sink): add internal events

fanatid:
Issue #4785 

It's only draft right now because I'm not sure that this is a best way for events from http requests. Before to fix everything I would like get feedback.

In http sinks (and others) we only need implement `HttpSink` trait and do not control request/retries/timeout. Since we have `Acker` I decide to replace it by success callback where we can emit event. This changes also requires `Batch` trait change, because we need to know size of batch in bytes.


MOZGIII:
Does this overlap with https://github.com/timberio/vector/issues/2766?

src/internal_events/http.rs
```diff
@@ -42,3 +42,40 @@ impl<'a> InternalEvent for HTTPBadRequest<'a> {
...
+
+    fn emit_metrics(&self) {
+        counter!("events_discarded_total", 1);
+    }
+}
+
+#[derive(Debug)]
+pub struct HTTPEventsSent {
+    pub events_count: usize,
+    pub byte_size: usize,
```
lukesteensen, line 65: This seems to be the source of most of the complications, and I think we could potentially just get rid of it? There will be situations like HTTP streaming bodies where we can't really get it anyway, so maybe simpler to just leave it out and get byte information from other instrumentation like batching. Then hopefully this is a bit simpler and doesn't require things to be passed around in the same way.

MOZGIII: Maybe do it like this?

https://github.com/timberio/vector/blob/f9bf655dd7aa7a3d08be104e67a9ed26424eeb22/src/http.rs#L119-L136

lukesteensen: True, but I worry about how we would represent the uncertainty there. If we just use those numbers naively, the resulting metrics would likely be "inaccurate".

MOZGIII: I would prefer if we avoid capturing the byte sizes of request/response unless it is made available to us by the underlying layers. Though realistically, we can't expect those to be available - nobody does that. I guess we should not pass the bytes here and remove them from the HTTP client logging. The way they're obtained is not intended for how we try to use it.

fanatid: If we remove `byte_size` we still need to find some way to emitting events when they are sent. Or we can emit an event when we encode data and put it in the queue?

MOZGIII: Assuming we emit an event at the HTTP client - I think we're covered, and there's no need to emit elsewhere.

src/sinks/util/sink.rs
```diff
@@ -520,16 +522,20 @@ where
...
+                    {
+                        total_items += num_items;
+                        total_bytes += num_bytes;
                         self.seq_tail += 1
                     }
-                    trace!(message = "Acking events.", acking_num = num_to_ack);
-                    self.acker.ack(num_to_ack);
+                    trace!(message = "Acking events.", acking_num = total_items);
+                    (self.on_success)(total_items, total_bytes);
                 }
```
bruceg, line 539: I think this logic points to the need for separate `acker` and `on_success` parameters. When an in flight request is complete, we should be emitting the event immediately (via `on_success`). The `pending_acks` logic is there to ensure the acknowledgements are issued strictly in sequence. I don't think those two parts should be joined.

fanatid:
Event moved from `ServiceSink` internals to `HttpSink::encode_event` (same as `elasticsearch`).

lukesteensen (approved):

bruceg (approved):

Issue #4948, opened 2020-11-10
Add internal events to prometheus_exporter sink


jszwedko:
This sink does not appear to be instrumented yet.

Requested by user in discord: https://discord.com/channels/742820443487993987/746070591097798688/775755348430684160

jszwedko:
Closing since this was added in the sweep of all sink components for instrumentation.

PR #5115, opened 2020-11-19
chore(prometheus sink): add internal events

fanatid:
Part of #4948 

I only add 1 event to `prometheus_exporter` and skip `prometheus_remote_write` because faced with the same problem as with HTTP sink in #5111 (we do not control the sending process and can not emit events in the right places).


src/sinks/prometheus/exporter.rs
```diff
@@ -255,6 +251,10 @@ impl PrometheusExporter {
                         )
                     });
 
+                    emit!(PrometheusServerRequestComplete {
+                        code: response.status().as_u16(),
```
bruceg, line 255: Could this be stored as the actual type of the status instead of converting to `u16`?

fanatid: Sure. Updated.

bruceg (approved):

PR #5117, opened 2020-11-19
fix(ci): add RUSTSEC-2020-0071 to deny.toml

fanatid:
CI fix for https://github.com/time-rs/time/issues/293

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -4,4 +4,7 @@ ignore = [
     "RUSTSEC-2020-0016",
     # failure: failure is officially deprecated/unmaintained
     "RUSTSEC-2020-0036",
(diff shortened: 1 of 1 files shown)


jszwedko:
@fanatid could we not upgrade `time` instead? It looks like they patched it: https://github.com/time-rs/time/security/advisories/GHSA-wcg3-cvx6-7396

fanatid:
Thanks for a fix in https://github.com/timberio/vector/pull/5119

jhpratt:
@jszwedko I wouldn't particularly call it a fix, but the issue doesn't exist in time 0.2.23. Literally all I did was remove the ability to determine the local offset on the affected operating systems.

PR #5118, opened 2020-11-19
fix(prometheus sink): Fix output timestamp

### LINKED ISSUE #5110: Ensure prometheus remote_write sink transmits data properly
While developing the Prometheus remote_write source, I discovered that the corresponding sink does not transmit timestamps properly. Further, the integration test only ensures that the correct *number* of metrics is transmitted, not any of the rest of the content (tags, timestamp, name, etc).

The timestamp transmission should be fixed and the integration tests should be enhanced to ensure there are no other data errors.

Blocks #4162 as the new source will be using this sink to test itself

bruceg:
The output timestamp is expressed in milliseconds since the UNIX epoch, and we were sending seconds since. This scales the output properly.

Closes #5110

### DIFF
--- a/src/sinks/influxdb/mod.rs
+++ b/src/sinks/influxdb/mod.rs
@@ -391,6 +391,7 @@ pub mod test_util {
 
     pub(crate) async fn onboarding_v1(endpoint: &str) -> String {
         let database = next_database();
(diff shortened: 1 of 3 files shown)


jszwedko (approved):
A few failing clippy checks, but otherwise this looks good!

What bizarre influxdb behavior.

PR #5120, opened 2020-11-19
chore(deps): Remove dirs

jszwedko:
It doesn't appear we were using it anywhere.

Resolves RUSTSEC-2020-0053


fanatid (approved):
:+1:

PR #5122, opened 2020-11-19
chore(deps): Remove a couple of more unused deps

jszwedko:
Found via cargo-udeps


fanatid (approved):
@jszwedko how did you find these deps? :laughing: I tried `undeps` but this requires nightly which produce a lot of errors :disappointed:

PR #5128, opened 2020-11-19
fix(tests): Fix flaky `configuration_path_recomputed` test

bruceg:
Also move all timeout values into constants.


jszwedko (approved):
Thanks for investigating these flakey tests!

It'd obviously be great to not need to rely on sleeps, but this is a definite improvement if it fixes the tests.

PR #5134, opened 2020-11-19
fix(host_metrics source): Fix test failure for empty metrics

bruceg:
The `assert_filtered_metrics` function assumed some metrics would be produced, but this may not be the case if permissions prohibit generating metrics.


jamtur01 (approved):
LGTM

PR #5139, opened 2020-11-20
chore(deps): upgrade to 1.48 toolchain

fanatid:
Announce in rust blog: https://blog.rust-lang.org/2020/11/19/Rust-1.48.html


MOZGIII:
I think using `future::ready` is a bit more expressive than `ready` (easier to guess what ready in without an extra lookup). If people agree, I think we could update the callsites with `std::future::ready` instead of `use`-aliasing it at the top and doing bare `ready` calls.

This is a minor nit though, not sure if we should care about it really.

fanatid:
Yeah, I'd like to see `future::ready` more than just `ready` too but often we already import `futures::future` (which makes it impossible to use `std::future`). Maybe we should use `std::future::ready`. I do not know.

jszwedko:
I agree, I'd also like to see `future::ready`. We could come up with a standard alias for `std::future` and use that to avoid the conflict though I'm fine with just `std::future`.

src/kubernetes/stream.rs
```diff
@@ -12,6 +12,8 @@ use snafu::{ResultExt, Snafu};
 
 /// Converts the HTTP response [`Body`] to a stream of parsed Kubernetes
 /// [`Response`]s.
+// Clippy rule, because `try_stream!`?
+#[allow(clippy::needless_return)]
```
jszwedko, line 16: I was also able to resolve this before in #4818 when I mistakenly ran into it running the nightly toolchain by adding a unit return at the end of `try_stream!`: https://github.com/timberio/vector/pull/4818/files/471caf28164db3599611494d68efbcb0ab090ae6#diff-38f65b460ca0ce88f0e24ff9791321090396872183b32b03c8a3435dbe4f45f3

jszwedko:
Actually, I might be coming around to `ready` and `pending`. I think it'll just take a slight mental adjustment, but given how core those methods are, I think we'll end up seeing them a lot and it'll become more natural.

jszwedko (approved):
This looks good to me, thanks @fanatid !

src/kubernetes/stream.rs
fanatid: Thanks @jszwedko! I rememer that I tried to fix this already but was not able issue/pr :disappointed:

JeanMertz (approved):

src/sinks/util/auto_concurrency/service.rs
```diff
@@ -317,12 +316,12 @@ mod tests {
                 // must be borrowed mutable with a non-static
                 // lifetime. Resolving it is more work than it's worth
                 // for this test, so the clippy lint allow remains above.
-                for i in 0..concurrent {
-                    reqs[i] = Some(svc.send(i < concurrent - 1).await);
+                for (i, req) in reqs.iter_mut().take(concurrent).enumerate() {
+                    *req = Some(svc.send(i < concurrent - 1).await);
```
bruceg, line 320: This is :fire: Thanks! You should probably drop the last part of the phrase about the clippy lint, since you *did* remove it.

bruceg (approved):

Hoverbear (approved):

FungusHumungus (approved):

src/kubernetes/stream.rs
```diff
@@ -35,6 +35,8 @@ where
             }
         }
         decoder.finish().map_err(|data| Error::UnparsedDataUponCompletion { data })?;
+
+        ()
```
lukesteensen, line 39: Why is this needed?

jszwedko: It's to satisfy clippy, though I'm not entirely sure why it is needed or if there is a better way to resolve that lint.

https://github.com/timberio/vector/pull/4818/files/471caf28164db3599611494d68efbcb0ab090ae6#diff-38f65b460ca0ce88f0e24ff9791321090396872183b32b03c8a3435dbe4f45f3

We could add a comment here about it though. I could see it being confusing.

lukesteensen: Hm, strange. I'm fine to leave it and hopefully a later clippy update will tell us it's unnecessary :smile:

PR #5141, opened 2020-11-20
chore(prometheus sink): less locks in StreamSink::run

fanatid:
Noticed yesterday that we have `lock` per each event in the loop but in source metrics, we usually push a lot of metrics at one time. As result, it should be possible to process all the metrics under one lock. It should not be a huge difference in performance, just a nicer way, imo.

Another approach is creating a helper like [StreamExt::buffered](https://docs.rs/futures/0.3.6/futures/stream/trait.StreamExt.html#method.buffered) which works with not futures items.


src/sinks/prometheus/exporter.rs
```diff
@@ -276,39 +278,66 @@ impl PrometheusExporter {
...
+                                        if reset_existing.is_none() {
+                                            let now = Utc::now().timestamp();
+                                            let interval =
+                                                now - *self.last_flush_timestamp.read().unwrap();
+                                            if interval > self.config.flush_period_secs as i64 {
+                                                *self.last_flush_timestamp.write().unwrap() = now;
+                                                reset_existing = Some(true);
+                                            } else {
+                                                reset_existing = Some(false);
+                                            }
```
bruceg, line 314:
```suggestion
                                            let expired = interval > self.config.flush_period_secs as i64;
                                            if expired {
                                                *self.last_flush_timestamp.write().unwrap() = now;
                                            }
                                            reset_existing = Some(expired);
```

src/sinks/prometheus/exporter.rs
```diff
@@ -276,39 +278,66 @@ impl PrometheusExporter {
...
+            let mut metrics = None;
+            let mut reset_existing = None;
+
+            let result = loop {
+                match Pin::new(&mut input).poll_next(cx) {
+                    Poll::Ready(Some(event)) => {
+                        received += 1;
+
+                        if metrics.is_none() {
+                            metrics.replace(self.metrics.write().unwrap());
```
bruceg, line 293: I also suggest replacing the `unwrap()`s with `expect(…)` which both gives better error messages and internally notes why it should never fail.

src/sinks/prometheus/exporter.rs
```diff
@@ -276,39 +278,66 @@ impl PrometheusExporter {
...
-                        existing.add(&item);
-                        metrics.insert(MetricEntry(existing));
-                    } else {
-                        metrics.insert(new);
-                    };
-                }
-                MetricKind::Absolute => {
-                    let new = MetricEntry(item);
-                    metrics.replace(new);
+                        let metrics = metrics.as_mut().unwrap();
```
bruceg, line 295: Suggestion:
```rust
    let metrics = metrics.get_or_insert_with(|| metrics.as_mut().unwrap());
```

bruceg (approved):
Looks ok. The `metrics.take`-then-`metrics.insert` logic looks a little alarming, but should be fast enough with the `IndexSet`.

lukesteensen (approved):
Agree with @bruceg's suggestions but overall looks good!

src/sinks/prometheus/exporter.rs
fanatid: Thanks! `get_or_insert_with` looks nice.

src/sinks/prometheus/exporter.rs
```diff
@@ -276,39 +278,60 @@ impl PrometheusExporter {
...
+                                        // sets need to be expired from time to time
+                                        // because otherwise they could grow infinitelly
+                                        let now = Utc::now().timestamp();
+                                        let interval =
+                                            now - *self.last_flush_timestamp.read().unwrap();
+                                        if interval > self.config.flush_period_secs as i64 {
+                                            *self.last_flush_timestamp.write().unwrap() = now;
+                                            existing.reset();
+                                        }
+                                    }
```
fanatid, line 310: @bruceg is this correct logic? Does it look like only one metric will be reset each flush period while we need to reset all the metrics, or I miss something?

bruceg: I agree with your reading, it looks broken, and would be a pre-existing bug with this sink. To be clear, only one *set* metric will be reset each flush period. This would be worth a test to prove out.

PR #5146, opened 2020-11-20
enhancement(file source): expire checkpoints

### LINKED ISSUE #5137: Expire file checkpoints shortly after file is removed
To avoid accumulating an unbounded number of checkpoints and slowing down the process, we should expire and remove checkpoints from our internal state once the corresponding file has been removed from disk. There should be a reasonable waiting period (e.g. 60 seconds) to ensure that the removal was not a partially visible move or something similar, and any updates during that period should cancel the removal.

lukesteensen:
Closes #5137

This clears out old checkpoints before writing to disk and adds some instrumentation around write timing. Checkpoints are considered old when the file they refer to was deleted more than 60 seconds ago, and any updates after "removal" will cancel the expiration.

### DIFF
--- a/lib/file-source/src/checkpointer.rs
+++ b/lib/file-source/src/checkpointer.rs
@@ -76,7 +76,7 @@ impl CheckpointsView {
                 let duration = now - *ts;
                 duration >= chrono::Duration::seconds(60)
             })
-            .map(|entry| entry.key().clone())
+            .map(|entry| *entry.key())
             .collect::<Vec<FileFingerprint>>();
 
         for fng in to_remove {
--- a/lib/file-source/src/checkpointer.rs
+++ b/lib/file-source/src/checkpointer.rs
@@ -45,16 +45,45 @@ pub struct Checkpointer {
 pub struct CheckpointsView {
     checkpoints: DashMap<FileFingerprint, FilePosition>,
     modified_times: DashMap<FileFingerprint, DateTime<Utc>>,
+    removed_times: DashMap<FileFingerprint, DateTime<Utc>>,
 }
 
 impl CheckpointsView {
     pub fn update(&self, fng: FileFingerprint, pos: FilePosition) {
         self.checkpoints.insert(fng, pos);
         self.modified_times.insert(fng, Utc::now());
+        self.removed_times.remove(&fng);
     }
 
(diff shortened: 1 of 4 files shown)


lib/file-source/src/checkpointer.rs
```diff
@@ -480,4 +514,41 @@ mod test {
             assert_eq!(chkptr.get_checkpoint(fingerprint), Some(position));
         }
     }
+
+    #[test]
+    fn test_checkpointer_expiration() {
+        let cases = vec![
+            // (checkpoint, position, seconds since removed)
```
jszwedko, line 521: :+1: thanks for this comment.

lib/file-source/src/file_server.rs
```diff
@@ -148,11 +148,11 @@ where
...
                 let checkpointer = Arc::clone(&checkpointer);
                 tokio::task::spawn_blocking(move || {
-                    checkpointer
-                        .write_checkpoints()
-                        .map_err(|error| emitter.emit_file_checkpoint_write_failed(error))
-                        .map(|count| emitter.emit_file_checkpointed(count))
-                        .ok()
+                    let start = time::Instant::now();
+                    match checkpointer.write_checkpoints() {
+                        Ok(count) => emitter.emit_file_checkpointed(count, start.elapsed()),
```
jszwedko, line 153: :+1: to removing this map usage. I always find it a bit awkward when map is used to "consume" values.

lib/file-source/src/file_server.rs
```diff
@@ -299,6 +299,7 @@ where
                                 Ok(()) => {
                                     self.emitter.emit_file_deleted(&watcher.path);
                                     watcher.set_dead();
+                                    checkpoints.set_dead(file_id);
```
jszwedko, line 302: Wouldn't the `watcher.set_dead()` that's already here cause it to be picked up by line 324?

jszwedko (approved):
I'm not super familiar with the file source code, but these changes make sense to me!

lib/file-source/src/file_server.rs
lukesteensen: Aha, yes it would. Not sure why this was a separate path in my mind. Will remove!

PR #5152, opened 2020-11-20
chore: update Source type to new futures

fanatid:
Wanted start `nginx` source and found that I need to use `.compat()` :cry: in the same time all sources already use new futures, so quick update.


lukesteensen (approved):
Nice! This is exciting :tada:

jszwedko (approved):
Agreed, this is exciting, thanks @fanatid !

Issue #5143, opened 2020-11-20
Include topic in events from Kafka source


spencergilbert:
### Current Vector Version

```
vector 0.11.0 (g9806442 x86_64-unknown-linux-musl 2020-11-17)
```

### Use-cases

We'd like to reference the Kafka topic an event originated from downstream, other log aggregators support similar metadata - [Logstash Kafka metadata](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-kafka.html#_metadata_fields) 

### Attempted Solutions

The only way it seems possible today is to create a sink per topic and based on the originating sink add a field for the topic. This is manageable for a small number of topics but can't scale.

### Proposal

Vector should know the topic it's retrieving messages from in the request, so it should be possible to extract that and include it in the event schema.

### References

[Logstash](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-kafka.html#_metadata_fields)
[Kafka source fields](https://vector.dev/docs/reference/sources/kafka/#fields)

jamtur01:
Also worth noting is #1150.

Issue #800, opened 2019-08-26
Fix and un-ignore kafka source integration test


lukesteensen:
See #774

PR #5153, opened 2020-11-20
enhancement(kafka source): include kafka metadata as optional keys

spencergilbert:
Fixes: #1150
Fixes: #5143 
Fixes: #800

Modelled after the existing `key_field` configuration I inserted the topic, partition, and offset provided by rdkafka into events when configured. These should be opt-in and not affect users unless explicitly configured.

### DIFF
--- a/docs/reference/components/sources/kafka.cue
+++ b/docs/reference/components/sources/kafka.cue
@@ -83,6 +83,36 @@ components: sources: kafka: {
 				examples: ["message_key"]
 			}
 		}
+		topic_key: {
+			common:      true
+			description: "The log field name to use for the Kafka topic key. If unspecified, the key would not be added to the log event."
+			required:    false
+			warnings: []
+			type: string: {
+				default: null
+				examples: ["topic_key"]
+			}
+		}
+		partition_key: {
+			common:      true
+			description: "The log field name to use for the Kafka partition key. If unspecified, the key would not be added to the log event."
+			required:    false
+			warnings: []
+			type: string: {
+				default: null
+				examples: ["partition_key"]
+			}
+		}
+		offset_key: {
+			common:      true
+			description: "The log field name to use for the Kafka offset key. If unspecified, the key would not be added to the log event."
+			required:    false
+			warnings: []
+			type: string: {
+				default: null
+				examples: ["offset_key"]
+			}
+		}
 		librdkafka_options: components._kafka.configuration.librdkafka_options
 		sasl: {
 			common:      false
--- a/docs/reference/components/sources/kafka.cue
+++ b/docs/reference/components/sources/kafka.cue
@@ -90,7 +90,7 @@ components: sources: kafka: {
 			warnings: []
 			type: string: {
 				default: null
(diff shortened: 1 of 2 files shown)


docs/reference/components/sources/kafka.cue
```diff
@@ -83,6 +83,36 @@ components: sources: kafka: {
...
 			}
 		}
+		topic_key: {
+			common:      true
+			description: "The log field name to use for the Kafka topic key. If unspecified, the key would not be added to the log event."
+			required:    false
+			warnings: []
+			type: string: {
+				default: null
+				examples: ["topic_key"]
```
jszwedko, line 93: I might just use `topic`, `partition`, etc. as the examples. I think `key_field`'s example was `message_key` because `key` could be ambiguous (I guess?).

jszwedko (commented):
Nice work! This looks good to me.

Could you add a few more assertions to the `kafka_source_consume_event` test to validate that these fields are set correctly?

docs/reference/components/sources/kafka.cue
spencergilbert: Sounds good

spencergilbert:
@jszwedko @jamtur01 @lukesteensen - I cleaned up the docs some and got the kafka integration test enabled

src/sources/kafka.rs
```diff
@@ -272,14 +296,14 @@ mod integration_test {
...
-    use chrono::Utc;
+    use chrono::{SubsecRound, Utc};
     use rdkafka::{
         config::ClientConfig,
         producer::{FutureProducer, FutureRecord},
         util::Timeout,
     };
 
-    const BOOTSTRAP_SERVER: &str = "localhost:9092";
+    const BOOTSTRAP_SERVER: &str = "localhost:9091";
```
lukesteensen, line 306: Just curious, why this change?

spencergilbert: The makefile is starting kafka with a platintext listener on 9091, 9092 is SSL.

```
 587   │      -e KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://localhost:9091,SSL://localhost:9092,SASL_PLAINTEXT://localhost:9093 \
```

I suppose we could also look to setup vector with a cert and use 9092 :thinking:

jszwedko: I personally think PLAINTEXT is fine for this test but we could add other tests later for SSL.

jszwedko (approved):
This looks good to me! Nice work.

src/sources/kafka.rs
spencergilbert: Probably worth another issue/pr to test the TLS config for the source

lukesteensen: Ah, that makes sense. Yeah, I'm fine to use plaintext for most tests and just have one that exercises TLS.

lukesteensen (approved):

binarylogic:
Thanks @spencergilbert.

jamtur01:
giphub

PR #5173, opened 2020-11-22
chore: use StreamExt::take_until instead tokio::select!

fanatid:
When removed old futures from sources I noticed that we use `tokio::select` for shutdown where `StreamExt::take_until` can be used. Also, correct mongo db event name with other metrics sources.


bruceg (approved):
LGTM

PR #5182, opened 2020-11-23
chore(external docs): Correct the Kubernetes installation docs

MOZGIII:
This PR fixes some issues with the Kubernetes installation docs, in both `helm` and `kubectl` installation methods.

It also establishes some internal structure in the `helm` documentation, so the logic there is easier to follow.


binarylogic (approved):
Nice

PR #5183, opened 2020-11-23
fix: Incorrect Log Level Message

### LINKED ISSUE #5084: Incorrect Log Level Message
### Vector Version

```
vector 0.11.0 (gb8f83d7 x86_64-unknown-linux-musl 2020-11-16) running on Docker
```

### Vector Configuration File

```bash
docker run \
  --network host \
  -v /etc/vector/vector.toml:/etc/vector/vector.toml:ro \
  --env LOG=debug \
  --restart always \
  timberio/vector:nightly-alpine
```

### Expected Behavior

Docker logs should indicate correct log level. e.g. debug 

### Actual Behavior

Docker logs indicate debug level as "info"

### Example Data

```
Nov 17 18:18:27.375  INFO vector::app: Log level is enabled. level=**"info"**
```

### Additional Context

https://discord.com/channels/742820443487993987/746070591097798688/778324625439588433

juchiast:
`level` was log-level from CLI options, therefore it did not take into account environment variables configuration.

close #5084

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -51,25 +51,21 @@ impl Application {
     pub fn prepare_from_opts(opts: Opts) -> Result<Self, exitcode::ExitCode> {
         openssl_probe::init_ssl_cert_env_vars();
 
-        let level = opts.log_level();
+        let level = std::env::var("LOG").unwrap_or_else(|_| match opts.log_level() {
+            "off" => "off".to_owned(),
+            level => [
+                format!("vector={}", level),
+                format!("codec={}", level),
+                format!("file_source={}", level),
+                "tower_limit=trace".to_owned(),
+                format!("rdkafka={}", level),
+            ]
+            .join(","),
+        });
+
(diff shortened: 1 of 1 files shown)


src/app.rs
```diff
@@ -51,26 +51,22 @@ impl Application {
...
+            level => [
+                format!("vector={}", level),
+                format!("codec={}", level),
+                format!("file_source={}", level),
+                "tower_limit=trace".to_owned(),
+                format!("rdkafka={}", level),
+            ]
+            .join(","),
+        });
+
```
fanatid, line 65: I think I miss something but isn't this code looks the same as it was? What is the difference?

juchiast: The difference is `level` and `levels`.

fanatid: Thanks. I opened the whole file and found that we log `level` later while applying`levels`. Now it's clear.

fanatid (approved):

PR #5184, opened 2020-11-23
chore: Improve the CI detection logic at cue formatting check

MOZGIII:
I have false positives locally and at Codespaces, this will fix it.

### DIFF
--- a/scripts/check-docs.sh
+++ b/scripts/check-docs.sh
@@ -19,7 +19,7 @@ if ! [ -x "$(command -v cue)" ]; then
   exit 1
 fi
 
-if [[ -z "${CI:-}" ]]; then
+if [[ "${CI:-"false"}" != "true" ]]; then
   echo "Skipping local formatting - reserved for CI"
 else
   echo "Validating ${DOCS_PATH}/**/*.cue formatting."


jszwedko (commented):
:thinking: do you know why this was needed?

If we do do this, we should probably update the other usages of `CI` (and possibly other checks for undefined vars).

MOZGIII:
> 🤔 do you know why this was needed?
> 
> If we do do this, we should probably update the other usages of `CI` (and possibly other checks for undefined vars).

Yeah, I don't know really, but I see the formatting check run locally.

jszwedko:
Very odd. I'm fine with this change, but could we update the other places `CI` appears to be the same? I imagine you'd run into the same issue with those scripts.

I still fail to see how this would happen unless you have `CI` defined though.

MOZGIII:
> Very odd. I'm fine with this change, but could we update the other places `CI` appears to be the same? I imagine you'd run into the same issue with those scripts.
> 
> I still fail to see how this would happen unless you have `CI` defined though.

We typically don't do this (we just have separate scripts - more UNIX-y way). I guess the actual proper fix would be to split this logic into two commands and only run the CI-relevant at CI.

jszwedko:
It looks like we have a couple of places, one of which is actually already doing what you have here:

```
scripts/ci-setup-minikube.sh
4:if [[ -z "${CI:-}" ]]; then

scripts/check-component-features.sh
38:if (${CI:-false}); then
46:if (${CI:-false}); then

scripts/check-docs.sh
22:if [[ -z "${CI:-}" ]]; then

scripts/environment/bootstrap-macos-10.sh
8:if [ -n "${CI-}" ] ; then
```

MOZGIII:
> scripts/check-component-features.sh
> 38:if (${CI:-false}); then
> 46:if (${CI:-false}); then

This is quite different, it'll exec the whatever the `CI` value is. It's good if it's `true` or ``, worse if it's `rm -rf /` :D

MOZGIII:
I guess I want to address the cue separately to provide a utility script, as I constantly find myself in need of some extra functionality. I'll submit a PR when I have some time between more pressing issues.

PR #5188, opened 2020-11-23
chore(deps): remove uom as direct dependency

fanatid:
We still have `uom` but as `heim` dependency.


bruceg (approved):

fanatid:
@bruceg do you have any ideas why linking fails on CI? I tested locally and everything compiled successfully.

bruceg:
> @bruceg do you have any ideas why linking fails on CI? I tested locally and everything compiled successfully.

I don't know. Note the error though: `collect2: fatal error: ld terminated with signal 9 [Killed]` The linker died from being sent the `KILL` signal. This is not a normal link error or bug, it was forcibly killed.

PR #5191, opened 2020-11-23
enhancement(http source): Add `vector_http_path` event key to `http` source

samgiles:
Implements https://github.com/timberio/vector/issues/5112

I wasn't sure what to call the event key but I settled on `vector_http_path` thinking it would be best to avoid potential collisions with any existing events with a `path` key which feels like it'd be quite a common one.  Also not sure where these are documented. I also decided to not add the key if the path is `"/"` or otherwise empty to make it slightly more backwards compatible.

I didn't add it to the schema, since it feels very specific to the http source and i'm not sure if there's a convention for source specific metadata keys. Maybe worth making it configurable somehow?

Also I tried not to change the `logplex` source, it does use the same `warp` filter setup, but the path is unused in the `logplex` source.


samgiles:
😨 the fmt checks look like they're failing for unrelated reasons. Happy to rebase this onto something if there's an inbound fix for that.

binarylogic:
Thanks @samgiles. Just noting, we'll need to add a field [here](https://github.com/timberio/vector/blob/master/docs/reference/components/sources/http.cue#L142) for our docs.

docs/reference/components/sources/http.cue
```diff
@@ -135,6 +135,11 @@ components: sources: http: {
 					type: string: examples: ["Hello world"]
 				}
 				timestamp: fields._current_timestamp
+				vector_http_path: {
```
samgiles, line 138: I have no idea what I'm doing

I have no idea what I'm doing with `cue` 😅

src/sources/logplex.rs
```diff
@@ -58,6 +58,7 @@ impl HttpSource for LogplexSource {
         body: Bytes,
         header_map: HeaderMap,
         query_parameters: HashMap<String, String>,
+        _full_path: &str,
```
jszwedko, line 61: It might be nice to add this support here in the `logplex` source too.

jszwedko (commented):
Thanks for this contribution @samgiles!

In general, I think it looks good. Historically, we've added a configuration option for configuring where a value like this is assigned. For example, see the recent addition of tagging events with their topic for the `kafka` source: https://github.com/timberio/vector/pull/5153 (`topic_key`)

Here, I think we could call that `path_key`. This would let users set it, alleviating any concerns of potentially overwriting a value in the event itself.

samgiles:
Thanks for the feedback and sorry for the delay I will address comments in the next couple of days 👌 , I took most of last week off 🌴 so just catching up.

jamtur01:
Hey @samgiles - just checking in to see how you were going?

binarylogic:
Thanks for your help @samgiles, closing since this has gone stale. If you'd like to make the above changes we'd be happy to review again.

jszwedko:
I'd actually like to see this across the finish line. It only needs a little tweaking.

prognant:
@samgiles The original work has been included with  https://github.com/timberio/vector/pull/6626 thanks again for your contribution !

PR #5203, opened 2020-11-24
chore(config): Rename `api.bind` -> `api.address`

leebenson:
Renames `api.bind` -> `api.address` to match other parts of our config.

### DIFF
--- a/src/api/server.rs
+++ b/src/api/server.rs
@@ -23,7 +23,7 @@ impl Server {
 
         let (_shutdown, rx) = oneshot::channel();
         let (addr, server) = warp::serve(routes).bind_with_graceful_shutdown(
-            config.api.bind.expect("Invalid socket address"),
+            config.api.address.expect("Invalid socket address"),
             async {
                 rx.await.ok();
             },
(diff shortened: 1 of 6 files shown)


binarylogic (approved):

src/api/server.rs
```diff
@@ -23,7 +23,7 @@ impl Server {
 
         let (_shutdown, rx) = oneshot::channel();
         let (addr, server) = warp::serve(routes).bind_with_graceful_shutdown(
-            config.api.bind.expect("Invalid socket address"),
+            config.api.address.expect("Invalid socket address"),
```
jszwedko, line 26: I realize this was already like this, but it seems like `config.api.address` is an `Option`? Would `.expect("No socket address")` be better here?

I'm imagining this is already validated before it gets to this point?

jszwedko (approved):
:+1:

src/api/server.rs
leebenson: Yeah, you're right. The perils of `Option<T>` and `Result<T>` sharing an unwrap API 😄 Well spotted 👍 Fixed.

Issue #5085, opened 2020-11-17
Unify compression handling for file fingerprints


lukesteensen:
From the [File Source RFC](https://github.com/timberio/vector/blob/master/rfcs/2020-10-15-3480-file-source-rework.md)

The current file source fingerprinting happens before we do any determination of whether the file is compressed or not. This means that the same file will have a different fingerprint and be reread if it's still visible after being rotated and compressed.

To fix this properly, we should evolve the current `FileWatcher` abstraction into a more general-purpose file wrapper. Instead of doing fingerprinting ahead of creating the watcher, we can create it first and pull fingerprints through the same interface. This should ensure that how we read lines and how we calculate fingerprints are unified in the way they read from files.

PR #5215, opened 2020-11-24
enhancement(file source): unify checksum methods

### LINKED ISSUE #5088: Unify checksum and first line checksum fingerprinting strategies
From the [File Source RFC](https://github.com/timberio/vector/blob/master/rfcs/2020-10-15-3480-file-source-rework.md)

We should evolve the checksum fingerprinting strategies to smoothly transition users from the previous byte-based method to the newer line-based method. This will involve a transition period where new fingerprints should include both when possible, to facilitate matching against older stored fingerprints and upgrading them.

lukesteensen:
Closes #5088

This change makes the file source use checksums of the first line of the file no matter which checksum strategy is configured. The old configuration is left in place for now to facilitate an automated migration on startup that upgrades existing bytes-based checksums to the new line-based version.

The one gotcha at this stage is around compressed files. If a gzipped file is smaller than the maximum allowed line length (default 100kb), we will currently not return a checksum and not read the file. This is because we need some fixed stopping point to make the checksum stable, and EOF is not that. This will be fixed by #5085.

### DIFF
--- a/lib/file-source/src/fingerprinter.rs
+++ b/lib/file-source/src/fingerprinter.rs
@@ -10,6 +10,7 @@ use std::{
 #[derive(Clone)]
 pub struct Fingerprinter {
     pub strategy: FingerprintStrategy,
+    pub max_line_length: usize,
     pub ignore_not_found: bool,
 }
 
@@ -20,7 +21,6 @@ pub enum FingerprintStrategy {
         ignored_header_bytes: usize,
     },
     FirstLineChecksum {
-        max_line_length: usize,
         ignored_header_bytes: usize,
     },
     DevInode,
@@ -29,7 +29,8 @@ pub enum FingerprintStrategy {
 #[derive(Debug, PartialEq, Eq, Hash, Clone, Copy, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
 pub enum FileFingerprint {
-    Checksum(u64),
+    #[serde(rename = "checksum")]
+    BytesChecksum(u64),
     FirstLineChecksum(u64),
     DevInode(u64, u64),
     Unknown(u64),
@@ -40,7 +41,7 @@ impl FileFingerprint {
         use FileFingerprint::*;
 
         match self {
-            Checksum(c) => *c,
+            BytesChecksum(c) => *c,
             FirstLineChecksum(c) => *c,
             DevInode(dev, ino) => {
                 let mut buf = Vec::with_capacity(std::mem::size_of_val(dev) * 2);
@@ -76,20 +77,12 @@ impl Fingerprinter {
             }
             FingerprintStrategy::Checksum {
                 ignored_header_bytes,
-                bytes,
-            } => {
-                buffer.resize(bytes, 0u8);
-                let mut fp = fs::File::open(path)?;
-                fp.seek(io::SeekFrom::Start(ignored_header_bytes as u64))?;
-                fp.read_exact(&mut buffer[..bytes])?;
-                let fingerprint = crc::crc64::checksum_ecma(&buffer[..]);
-                Ok(Checksum(fingerprint))
+                bytes: _,
             }
-            FingerprintStrategy::FirstLineChecksum {
-                max_line_length,
+            | FingerprintStrategy::FirstLineChecksum {
                 ignored_header_bytes,
             } => {
-                buffer.resize(max_line_length, 0u8);
+                buffer.resize(self.max_line_length, 0u8);
                 let mut fp = fs::File::open(path)?;
                 fp.seek(SeekFrom::Start(ignored_header_bytes as u64))?;
                 fingerprinter_read_until(fp, b'\n', buffer)?;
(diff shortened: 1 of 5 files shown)


lib/file-source/src/fingerprinter.rs
```diff
@@ -76,20 +77,12 @@ impl Fingerprinter {
...
             FingerprintStrategy::Checksum {
                 ignored_header_bytes,
-                bytes,
-            } => {
-                buffer.resize(bytes, 0u8);
-                let mut fp = fs::File::open(path)?;
-                fp.seek(io::SeekFrom::Start(ignored_header_bytes as u64))?;
-                fp.read_exact(&mut buffer[..bytes])?;
-                let fingerprint = crc::crc64::checksum_ecma(&buffer[..]);
-                Ok(Checksum(fingerprint))
```
MOZGIII, line 86: This seems a bit confusing. Is this intended for a short transitional period? If not, we probably want to make the `FingerprintStrategy::Checksum` `unimplemented`, and rewrite the attempts to use it at the call sites (upper in the callstack). It will allow us to iteratively phase it out.

src/sources/file.rs
```diff
@@ -1450,6 +1451,13 @@ mod tests {
         let dir = tempdir().unwrap();
         let config = file::FileConfig {
             include: vec![PathBuf::from("tests/data/gzipped.log")],
+            // TODO: remove this once files are fingerprinted after decompression
+            //
+            // Currently, this needs to be smaller than the total size of the compressed file
+            // because the fingerprinter tries to read until a newline, which it's not going to see
+            // in the compressed data, or this number of bytes. If it hits EOF before that, it
+            // can't return a fingerprint because the value would change once more data is written.
+            max_line_bytes: 100,
```
MOZGIII, line 1460: This is because we check this value before the decompression, right? If we were checking it after - it would just work. I presume this is tracked by https://github.com/timberio/vector/issues/5085.

MOZGIII (approved):
Looks good! I think we'll need a clean-up pass after we complete the planned upgrades to regain the type-safety (eliminate the legacy enum variants, etc) - but seems like it doesn't have to happen as part of this PR.

MOZGIII:
Does this PR close #2163?

lib/file-source/src/fingerprinter.rs
lukesteensen: Yes, this is a temporary state. Once we've released a version like this that knows how to do both, we can clean up the code to only do the one we want and instruct users that any migration path needs to go through the one with this upgrade mechanism.

src/sources/file.rs
lukesteensen: Correct, that's what I tried to explain with the comment.

lib/file-source/src/fingerprinter.rs
```diff
@@ -125,6 +118,27 @@ impl Fingerprinter {
...
+    ) -> Result<Option<FileFingerprint>, io::Error> {
+        match self.strategy {
+            FingerprintStrategy::Checksum {
+                bytes,
+                ignored_header_bytes,
+            } => {
+                buffer.resize(bytes, 0u8);
+                let mut fp = fs::File::open(path)?;
+                fp.seek(io::SeekFrom::Start(ignored_header_bytes as u64))?;
+                fp.read_exact(&mut buffer[..bytes])?;
```
Hoverbear, line 135: I wonder if it would be better to read what is available, and allow the trailing 0s? We could call `read()` iteratively until we had the bytes or an EOF?

Hoverbear (approved):

jszwedko:
I stumbled upon this because a user ran into the lack of support for `fingerprint.bytes`: https://discord.com/channels/742820443487993987/746070591097798688/806520661702344715

I feel like there were a few issues with this PR:

* I feel it should have been marked as a breaking change as it drops support for functionality: byte based fingerprinting (`fingerprint.bytes`)
* The docs don't seem to have been updated to reflect this change

lukesteensen:
> The docs don't seem to have been updated to reflect this change

This was definitely a miss, so thank you for taking care of it!

> I feel it should have been marked as a breaking change as it drops support for functionality: byte based fingerprinting

This was also a miss, but a bit more complicated and I'm still not sure of the correct solution. We took quite a few extra steps between this PR and #4799 to make the transition between bytes-based and line-based checksums transparent to users, at least from the perspective of existing files and checkpoints. What we seem to have overlooked are these cases where the first line is less data than we were taking into account with bytes and that decrease is enough to cause uniqueness issues.

It's pretty obvious in retrospect, but somewhere during the long, winding fingerprint discussion, we seem to have picked up an assumption that the transition wouldn't cause any use cases to move from sufficient data to insufficient data. This is true for the motivating use cases of k8s and docker (long lines with uniqueness at the end), as well as the example use case of lines <256 bytes _with_ sufficient uniqueness, but clearly isn't true in general.

The scenarios roughly break down as follows:

| First 256B unique | First line unique | Example | Status |
| --- | --- | --- | --- |
| False | True | k8s/docker, small manual testing lines | Improved |
| False | False | Files w/ large common headers | No change |
| True | True | Syslog w/ pid and timestamp | No change |
| True | False | Files w/ short common headers or oft-repeated lines | Broken |

The first row is what we intended to fix, and the next two we knew should still work fine. The last row is the one we didn't consider.

I haven't (yet?) come up with an obvious way to address this without making more tradeoffs, but it's possible we could do something like collect both types and compare in a way that remains stable as the file is written to. That could be relatively complicated, so it's also worth considering temporary fixes like letting people specify byte lengths explicitly. The problem is clearly pretty thorny, so if anyone has any suggestions I'm all ears :slightly_smiling_face:

jszwedko:
Thanks for the detailed response! The problem is definitely thorny :slightly_smiling_face: 

Yeah, I see, scenario 4 there was the only one that would have broken. I could see supporting it with, rather than `fingerprint.bytes`, having `fingerprint.lines` which would read the first N lines (defaulting to 1). That might help in the situation where there are oft repeated lines.

In this particular case, the user had files that had common header so I'm wondering if we could also have a `fingerprint.ignore_lines` that would ignore the first N lines, similar to `fingerprint.ignore_bytes`. This is mostly a UX enhancement to avoid making users count the actual number of bytes in the common static header.

This is also making me think that we may want to store the parameters that were used for a given checkpoint in the checkpoint itself so that the checkpoint is valid even if the user reloads vector with a different config. It would use the old checkpoint to identify the file, and then write a new checkpoint using the new parameters. Would that make sense? Or are we already doing something like that?

I assume we've surveyed other tools already, but I'm curious how they deal with this issue (like fluentd, rsyslog, etc.). I couldn't find any RFCs which makes sense since this was one of the very first sources.

PR #5234, opened 2020-11-25
chore(ci): Test Linux self-hosted runner

jamtur01:


Hoverbear (approved):
As long as we're comfortable with the risks detailed in https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories, I support this!

Cargo.toml
```diff
@@ -213,6 +213,9 @@ security-framework = "0.4"
 atty = "0.2"
 nix = "0.17.0"
 
+[build]
+jobs = 2
```
Hoverbear, line 217: This particular setting is not valid here I think? Try at `.cargo/config.toml` on the host. https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure

jamtur01: This is weird because it works with this but not without? :) I'll try your suggestion too. Computers. Why do we even? :)

Hoverbear: Hmmmm I think it's better if we don't have it in the cargo.toml anywho since some folks have more than 2 threads to use. :)

jamtur01: Yeah - I was planning to add some configurable "stuff" to do this but I am happy to leave it in the runner for now.

PR #5243, opened 2020-11-26
chore(observability): Top throughput

leebenson:
Refactors `vector top` columns, to group events/byte counts and throughput.

Throughput is now displayed per second, irrespective of sample interval.

Updates columns without metrics to `N/A` to make it clearer that no metrics have been raised. Retains `--` where metrics are available but the sample interval hasn't elapsed.

image

### DIFF
--- a/src/top/dashboard.rs
+++ b/src/top/dashboard.rs
@@ -73,14 +73,7 @@ impl HumanFormatter for i64 {
     }
 }
 
-static COMPONENT_HEADERS: [&str; 6] = [
-    "Name",
-    "Kind",
-    "Type",
-    "Events (Total/Sec)",
-    "Bytes (Total/Sec)",
-    "Errors",
-];
+static COMPONENT_HEADERS: [&str; 6] = ["Name", "Kind", "Type", "Events", "Bytes", "Errors"];
 
 struct Widgets<'a> {
     constraints: Vec<Constraint>,
--- a/src/top/dashboard.rs
+++ b/src/top/dashboard.rs
@@ -73,11 +73,19 @@ impl HumanFormatter for i64 {
     }
 }
 
+static COMPONENT_HEADERS: [&str; 6] = [
+    "Name",
+    "Kind",
+    "Type",
+    "Events (Total/Sec)",
+    "Bytes (Total/Sec)",
+    "Errors",
+];
+
 struct Widgets<'a> {
     constraints: Vec<Constraint>,
     url_string: &'a str,
     opts: &'a super::Opts,
-    component_headers: [String; 8],
 }
 
 impl<'a> Widgets<'a> {
@@ -89,23 +97,10 @@ impl<'a> Widgets<'a> {
             Constraint::Length(3),
         ];
 
-        // Create component headers
-        let component_headers = [
-            "Name".to_string(),
-            "Kind".to_string(),
-            "Type".to_string(),
-            "Events".to_string(),
-            format!("I/O @ {}ms", opts.interval),
-            "Bytes".to_string(),
-            format!("I/O @ {}ms", opts.interval),
-            "Errors".to_string(),
-        ];
-
         Self {
             constraints,
             url_string,
             opts,
-            component_headers,
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/top/dashboard.rs
```diff
@@ -73,11 +73,19 @@ impl HumanFormatter for i64 {
     }
 }
 
+static COMPONENT_HEADERS: [&str; 6] = [
+    "Name",
+    "Kind",
+    "Type",
+    "Events (Total/Sec)",
```
jszwedko, line 80: I personally think the `(Total/Sec)` part of this is noise given I think the values in the column, like `36.66k (2.98k / s)` make it pretty clear.

src/top/state.rs
```diff
@@ -50,10 +52,11 @@ pub async fn updater(mut state: State, mut event_rx: EventRx) -> StateRx {
                             }
                         }
                     }
-                    EventType::ProcessedEventsThroughputs(rows) => {
+                    EventType::ProcessedEventsThroughputs(interval, rows) => {
```
jszwedko, line 55: Nit: I might call these `interval_ms` just for clarity (or change `interval` to a `Duration`).

src/top/state.rs
```diff
@@ -6,9 +6,11 @@ type NamedMetric = (String, i64);
 #[derive(Debug)]
 pub enum EventType {
     ProcessedEventsTotals(Vec<NamedMetric>),
-    ProcessedEventsThroughputs(Vec<NamedMetric>),
+    /// Interval + named metric
```
jszwedko, line 9:
```suggestion
    /// Interval in ms + named metric
```

jszwedko (approved):
Nice work! I feel like this looks much cleaner.

src/top/dashboard.rs
```diff
@@ -136,41 +131,52 @@ impl<'a> Widgets<'a> {
...
+                },
+                if self.opts.human_metrics {
+                    r.errors.human_format()
+                } else {
+                    r.errors.thousands_format()
+                },
+            ];
 
             data.extend_from_slice(&formatted_metrics);
             Row::StyledData(data.into_iter(), Style::default().fg(Color::White))
```
sghall, line 167: Hey, off topic here, but noticed yesterday that  the CLion terminal does not like this `Color::White` here when you run `vector top`.  Shows up like below with the stats not readable.  "White" seems non-controversial and I would think it be handled everywhere, but on my machine the text is not visible...

Screen Shot 2020-11-26 at 4 14 04 PM

Have you noticed this?  If you switch the CLion theme to MacOS light, the text goes to white...

Screen Shot 2020-11-26 at 4 15 18 PM

If you switch the line to just the "default" the text is visible again ("gray" also works and is used in other places in the dashboard).

```rs
Row::StyledData(data.into_iter(), Style::default())
```

Darcula theme:

Screen Shot 2020-11-26 at 4 20 09 PM

MacOS Light:

Screen Shot 2020-11-26 at 4 20 35 PM

Not sure what's different about the terminal in CLion to cause this.  Also, this part about dealing with the theme the person has chosen, was something I had not thought about before.  Thinking that TUI abstracts this away a bit and re-maps the colors somehow (?) , but haven't really looked at it.   Maybe it would be better to reports this issue in a ticket.

sghall (approved):
Nice @leebenson.  I think this is much clearer. 👍

src/top/dashboard.rs
leebenson: Agreed, I think it's pretty clear without.

src/top/state.rs
leebenson: This models the `interval` in the API schema (an `Int!`), and the `--interval` CLI switch. I'd normally pass around as a duration, but in this case, it'd be an i64 -> Duration -> i64, for the sake of a cleaner signature. I think a comment is probably sufficient.

src/top/dashboard.rs
leebenson: Interesting! I'm not sure if tui can detect a custom theme, or whether that's just a term post-filter after rendering. In any case, it seems like we should probably drop `.fg(Color::White))` and side-step this.

leebenson: Looking at this further, I found this paragraph at https://perldoc.perl.org/Term::ANSIColor:

> Unfortunately, interpretation of colors 0 through 7 often depends on whether the emulator supports eight colors or sixteen colors. Emulators that only support eight colors (such as the Linux console) will display colors 0 through 7 with normal brightness and ignore colors 8 through 15, treating them the same as white. Emulators that support 16 colors, such as gnome-terminal, normally display colors 0 through 7 as dim or darker versions and colors 8 through 15 as normal brightness. On such emulators, the "normal" white (color 7) usually is shown as pale grey, requiring bright white (15) to be used to get a real white color. Bright black usually is a dark grey color, although some terminals display it as pure black. Some sixteen-color terminal emulators also treat normal yellow (color 3) as orange or brown, and bright yellow (color 11) as yellow.

I don't know whether that's relevant here, but crossterm interprets `Color::White` as:

```rust
Color::White => f.write_str("5;15")
```

Which would put it in the upper 16 colours. Wonder if that's why it's reverting to black?

Given that light themes are likely to be a popular option, I'm thinking opinionated colours may not be the best option, anyway. The use of MacOS Light highlighted that. Grey seems like a pretty universal/safe-ish default for asides, like the sample rate/quit instructions. Everything else, I'd probably just leave to the theme default.

Examples:

image

image

image

image

PR #5258, opened 2020-11-27
fix(prometheus sink): fix set metric reset on flush period

fanatid:
Found in https://github.com/timberio/vector/pull/5141#discussion_r530239394. This PR adds a time point to each metric so on the update we can check if this set metric requires reset or not.


src/sinks/prometheus/exporter.rs
```diff
@@ -151,16 +152,44 @@ struct PrometheusExporter {
     server_shutdown_trigger: Option<Trigger>,
     config: PrometheusExporterConfig,
     metrics: Arc<RwLock<IndexSet<MetricEntry>>>,
-    last_flush_timestamp: Arc<RwLock<i64>>,
     acker: Acker,
 }
 
+#[derive(Debug)]
+struct MetricEntry {
```
bruceg, line 159: Consider naming it `ExpiringMetric` or similar to avoid collision with the `MetricEntry` above.

src/sinks/prometheus/exporter.rs
```diff
@@ -369,6 +404,60 @@ mod integration_tests {
...
+        // Wait a bit for expired metrics
+        time::delay_for(time::Duration::from_secs(2)).await;
+
+        let (name1, event) = create_metric_set(Some(name1), vec!["6", "7"]);
+        tx.send(event).expect("Failed to send.");
+        let (name2, event) = create_metric_set(Some(name2), vec!["8", "9"]);
+        tx.send(event).expect("Failed to send.");
+
+        // Wait a bit for the prometheus server to scrape the metrics
+        time::delay_for(time::Duration::from_secs(2)).await;
```
bruceg, line 446: Is this enough time to *guarantee* they are expired with a 3 second expiry?

src/sinks/prometheus/exporter.rs
```diff
@@ -326,41 +351,51 @@ mod tests {
...
 
     #[tokio::test]
-    async fn prometheus_scrapes_metrics() {
-        crate::test_util::trace_init();
+    async fn prometheus_metrics() {
+        trace_init();
+
+        prometheus_scrapes_metrics().await;
+        time::delay_for(time::Duration::from_millis(500)).await;
+        reset_on_flush_preiod().await;
```
bruceg, line 370: This looks like the only call to this function. If you don't anticipate additional uses of it, I suggest integrating it into this test case.

src/sinks/prometheus/exporter.rs
```diff
@@ -369,6 +404,60 @@ mod integration_tests {
         assert_eq!(data["value"][1], Value::String("123.4".into()));
     }
 
+    async fn reset_on_flush_preiod() {
```
bruceg, line 407:
```suggestion
    async fn reset_on_flush_period() {
```

src/sinks/prometheus/exporter.rs
```diff
@@ -151,16 +152,44 @@ struct PrometheusExporter {
...
+    }
+
+    fn is_expired(&self, time: Instant) -> bool {
+        self.next_flush < time
+    }
+}
+
+impl Hash for MetricEntry {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        BufferMetricEntry::metric_hash(&self.metric, state)
```
bruceg, line 176: Can this not use the `hash` method directly?
```suggestion
        self.metric.hash(state)
```

src/sinks/util/buffer/metrics.rs
```diff
@@ -43,22 +40,34 @@ impl Hash for MetricEntry {
...
     }
 }
 
+impl Eq for MetricEntry {}
+
+impl Hash for MetricEntry {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        Self::metric_hash(&self.0, state)
+    }
+}
```
bruceg, line 63: This shouldn't be necessary, see my comments on the exporter.

src/sources/statsd/mod.rs
```diff
@@ -209,19 +209,15 @@ mod test {
...
-        };
+        let (sender, mut receiver) = mpsc::channel(200);
+        tokio::spawn(async move {
+            let bind_addr = next_addr();
+            let mut socket = UdpSocket::bind(bind_addr).await.unwrap();
+            socket.connect(in_addr).await.unwrap();
+            while let Some(bytes) = receiver.recv().await {
+                socket.send(bytes).await.unwrap();
+            }
+        });
```
bruceg, line 220: I'm not clear how this is related to the prometheus_exporter issue.

bruceg (commented):
The approach looks reasonable, but I have some questions about the details. I'd also want to get comment from @lukesteensen before merging.

src/sinks/prometheus/exporter.rs
fanatid: Thanks for the good suggestion @bruceg! Will rename and will use `MetricEntry` from buffer as inner.

src/sinks/prometheus/exporter.rs
fanatid: I think yes, this should be enough. With **2** seconds above we have **4** in total. I'll add **1** more second to be sure that we are ok :slightly_smiling_face:

lukesteensen (commented):
I'm curious if we really need to track this on every metric. From what I can see, the flush period configuration is for the whole sink. Would it make sense to have a single interval on which we go through and flush all the metrics?

fanatid:
:thinking: I think yes, we can flush all metrics at once and this will be much easier than tracking every metric.

src/sinks/prometheus/exporter.rs
```diff
@@ -271,29 +274,35 @@ impl StreamSink for PrometheusExporter {
...
 
+            // sets need to be expired from time to time
+            // because otherwise they could grow infinitelly
+            let now = Utc::now().timestamp();
+            let interval = now - metrics.last_flush_timestamp;
+            if interval > self.config.flush_period_secs as i64 {
+                metrics.last_flush_timestamp = now;
+
+                let mut map = IndexMap::new();
+                for (MetricEntry(mut metric), is_incremental_set) in metrics.map.drain(..) {
```
fanatid, line 285: Not sure that `drain` is the best way to reset keys, but do not see any other options.

fanatid:
@bruceg @lukesteensen sorry for the delay :disappointed: Updated.

src/sinks/prometheus/exporter.rs
bruceg: I think you could also do this with:
```rust
let mut map = std::mem::replace(&mut metrics.map, IndexMap::new());
for (MetricEntry(mut metric), is_incremental_set) in &mut map {
    if is_incremental_set {
        metric.reset()
    }
}
```

bruceg: Never mind my last comment, I missed that you assign back to `metrics.map`.

I think you're right that this needs a `drain`. The only other way would be to modify the keys in place, and that doesn't work well with iterating. I'd approach it like this:
```rust
let map = metrics.map
    .drain()
    .map(|(MetricEntry(mut metric), is_incremental_set))| {
        if is_incremental_set {
            metric.reset();
        }
        (MetricEntry(metric), is_incremental_set)
    })
    .collect();
metrics.map = map;
```

bruceg (approved):

PR #5264, opened 2020-11-27
enhancement(host_metrics source): integrate upstream container support

eeyun:
This one is a tough one to test. Theoretically this shouldn't alter any behavior whatsoever _currently_

In order to test the behavior we have to be running a build of the branch in a container with the appropriate environment variables set. I'll post some output from my local testing here and ultimately @MOZGIII is adding e2e tests for this in our kubernetes testing components.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -188,7 +188,7 @@ sha2 = "0.9"
 sha3 = "0.9"
 md-5 = "0.9"
 hex = "0.4.2"
-heim = { version = "0.1.0-beta.3", optional = true, features = ["full"] }
+heim = { version = "0.1.0-rc.1", optional = true, features = ["full"] }
 rust_decimal = "1.8.1"
 mongodb = { version = "1.1.1", optional = true }
 anyhow = { version = "1.0.28" }
--- a/src/sources/host_metrics.rs
+++ b/src/sources/host_metrics.rs
@@ -28,6 +28,7 @@ use heim::{
     units::{information::byte, time::second},
     Error,
 };
+
 use serde::{
     de::{self, Visitor},
     Deserialize, Deserializer, Serialize, Serializer,
@@ -690,6 +691,33 @@ fn add_collector(collector: &str, mut metrics: Vec<Metric>) -> Vec<Metric> {
         (metric.tags.as_mut().unwrap()).insert("collector".into(), collector.into());
     }
     metrics
+}
+
+pub fn init_roots() {
+    #[cfg(target_os = "linux")]
+    {
+        match std::env::var_os("PROCFS_ROOT") {
+            Some(procfs_root) => {
+                info!(
+                    message = "PROCFS_ROOT is set in envvars. Using custom for procfs.",
+                    custom = ?procfs_root
+                );
+                heim::os::linux::set_procfs_root(std::path::PathBuf::from(&procfs_root));
+            }
+            None => info!("PROCFS_ROOT is unset. Using default '/proc' for procfs root."),
+        };
+
+        match std::env::var_os("SYSFS_ROOT") {
+            Some(sysfs_root) => {
+                info!(
+                    message = "SYSFS_ROOT is set in envvars. Using custom for sysfs.",
+                    custom = ?sysfs_root
+                );
+                heim::os::linux::set_sysfs_root(std::path::PathBuf::from(&sysfs_root));
+            }
+            None => info!("SYSFS_ROOT is unset. Using default '/sys' for sysfs root."),
+        }
+    };
 }
 
 impl FilterList {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -172,12 +172,23 @@ version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d373d78ded7d0b3fa8039375718cde0aace493f2e34fb60f51cbf567562ca801"
 dependencies = [
- "async-task 4.0.2",
+ "async-task",
  "concurrent-queue",
  "fastrand",
- "futures-lite 1.11.1",
+ "futures-lite",
  "once_cell",
  "vec-arena",
+]
+
+[[package]]
+name = "async-fs"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b3ca4f8ff117c37c278a2f7415ce9be55560b846b5bc4412aaa5d29c1c3dae2"
+dependencies = [
+ "async-lock",
+ "blocking",
+ "futures-lite",
 ]
 
 [[package]]
@@ -192,7 +203,7 @@ dependencies = [
  "async-mutex",
  "async-stream",
  "async-trait",
- "blocking 1.0.2",
+ "blocking",
  "bson",
  "chrono",
  "chrono-tz",
@@ -277,11 +288,11 @@ checksum = "b5bfd63f6fc8fd2925473a147d3f4d252c712291efdde0d7057b25146563402c"
 dependencies = [
  "concurrent-queue",
  "fastrand",
- "futures-lite 1.11.1",
+ "futures-lite",
  "log",
  "nb-connect",
  "once_cell",
- "parking 2.0.0",
+ "parking",
  "polling",
  "vec-arena",
  "waker-fn",
@@ -312,9 +323,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "06de475c85affe184648202401d7622afb32f0f74e02192857d0201a16defbe5"
 dependencies = [
  "async-io",
- "blocking 1.0.2",
+ "blocking",
  "fastrand",
- "futures-lite 1.11.1",
+ "futures-lite",
+]
+
+[[package]]
+name = "async-process"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4c8cea09c1fb10a317d1b5af8024eeba256d6554763e85ecd90ff8df31c7bbda"
+dependencies = [
+ "async-io",
+ "blocking",
+ "cfg-if 0.1.10",
+ "event-listener",
+ "futures-lite",
+ "once_cell",
+ "signal-hook",
+ "winapi 0.3.9",
 ]
 
 [[package]]
@@ -323,7 +350,7 @@ version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c238bd34d425674d8003b8d674cc04baf74e1b71802f3c62451e3bf86f2858ef"
 dependencies = [
- "futures-lite 1.11.1",
+ "futures-lite",
  "rustls 0.18.1",
  "webpki",
 ]
@@ -348,12 +375,6 @@ dependencies = [
  "quote 1.0.7",
  "syn 1.0.54",
 ]
-
-[[package]]
-name = "async-task"
-version = "3.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c17772156ef2829aadc587461c7753af20b7e8db1529bc66855add962a3b35d3"
 
 [[package]]
 name = "async-task"
@@ -597,29 +618,15 @@ checksum = "8d696c370c750c948ada61c69a0ee2cbbb9c50b1019ddb86d9317157a99c2cae"
 
 [[package]]
 name = "blocking"
-version = "0.4.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2468ff7bf85066b4a3678fede6fe66db31846d753ff0adfbfab2c6a6e81612b"
-dependencies = [
- "async-channel",
- "atomic-waker",
- "futures-lite 0.1.11",
- "once_cell",
- "parking 1.0.6",
- "waker-fn",
-]
-
-[[package]]
-name = "blocking"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c5e170dbede1f740736619b776d7251cb1b9095c435c34d8ca9f57fcd2f335e9"
 dependencies = [
  "async-channel",
- "async-task 4.0.2",
+ "async-task",
  "atomic-waker",
  "fastrand",
- "futures-lite 1.11.1",
+ "futures-lite",
  "once_cell",
 ]
 
@@ -1444,9 +1451,9 @@ dependencies = [
 
 [[package]]
 name = "darwin-libproc"
-version = "0.1.2"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9fb90051930c9a0f09e585762152048e23ac74d20c10590ef7cf01c0343c3046"
+checksum = "cc629b7cf42586fee31dae31f9ab73fa5ff5f0170016aa61be5fcbc12a90c516"
 dependencies = [
  "darwin-libproc-sys",
  "libc",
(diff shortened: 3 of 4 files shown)


eeyun:
To test this and prove its working you can do the following:

Make a local build of the branch.
`make package-deb-x86_64`

In my case I created a `custom.toml` file and placed it in `target/artifacts`.
```
[sources.generator]
type = "host_metrics"
collectors = ["cpu", "disk", "filesystem", "load", "memory", "network"]
scrape_interval_secs = 5

[sinks.stdout]
type = "console"
inputs = ["generator"]
target = "stdout"
encoding.codec = "json"
```

Then I modified `distribution/docker/debian/Dockerfile` to specifically test this usecase.

```
FROM debian:buster-slim AS builder

COPY vector-*.deb ./

RUN dpkg -i vector-$(dpkg --print-architecture).deb

RUN rm /etc/vector/vector.toml

COPY custom.toml /etc/vector/vector.toml

FROM debian:buster-slim

RUN apt-get update && apt-get install -y ca-certificates tzdata systemd && rm -rf /var/lib/apt/lists/*

COPY --from=builder /usr/bin/vector /usr/bin/vector
COPY --from=builder /usr/share/doc/vector /usr/share/doc/vector
COPY --from=builder /etc/vector /etc/vector

RUN mkdir /host

VOLUME /var/lib/vector/

# Smoke test
RUN ["vector", "--version"]

ENTRYPOINT ["/usr/bin/vector"]
```
Built the container
```
❯❯❯ docker build --tag procroot target/artifacts -f distribution/docker/debian/Dockerfile
```

Finally, ran the container with environment variables set, networking configured to host and privileged mode:
```
❯❯❯ sudo docker run --privileged -e PROCFS_ROOT="/host/proc" -e SYSFS_ROOT="/host/sys" -v /proc:/host/proc:rw -v /sys:/host/sys:rw --network host --name mycontainer -it --rm procroot
```

Should net you some output that is distinct to the host:
```
{"name":"network_receive_packets_total","namespace":"host","timestamp":"2020-12-08T22:44:16.700524279Z","tags":{"collector":"network","device":"docker0","host":"curse"},"kind":"absolute","counter":{"value":136452.0}}
{"name":"network_transmit_bytes_total","namespace":"host","timestamp":"2020-12-08T22:44:16.700524279Z","tags":{"collector":"network","device":"docker0","host":"curse"},"kind":"absolute","counter":{"value":2203702506.0}}
{"name":"network_transmit_errs_total","namespace":"host","timestamp":"2020-12-08T22:44:16.700524279Z","tags":{"collector":"network","device":"docker0","host":"curse"},"kind":"absolute","counter":{"value":0.0}}
{"name":"network_transmit_packets_drop_total","namespace":"host","timestamp":"2020-12-08T22:44:16.700524279Z","tags":{"collector":"network","device":"docker0","host":"curse"},"kind":"absolute","counter":{"value":0.0}}
{"name":"network_transmit_packets_total","namespace":"host","timestamp":"2020-12-08T22:44:16.700524279Z","tags":{"collector":"network","device":"docker0","host":"curse"},"kind":"absolute","counter":{"value":221988.0}}
```

src/sources/host_metrics.rs
```diff
@@ -692,6 +694,22 @@ fn add_collector(collector: &str, mut metrics: Vec<Metric>) -> Vec<Metric> {
     metrics
 }
 
+#[cfg(target_os = "linux")]
+pub fn maybe_set_roots() {
```
bruceg, line 698: I suggest something like `init_roots`, and call with `host_metrics::init_roots` in `app.rs`.

src/sources/host_metrics.rs
```diff
@@ -692,6 +694,22 @@ fn add_collector(collector: &str, mut metrics: Vec<Metric>) -> Vec<Metric> {
...
+    let virt = futures::executor::block_on(heim::virt::detect());
+
+    if let Some(virtualized) = virt {
+        if virtualized.is_container() {
+            if std::env::var_os("PROCFS_ROOT").is_some() {
+                heim_runtime::linux::set_procfs_root("PROCFS_ROOT".to_string());
+            }
```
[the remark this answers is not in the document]
eeyun: `set_procfs_root` takes a path value and configures heim to use it rather than the default procfs root path.

Cargo.toml
```diff
@@ -187,6 +187,7 @@ sha3 = "0.9"
 md-5 = "0.9"
 hex = "0.4.2"
 heim = { version = "0.1.0-beta.3", optional = true, features = ["full"] }
+heim-runtime = { git = "https://github.com/eeyun/heim", branch = "eeyun/proc_sys_cfg" }
```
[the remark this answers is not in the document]
eeyun: `heim` doesn't include `heim_runtime` in it's root so it has to be depended upon separately.

src/sources/host_metrics.rs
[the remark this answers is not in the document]
bruceg: That's what I thought, but this code appears to pass the literal path `PROCFS_ROOT` instead of the value of that environment variable.

src/sources/host_metrics.rs
```diff
@@ -692,6 +694,22 @@ fn add_collector(collector: &str, mut metrics: Vec<Metric>) -> Vec<Metric> {
...
+            if std::env::var_os("SYSFS_ROOT").is_some() {
+                heim_runtime::linux::set_sysfs_root("PROCFS_ROOT".to_string());
```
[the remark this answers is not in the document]
eeyun: Oh yeah apologies this is an old commit! whoops!

src/sources/host_metrics.rs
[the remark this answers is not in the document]
eeyun: Ah I see what happened - this was the wrong commit. Should be resolved!

src/sources/host_metrics.rs
```diff
@@ -692,6 +694,22 @@ fn add_collector(collector: &str, mut metrics: Vec<Metric>) -> Vec<Metric> {
     metrics
 }
 
+#[cfg(target_os = "linux")]
+pub fn init_roots() {
```
bruceg, line 698: I wonder if the `#[cfg]` shouldn't be inside this function to reduce the complexity upstream.

src/sources/host_metrics.rs
```diff
@@ -692,6 +694,22 @@ fn add_collector(collector: &str, mut metrics: Vec<Metric>) -> Vec<Metric> {
...
+    if let Some(virtualized) = virt {
+        if virtualized.is_container() {
+            if let Some(procfs_root) = std::env::var_os("PROCFS_ROOT") {
+                heim_runtime::linux::set_procfs_root(PathBuf::from(&procfs_root));
+            }
+            if let Some(sysfs_root) = std::env::var_os("SYSFS_ROOT") {
+                heim_runtime::linux::set_sysfs_root(PathBuf::from(&sysfs_root));
+            }
+        }
+    }
```
bruceg, line 710: Is this feature useful for non-virtual environments, like for doing testing? That is, would there be any problems with removing the `virtualized` guard here?

src/sources/host_metrics.rs
```diff
@@ -692,6 +694,22 @@ fn add_collector(collector: &str, mut metrics: Vec<Metric>) -> Vec<Metric> {
     metrics
 }
 
+#[cfg(target_os = "linux")]
+pub fn init_roots() {
+    let virt = futures::executor::block_on(heim::virt::detect());
```
bruceg, line 699: Should this be an `async fn` to just use `.await` here?

src/sources/host_metrics.rs
eeyun: I can't imagine that it would be super useful? But, I could be wrong!

src/sources/host_metrics.rs
eeyun: We _can_ make this async, but then any synchronous function that calls this one like we're doing in `app.rs` has an unused implementer of `futures::Future` which breaks our lints

src/sources/host_metrics.rs
eeyun: In heim these functions can _only_ be implemented for linux so I put the guard here avoid this call site breaking at runtime on windows or darwin.

bruceg: Right. I was suggesting moving the `#[cfg]` inside the function, so that the function isn't omitted on non-Linux. As it stands now, it is always referenced in `fn prepare_from_opts`, so it won't compile on other OS.

src/sources/host_metrics.rs
bruceg: Right, we would have to move the blocking into the caller(s), which would require them to either become async themselves or move the `block_on` there.

eeyun: Ah indeed - I more meant we have to block on it somewhere, so if you'd prefer to do so at the call site just say the word.

src/sources/host_metrics.rs
eeyun: Oh yes! My apologies, I completely misunderstood what you were saying 😅

src/sources/host_metrics.rs
bruceg: Hmm, looking at it further, I'm surprised this works. AFAIK `block_on` depends on an async runtime being running, but the runtime isn't created until later in `fn prepare_from_opts`. I would have expected this to throw an error.

In any case, this kind of issue is why I would tend to want to move the blocking into the caller, particularly for early init functions, to make it obvious that a runtime must be running.

eeyun: I think that's a very good call @bruceg ! I'll push a fix first thing in the am

src/sources/host_metrics.rs
MOZGIII: Let’s not apply additional constraints - it’s already an opt-in - it makes sense to keep it simple to make it more predictable.

src/sources/host_metrics.rs
MOZGIII: The reason it works is it’s not a `tokio` executor, but `futures` - it simply run the future to completion on the current thread (without `spawn` support, but with the ability to manage concurrent execution within one feature).
@eeyun I didn’t expect this to involve async operation, I assumed we’d just assign values from the env syncronously. I still think we don’t need to go beyond that.

src/sources/host_metrics.rs
MOZGIII: Btw, containers are sort of a well known use case for this, but I’ve seen special dev environments with custom linux-based OS where the procfs and sysfs were mounted in non-standard locations. It’s hardly a case for Vector production use, but there might be users with such a setup out there.

src/sources/host_metrics.rs
```diff
@@ -28,13 +28,15 @@ use heim::{
     units::{information::byte, time::second},
     Error,
 };
+#[cfg(target_os = "linux")]
+use heim_runtime;
```
MOZGIII, line 32: What if we move this use to the function itself so it doesn’t have to be kept track of separately?

MOZGIII:
I'm running with `PROCFS_ROOT=/qwerty` (invalid path) and getting no errors. Something must be broken.
It's an early report, I'll continue my testing in the meantime.

Testing with the following settings:

```shell
cat <<'EOF' > cfg.toml
[sources.host_metrics]
type = "host_metrics"
scrape_interval_secs = 5

[sinks.prometheus]
type = "prometheus"
inputs = ["host_metrics"]
address = "0.0.0.0:9090"
EOF
```

Without fs passthrough:

```shell
docker run --rm -it -v "$PWD:$PWD" --cpus 2 -p 9090:9090 test --config "$PWD/cfg.toml"
```

With fs passthrough:

```shell
docker run --rm -it -v "$PWD:$PWD" --cpus 2 -p 9090:9090 -v /proc:/host/proc:ro -e PROCFS_ROOT=/host/proc -v /sys:/host/sys:ro -e SYSFS_ROOT=/host/sys test --config "$PWD/cfg.toml"
```

I Will post updates here as I go.

src/sources/host_metrics.rs
```diff
@@ -692,6 +694,23 @@ fn add_collector(collector: &str, mut metrics: Vec<Metric>) -> Vec<Metric> {
...
 }
 
+pub async fn init_roots() {
+    #[cfg(target_os = "linux")]
+    {
+        let virt = heim::virt::detect().await;
+        if let Some(virtualized) = virt {
+            if virtualized.is_container() {
+                if let Some(procfs_root) = std::env::var_os("PROCFS_ROOT") {
+                    heim_runtime::linux::set_procfs_root(PathBuf::from(&procfs_root));
```
MOZGIII, line 704: Let's add a debug level print here (and below) to notify users that the non-standard path is picked up for use.

src/sources/host_metrics.rs
eeyun: Not sure what to tell you about that. Nothing in the heim functions we added is explicitly async so I was surprised by this myself.

src/sources/host_metrics.rs
eeyun: This can actually be completely removed. It's a redundant inclusion.

src/sources/host_metrics.rs
bruceg: `fn heim::virt::detect` is async, and digging a bit, what it calls involves some async filesystem calls (reads from proc and sys, etc). Since you removed that bit, there is no longer any need for this to be async, and so no `block_on`.

@MOZGIII I'm confused then. Without tokio, what handles things like non-blocking I/O that are invoked by the futures?

bruceg (approved):
LGTM once CI checks clear

src/sources/host_metrics.rs
MOZGIII: In my local testing, I figured this branch is likely not executing at all, presumably because the `heim::virt::detect()` is broken.

MOZGIII:
Please wait a bit with the merge, I'd like to finish testing.

MOZGIII:
Btw, here's a tip on what to expect from the metrics (i.e. how to compare on an obvious examle):

```shell
$ docker run --rm -i busybox cat /sys/fs/cgroup/memory/memory.limit_in_bytes
9223372036854771712
$ docker run --rm -i -m 512m busybox cat /sys/fs/cgroup/memory/memory.limit_in_bytes
WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
536870912
$ docker run --rm -i -m 512m -v /sys:/host/sys busybox cat /host/sys/fs/cgroup/memory/memory.limit_in_bytes
WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
9223372036854771712
```

eeyun:
Aside from our testing - this won't be ready to merge until the upstream module re-exports pass CI and we alter our consumption here. Should be momentarily though.

src/sources/host_metrics.rs
MOZGIII: Non-blocking io can still function as long as there's a reactor loop. (I suspect the terminology I'm used to is different from the naming conventions used in `tokio` - so take this with a grain of salt).
Reactor loop is not necessarily a scheduler and doesn't necessarily handle "task" management and lifecycle - it just has to poll the async function when the Waker signals, and park the thread when the future returns pending. Afaik, In the case of async io, the Waker abstraction is enough to handle the progress.

We essentially only need something like tokio if we want to schedule tasks over a thread pool with a scheduler, or if we need a timer (which doesn't have to be handled at the reactor level, but in the current ecosystem it is). IO operations are portable across the "async runtimes".

src/sources/host_metrics.rs
```diff
@@ -692,6 +693,33 @@ fn add_collector(collector: &str, mut metrics: Vec<Metric>) -> Vec<Metric> {
...
 
+pub fn init_roots() {
+    #[cfg(target_os = "linux")]
+    {
+        match std::env::var_os("PROCFS_ROOT") {
+            Some(procfs_root) => {
+                info!(
+                    "PROCFS_ROOT is set in envvars. Using {:?} for procfs.",
+                    procfs_root
+                );
```
MOZGIII, line 704: I think there was an overhaul of logs in which we moved things to structured logging form from the formatting form as much as possible. Not sure if it's a strict rule now to use the structured form, cc @jamtur01

```suggestion
                info!(
                    message = "PROCFS_ROOT is set in envvars. Using custom for procfs.",
                    ?procfs_root
                );
```
It would look something like this if we were to used structured form.

The same should probably be applied to the logs below.

MOZGIII (approved):

PR #5277, opened 2020-11-28
enhancement(remap): support path queries for variables

### LINKED ISSUE #5094: Remap lang: support querying paths in variables
Currently, this works:

```rust
.foo.bar.(baz | qux)[2]
```

This does not:

```rust
$foo = .foo				// ok
$foo.bar.(baz | qux)[2]	// syntax error
```

This is because the parser doesn't allow the syntax in which a variable (`$foo`) is followed by a query path (`.bar.(baz | qux)[2]`).

This is a gap in usability of the language, and something we need to tackle.

JeanMertz:
Before, this didn't work, but now it does:

```json
{ "foo": { "bar": "baz" } }
```

```rust
$foo = .foo
$foo.bar // "baz"
```

That is to say, you can now append a query path to a variable to fetch nested values, if any.

This behaves similar to regular paths.

We currently don't (and maybe never) support this syntax for variables:

_update: there seems to be a desire to **do** support this, so I've filed #5350 to track this feature_

```rust
$foo.bar = .foo.bar
// ^ remap error: parser error: paths in variable assignment not supported. Use "$foo" without ".bar"
```

Technically, we could treat variables the same as paths, but whereas a path references an object provided by an external source (e.g. for vector this is the `LogEvent`), a variable would point to an internally managed store. Both would implement `remap::Object`. This would mean that we could unify pathing behaviour for both `$foo` and `.foo`, the only difference being `$` referencing the internal store and `.` referencing the `LogEvent`.

I opted not to do this in this PR, as that behaviour might be surprising to people (it is less surprising if you consider `.foo` and `$foo` to be similar, but one pointing to _internal temporary state_ and the other to _external state_, but since the behaviour differs from how variables regularly work, I still think it's _more_ confusing in the end.

---

To add the above described support, I did had to make some significant changes to our pathing logic, as it was getting more difficult to work with, and also had a number of untested bugs. I opted _not_ to optimize the intergration between Vector and Remap too much, as we'll have to refactor this anyway, once the `Lookup` work by Ana lands.

I did add a heap of tests to make sure pathing works as expected now, and any changes in the future can rely on these tests to validate the updated implementation.

closes #5094


lib/remap-lang/src/parser.rs
```diff
@@ -504,7 +693,7 @@ mod tests {
             (
                 // Regexes can't be parsed as part of a path
                 r#".foo = split(.foo, ./[aa]/)"#,
-                vec![" 1:21\n", "= expected path_segment"],
+                vec![" 1:22\n", "= expected not"],
```
JeanMertz, line 696: This error is somewhat cryptic (better error messages is still on the TODO list), but the change is caused because you could technically do `. / 5` (in reality you can't because for Vector, `.` will always point to a map, so we might introduce a custom error for this case).

src/event/util/log/remove.rs
```diff
@@ -27,10 +27,10 @@ fn remove_array(
 ) -> Option<(Value, bool)> {
     match path.next()? {
         PathComponent::Index(index) => match path.peek() {
-            None => array_remove(array, index).map(|v| (v, false)),
+            None => array_remove(array, index).map(|v| (v, array.is_empty())),
             Some(_) => array
                 .get_mut(index)
-                .and_then(|value| Some((remove_rec(value, path, prune)?.0, false))),
+                .and_then(|value| remove_rec(value, path, prune)),
```
JeanMertz, line 33: This makes it so that arrays can also be pruned, similar to maps.

tests/behavior/transforms/remap.toml
```diff
@@ -215,14 +215,14 @@
       "b.equals" = "10"
       "c.equals" = "10"
       "d.equals" = "10"
-      "e.equals" = "20"
-      "f.equals" = "20"
+      "e.equals" = ""
+      "f.equals" = ""
```
JeanMertz, line 219: Because `to_string(null)` now returns `""`, these tests had to be updated.

However, I'm pretty sure that change was made in earlier PRs that are in master now, so I'm unsure why the CI didn't fail?

jszwedko (commented):
Nice!

I feel like I'm still not quite up-to-speed enough on remap internals to add much of value there, but nothing jumped out to me.

However, I do think we should allow assignment to. If we can do, for example:

```
$foo = .foo
$foo.bar // "baz"
```

I would have also expected to be able to assign via `$foo.bar = "something else"`. If we don't do that as part of this PR, I'd personally advocate for it to be a quick follow-up as the symmetry of allowing assignment feels very natural to me and potentially useful.

It looks like there are some failing checks.

lib/remap-lang/src/value/object.rs
```diff
@@ -0,0 +1,327 @@
...
+                vec![
+                    Field(Regular("foo".to_owned())),
+                    Field(Quoted("bar baz".to_owned())),
+                    Index(0),
+                ],
+                false,
+                Some(map!["foo": map!["bar baz": Value::Array(vec![])], "bar": "baz"].into()),
+            ),
+            (
+                map!["foo": map!["bar baz": vec![0]], "bar": "baz"].into(),
```
Hoverbear, line 245: Are you able to do like `map! { "foo": map! { "bar baz": true } }`? I notice the `[]` brackets make me think array. 😅

src/event/mod.rs
```diff
@@ -640,4 +598,279 @@ mod test {
...
+        for (object, segments, value, expect, result) in cases {
+            let object: BTreeMap<String, Value> = object;
+            let mut event = Event::from(object);
+            let expect = Event::from(expect);
+            let value: remap::Value = value;
+            let path = Path::new_unchecked(segments);
+
+            assert_eq!(event.insert(&path, value.clone()), result);
+            assert_eq!(event, expect);
+            assert_eq!(event.get(&path), Ok(Some(value)));
```
Hoverbear, line 757: I've noticed you tend to use this pattern often, during debugging I've noticed this can be hard to identify the case and intent of the test. Do you think we could help future authors by better describing the cornercases being tested when they are not obvious?

src/event/util/log/remove.rs
Hoverbear: Ah I did this in the lookup work too. :) Good choice.

Hoverbear (approved):
Nothing blocking. :) Good improvements.

lib/remap-lang/src/value/object.rs
```diff
@@ -0,0 +1,327 @@
+use crate::{Object, Path, Value};
+
+impl Object for Value {
+    fn insert(&mut self, path: &Path, value: Value) -> Result<(), String> {
+        Ok(self.insert_by_path(path, value))
+    }
+
+    fn get(&self, path: &Path) -> Result<Option<Value>, String> {
+        Ok(self.get_by_path(path).cloned())
```
FungusHumungus, line 9: Probably out of scope for this PR, but it would be nice if we could avoid this `clone` here.

As far as I'm aware the lifetime of the returned Value is always going to be the same as the Object, so it shouldn't be too hard to make it able to return a reference to the Value instead.

lib/remap-lang/src/path.rs
```diff
@@ -0,0 +1,445 @@
...
+
+#[derive(thiserror::Error, Clone, Debug, PartialEq)]
+pub enum Error {
+    #[error("unable to create path from alternative string: {0}")]
+    Alternative(String),
+}
+
+/// Provide easy access to individual [`Segment`]s of a path.
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct Path {
```
FungusHumungus, line 13: I do get confused with two different `Path` structs in the same project. It's hard to keep track of which instance is being referred to. Is there a different naming strategy we could use? Perhaps `PathSegments`?

JeanMertz:
> I would have also expected to be able to assign via `$foo.bar = "something else"`.

I wrote my motivation for not supporting this (yet) in the PR description. The TL;DR version is that I'm worried it might confuse people coming from other languages that don't support this pattern.

That is, if one were to write:

```rust
$foo = "bar"
$foo.baz = "qux"
```

Now, `$foo` would actually be `{ "baz": "qux" }` instead of `"bar"`.

One the one hand, I agree with you that the "cleanness" of having `.` and `$` be similar, as I wrote in the PR description:

> it is less surprising if you consider `.foo` and `$foo` to be similar, but one pointing to _internal temporary state_ and the other to _external state_

There are definitely upsides to treating both the same:

* We would be able to share almost all code between the two target types, you're either assigning to the internal state using `$`, or the external state using `.`, and then can use paths to assign to map fields/array indices.
* Conceptually both would be treated and documented equally
* You can use both interchangeably.

There are also downsides:

* As mentioned above, it might be confusing to newcomers
* If we _truly_ want to treat them equally, then `$ = { "foo": "bar" }` would also be allowed, which might be even weirder for newcomers?

As the implementer of this feature, it's definitely tempting to go this route, as it'll unify a lot of logic, but I'm still not convinced it won't be confusing to users (although I can also see how this might actually be _less convincing_ to users once you realize the difference between internal and external state).

I'll go ahead and merge as-is, but I'm happy to discuss this further (I'll file an issue) and perhaps make the required change.

lib/remap-lang/src/path.rs
JeanMertz: I'm probably going to merge the expression and the `Path`, since there isn't really any reason to keep both around.

lib/remap-lang/src/parser.rs
```diff
@@ -301,65 +353,103 @@ impl Parser<'_> {
...
+            .parse::<usize>()
+            .map_err(|_| e(R::path_index_inner))?;
+
+        Ok(path::Segment::Index(index))
     }
 
     /// Parse a [`Variable`] value, e.g. "$foo"
-    fn variable_from_pair(&self, pair: Pair<R>) -> Result<Expr> {
-        let ident = pair.into_inner().next().ok_or(e(R::variable))?;
+    fn variable_from_pair(&self, pair: Pair<R>) -> Result<Variable> {
```
FungusHumungus, line 430: I'm not sure I understand why this needs to work differently from `path_from_pair`?

Is the plan to tidy this up with #5350?

JeanMertz: It differs exactly because of the reasons listed in the issue you linked.

That is to say, a variable right not _has_ to include a top-level identifier: `$foo`, you can't write `$`, whereas `.` is a valid path.

So, this means we have to parse paths attached to a variable a bit differently from paths.

It would be cleaned up if we did what I described in #5350, yes, but I we'll have to reach consensus first.

FungusHumungus: So, they are both essentially just a list of segments, but they differ in that one requires at least one `path_segment` segment, and the other is ok to have none?

Couldn't the code to parse a `Vec` of segments be extracted out and shared between both?

Pest code is not the easiest to follow, so there's a good chance I'm missing something! But I think it is important to aim for as much simplicity here as possible.

JeanMertz: This is related to my comment here: https://github.com/timberio/vector/pull/5277#discussion_r532074268

I'll probably do another pass in a follow-up PR, but right now these two differ too much to unify more without changing the grammar (or without making the parser even more complex).

It all stems from these two rules that differ:

```
variable = ${ "$" ~ ident ~ path_index* ~ ("." ~ path_segments)?  }
```

```
path             = ${ "." ~ path_segments? ~ !(WHITESPACE* ~ ("." | "\"")) }
path_segments    = ${
    (
        ((path_segment | path_index+) ~ path_index*)
        ~ ("." ~ path_segment ~ path_index*)*
    )
    | path_index+
}
path_segment     = ${ path_field | path_coalesce }
```

FungusHumungus: Got it.

At the risk of over labouring a point, is there a reason they couldn't share the rules:

```
path             = ${ ("." | "$") ~ path_segments? ~ !(WHITESPACE* ~ ("." | "$" | "\"")) }
path_segments    = ${
    (
        ((path_segment | path_index+) ~ path_index*)
        ~ (("." | "$") ~ path_segment ~ path_index*)*
    )
    | path_index+
}
path_segment     = ${ path_field | path_coalesce }
```

Validation dissallowing a lone `$` could still occur later on in code?

Yeah, I'm all for #5350.

JeanMertz: As long as we don't have #5350 yet, I like to keep a conceptual difference between paths and variables. If we do the check later in the code, then we lose the useful error message we get from Pest.

We do still want to improve error messages in general, but we haven't done so yet, unfortunately.

FungusHumungus: Yes, good point on the error messages.

StephenWakely (approved):
Really nice work! A number of the bugs fixed here are much needed!

lib/remap-lang/src/value/object.rs
JeanMertz: Yeah, I've been thinking of building up some flame charts and checking where TRL currently spends the most time. I would love to remove a bunch of clones we do right now by starting to introduce lifetimes (we currently don't borrow any data anywhere, so there's definitely room for improvement here).

PR #5278, opened 2020-11-28
enhancement(remap): add support for literal arrays

### LINKED ISSUE #5219: remap-lang: add support for array literals
Currently, this works:

```rust
.foo[0] = "foo"
.foo[1] = "bar"
.foo[2] = "baz"
```

But this doesn't:

```rust
.foo = ["foo", "bar", "baz"]
```

We want to fix that.

_(it should be noted that the outcome of these two examples can differ if `.foo` is an existing array, in which case the second example \*overwrites\* the array, whereas the first only changes (or adds) the first three elements in the array)_

Aside from the obvious UX improvements, this feature will make it possible for functions to take in a list of values. E.g.:

```rust
redact(.foo, patterns = ["address", "credit_card"])
```

ref #4519

JeanMertz:
Add support for literal arrays:

```rust
.foo = ["foo", 5, ["bar"]]
```

closes #5219

### DIFF
--- a/lib/remap-lang/src/lib.rs
+++ b/lib/remap-lang/src/lib.rs
@@ -180,6 +180,19 @@ mod tests {
                 Err(r#"remap error: parser error: paths in variable assignment not supported. Use "$foo" without ".[0]""#),
                 Ok(().into()),
             ),
+            (r#"["foo", "bar", "baz"]"#, Ok(()), Ok(vec!["foo", "bar", "baz"].into())),
+            (
+                r#"
+                    .foo = [
+                        "foo",
+                        5,
+                        ["bar"],
+                    ]
+                    .foo
+                "#,
+                Ok(()),
+                Ok(vec!["foo".into(), 5.into(), Value::Array(vec!["bar".into()])].into()),
+            ),
         ];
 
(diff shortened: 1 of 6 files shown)


bruceg (approved):

lib/remap-lang/src/lib.rs
```diff
@@ -240,6 +240,19 @@ mod tests {
...
+                r#"
+                    .foo = [
+                        "foo",
+                        5,
+                        ["bar"],
+                    ]
+                    .foo
+                "#,
+                Ok(()),
+                Ok(vec!["foo".into(), 5.into(), Value::Array(vec!["bar".into()])].into()),
```
FungusHumungus, line 254: I added an `array!` macro that saves you from typing `.into()` for every element. Would be useful here..

https://github.com/timberio/vector/blob/master/src/test_util/mod.rs#L542

StephenWakely (approved):
Nice.

lib/remap-lang/src/lib.rs
JeanMertz: Sweet 👌

PR #5279, opened 2020-11-28
enhancement(remap): add support for literal maps

JeanMertz:
Similar to #5278, but for maps:

```rust
$bar = "bar"
.foo = { "foo": $bar }
```

Note that blocks are still working as expected:

```rust
{ "foo" } 			// block, result "foo"
{ "foo": "bar" } 	// map, result { "foo": "bar" }
```

This PR also removes the support for empty blocks, and instead `{}` now resolves to an empty map.

There is no tracking issue for this, as it hasn't come up yet in discussions, but after working on #5278, I didn't expect this to be difficult to support, and went ahead to see if I was right. Now that I know I was, we can still decide we don't have a need for map literals at this point, although it _is_ somewhat surprising that we support all other value types (except timestamps), but not this one.

I can see an _eventual_ use-case for being able to dynamically set map fields, but we can either do that through a function (e.g. `rename(.foo.bar, "baz")`), or add direct support within the language grammar at a later point (e.g. `{ .foo: $bar }`). Or we cannot do it at all, if there is no feedback from users requesting this feature.


StephenWakely (approved):
I think this is a good idea!

It would be nice to add a test to see how an empty map/block parses - `{ }`. It's not clear which way it would go. If it is a block it would ultimately resolve to `Null`, if a map it would be an empty `Map`. Should an empty map resolve to `Null` anyway?

JeanMertz:
> It would be nice to add a test to see how an empty map/block parses - `{ }`. It's not clear which way it would go. If it is a block it would ultimately resolve to `Null`, if a map it would be an empty `Map`. Should an empty map resolve to `Null` anyway?

Good point. It would parse as a block (higher precedence).

I remember adding support for empty blocks "since there's no reason not to support them", but with this change, we might just disallow empty blocks in the grammar, as there's no use-case for them, and it would disambiguate `.foo = {}` to resolve to an empty map (I don't think we should treat empty maps as `null`).

Issue #5221, opened 2020-11-25
remap-lang: add initial `redact` function


JeanMertz:
This is based on #4519, and builds on top of #5219 and #5220. 

The goal is to have an initial version of the `redact` function, with support for the generic `pattern` filter:

```rust
/// The type of filter to be applied to the provided string value.
enum Filter {
	Pattern
	// Address
	// Creditcard
	// ...
}

/// The recipe for redacting the matched filters.
enum Redactor {
	Full
	// Partial
	// Descriptive
	// Custom
}

fn redact(
	value = String,
	filters = [Filter],
	redactor = Redactor,
	patterns = [String or Regex],
) -> String
```

More `filter` and `redactor` variants will be added in future iterations.

The `patterns` parameter is used when `filters` contains the `pattern` variant.

```json
{ "sensitive": "foo bar baz qux quux"
```

```rust
.sensitive = redact(.sensitive, filters = ["pattern"], redactor = "full", patterns = ["bar", /qu+x/])
```

```json
{ "sensitive": "foo **** baz ****" }
```

ref #4519

PR #5283, opened 2020-11-29
enhancement(remap): accept array of arguments for function parameters

### LINKED ISSUE #5220: remap-lang: support enum-based array parameters
Building on #5219, we want functions to require a list of enum variants for a given parameter.

Given this example:

```rust
redact(.foo, patterns = ["address", "credit_card", "non-existent"])
```

The `redact` function should be able to qualify its `patterns` parameter as taking only a list of string literals, which can then be checked at compile-time.

The above program would stop Vector from booting with a message similar to:

> error for function "redact": invalid enum variant "non-existent" for parameter "patterns"

ref #4519

JeanMertz:
This PR adds support for providing a list of arguments for a single function parameter.

```rust
my_func(values = ["foo", /bar/, true])
```

It should be noted that these _differ_ from expression-based arrays, which already support passing in a `Value::Array`:

```rust
.foo = ["foo", "bar", true]
my_func(values = .foo)

// or

my_func(values = ["foo", "bar", true])
```

That is to say, the former supports mixing expressions and **regex patterns**, whereas the  latter only supports expressions.

The reason for this is that Remap does not support using regex patterns as expressions, as we have no (acceptable) way of representing them as a `Value`.

So, this works:

```rust
.baz = "baz"
.foobar = ["foo", "bar"]
my_func(value = .baz)
my_func(value = /foo/)
my_func(values = .foobar)
my_func(values = [/foo/, /bar/])
my_func(values = [/foo/, /bar/, .baz])
```

This does not:

```rust
.baz = /baz/ 			// ERROR: expected expression
.baz = [/baz/, "bar"] 	// ERROR: expected expression
```

It should be noted that this is unlikely to ever really cause any confusion/issues, but we definitely will want to document this behaviour.

---

Additionally, the PR also supports requiring a list of enum variants for a given function parameter:

```rust
my_func(variants = ["foo", "qux"])
// remap error: function error: unknown enum variant: qux, must be one of: foo, bar, baz
```

This feature will be used in a follow-up PR implementing #5221.

closes #5220

### DIFF
--- a/src/remap/function/replace.rs
+++ b/src/remap/function/replace.rs
@@ -35,7 +35,7 @@ impl Function for Replace {
 
     fn compile(&self, mut arguments: ArgumentList) -> Result<Box<dyn Expression>> {
         let value = arguments.required_expr("value")?;
-        let pattern = arguments.required("pattern")?;
+        let pattern = arguments.required_expr_or_regex("pattern")?;
         let with = arguments.required_expr("with")?;
         let count = arguments.optional_expr("count")?;
 
@@ -108,6 +108,7 @@ impl Expression for ReplaceFn {
 
                 Ok(replaced)
             }
+            Argument::Array(_) => unreachable!(),
         }
     }
 
@@ -124,6 +125,7 @@ impl Expression for ReplaceFn {
         let pattern_def = match &self.pattern {
             Argument::Expression(expr) => Some(expr.type_def(state).fallible_unless(Kind::Bytes)),
             Argument::Regex(_) => None, // regex is a concrete infallible type
+            Argument::Array(_) => unreachable!(),
         };
 
         self.value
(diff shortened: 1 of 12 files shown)


Hoverbear:
Oh heck this is neat, Jean!

src/remap/function/replace.rs
```diff
@@ -108,6 +108,7 @@ impl Expression for ReplaceFn {
 
                 Ok(replaced)
             }
+            Argument::Array(_) => unreachable!(),
```
Hoverbear, line 111: It'd be nice to put a small message alongside these unreachables (eg "function does not accept this arg type" for future readers.

Hoverbear (approved):

src/remap/function/replace.rs
JeanMertz: I changed how arrays are handled in https://github.com/timberio/vector/pull/5308, so this will be removed once that PR is merged.

PR #5291, opened 2020-11-30
chore(config): Reject unknown API fields

### LINKED ISSUE #5239: Invalid `[api]` options should be rejected
We had a situation recently where `api.bind` was renamed to `api.address`, with vector.toml still using the old setting. Rather than fail, it silently fell back to the default 127.0.0.1:8686 `address` valid.

Unknown keys should be rejected from the config, as they are in topology configuration sections.

leebenson:
Closes #5239.


jamtur01 (approved):
LGTM

PR #5292, opened 2020-11-30
chore(graphql api): Async-graphql 2.1.6

leebenson:
Bumps to async-graphql 2.1.6.

Some fixes for fragments, Apollo Tracing (which we're not currently using, but might be useful later) and general bug fixes.

@sghall - tagged you for a quick sanity check that the web client still works as expected.


sghall (approved):
Tried it out against the app.  Looks good 👍

PR #5296, opened 2020-11-30
enhancement(observability): Enable TLS subscription connections in vector top

sghall:
ping @leebenson 

This PR enables TLS subscription connections and allows users to get the error message on connection failures. 

I noticed that I could not connect to a remote instance using secure sockets in `vector top --url <some url>`.    I first added a log of the actual error which showed like this:

```
Couldn't connect to Vector API via WebSockets: Url("TLS support not compiled in.")
```

I think we should keep that so users can get a little more info when something goes wrong (optional).  We could also, use the same strategy for the other errors in this file (glad to add here).   So the fix was to enable the TLS feature in `tokio-tungstenite` and it now works against the remote instance.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6395,8 +6395,10 @@ checksum = "6d9e878ad426ca286e4dcae09cbd4e1973a7f8987d97570e2469703dd7f5720c"
 dependencies = [
  "futures-util",
  "log",
+ "native-tls",
  "pin-project 0.4.27",
  "tokio",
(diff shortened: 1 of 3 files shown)


sghall:
We touched on this in discord, but permission to merge, Cap'n @leebenson?  Need that validation of a nice ✔️ on there 😄

leebenson (approved):
Apologies, the Discord comment wasn't very definitive. Consider this ✅

PR #5297, opened 2020-11-30
feat(remap): add `redact` function

### LINKED ISSUE #5221: remap-lang: add initial `redact` function

JeanMertz:
This PR implements the first iteration of the `redact` function:

```rust
.input = "hello world, hello universe"

redact(.input, filters = ["pattern"], patterns = ["hello"])
// **** world, **** universe

redact(.input, filters = ["pattern"], patterns = ["hello", "wor"])
// **** ****ld, **** universe

redact(.input, filters = ["pattern"], patterns = [/world|universe/])
// hello ****, hello ****

redact(.input, filters = ["pattern"], patterns = [])
// hello world, hello universe

redact(.input, filters = ["pattern"], patterns = ["hello", /[uieao]/])
// **** w****rld, **** ****n****v****rs****
```

I added a bunch of behaviour tests, but opted not to add unit tests yet, as the implementation is likely to change over the next couple of PRs as we add more filters and redactors (see #4519)

closes #5221


jszwedko (commented):
Do we have an idea of what the ultimate form of this will look like? Is it https://github.com/timberio/vector/issues/4519#issuecomment-732220424 ?

I ask because the current API seems a bit awkward where certain options (i.e. `patterns`) only apply to certain filters (i.e.`pattern`). Rather than having these ignored parameters, I might have separate functions (like `redact_patterns`).

Alternatively, we could consider allowing regexes directly in `filters`. For example: ` redact(.input, filters = ["pattern"], patterns = ["hello"])` would become` redact(.input, filters = [/hello/])`. This also removes the dependent parameter.

JeanMertz:
> Do we have an idea of what the ultimate form of this will look like? Is it [#4519 (comment)](https://github.com/timberio/vector/issues/4519#issuecomment-732220424)

Yes, that's the one.



> I ask because the current API seems a bit awkward where certain options (i.e. `patterns`) only apply to certain filters (i.e.`pattern`). Rather than having these ignored parameters, I might have separate functions (like `redact_patterns`).

I considered this, but the current thinking is that the `pattern` filter will not be part of the filters that are enabled by default, and should be the _only_ filter that requires an extra argument.

In a sense, it's an "escape hatch" filter that allows you to use custom regex logic while still taking advantage of the other features this function provides (e.g. the custom redacting patterns).

So my assessment would be that _currently_ this looks awkward, but once we have many more filters, none of those require extra configuration, and the `pattern` filter isn't on by default but documented as "in case you need it", I think the current interface makes more sense.

I would advocate against introducing a separate function specifically for this one filter.

jszwedko:
:+1: Gotcha, thanks for the explanation.

What do you think of my suggestion to allow either an enum value (like `credit_card`) or a regex in `filters`? This would alleviate the need for the extra `patterns` parameter. That is: allowing ` redact(.input, filters = [/hello/])`.

FungusHumungus:
I'm not sure I fully understand what is going on.  What other filters to you anticipate adding?

jszwedko:
@FungusHumungus There are other filters mentioned in https://github.com/timberio/vector/issues/4519#issuecomment-732220424 like `credit_card`, `phone_number`, `home_address`.

FungusHumungus:
>  There are other filters mentioned in [#4519 (comment)](https://github.com/timberio/vector/issues/4519#issuecomment-732220424) like `credit_card`, `phone_number`, `home_address`.

How will these work? Are they just set regexes, or will there potentially be more complicated parsing involved?

lukesteensen:
> How will these work? Are they just set regexes, or will there potentially be more complicated parsing involved?

Regex is likely to be the vast majority, but I think it'd be a good idea to be a little bit more abstract so that we can do more intelligent things in the future. For example, [presidio](https://github.com/microsoft/presidio/blob/7935adf03c80a0ba5d53c6affb7d7c78198affaa/presidio-analyzer/presidio_analyzer/predefined_recognizers/credit_card_recognizer.py) will use the combination of a regex and Luhn checksum to more accurately identify credit card numbers.

JeanMertz:
> Regex is likely to be the vast majority, but I think it'd be a good idea to be a little bit more abstract so that we can do more intelligent things in the future.

Correct. The approach I'm taking is that each filter is an enum variant, which can have a pre-compiled regex alongside a Luhn checksum or any other input that helps identify patterns.

FungusHumungus:
So does that mean the `patterns` parameter is only relevant if the `filter` is set to `pattern`? If you have say two filters specified, the `patterns` would only be relevant if `pattern` was one of the filters.

If so, it feels a little bit wrong. I think we would somehow need to specify the patterns with the `pattern` parameter..

```coffeescript
redact(.input, filters = [ { type = "pattern", patterns = [ "\b{3}", "\b{6}" ] } , "credit_card" ])
```

I'm not totally keen on that either, but it feels more correct.

JeanMertz:
> So does that mean the `patterns` parameter is only relevant if the `filter` is set to `pattern`? If you have say two filters specified, the `patterns` would only be relevant if `pattern` was one of the filters.

That is correct.

However, I honestly think this discussion will be moot in a few iterations, as there will be less and less reason to use this filter. It is there as an escape hatch, will not be enabled by default, and won't be prominently promoted since we should strive to have non-configurable filters that "just work" (and most are enabled by default).

So, my preference would be to bench the function API discussion for now, until we're a few iterations further. We can keep tracking this in the relevant issue that we created, as we'll definitely have a final "clean up and correct API warts" task before we release 0.12 (and even in 0.12 we'll release this as a beta feature, so we could still make changes if we still think this interface is a mistake).

---

In the interest of moving forward, I'm going to merge this PR, but don't mistake that for me trying to cut off any discussion, I definitely think there's merit to the thoughts put in this PR, I just have a feeling some of those thoughts might change as we iterate a bit more on this function.

JeanMertz:
> What do you think of my suggestion to allow either an enum value (like `credit_card`) or a regex in `filters`? This would alleviate the need for the extra `patterns` parameter. That is: allowing ` redact(.input, filters = [/hello/])`.

Sorry, @jszwedko I forgot to reply to this proposal. I definitely think there's merit to trying this out 👍

jszwedko:
:+1: sounds good. Thanks @JeanMertz . I mentioned this discussion over on https://github.com/timberio/vector/issues/4519

I kinda like @FungusHumungus's suggestion of:

```
redact(.input, filters = [ { type = "pattern", patterns = [ "\b{3}", "\b{6}" ] } , "credit_card" ])
```

as it would set a precedent in case we have other filters that have configuration (it reminds me a little of `Cargo.toml` allowing for both "foo = 1" and `foo = { version = 1, ... }`).

PR #5308, opened 2020-12-01
chore(remap): improve regex handling

JeanMertz:
_PR is best reviewed per commit_

Regex support has some history in Remap (even in its short lifetime).

We discussed the implementation in #4091, then had an initial implementation in #4290, a refactor of that implementation in #4695, and added support for arrays with regexps in #5283.

Since the start, regexps have been treated in a special way, compared to other expressions. That is to say, there is no easy/correct way to represent a regexp as an `event::Value`, and so we didn't model regexps as a `remap::Value` (and thus we couldn't treat them as expressions), instead opting to add special parsing rules that allow regexps in argument positions, but nowhere else.

This "works", but it came with a bag full of special/edge-cases that need to be handled, and an entire set of infrastructure to handle function arguments differently from other expressions in Remap.

After adding arrays, it became clear to me this was 1) becoming more difficult to work with, and 2) didn't scale well to container types (such as arrays, but also maps, eventually).

This PR proposes a way to address those issues.

After this PR:

* A new `Value::Regex` variant is added to Remap (not Vector!).
* Remap now _symantically_ supports regexp expressions, but _syntactically_ rejects assigning regexp values to paths, or to return regexp values as a root expression (i.e. the `Value::Regex` type is self-contained within Remap and cannot leak out to Vector or other users of Remap).
* There is **no† difference in behaviour to the end-user**.
* Functions no longer take `Argument`s, but `Expression` types, removing a host of special-casing logic in the process.
* An `expression::Literal` no longer stores `Value::Array`, instead this is handled by `expression::Array` (there is a debug assert to make sure this invariant is upheld).
* It is now **allowed** to assign a regexp (or an array containing regexps) to a variable.

†: the only noticeable difference is a slightly updated compiler error when using regexps inappropriately

---

Here's an example of what now works and doesn't work:

```rust
$foo = /foo/ 	// ok
.foo = /foo/ 	// remap error: parser error: cannot assign regex to object
.foo = [/foo/] 	// remap error: parser error: cannot assign regex to object
my_func(/foo/) 	// ok
my_func($foo) 	// ok
// etc...
```

---

There is still some work to do in a follow-up PR. Namely, we should extend our `TypeDef` infrastructure to also encode which types are embedded in a container type.

If we did this, we can always statically know if there is a chance of a regexp being returned in an expression, which would make our checks to disallow regexps in path assignments or as root expressions more robust.

I left this out of this PR, but will follow up on this in the next iteration.

### DIFF
--- a/lib/remap-lang/src/parser.rs
+++ b/lib/remap-lang/src/parser.rs
@@ -321,8 +321,6 @@ impl Parser<'_> {
             match pair.as_rule() {
                 // This matches first, if a keyword is provided.
                 R::ident => ident = Some(pair.as_str().to_owned()),
-                R::regex => return Ok((ident, Argument::Regex(self.regex_from_pair(pair)?))),
-                R::argument_array => return Ok((ident, self.argument_array_from_pair(pair)?)),
                 _ => {
                     let expr = self.expression_from_pair(pair)?;
                     return Ok((ident, Argument::Expression(expr)));
@@ -331,16 +329,6 @@ impl Parser<'_> {
         }
 
         Err(e(R::argument))
-    }
-
-    fn argument_array_from_pair(&mut self, pair: Pair<R>) -> Result<Argument> {
-        pair.into_inner()
-            .map(|pair| match pair.as_rule() {
-                R::regex => self.regex_from_pair(pair).map(Argument::Regex),
-                _ => self.expression_from_pair(pair).map(Argument::Expression),
-            })
-            .collect::<Result<Vec<_>>>()
-            .map(Argument::Array)
     }
 
     /// Parse a [`Regex`] value
--- a/lib/remap-lang/src/parser.rs
+++ b/lib/remap-lang/src/parser.rs
@@ -278,6 +278,7 @@ impl Parser<'_> {
                 Literal::from(pair.as_str().parse::<f64>().map_err(|_| e(R::float))?).into()
             }
             R::array => self.array_from_pair(pair)?.into(),
+            R::regex => Literal::from(self.regex_from_pair(pair)?).into(),
             _ => return Err(e(R::value)),
         })
     }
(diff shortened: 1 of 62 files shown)


jszwedko (approved):
Nice! This looks much cleaner.

Thanks, as usual, for the thorough PR description.

lib/remap-lang/src/parser.rs
```diff
@@ -602,14 +654,15 @@ mod tests {
...
             i += 1;
 
-            let pairs = Parser::parse(Rule::program, source);
+            let mut parser = Parser::new(&[]);
+            let pairs = parser.program_from_str(source).map_err(RemapError::from);
 
-            match pairs {
-                Ok(pairs) => {
-                    let got = Parser::default().pairs_to_expressions(pairs);
+            dbg!(&pairs);
```
FungusHumungus, line 660: This is probably stray..

FungusHumungus (approved):
One possible issue - it is possible for regexes to leak into the object a number of ways by going eg.

```coffeescript
    .erk = flatten([1, [/[a-f]{3}/, 4]])

    $$regex = /[a-f]/
    .ook = $$regex
```

The regex is just converted to bytes in the final output. It is not possible to use this regex, or indeed use any field from the object as a dynamic regex - which is the critical issue we need to avoid, so I don't think it is a critical issue

```coffeescript
    .erk = flatten([1,/[a-f]/, 4])
    .regex = .erk[1]
    .oog = replace(.message, .regex, "hi")
```

I do really like that we can assign them to variables though!

JeanMertz:
> One possible issue - it is possible for regexes to leak into the object a number of ways by going eg.

Correct, I should have linked to another PR I made to solve this problem:

https://github.com/timberio/vector/pull/5375

FungusHumungus:
> Correct, I should have linked to another PR I made to solve this problem:
> 
> #5375

You are always one step ahead! :-)

PR #5309, opened 2020-12-01
feat(remap): support multiline expressions

### LINKED ISSUE #4946: Support newlines before right-hand operator side in Remap
This works:

```rust
.very_long_expression_here || .another_long_expression
```

This works as well:

```rust
$expr1 = .very_long_expression_here
$expr2 = .another_long_expression

$expr1 || $expr2
```

But this doesn't:

```rust
.very_long_expression_here ||
	.another_long_expression
```

We should support adding one or more newlines after an operator in Remap, to make scripts with long expressions (for example, complex regular expressions) more readable.

---

The one reason we might not want to support this, is to avoid unexpected results if someone forgets to add a right-hand-side expression, e.g.:

```rust
true || /* forgot to add rhs expression here */
$foo = "bar"
```

The above is equal to:

```rust
true || $foo = "bar"
```

Which resolves to `true`, since `$foo = "bar"` is a valid rhs expression.

But I believe this to be an unlikely scenario for someone to run into.

---

Alternatively, we could add support to break up expressions over multiple lines using `\`, which would make it more clear:

```rust
.very_long_expression_here || \
	.another_long_expression
```

---

Or we could do nothing here, avoiding adding more complexity to the language, but also making some remap sources harder to read for users (example 1, single line), or more complex to write (example 2, variables).

JeanMertz:
While looking at #4946, I wanted to see how easy it was to add support for breaking up expressions over multiple lines.

Turns out, quite easy.

As described in the linked issue, we've had people write long checks, for example:

```rust
match(.log, /^\\d{4}-\\d{2}-\\d{2}(T|\\s)?\\d{2}:\\d{2}:\\d{2}/) || match(.log,/^\\[?\\d{2}\\S\\w{3}\\S\\d{4}:\\d{2}:\\d{2}:\\d{2}/) || match(.log,/^\\d{4}\\S\\d{2}\\S\\d{2}\\s\\d{2}:\\d{2}:\\d{2}/)
```

After this PR, it can be written as:

```rust
match(.log, /^\\d{4}-\\d{2}-\\d{2}(T|\\s)?\\d{2}:\\d{2}:\\d{2}/) 			\
	|| match(.log,/^\\[?\\d{2}\\S\\w{3}\\S\\d{4}:\\d{2}:\\d{2}:\\d{2}/) 	\
	|| match(.log,/^\\d{4}\\S\\d{2}\\S\\d{2}\\s\\d{2}:\\d{2}:\\d{2}/)
```

That is to say, anywhere in an expression where _whitespace_ is supported, you can now add `\<newline>` in addition to the already supported space or tab characters.

This _seems_ useful to me, and I can't think of any major downsides (the only one being "we need to document this"). It also matches closely with other languages that support similar syntax.

Happy to _not_ do this, but curious what others think.

closes #4946

### DIFF
--- a/lib/remap-lang/grammar.pest
+++ b/lib/remap-lang/grammar.pest
@@ -90,4 +90,4 @@ char         =  { !("\"" | "\\") ~ ANY | "\\" ~ ("\"" | "\\" | "n" | "t") }
 regex_inner = @{ regex_char* }
 regex_char  =  { !("/" | "\\") ~ ANY | "\\" ~ ANY }
 
-WHITESPACE  = _{ " " | "\t" }
+WHITESPACE  = _{ " " | "\t" | ("\\" ~ NEWLINE) }
--- a/lib/remap-lang/src/lib.rs
+++ b/lib/remap-lang/src/lib.rs
@@ -233,6 +233,16 @@ mod tests {
                 Err("remap error: unexpected expression: expected Array, got Literal"),
                 Ok(().into()),
             ),
+            (
+                r#"
+                    .foo        \
+                        =       \
+                        null || \
+                        "bar"
+                "#,
+                Ok(()),
+                Ok("bar".into()),
+            ),
         ];
 
         for (script, compile_expected, runtime_expected) in cases {


JeanMertz:
I should note that some expressions have explicit support for newlines within them, for example:

```rust
if true {
	"bar"
} else {
	"baz"
}

.foo = [
	"foo",
	"bar",
]
```

I would vote to leave those as they are, as it makes little sense to require the special `\` character in those cases. We can continue to special case expressions that feel more _natural_ with newlines inside them.

JeanMertz:
Of course, another alternative is to no longer use newlines as the expression termination character, instead using a semicolon:

```rust
 match(.log, /^\\d{4}-\\d{2}-\\d{2}(T|\\s)?\\d{2}:\\d{2}:\\d{2}/)
	|| match(.log,/^\\[?\\d{2}\\S\\w{3}\\S\\d{4}:\\d{2}:\\d{2}:\\d{2}/)
	|| match(.log,/^\\d{4}\\S\\d{2}\\S\\d{2}\\s\\d{2}:\\d{2}:\\d{2}/);
```

I would definitely **not** be favouring this, as it clutters an otherwise clean language syntax.

_We do intend to optionally allow semicolons to terminate an expression, but that's because sometimes you just want to write a quick single line program such as `$foo = "bar"; $foo`, and we want to allow multiple expressions in an if-statement such as `if ($foo = my_func(); $foo.bar)`_, we don't indent to make it mandatory.

binarylogic (approved):

jszwedko (commented):
Does this also allow multi-line string literals?

Like:

```
.foo = "hello \
 bar"
```
?

If so, what is the behavior there? To join the lines?

As an alternative to this, I think we could also expand the places where we support multiline strings without requiring `\`s. For example, ending a line with a `||`, a `+`, or any other binary operator. Also in the middle of argument lists. I think the only time a \ would be needed is if it is ambiguous -- I'm actually having a hard time imagining when it'd be needed outside of multi-line strings and regexes. I'm curious what you think.

bruceg (approved):
LGTM

bruceg:
> Does this also allow multi-line string literals?

AFAICT from the grammar, multi-line string literals are already allowed *without* the back-slash escape. Any character is allowed within the quotes.

StephenWakely (approved):

JeanMertz:
> I'm actually having a hard time imagining when it'd be needed outside of multi-line strings and regexes. I'm curious what you think.

Currently, a newline indicates the end of an expression. If we were to go this route, we'd have to manually update each expression to allow newlines in specific places.

I'm not against this, but it does significantly bloat our syntax definition, and also means we have to remember to add newline capabilities to every new expression we add.

I'll admit that I'm not the biggest fan of forcing `\` to split up expressions, but I'm also not sure it's worth allowing newlines without `\` anywhere we can, as it'll clutter the syntax for readers of TRL scripts. Another alternative would be to _only_ allow it in a few places (such as after operators), but then people might expect it to work in one place, but find out it doesn't.

The `\` does make it more explicit that the intent of the writer of the script was to split up a long expression over multiple lines, which is a property I do like.

---

In the end, there's up and downsides to all of these proposals. I think from a teachability perspective, mentioning "newlines terminate an expression" is the easiest to teach, and we can have a less prominent "if you really need to split an expression over multiple lines, use `\`" message for those that need it.

Writing the above, I do realize that — as I mentioned before — we already have exceptions to this, but those cases contain either blocks, or are collection literals such as arrays or maps, so I feel those decisions are somewhat justified.

jszwedko:
:+1: 

> Writing the above, I do realize that — as I mentioned before — we already have exceptions to this, but those cases contain either blocks, or are collection literals such as arrays or maps, so I feel those decisions are somewhat justified.

I think having any exceptions is a bit confusing, but I've come around to the idea of allowing `\`s to allow breaking up expressions even if we, in the future, expand the places where expressions can flow over multiple lines unambiguously due to ending in an operator or other symbol that requires more input. This would let users break up the line how they like.

PR #5311, opened 2020-12-01
chore(external docs): Remap documentation fixes

FungusHumungus:
This fixes the documentation errors noticed by [Steve](https://discordapp.com/channels/742820443487993987/751109034827972709/782782924541394965).

In the process I noticed that documentation was missing for a number of functions. So this also adds docs for:

- `ceil`
- `floor`
- `round`
- `parse_syslog`


JeanMertz (approved):
Looks good to me 👍 

It would be nice if we could parse these cue files and actually run their examples as tests on our CI.

binarylogic (approved):

PR #5315, opened 2020-12-01
chore(ci): Allow benchmarks to run without baseline

jszwedko:
I neglected that this would always be the case on the first run. I'm
coming around to allowing them to run without anyway.


jszwedko:
CI failure (500 on upload) appears to be due to service issue with Github Actions.

Issue #579, opened 2019-07-03
Use `tokio_sink::mpsc` over `futures::sync::mpsc`


LucioFranco:
`tokio-sync` has a more performant channel that we may want to consider moving too. This change should be benchmarked and proven that it does indeed give us the perf increase.

cc @lukesteensen

LucioFranco:
Related #2184

binarylogic:
Noting, this will be covered in https://github.com/timberio/vector/issues/2625

lukesteensen:
This is blocked on #2935

kit70669:
Can someone give me info on all this I can donate

lukesteensen:
@kit70669 Thank you for the offer! I actually believe this will tie into some very core systems of Vector, so it is probably best for a core team member to tackle this particular issue. I am actually working towards solving it right now.

lukesteensen:
Closing in favor of the four more specific issues referenced above.

PR #5321, opened 2020-12-02
chore(sources): upgrade Pipeline to futures 0.3 and tokio channels

lukesteensen:
Ref #579 

This is the front end of the topology channel upgrade, covering just `Pipeline` and sources. Next up will be `Fanout`.

Most of this diff is pretty mechanical, but I'd appreciate some extra eyes on the modified `Pipeline` internals. It is mostly the same implementation as before but rearranged a bit to fit the 0.3 `Sink` interface.


src/pipeline.rs
```diff
@@ -127,15 +151,16 @@ mod test {
...
             Pipeline::new_with_buffer(100, vec![Box::new(transform_1), Box::new(transform_2)]);
 
         let event = Event::try_from(json!({
             "message": "MESSAGE_MARKER",
         }))?;
 
-        pipeline.send(event).compat().await?;
-        let out = reciever.wait().collect::<Result<Vec<_>, ()>>().unwrap();
+        pipeline.send(event).await?;
+        pipeline.flush().await?;
```
fanatid, line 162: I do not think that we really need `flush` because [send](https://docs.rs/futures/0.3.8/futures/sink/trait.SinkExt.html#method.send) already include flushing:
> A future that completes after the given item has been fully processed into the sink, **including flushing**.

src/sinks/statsd.rs
```diff
@@ -448,21 +448,17 @@ mod test {
...
-            let mut tx = tx
-                .sink_compat()
-                .sink_map_err(|error| error!(message = "Error sending event.", %error));
-
-            stream.forward(&mut tx).await.unwrap();
-
-            tx.flush().await.unwrap();
+            while let Some(Ok(item)) = stream.next().await {
+                tx.send(item).await.unwrap();
+            }
```
fanatid, line 461: This looks like `forward`/`send_all` :thinking:

src/sources/splunk_hec.rs
```diff
@@ -200,10 +200,12 @@ impl SplunkSource {
                     let out = out.clone();
                     async move {
                         // Construct event parser
-                        futures01::stream::once(raw_event(body, gzip, channel, host))
-                            .forward(out.clone().sink_map_err(|_| ApiError::ServerShutdown))
-                            .map(|_| ())
-                            .compat()
+                        let event = futures::future::ready(raw_event(body, gzip, channel, host));
```
fanatid, line 203: `std::future::ready`?

src/sinks/statsd.rs
lukesteensen: Yes, this is because `tokio::sync::mpsc::Sender` does not implement `Sink` itself. In most places like this, we use `Pipeline`, which does, but in this case, we are sending `Bytes`, not events. It's a little bit weird but seemed like the simplest thing to do.

fanatid (approved):
:tada:

Hoverbear (approved):

Hoverbear:
Got curious, did benches.

Here's the `transforms/transforms` bench:

image

Here's `interconnected/interconnected`: 

image

Here's `pipe/pipe_big_lines`:

image

Here's `pipe/pipe_multiple_writers`:

image

src/sources/socket/udp.rs
```diff
@@ -84,9 +83,7 @@ pub fn udp(
                         emit!(SocketEventReceived { byte_size,mode:SocketMode::Udp });
 
                         tokio::select!{
-                            result = out.send(event).compat() => {
-                                out = result?;
-                            }
+                            _ = out.send(event) => { }
```
ktff, line 86: Can this return `ClosedError`?

src/pipeline.rs
```diff
@@ -1,87 +1,115 @@
...
+                .inner
+                .poll_ready(cx)
+                .map_err(|_: mpsc::error::ClosedError| ClosedError))?;
+
+            match self.inner.try_send(event) {
+                Ok(()) => {
+                    // we good, keep looping
+                }
+                Err(Full(item)) => {
+                    // TODO: can we get here after the poll_ready?
```
ktff, line 47: Nope, `poll_ready` [guarantees](https://docs.rs/futures/0.3.8/futures/channel/mpsc/struct.Sender.html#method.poll_ready) that. So instead of `try_send` we can use `start_send`.

src/pipeline.rs
```diff
@@ -1,87 +1,115 @@
...
+        use mpsc::error::TrySendError::*;
+
         while let Some(event) = self.enqueued.pop_front() {
-            if let AsyncSink::NotReady(item) = self.inner.start_send(event)? {
-                self.enqueued.push_front(item);
-                return Ok(Async::NotReady);
+            futures::ready!(self
+                .inner
+                .poll_ready(cx)
+                .map_err(|_: mpsc::error::ClosedError| ClosedError))?;
```
ktff, line 40: If this returns `Pending`, the event will not be pushed back to queue and will be lost.

ktff (changes requested):
I think we are loosing events in one scenario, other than that this is quite nice.

src/pipeline.rs
lukesteensen: Great catch! You are totally right.

src/pipeline.rs
lukesteensen: Unfortunately, there is no `start_send` on `Sender`, so I think our best option is to keep using `try_send` and consider this branch unreachable. I'll make that change.

src/sources/socket/udp.rs
lukesteensen: Yes it can! I will change it to handle that.

ktff (approved):

ktff:
It seams a lot of topology tests are failing, most of them if not all with error
> mpsc bounded channel requires buffer > 0

lukesteensen:
**Status update:** Our test suite uncovered a small bug in the 0.2 tokio channel implementation (not present in 0.3, for the record), so I am waiting to merge this until https://github.com/tokio-rs/tokio/pull/3227 lands and we confirm that CI is green.

PR #5328, opened 2020-12-02
chore: unify metric names

fanatid:
@jszwedko should we also rename `request_error_total` => `requests_error_total` and `request_read_errors_total` => `requests_read_errors_total` (because we have `requests_completed_total` / `requests_received_total`)?

```
$ grep 'counter!\("request.*?"' -rhoP src/internal_events/ | sort -u
counter!("request_error_total"
counter!("request_read_errors_total"
counter!("requests_completed_total"
counter!("requests_received_total"
```

edit:
**Changes**:
- `events_processed_total` => `processed_events_total` (prometheus)
- `request_received_total` => `requests_received_total` (splunk_hec)
- `request_error_total` => `request_errors_total` (mongo, nginx)

### DIFF
--- a/src/sinks/elasticsearch.rs
+++ b/src/sinks/elasticsearch.rs
@@ -3,7 +3,7 @@ use crate::{
     emit,
     event::Event,
     http::{Auth, HttpClient},
-    internal_events::{ElasticSearchEventReceived, ElasticSearchMissingKeys},
+    internal_events::{ElasticSearchEventEncoded, ElasticSearchMissingKeys},
     rusoto::{self, region_from_endpoint, RegionOrEndpoint},
     sinks::util::{
         encoding::{EncodingConfigWithDefault, EncodingConfiguration},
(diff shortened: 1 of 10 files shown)


src/sinks/elasticsearch.rs
```diff
@@ -196,7 +196,7 @@ impl HttpSink for ElasticSearchCommon {
         serde_json::to_writer(&mut body, &event.into_log()).unwrap();
         body.push(b'\n');
 
-        emit!(ElasticSearchEventReceived {
+        emit!(ElasticSearchEventEncoded {
```
jszwedko, line 199: I think you need to update the imports too.

jszwedko (commented):
Thanks for cleaning these up!

I would suggest `request_error_total` -> `request_errors_total` . I think the noun in that case is "request error". Alternatively, it could be `requests_errored_total`. Similarly I think `request_read_errors_total` is fine as it is, the noun being "request read error".

I do think the API has a dependency on error counters ending in `errors_total` right now so would lean towards keeping them consistent with that. cc/ @leebenson

leebenson:
> I do think the API has a dependency on error counters ending in errors_total right now so would lean towards keeping them consistent with that. cc/ @leebenson

This, although it looks like it might not be reporting as expected currently (#5329). Relying on a suffix is brittle. Probably makes sense to switch that out for a tag.

In any case, +1 for consistency.

bruceg (approved):

PR #5344, opened 2020-12-02
enhancement(topology): Reuse buffers

### LINKED ISSUE #5147: Check for conflicting buffer usage when reloading config
### Expected Behavior
Should be able to edit config of a sink that uses the disk buffer.

### Actual Behavior
Vector crashes out - can't build new config, and can't rebuild old config.

Ran into an issue related to #4031. Basically if you are using disk buffer on any sink, and then change the sink, the reload will fail. What happens is the new (changed) sink tries to open the data_dir again, but the old sink hasn't ended yet, so still has data_dir open. So the reload fails. Then Vector tries to rebuild the old configuration, but again cannot open the data_dir, so it fails to recover and stops the program.

This happens on any change to the sink -- e.g. in the below config, if I change the encoding from text to json, the above will happen and it will crash.

### Vector Configuration File

```
data_dir = "tmp_data/"

[sources.in]
  type = "stdin"

# Output data
[sinks.out]
  inputs   = ["in"]
  type     = "console"
  encoding = "text"
  buffer.type = "disk"
  buffer.max_size = 1400000
  buffer.when_full = "drop_newest"
```

### Output
```
Nov 20 09:12:10.975  INFO vector::app: Log level is enabled. level="info"
Nov 20 09:12:10.976  INFO vector::app: Loading configs. path=["config/vector.toml"]
Nov 20 09:12:10.991  INFO vector::sources::stdin: Capturing STDIN.
Nov 20 09:12:16.041  INFO vector::topology: Running healthchecks.
Nov 20 09:12:16.042  INFO vector::topology: Starting source. name="in"
Nov 20 09:12:16.042  INFO vector::topology::builder: Healthcheck: Passed.
Nov 20 09:12:16.042  INFO vector::topology: Starting sink. name="out"
Nov 20 09:12:16.043  INFO vector: Vector has started. version="0.11.0" git_version="v0.9.0-1391-ge769bcc" released="Fri, 20 Nov 2020 16:45:10 +0000" arch="x86_64"
Nov 20 09:12:31.642 ERROR vector::topology: Configuration error. error=Sink "out": Unable to open data_dir "tmp_data"
Nov 20 09:12:31.642  INFO vector::topology: Rebuilding old configuration.
Nov 20 09:12:31.642 ERROR vector::topology: Configuration error. error=Sink "out": Unable to open data_dir "tmp_data"
Nov 20 09:12:31.643 ERROR vector::topology: Failed in rebuilding the old configuration.
Nov 20 09:12:31.643 ERROR vector: Reload was not successful.
Nov 20 09:12:31.643 ERROR vector: Vector has failed to recover from a failed reload.
Nov 20 09:12:31.643  INFO vector: Vector has stopped.
Nov 20 09:12:31.644  INFO source{component_kind="source" component_name=in component_type=stdin}: vector::sources::stdin: Finished sending.
Nov 20 09:12:31.644  INFO vector::topology: Shutting down... Waiting on running components. remaining_components="in, out" time_remaining="59 seconds left"
```

### Related To
#4031, #4735 - basically the same issue, just with the data_dir being the resource in conflict instead of the port.
 
Fix could be to add data_dir as a `Resource` somehow. If I force the reload to wait for the sink, then the reload works correctly. But the buffer is in the `SinkOuter`, and resources are currently only in the inner `SinkConfig`. So not sure how that should look.

ktff:
Closes #5147 

With this PR, the behavior for a changed sink is:
- If buffer configuration wasn't changed, buffer will be reused in the new sink.
- Else if both new and old buffer configurations use disk, we will wait for the old sink to finish. That includes draining the buffer. 
- Else sink is detached with its buffer and left to run in the background.

### Todo

- [x] Open issue for: instead of waiting for old sink to drain disk buffer, we can adjust old disk buffer to new disk buffer configuration.

### DIFF
--- a/src/topology/builder.rs
+++ b/src/topology/builder.rs
@@ -1,7 +1,7 @@
 use super::{
     fanout::{self, Fanout},
     task::{Task, TaskOutput},
-    BuildedBuffer, ConfigDiff,
+    BuiltBuffer, ConfigDiff,
 };
 use crate::{
     buffers,
@@ -38,7 +38,7 @@ pub struct Pieces {
 pub async fn build_pieces(
     config: &super::Config,
     diff: &ConfigDiff,
-    mut buffers: HashMap<String, BuildedBuffer>,
+    mut buffers: HashMap<String, BuiltBuffer>,
 ) -> Result<Pieces, Vec<String>> {
     let mut inputs = HashMap::new();
     let mut outputs = HashMap::new();
@@ -188,7 +188,7 @@ pub async fn build_pieces(
                 errors.push(format!("Sink \"{}\": {}", name, error));
                 continue;
             }
-            Ok(builded) => builded,
+            Ok(built) => built,
         };
 
         let (trigger, tripwire) = Tripwire::new();
--- a/src/topology/builder.rs
+++ b/src/topology/builder.rs
@@ -1,6 +1,6 @@
 use super::{
     fanout::{self, Fanout},
-    task::{Task, TaskBuffer},
+    task::{Task, TaskOutput},
     BuildedBuffer, ConfigDiff,
 };
 use crate::{
@@ -75,7 +75,7 @@ pub async fn build_pieces(
         };
 
         let (output, control) = Fanout::new();
-        let pump = rx.forward(output).map(|_| TaskBuffer::Other).compat();
+        let pump = rx.forward(output).map(|_| TaskOutput::Source).compat();
         let pump = Task::new(name, typetag, pump);
 
         // The force_shutdown_tripwire is a Future that when it resolves means that this source
@@ -86,7 +86,7 @@ pub async fn build_pieces(
         let server = future::try_select(server, force_shutdown_tripwire.unit_error().boxed())
             .map_ok(|_| {
                 debug!("Finished.");
-                TaskBuffer::Other
+                TaskOutput::Source
             })
             .map_err(|_| ());
         let server = Task::new(name, typetag, server);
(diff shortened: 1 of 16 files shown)


src/topology/builder.rs
```diff
@@ -152,33 +166,64 @@ pub async fn build_pieces(
...
-                filter_event_type(rx, input_type)
+        let (trigger, tripwire) = Tripwire::new();
+
+        let sink = async move {
+            // Why is this Arc<Mutex<Option<_>>> needed you ask.
+            // In case when this function build_pieces errors
+            // this future won't be run so this rx won't be taken
+            // which will enable us to reuse rx to rebuild
+            // old configuration by passing this Arc<Mutex<Option<_>>>
+            // yet again.
```
lukesteensen, line 202: How will we pass it again? I see that it won't be taken unless the sink is actually run, but it does seem that every time we call `build_pieces` it is either `remove`'d from `buffers` of newly built. Either way, it doesn't seem like we would still have it if we fail and call the function again.

lukesteensen (commented):
Overall this looks good! Just one question about the failure and rebuild case.

src/topology/builder.rs
ktff: That's all true, and the missing piece is that we are cloning `buffers` when calling `build_pieces`. 
https://github.com/timberio/vector/blob/437cb648fe45adb5cc9ab20739b5c8b766b0d0e0/src/topology/mod.rs#L248

lukesteensen (approved):

PR #5352, opened 2020-12-03
fix(es sink): enable content-encoding header on AWS ES sink

smlgbl:
Following error is returned when using compression on AWS ES:

`400 Bad Request: {\"error\":{\"root_cause\":[{\"type\":\"json_parse_exception\",\"reason\":\"Illegal character ((CTRL-CHAR, code 31)): only regular white space (\\\\r, \\\\n, \\\\t) is allowed between tokens\\n at [Source: (org.elasticsearch.common.bytes.AbstractBytesReference$MarkSupportingStreamInputWrapper); line: 1, column: 2]\"}]
`
I suspect the content-encoding header is missing, so the payload is gzip compressed, but ES doesn't expect it to be.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7118,7 +7118,7 @@ checksum = "f1bddf1187be692e79c5ffeab891132dfb0f236ed36a43c7ed39f1165ee20191"
 
 [[package]]
 name = "vector"
-version = "0.11.0"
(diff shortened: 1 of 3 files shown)


smlgbl:
Messed up commits. Will retry in new PR. Sorry.

PR #5355, opened 2020-12-03
fix(elasticsearch sink): set content encoding header when compression is on

smlgbl:
Following error is returned when using compression on AWS ES:

400 Bad Request: {\"error\":{\"root_cause\":[{\"type\":\"json_parse_exception\",\"reason\":\"Illegal character ((CTRL-CHAR, code 31)): only regular white space (\\\\r, \\\\n, \\\\t) is allowed between tokens\\n at [Source: (org.elasticsearch.common.bytes.AbstractBytesReference$MarkSupportingStreamInputWrapper); line: 1, column: 2]\"}]

I suspect the content-encoding header is missing, so the payload is gzip compressed, but ES doesn't expect it to be.

### DIFF
--- a/src/sinks/elasticsearch.rs
+++ b/src/sinks/elasticsearch.rs
@@ -713,7 +713,7 @@ mod integration_tests {
             },
             false,
         )
-            .await;
+        .await;
     }
 
     #[tokio::test]
@@ -729,7 +729,7 @@ mod integration_tests {
             },
             false,
         )
-            .await;
+        .await;
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


fanatid:
Please run `cargo-fmt` locally and push changes. Thanks!

fanatid (approved):

smlgbl:
Thank you very much for the quick merge, it is working in our production environment.
Nevertheless, I think the "Accept-Encoding" header is still missing.
In the logstash-code I found that there's an underlying lib that automatcally adds that, but I guess this is not happening here.
I'll open a new PR for that, unless you have a different idea, @fanatid?

fanatid:
We currently do not support HTTP requests where the response compressed. It's related to our [HTTP client code](https://github.com/timberio/vector/blob/f25d0420afb7e6c5321d62d75bc6c3ca28930145/src/http.rs), not with elasticsearch sink, so #5439 is not correct. Vector will not able decode response.
We do not have an issue exactly with this feature, I'll create a new one and will close #5439.

PR #5357, opened 2020-12-03
chore: Add `namespace` alias for `default_namespace`

ktff:
Adds `namespace` alias to `default_namespace` options that replaced `namespace` options in #4806.


bruceg (commented):
This looks fine, but it would probably be worthwhile to add a config parsing test to each of these to ensure this does what it's supposed to.

PR #5374, opened 2020-12-03
enhancement: Use new Lookup accessors on LogEvents and entire internal API.

Hoverbear:
Closes https://github.com/timberio/vector/issues/2843
Closes https://github.com/timberio/vector/issues/2845
Closes https://github.com/timberio/vector/issues/4005

This is a ... chonky PR.

Most of it is trivial (and largely mechanized) changes to some basic interfaces around Event types, as well as a bunch of new code. Much of the old code remains in deprecated state for `remap-lang` to fully adopt.

# Notable developer facing changes

* All `Event`, `Value`, and the new `Lookup`/`LookupBuf` types are now fully documented and tested, including several corner cases found during development.
* `From<String>` and `From<&str>` are no longer implemented on `Event`.
  + The `Event` type was moved into `shared` and can no longer access the global `log_schema`. *Yes, it was a lot of work.*
* `LookupBuf`, a new `String` analog which pre-parses lookup paths and allows rich remap-syntax for APIs, has been added.
* `Lookup`, a new `&str` analog which pre-parses lookup paths and allows rich remap-syntax for APIs, has been added. It can be trivially used from `LookupBuf`, like a `&str` and a `String`.
* The new `LookupBuf` type is deserializable, as well as serializable (like `Lookup`). This means `Config` structs can freely contain them and use them, without a conversion step.
* The `log_event! {}` macro is much more prominent, use it, please!
* `Event` and `Value` have (nearly) matching APIs around `insert`/`remove` etc. On `Value` these return a result, as the action is not valid on 'primitives'.
  + This was required to implement the recursive API the new `Event` API uses cleanly, and gives us a lot more options around working with things.
* `log_event.insert(String::from("foo.bar"), 1)` will now insert at `"foo.bar"` the value 1, where prior it would insert `"bar"`, set to 1, into `"foo"`.
* `log_event.insert(LookupBuf::from_str("foo.bar")?, 1)` will now insert `"bar"`, set to 1, into `"foo"`, as strings did prior.
* The `Event` and `Value` types have been moved into `shared` to facilitate sharing between all our crates, making the type more globally accessible. *They are re-exported in the old places, you shouldn't notice.*
* A `Look` and `LookSegment` trait are present to enforce parity between types, but it's preferred that the types implement functions directly, to ensure users always have function access (even without the traits imported).

# Reviewer Notes

> @JeanMertz and I agreed to make a conversion layer for this initial PR between Remap Paths and Lookups, this will be removed later.

I suggest treating this PR in three separate chunks:

* **Review the changes to `shared`.** It will introduce to you the new `Event` type and the related API.
* **Help comb over the changes inside `vector`** to ensure we are not reading any raw data as `Lookup` types.
* **Suggest some new test cases for lookups or the API in general**.

# TODOs

* [ ] I want to do some more benching/profiling. I am unsure about the current `Value::deserialize` impl.
* [x] Typify the errors.
* [ ] Resolve last remap<->lookup inconsistencies


Hoverbear:
@JeanMertz I'd love it if you could help me replace the remap language path with Lookup! I think the first step might be putting the Lookup types in the remap lang?

JeanMertz:
> @JeanMertz I'd love it if you could help me understand how best to replace the remap language path with Lookup!

I think there're two approaches we can take.

First, we can update [impl remap::Object for vector::Event](https://github.com/timberio/vector/blob/2082bd917746875fa78036d8214e9937352bf4a5/src/event/mod.rs#L427-L505) to use the new `Lookup` API and convert the result to a [`remap::Path`](https://github.com/timberio/vector/blob/2082bd917746875fa78036d8214e9937352bf4a5/lib/remap-lang/src/path.rs#L11-L15). This will likely be the least efficient but maybe most simple first step.

After that, we could follow up with a separate PR to actually swap out `Path` for `Lookup` entirely, which brings all the performance benefits we're looking for, at the cost of some extra work.

The second approach is basically to do all the above in one step, which might be worth it, if converting from `Lookup` to `Path` in the intermediate step turns out to be too difficult/cumbersome to do.

> I think the first step might be putting the Lookup types in the remap lang?

Perhaps it makes sense to put all lookup related logic into its own `path-lookup` crate or something similar? That way, we can isolate its implementation and the test suite as well. I don't see any downsides in terms of slowing our development cycle down, as it'll still live in this repository, and is linked to by path instead of through crates.io.

But if we don't go that route, then yes, I think we have no choice but to put it in the `remap-lang` crate, as it has no dependency on Vector, and the `remap-functions` crate (might) also need access and also doesn't depend on the Vector crate.]

Hoverbear:
Doing a rebase has lead me to need to move all the event/lookup stuff into shared, so this kind of solves the problem for us?

Hoverbear:
@FungusHumungus I've marked this Ready for Review, but there is a failing test with remap inserts. I'm hoping you could take a look at it and let me know what you think the best way to approach it would be?

image

https://github.com/timberio/vector/blob/7e7b46f8f68a4eb8ebf7e3d75dc9a929262bb686/lib/shared/src/event/log_event/test.rs#L616-L623

You can see it's because we're seeking into a primitive field in order to insert:
image


t I expect it will require at least a bit of feedback reflection before it's good to merge.

FungusHumungus:
> I've marked this Ready for Review, but there is a failing test with remap inserts. I'm hoping you could take a look at it and let me know what you think the best way to approach it would be?

This is a design issue that we need to come to a decision about, and then apply it consistently throughout the app. With the Remap value type if you insert a value into a path that results in a type on the path changing (ie.. it was a map and now needs to become an array) it will change that type.

So, for example in the Repl:

```
$ .foo.bar = 3
3

$ .
{ "foo": { "bar": 3 } }

$ .foo[2] = 4
4

$ .
{ "foo": [null, null, 4] }
```

If we run this script in the Remap transform using the new lookup accessors:

```
	del(.)
	.foo.bar = 3
	.foo[2] = 4
```

The result of `.` is `{"foo":{"bar":3}}`. As in, it hasn't changed the type, which is why this test is failing.

So I think we need to decide:

1. Should it change the type?
2. If not, should it fail silently, or should it error?

I think you were saying there were some problems with raising an error here? How insurmountable are they? We could perhaps defer that to a separate issue.

If we choose not to change the type, we can delete this failing test and should raise an issue to change Remap inline with this (I think this just affects the repl, so is not too high priority).

lib/shared/src/event/log_event/lua.rs
```diff
@@ -31,16 +32,22 @@ impl<'a> FromLua<'a> for LogEvent {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::event::Event;
+    use rlua::Lua;
 
-    #[test]
+    #[test_env_log::test]
```
FungusHumungus, line 37: What does this do?

Hoverbear:
@FungusHumungus I was hoping you could help me with this one?

I see that quotes are used here? https://github.com/timberio/vector/blob/86704643385b1175a71c4192c5921b24597dc55b/lib/shared/src/event/log_event/test.rs#L861 but it seems the `path.to_string()` doesn't print them?

```
ana@autonoma:~/git/timberio/vector$ cargo test --package shared object_paths
    Finished test [unoptimized + debuginfo] target(s) in 0.20s
     Running target/debug/deps/shared-4c914a4dd4cb828c

running 1 test
test event::log_event::test::remap::object_paths::f ... FAILED

failures:

---- event::log_event::test::remap::object_paths::f stdout ----
thread 'event::log_event::test::remap::object_paths::f' panicked at 'assertion failed: `(left == right)`
  left: `Err("Parser(Pest(Error { variant: CustomError { message: \"unknown parsing error\" }, location: Pos(5), line_col: Pos((1, 6)), path: None, line: \".foo bar baz\", continued_line: None }))")`,
 right: `Ok([Path { segments: [Field(Quoted("foo bar baz"))] }])`', lib/shared/src/event/log_event/test.rs:896:13


failures:
    event::log_event::test::remap::object_paths::f

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 125 filtered out

error: test failed, to rerun pass '-p shared --lib'
```

lib/shared/src/event/log_event/lua.rs
Hoverbear: This is https://crates.io/crates/test-env-log, since we don't have `vector::test_util::trace_init()` this accomplishes similar.

Hoverbear:
@FungusHumungus and I connected earlier about the insert issues and surmounted them. :)

Hoverbear:
The other problem I've found:

Looking at this: https://github.com/timberio/vector/blob/6c0286da0dc34d2ebb8086ffa8c6eddc0a6a8cf5/src/mapping/query/path.rs#L157-L166

And this error:

```
ana@autonoma:~/git/timberio/vector$ cargo test --lib check_path_query
    Finished test [unoptimized + debuginfo] target(s) in 0.32s
     Running target/debug/deps/vector-886c6f1e7df5ddd0

running 1 test
test mapping::query::path::tests::check_path_query ... FAILED

failures:

---- mapping::query::path::tests::check_path_query stdout ----
thread 'mapping::query::path::tests::check_path_query' panicked at 'assertion failed: `(left == right)`
  left: `Err("path .foo\\.bar[0] not found in event")`,
 right: `Ok(Value(Integer(20)))`: Query path failed: Path { path: [["foo\\.bar[0]"], ["baz"]] }', src/mapping/query/path.rs:228:13


failures:
    mapping::query::path::tests::check_path_query

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 981 filtered out

```

You can see that the `Path::to_string()` seems to not be escaping the `[` and `]` properly?

Hoverbear:
Some benches:

Pretty good:
image
image

Not so good:
image

Ambivalent:
image
image

lib/shared/src/event/error.rs
```diff
@@ -0,0 +1,48 @@
...
+        primitive_at: LookupBuf,
+        original_target: LookupBuf,
+        original_value: Option<Value>,
+    },
+    LookupError(crate::lookup::LookupError),
+    EmptyCoalesceSubSegment,
+    RemovingSelf,
+}
+
+impl fmt::Display for EventError {
```
FungusHumungus, line 17: Could you use Snafu here to do all this?

lib/shared/src/event/log_event/mod.rs
```diff
@@ -0,0 +1,831 @@
...
+    #[instrument(level = "trace", skip(self))]
+    pub fn inner_mut(&mut self) -> &mut BTreeMap<String, Value> {
+        &mut self.fields
+    }
+}
+
+impl remap_lang::Object for LogEvent {
+    fn get(&self, path: &remap_lang::Path) -> Result<Option<remap_lang::Value>, String> {
+        let path_string = path.to_string();
+        if path_string == "." {
```
FungusHumungus, line 665: We could call `path.is_root` here to see if we are root, which would save the allocation. It gets in the way of the `trace!` below. Could you change the trace to be: `trace!(path = %path.to_string()..` which would then only build in the allocation for debug builds?

lib/shared/src/event/log_event/mod.rs
```diff
@@ -0,0 +1,831 @@
...
+/// assert!(event.contains("foo"));
+/// assert!(event.contains(Lookup::from_str("foo").unwrap()));
+/// ```
+#[derive(PartialEq, Debug, Clone, Default, Serialize, Deserialize)]
+pub struct LogEvent {
+    #[serde(flatten)]
+    fields: BTreeMap<String, Value>,
+}
+
+impl LogEvent {
```
FungusHumungus, line 79: Given that `LogEvent` is essentially a wrapper over `BTreeMap` and `Value::Map` is also a wrapper over `BTreeMap`, is there any reason why we can't share more lookup code between this and `Value`?

Hoverbear: That could absolutely be done! I actually did this at one point along my refactoring. 
Here's why I didn't:

* `LogEvent`, as Vector currently works, **must** always be a map, making `LogEvent` a `Value` would mean we can't guarantee that `LogEvent` is indeed a value. We'd have to use things like `debug_assert!()` and `Value::is_map()` quite often. It felt kind of unsafe.
* At various states the code did in fact differ, so they were differing to allow that. I should audit to see if I can combine them now.

I agree with you that this is a worthwhile exploration to take, and I'll have a try at it.

lib/shared/src/event/log_event/mod.rs
Hoverbear: Heck yes, good idea!

lib/shared/src/event/error.rs
Hoverbear: Indeed! I didn't really want to add in a new dependency to shared just for that though. Snafu's `ErrorCompat` enables these to use it's features anyways: https://docs.rs/snafu/0.3.1/snafu/trait.ErrorCompat.html

Would you prefer we use the lib here? Would be happy to.

FungusHumungus: Yeah. It has been pulled in anyway in `metric/mod.rs`.

lib/shared/src/event/log_event/mod.rs
FungusHumungus: Yeah, I think making `LogEvent` a `Value` would likely be a mistake for the reasons you outlined.  I'm thinking more things like:

```rust
  impl LogEvent {
    pub fn get<'a>(&self, lookup: impl Into<Lookup<'a>> + Debug) -> Option<&Value> {
       self.as_value().get(lookup)
    }
 
  ....
```

And then `as_value` just wraps the `LogEvent::fields` in a `Value::Map`.

I realise there are probably 100's of gotchas to this I haven't thought through..

lib/shared/src/event/error.rs
Hoverbear: Oh perfect!

lib/shared/src/event/log_event/mod.rs
Hoverbear: In order to do this, `LogEvent` would have to store a `Value` itself, or clone the BtreeMap and transmute it into a value each time (which it then would not be able to return since it can't return a borrow to a value owned by the current scope). The prior is kind of problematic for the reasons outlined, the former is hecking expensive.

FungusHumungus:
This test fails:

```rust
    #[test]
    fn test_indexed_coalesce_from_string() {
        let parsed = LookupBuf::from_str("(a | b)[2]").unwrap();
        assert_eq!("(a | b)[2]", parsed.to_string());
    }
```

It doesn't seem to parse the Index after a coalesce, which it should.

lib/shared/src/event/metric/mod.rs
```diff
@@ -326,6 +327,156 @@ impl Metric {
     }
 }
 
+impl remap_lang::Object for Metric {
```
Hoverbear, line 330: This looks like a bad merge. Removing it.

src/config/unit_test.rs
```diff
@@ -45,7 +45,11 @@ async fn build_unit_tests(builder: ConfigBuilder) -> Result<Vec<UnitTest>, Vec<S
         match build_unit_test(test, &config).await {
             Ok(t) => tests.push(t),
             Err(errs) => {
-                let mut test_err = errs.join("\n");
+                let mut test_err = errs
```
Hoverbear, line 48: Generally it's not good practice to turn errors into strings, but here the caller returns strings, so it's warranted.

src/mapping/mod.rs
```diff
@@ -1,4 +1,6 @@
-use crate::event::{Event, Value};
+#![allow(deprecated)] // TODO: This uses things deprecated in https://github.com/timberio/vector/pull/5374
```
Hoverbear, line 1: I am under the impression all this code is going away and so all the marked deprecated code can when that does.

src/sources/http.rs
```diff
@@ -160,75 +159,60 @@ fn body_to_lines(buf: Bytes) -> impl Iterator<Item = Result<Bytes, ErrorMessage>
...
         Encoding::Text => body_to_lines(body)
-            .map(|r| Ok(Event::from(r?)))
+            .map(|r| {
+                Ok(log_event! {
+                    crate::config::log_schema().message_key().clone() => r?,
+                    crate::config::log_schema().timestamp_key().clone() => chrono::Utc::now(),
+                })
+            })
             .collect::<Result<_, _>>(),
         Encoding::Ndjson => body_to_lines(body)
```
Hoverbear, line 185: Some behavior changes here that are worth reviewing carefully.

lukesteensen (commented):
Still working through this but have a few comments so far.

lib/shared/src/event/log_event/mod.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// assert_eq!(event.get(plain_key), Some(&Value::from(1)));
+    /// assert_eq!(event.get(&lookup_key), Some(&Value::from(2)));
+    /// ```
+    pub fn get<'a>(&self, lookup: impl Into<Lookup<'a>> + Debug) -> Option<&Value> {
+        let mut working_lookup = lookup.into();
+        let span = trace_span!("get", lookup = %working_lookup);
+        let _guard = span.enter();
+
+        // The first step should always be a field.
+        let this_segment = working_lookup.pop_front().unwrap();
```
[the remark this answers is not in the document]
Hoverbear: Yes!

lib/shared/src/event/log_event/mod.rs
```diff
@@ -0,0 +1,828 @@
...
+        // This is good, since the first step into a LogEvent will also be a field.
+
+        // This step largely exists so that we can make `cursor` a `Value` right off the bat.
+        // We couldn't go like `let cursor = Value::from(self.fields)` since that'd take the value.
+        match this_segment {
+            Segment::Coalesce(sub_segments) => {
```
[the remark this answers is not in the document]
Hoverbear: I was thinking about this too! At the start I was just using a vector and iterating, but it ended up being that there was need to reconstruct the lookup in some error paths like coalesce where we attempt to insert into a primitive and need to do some backtracking and fixups.

lib/shared/src/event/log_event/mod.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// let plain_key = "foo";
+    /// let lookup_key = LookupBuf::from_str("bar.baz").unwrap();
+    /// let mut event = log_event! {
+    ///     plain_key => 1,
+    ///     lookup_key.clone() => 2,
+    /// }.into_log();
+    /// assert_eq!(event.get_mut(plain_key), Some(&mut Value::from(1)));
+    /// assert_eq!(event.get_mut(&lookup_key), Some(&mut Value::from(2)));
+    /// ```
+    pub fn get_mut<'a>(&mut self, lookup: impl Into<Lookup<'a>> + Debug) -> Option<&mut Value> {
```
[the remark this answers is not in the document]
Hoverbear: I absolutely agree, and I've been pondering how exactly to do it.

There's a few complications:

* It'd need to be generic over `&self` and `&mut self` if it were to work for get and get_mut/insert/remove
   + Having a `descent` and `descent_mut` option might be a cozy middle ground.
   + It would need to handle several possible cases, which might end up making it not much nicer than it is now.

lib/shared/src/event/log_event/mod.rs
```diff
@@ -0,0 +1,828 @@
...
+            Segment::Field {
+                name,
+                requires_quoting: _,
+            } => {
+                if working_lookup.len() == 0 {
+                    // Terminus: We **must** insert here or abort.
+                    trace!(field = %name, "Getting from root.");
+                    let retval = self.fields.remove(name);
+                    if prune && self.fields.get(name) == Some(&Value::Null) {
+                        self.fields.remove(name);
```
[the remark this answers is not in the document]
Hoverbear: I think this is extra cruft. I'll try to remove it. 😊

lib/shared/src/event/log_event/mod.rs
```diff
@@ -0,0 +1,828 @@
...
+
+    /// Returns the entire event as a `Value::Map`.
+    ///
+    /// ```rust
+    /// use shared::{event::*, lookup::*};
+    /// let event = LogEvent::default();
+    /// assert_eq!(event.take(), std::collections::BTreeMap::default());
+    /// ```
+    #[instrument(level = "trace", skip(self))]
+    pub fn take(self) -> BTreeMap<String, Value> {
```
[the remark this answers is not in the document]
Hoverbear: Ah heck. >.< Thank you.

I'm not really sure what would be best. I think returning a BTreeMap is ok? I also don't think returning a `Value` is a bad thing since users could then continue to use the same APIs.

lib/shared/src/event/log_event/mod.rs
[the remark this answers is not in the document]
Hoverbear: `remap_lang::Path` has these:

https://github.com/timberio/vector/blob/3b16679f07193b05fb5f9bd69b40b4b1e4bb2dd2/lib/remap-lang/src/path.rs#L65-L135

Maybe we could offer a similar API?

lib/shared/src/event/log_event/mod.rs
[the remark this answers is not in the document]
Hoverbear: A few other reasons we might want `LogEvent` to hold a `Value`:

* In `get` and `get_mut` we could support `event.get(Lookup::from_str(".").unwrap())`, instead of hitting an `.expect`
* In `paths` we could return `(., self.fields)` instead of being unable to return this.

lib/shared/src/event/log_event/mod.rs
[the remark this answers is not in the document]
lukesteensen: Yeah, that's similar to what I had in mind, just instead of `.foo.bar.(baz | qux)` turning into `[[".foo"], [".bar"], [".baz", ".qux"]]`, it'd let you iterate over `[[".foo", ".bar", ".baz"], [".foo", ".bar", ".qux"]]`. Both expose that there are multiple options, but the second does the work of materializing them for you. There is the theoretical danger of blowups, but real-world lookups should not be complex enough to cause problems (and being a lazy iterator would help).

I've been trying to work out if/when there's a difference in behavior between doing that kind of full materialization for paths with coalesces vs going segment-by-segment with backtracking. It _seems_ like there could be in the insert case at least, but it's complicated enough that I'm not sure.

lib/shared/src/event/log_event/mod.rs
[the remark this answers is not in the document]
lukesteensen: That makes sense to me. The top-level is special in that we know which type it is, but that doesn't make it not a `Value`.

lib/shared/src/event/log_event/mod.rs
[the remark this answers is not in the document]
lukesteensen: The downside of returning a `Value` seems to be that you'd lose the information that it's a map and not a string/int/etc, so I don't think that should be the only API. Maybe we return one for `take`, one for `into_inner` or something like that?

lib/shared/src/event/log_event/mod.rs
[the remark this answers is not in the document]
Hoverbear: Let me investigate doing that. :)

lib/shared/src/event/log_event/mod.rs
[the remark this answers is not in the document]
Hoverbear: You'd be able to do `take().into_map()` and have the exact map.

lib/shared/src/event/log_event/mod.rs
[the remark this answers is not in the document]
Hoverbear: Yeah, maybe it's better to leave this for future perf exploration?

lib/shared/src/event/log_event/mod.rs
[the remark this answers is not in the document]
Hoverbear: Looks to be mostly working! I have a few doc tests to fix then I want to add tests for the cases I listed.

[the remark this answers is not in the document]
Hoverbear: Totally worked!

Hoverbear:
@FungusHumungus could you help me explore this failure in `make test-behavior`?

https://github.com/timberio/vector/blob/4d97792cc4e7fc9fa79248ed08916d3da0671935/tests/behavior/transforms/remap.toml#L189-L206

```
failures:

--- tests/behavior/transforms/remap.toml ---

test 'remap_quoted_path':

check transform 'remap_quoted_path' failed conditions:
  condition[0]: predicates failed: [ a.b\.c.equals: "baz" ]
payloads (events encoded as JSON):
   input: {"a":{"b\\.c":"bar"},"d\\.e":"baz","message":"","timestamp":"2020-12-31T18:18:36.372670700Z"}
  output: {"a":{"b.c":null,"b\\.c":"bar"},"d\\.e":"baz","message":"","timestamp":"2020-12-31T18:18:36.372670700Z"}
```

FungusHumungus:
> could you help me explore this failure in `make test-behavior`? (It seems to be the last straggler)
> 

This seems to have highlighted a regression. It looks like at the moment, if you have a dot in your field name you can escape it with `\\`. 

For example, take this transform:

```
[transforms.add]
  type = "add_fields"
  inputs = ["stdin"]
  fields."zork\\.zonk" = 32
```

This adds the field: `"zork.zonk":32`.

In this PR, the field that gets added is: `"zork\\.zonk":32`.


It is possible to fix the failing test like:

```toml
[transforms.remap_quoted_path]
  inputs = []
  type = "remap"
  source = """
    .a."b.c" = ."d.e"
  """
[[tests]]
  name = "remap_quoted_path"
  [tests.input]
    insert_at = "remap_quoted_path"
    type = "log"
    [tests.input.log_fields]
      "a.\"b.c\"" = "bar"
      "\"d.e\"" = "baz"
  [[tests.outputs]]
    extract_from = "remap_quoted_path"
    [[tests.outputs.conditions]]
      "a.\"b.c\".equals" = "baz"
```

But, I'm not sure that would be correct. We probably need to update the path parsing so that it can escape `.` in the path segment if we want to maintain backward compatibility.

src/sinks/kafka.rs
```diff
@@ -285,7 +286,7 @@ impl Sink<Event> for KafkaSink {
         let timestamp_ms = match &item {
             Event::Log(log) => log
                 .get(log_schema().timestamp_key())
-                .and_then(|v| v.as_timestamp()),
+                .map(|v| v.as_timestamp()),
```
FungusHumungus, line 289: `as_timestamp` will panic if the field is not a timestamp. Are we absolutely sure that this will never happen here?

FungusHumungus: Nope. This can happen. Just send this message to the sink to get a panic:

```
{"message": "noog", "timestamp": "noog"}
```

lib/remap-lang/grammar.pest
```diff
@@ -109,7 +109,7 @@ lookup         = ${
   // Other paths (internal to Vector) do not necessarily have this distinction, despite mapping the same way.
   | (lookup_segment ~ (LOOKUP_PERIOD ~ lookup_segment)*)
 }
-lookup_segment = ${ ((lookup_field | lookup_field_quoted | lookup_array) ~ lookup_array*) | lookup_coalesce }
```
Hoverbear: Ahh omg I hadn't considered this! Does it just slip right in?

FungusHumungus: Yeah. just works. Thinking about it, it could probably be simplified even further..

jszwedko:
I ran the test-harness tests on commit `633a0fb5d4fac29b79d6890abcc666886d967043` of this branch vs. the `nightly/2021-01-08`. Results are here: 
[compare.txt](https://github.com/timberio/vector/files/5787694/compare.txt)

It does look like the "real-world" performance tests show some degradation: around ~20%. For example:

```

                                   __   __  __
                                   \ \ / / / /
                                    \ V / / /
                                     \_/  \/

                                   V E C T O R

Test Comparison
Test name: real_world_1_performance
Test configuration: high_concurrency
Subject: vector
Versions: nightly/2021-01-07 dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
Metric          | dev-new-accessors-633a0fb5d... | nightly/2021-01-07
----------------|--------------------------------|-------------------
Test count      | 1                              | 1                 
Duration (avg)  | 69s                            | 69s               
Duration (max)  | 69s                            | 69s               
CPU sys (max)   | 2 W                            | 2.5 (+23%)        
CPU usr (max)   | 27.9 W                         | 31.8 (+14%)       
Load 1m (avg)   | 3.7 W                          | 4.1 (+9%)         
Mem used (max)  | 579.3 MiB W                    | 587.6 MiB (+1%)   
Disk read (avg) | 800.1 kib/s (-8%)              | 877.3 kib/s W     
Disk read (sum) | 53.9 MiB W                     | 59.1 MiB (+9%)    
Disk writ (sum) | 12 MiB W                       | 27.2 MiB (+126%)  
Net recv (avg)  | 11 MiB/s (-17%)                | 13.4 MiB/s W      
Net recv (sum)  | 758.6 MiB (-17%)               | 922.4 MiB W       
Net send (sum)  | 251.5 MiB                      | 597.2 MiB         
TCP estab (avg) | 614                            | 611               
TCP syn (avg)   | 36                             | 8                 
TCP close (avg) | 0                              | 0                 
W = winner
vector = dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
vector = nightly/2021-01-07
```

Some of the tests didn't run for both or one of the versions. I need to dig into that a bit more.

Hoverbear:
@jszwedko are these using nested lookups or flat keys? @lukesteensen added some regex fastpaths to the original Path which let us skip a lot of the work. I'd be curious if this is resulting from that.

Hoverbear:
This one is negligible:

```

                                   __   __  __
                                   \ \ / / / /
                                    \ V / / /
                                     \_/  \/

                                   V E C T O R

Test Comparison
Test configuration: high_concurrency
Subject: vector
Versions: nightly/2021-01-07 dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
Metric          | dev-new-accessors-633a0fb5d... | nightly/2021-01-07
----------------|--------------------------------|-------------------
Test count      | 1                              | 2                 
Duration (avg)  | 75s                            | 71.5s             
Duration (max)  | 75s                            | 72s               
CPU sys (max)   | 2.4 (+2%)                      | 2.3 W             
CPU usr (max)   | 27.8 W                         | 33.2 (+19%)       
Load 1m (avg)   | 4.8 (+4%)                      | 4.6 W             
Mem used (max)  | 530.6 MiB W                    | 545.1 MiB (+2%)   
Disk read (avg) | 609 kib/s (-1%)                | 618 kib/s W       
Disk read (sum) | 44.6 MiB (+3%)                 | 43.1 MiB W        
Disk writ (sum) | 11.1 MiB (+39%)                | 7.9 MiB W         
Net recv (avg)  | 54.3 MiB/s (-7%)               | 58.4 MiB/s W      
Net recv (sum)  | 4 gib (-2%)                    | 4.1 gib W         
Net send (sum)  | 4 gib                          | 4.1 gib           
TCP estab (avg) | 1347                           | 1340.5            
TCP syn (avg)   | 0                              | 0                 
TCP close (avg) | 4                              | 1                 
W = winner
vector = dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
vector = nightly/2021-01-07
```

Similar:

```

                                   __   __  __
                                   \ \ / / / /
                                    \ V / / /
                                     \_/  \/

                                   V E C T O R

Test Comparison
Test name: file_to_tcp_performance
Test configuration: single_file
Subject: vector
Versions: nightly/2021-01-07 dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
Metric          | dev-new-accessors-633a0fb5d... | nightly/2021-01-07
----------------|--------------------------------|-------------------
Test count      | 1                              | 2                 
Duration (avg)  | 65s                            | 63.5s             
Duration (max)  | 65s                            | 64s               
CPU sys (max)   | 5.4 W                          | 7.9 (+45%)        
CPU usr (max)   | 19.7 (+11%)                    | 17.7 W            
Load 1m (avg)   | 0.2 W                          | 0.2 (+0%)         
Mem used (max)  | 216.1 MiB W                    | 218.5 MiB (+1%)   
Disk read (avg) | 752 kib/s (-3%)                | 776.5 kib/s W     
Disk read (sum) | 47.7 MiB W                     | 48.1 MiB (+0%)    
Disk writ (sum) | 19.9 MiB W                     | 19.9 MiB (+0%)    
Net recv (avg)  | 1.3 kib/s W                    | 288.9 b/s (-78%)  
Net recv (sum)  | 87 kib W                       | 17.9 kib (-79%)   
Net send (sum)  | 7.6 kib                        | 5.2 kib           
TCP estab (avg) | 66                             | 64.5              
TCP syn (avg)   | 0                              | 0                 
TCP close (avg) | 0                              | 0                 
W = winner
vector = dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
vector = nightly/2021-01-07
```

Also (but what's going on with disk write here??? ):

```
                                   __   __  __
                                   \ \ / / / /
                                    \ V / / /
                                     \_/  \/

                                   V E C T O R

Test Comparison
Test name: real_world_1_performance
Test configuration: without_lua
Subject: vector
Versions: nightly/2021-01-07 dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
Metric          | dev-new-accessors-633a0fb5d... | nightly/2021-01-07
----------------|--------------------------------|-------------------
Test count      | 1                              | 2                 
Duration (avg)  | 68s                            | 66.5s             
Duration (max)  | 68s                            | 67s               
CPU sys (max)   | 2.5 (+4%)                      | 2.4 W             
CPU usr (max)   | 99.7 (+0%)                     | 99.4 W            
Load 1m (avg)   | 1.5 W                          | 1.6 (+7%)         
Mem used (max)  | 305.7 MiB W                    | 306.6 MiB (+0%)   
Disk read (avg) | 835.1 kib/s W                  | 758.1 kib/s (-9%) 
Disk read (sum) | 55.5 MiB (+12%)                | 49.2 MiB W        
Disk writ (sum) | 14.1 MiB (+361%)               | 3.1 MiB W         
Net recv (avg)  | 4.5 MiB/s (-26%)               | 6.1 MiB/s W       
Net recv (sum)  | 304.7 MiB (-24%)               | 405.7 MiB W       
Net send (sum)  | 181.4 MiB                      | 264.4 MiB         
TCP estab (avg) | 611                            | 551               
TCP syn (avg)   | 0                              | 0.5               
TCP close (avg) | 1                              | 0.5               
W = winner
vector = dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
vector = nightly/2021-01-07
```

This is also negligible?

```
                                   __   __  __
                                   \ \ / / / /
                                    \ V / / /
                                     \_/  \/

                                   V E C T O R

Test Comparison
Test name: real_world_1_performance
Test configuration: one_sink
Subject: vector
Versions: nightly/2021-01-07 dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
Metric          | dev-new-accessors-633a0fb5d... | nightly/2021-01-07
----------------|--------------------------------|-------------------
Test count      | 1                              | 1                 
Duration (avg)  | 68s                            | 75s               
Duration (max)  | 68s                            | 75s               
CPU sys (max)   | 2 W                            | 3.5 (+75%)        
CPU usr (max)   | 89.6 W                         | 97.7 (+9%)        
Load 1m (avg)   | 1.1 W                          | 1.6 (+41%)        
Mem used (max)  | 317.2 MiB (+3%)                | 305.7 MiB W       
Disk read (avg) | 835.3 kib/s W                  | 797.8 kib/s (-4%) 
Disk read (sum) | 55.5 MiB W                     | 58.4 MiB (+5%)    
Disk writ (sum) | 13.8 MiB W                     | 25.2 MiB (+82%)   
Net recv (avg)  | 14.4 MiB/s (-35%)              | 22.4 MiB/s W      
Net recv (sum)  | 978.1 MiB (-41%)               | 1.6 gib W         
Net send (sum)  | 36.9 MiB                       | 64.4 MiB          
TCP estab (avg) | 424                            | 432               
TCP syn (avg)   | 25                             | 1                 
TCP close (avg) | 0                              | 0                 
W = winner
vector = dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043
vector = nightly/2021-01-07
```

Hoverbear:
For `real_world_1_performance/high_concurrency` I presume that the slowdown is the lua transform:

 https://github.com/timberio/vector-test-harness/blob/3f18b986e84cf6c8cbc18341c50b6e6f9fa2cbe7/cases/real_world_1_performance/ansible/config_files/vector.toml#L139-L145

This would need to re-parse those accessors each time. I wonder if building a little cache on the lua transform would fix this.

lukesteensen:
As discussed earlier, the plan here is to maintain this PR and merge shortly after the 0.12 release is cut. As part of that, there are a few concerns we want to make sure get addressed:

### Behavior compatibility

The primary change in this PR is that strings will no longer automatically be parsed into lookup segments. This is good because we want to encourage the pre-parsing of lookups for performance reasons, but the fact that we still accept strings means that it's possible we missed an area where we should be doing that pre-parsing.

This is the biggest reason for delaying merge until after the release, to give us more time to flush out these kinds of issues in nightlies. One potentially complementary strategy would be to further limit the API to not accept raw strings and let the compiler point us to areas where we are using them. This may or may not be a practical change to merge, but trying would give us a list of callsites to investigate. We could also look into only allowing raw string lookups in tests, which I believe is the bulk of their usage.

A secondary behavior concern is how we handle escaping input paths from the `{add,remove,rename}_fields` transforms. Given that these are already deprecated, it would be nice to avoid any potentially breaking change to their configuration. It has been suggested that a small compatibility shim between their configs and lookup creation could accomplish this.

### Performance

We need to dig into the benchmark results posted above and determine the degree to which they'll represent user experience when this change is released. Given that the API is being moved in the direction of pre-parsing paths, this change _should_ but us in a better position performance-wise, but that's obviously not the case yet. We should not ship an actual ~20% reduction in user-facing performance, but I'm skeptical that we won't be able to avoid that. The API is what we want, and we can reduce the complexity of the implementation as needed to make it fast.

### Feature reduction

This area of functionality has grown in complexity over time to a concerning degree. In order to minimize potential edge cases, performance issues, and ongoing maintenance cost, we should make an effort to simplify down to the features we are confident we want to support long term. A prime example of something to be cut is coalescing. While a valuable feature in the remap language itself, it should be implemented there and kept out of our core event API. This can be done in a similar manner to the existing `to_alternative_strings` function.

We should also revisit the use cases for lookups in light of our new focus on remap. Their original purpose was primarily to structure user input from TOML configs for components operating on specific keys. This is still valid, but many of the more complex cases may now be covered by remap. Another big use of the API is building and inspecting events in tests, which has very different requirements from both pre-configured component usage and programmable usages like Lua or WASM. We should drop any feature that doesn't support these use cases and potentially look into splitting them from each other in order to provide simpler and more targeted implementations for each.

jszwedko:
Complete set of benches comparing `nightly/2021-01-07` and `dev-new-accessors-633a0fb5d4fac29b79d6890abcc666886d967043`: [compare.txt](https://github.com/timberio/vector/files/5797606/compare.txt).

MOZGIII:
#6274 has been merged, and should be available in the CI automatically once the `master` is merged into this PR.

FungusHumungus:
Ok, I have merged master into this and have it compiling, but I see already there are more conflicts!!! I will get to these shortly, they can't be too arduous..

The code I have pushed is not perfect, there's an outstanding question on the kubernetes source I have fired at @eeyun and a number of places where it does `use shared::lookup::LookupBuf` where to keep it consistent it should do `use crate::event::LookupBuf`.

FungusHumungus:
I should note that currently this is pulling in the entire legacy Remap pest parser to parse the paths. This needs to be tidied up. It seems extravagant to have three different parsing libraries in a single project, so it would be useful if we could reuse some of the lalrpop code -  or since it is quite a simple parsing task, perhaps we could use nom, which may also provide some performance benefits..

blt:
> I should note that currently this is pulling in the entire legacy Remap pest parser to parse the paths. This needs to be tidied up. It seems extravagant to have three different parsing libraries in a single project, so it would be useful if we could reuse some of the lalrpop code - or since it is quite a simple parsing task, perhaps we could use nom, which may also provide some performance benefits..

This strikes me as something that would be best tackled in a follow-up issue, to avoid this PR ballooning any further.

blt:
Ah, hmm. #7014 collides with this change set as the metrics stuff is no longer in the top-level crate but `shared`.

FungusHumungus:
> Ah, hmm. #7014 collides with this change set as the metrics stuff is no longer in the top-level crate but `shared`.

Yeah, a lot of the conflicts this gets are caused by the move to shared.

blt:
> > Ah, hmm. #7014 collides with this change set as the metrics stuff is no longer in the top-level crate but `shared`.
> 
> Yeah, a lot of the conflicts this gets are caused by the move to shared.

Indeed. This is a change I'd like to see unwound actually. My proposal for the core extraction is that core be the top-level crate and these are very much core-level concerns, to my mind.

blt:
Alright, well. I've reverted the merges that I've done. My hope was to bring #7014 in so I can compare the runtime of current master with this. Since I cannot easily merge with master I'm going to compare the tip of the PR with 6c0df5614a10c4ef7d8c5e18d3bba289c6a7a0b5 instead.

blt:
Okay, benchmarking process started. Should have numbers in a few hours.

blt:
The results so far are not uniformly positive. I've made a comparison between 6c0df5614a10c4ef7d8c5e18d3bba289c6a7a0b5 and cb12351ba2a039db1ea962ab75f990b3d149f37e with [this run script](https://gist.github.com/blt/21a67a48134619eee6a99d6cca161ebd#file-vector-sh). The following [gist](https://gist.github.com/blt/21a67a48134619eee6a99d6cca161ebd) has the vector configs referenced in the below. 

# Hyperfine 

The first phase of the this benchmark uses the [hyperfine](https://github.com/sharkdp/hyperfine) tool to compare vector runtime versions in a stable way. The results:

``` _                            __ _
| |__  _   _ _ __   ___ _ __ / _(_)_ __   ___
| '_ \| | | | '_ \ / _ \ '__| |_| | '_ \ / _ \
| | | | |_| | |_) |  __/ |  |  _| | | | |  __/
|_| |_|\__, | .__/ \___|_|  |_| |_|_| |_|\___|
       |___/|_|
PROCESSING: remap-multihop.toml
Benchmark #1: zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-6c0df5614a10c4ef7d8c5e18d3bba289c6a7a0b5 -qq --config /home/blt/projects/com/datadog/timberio
/vector-test-harness/cases/performance/klog_remap_to_blackhole/configs/remap-multihop.toml
  Time (mean ± σ):     16.608 s ±  0.359 s    [User: 75.169 s, System: 10.587 s]
  Range (min … max):   16.151 s … 17.099 s    10 runs

Benchmark #2: zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-cb12351ba2a039db1ea962ab75f990b3d149f37e -qq --config /home/blt/projects/com/datadog/timberio
/vector-test-harness/cases/performance/klog_remap_to_blackhole/configs/remap-multihop.toml
  Time (mean ± σ):     13.808 s ±  0.403 s    [User: 71.355 s, System: 4.251 s]
  Range (min … max):   13.355 s … 14.671 s    10 runs

Summary
  'zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-cb12351ba2a039db1ea962ab75f990b3d149f37e -qq --config /home/blt/projects/com/datadog/timberio/vector-tes
t-harness/cases/performance/klog_remap_to_blackhole/configs/remap-multihop.toml' ran
    1.20 ± 0.04 times faster than 'zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-6c0df5614a10c4ef7d8c5e18d3bba289c6a7a0b5 -qq --config /home/blt/projects
/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/configs/remap-multihop.toml'


PROCESSING: remap-nohop.toml
Benchmark #1: zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-6c0df5614a10c4ef7d8c5e18d3bba289c6a7a0b5 -qq --config /home/blt/projects/com/datadog/timberio
/vector-test-harness/cases/performance/klog_remap_to_blackhole/configs/remap-nohop.toml
  Time (mean ± σ):     29.038 s ±  0.960 s    [User: 77.805 s, System: 10.876 s]
  Range (min … max):   27.844 s … 30.734 s    10 runs

Benchmark #2: zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-cb12351ba2a039db1ea962ab75f990b3d149f37e -qq --config /home/blt/projects/com/datadog/timberio
/vector-test-harness/cases/performance/klog_remap_to_blackhole/configs/remap-nohop.toml
  Time (mean ± σ):     37.095 s ±  0.685 s    [User: 93.243 s, System: 12.653 s]
  Range (min … max):   36.222 s … 38.132 s    10 runs

Summary
  'zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-6c0df5614a10c4ef7d8c5e18d3bba289c6a7a0b5 -qq --config /home/blt/projects/com/datadog/timberio/vector-tes
t-harness/cases/performance/klog_remap_to_blackhole/configs/remap-nohop.toml' ran
    1.28 ± 0.05 times faster than 'zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-cb12351ba2a039db1ea962ab75f990b3d149f37e -qq --config /home/blt/projects
/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/configs/remap-nohop.toml'


PROCESSING: stdin_to_blackhole.toml
Benchmark #1: zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-6c0df5614a10c4ef7d8c5e18d3bba289c6a7a0b5 -qq --config /home/blt/projects/com/datadog/timberio
/vector-test-harness/cases/performance/klog_remap_to_blackhole/configs/stdin_to_blackhole.toml
  Time (mean ± σ):      6.745 s ±  0.162 s    [User: 12.504 s, System: 2.282 s]
  Range (min … max):    6.502 s …  6.943 s    10 runs

Benchmark #2: zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-cb12351ba2a039db1ea962ab75f990b3d149f37e -qq --config /home/blt/projects/com/datadog/timberio
/vector-test-harness/cases/performance/klog_remap_to_blackhole/configs/stdin_to_blackhole.toml
  Time (mean ± σ):      6.741 s ±  0.155 s    [User: 12.425 s, System: 2.235 s]
  Range (min … max):    6.477 s …  7.008 s    10 runs

Summary
  'zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-cb12351ba2a039db1ea962ab75f990b3d149f37e -qq --config /home/blt/projects/com/datadog/timberio/vector-tes
t-harness/cases/performance/klog_remap_to_blackhole/configs/stdin_to_blackhole.toml' ran
    1.00 ± 0.03 times faster than 'zcat baked.log.gz | /home/blt/projects/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/bin/vector-6c0df5614a10c4ef7d8c5e18d3bba289c6a7a0b5 -qq --config /home/blt/projects
/com/datadog/timberio/vector-test-harness/cases/performance/klog_remap_to_blackhole/configs/stdin_to_blackhole.toml'

```

In the case of [`remap-multihop`](https://gist.github.com/blt/21a67a48134619eee6a99d6cca161ebd#file-remap-multihop-toml) this PR is 20% +/- 4% faster than master. For [`remap-nohop`](https://gist.github.com/blt/21a67a48134619eee6a99d6cca161ebd#file-remap-nohop-toml) master is 28% +/-  5% faster than this PR. For [`stdin-to-blackhole`](https://gist.github.com/blt/21a67a48134619eee6a99d6cca161ebd#file-stdin_to_blackhole-toml) there is no change. 

## Summary 

In two of the three cases tested this PR is dramatically different than master. Unfortunately, in one case it is dramatically slower. The case in which this PR is dramatically slower is the "nohop" variant of a remap pipeline, likely to be the manner in which users configure their vector (as we've instructed them to). At least, this "nohop" configuration is not _obviously wrong_ and the 28+/-5% difference is steep. 

# Perf / Flamegraph 

This flamegraphs generated from perf are [here](https://gist.github.com/blt/c32f22c079b44fd3c10f8206230f6df6). The overall shape of the graphs for each variant are similar -- good -- although the timings are different. However, since the structure of the code has changed so dramatically between this PR and master it's quiet hard, overall, to draw any conclusions here. 

## Summary 

Inconclusive. Comparisons of this kind are best with small differences. 

# Massif

The massif results for each variant are available in this zip file: [massif-results.zip](https://github.com/timberio/vector/files/6282337/massif-results.zip) The programs do not exhibit the leak fixed in #7014. In each variant across all configurations there is very little difference in the allocation pattern presented. 

## Summary 

No difference between variants.

blt:
Looking at the data here I'm inclined to say that we should close this PR out. While it's clear that there are performance gains to be made it's also clear that this PR causes a significant pessimization in a remap heavy configuration, something we're pushing for. I consider it likely that this pessimization is fixable and that the approach taken here has legs but this PR is too big. I do not have a hypothesis on where the pessimization is and while it's possible to get set up to make a hypothesis that in itself will be a non-trivial amount of work. As @FungusHumungus noted keeping this PR up to date with master branch is a daily feat. Further, but less concerning, we have not significantly reduced allocations in this PR relative to master, with regard to #6831. Saving allocations were part of this PR's goal set, I do believe. If we were to continue with this PR -- and I would like to acknowledge that _a lot_ of work has gone into this PR -- I think we will need to:

* pull `shared` back into the root crate to avoid conflicts with #7014 and my future recommendation in #7027
* address the use of the [legacy parser](https://github.com/timberio/vector/pull/5374#issuecomment-815912685) 
* en-cheapen the master merge cycle enough to enable performance interrogation 

The last point seems the hardest to achieve to me, though it's potentially addressed by the first. @FungusHumungus you've spent the most time wrestling with this code lately. Do you see the situation here differently than I do?

FungusHumungus:
Those are interesting findings. So it seems that the PR is significantly faster, unless it has to deal with Remap. This suggests that the issue is in the interface between Remap and the Lookup api. 

I would put my money on the culprit being here: https://github.com/timberio/vector/blob/new-accessors/lib/shared/src/lookup/lookup_buf/mod.rs#L107-L114

```rust
impl TryFrom<&vrl::Path> for LookupBuf {
    type Error = LookupError;
    fn try_from(target: &vrl::Path) -> Result<Self, Self::Error> {
        let path_string = target.to_string();
        trace!(path = %path_string, "Converting to LookupBuf.");
        LookupBuf::from_str(&path_string)
    }
}
```

Any time Remap has to access a path in the event it needs to convert the Remap path into a Lookup path. It does this by dumping it to a string and then reparsing it. The main objective behind this lookup was to reduce these parsing costs. However, when using Remap, it actually now increases the amount of parsing that occurs.

JeanMertz:
> While it's clear that there are performance gains to be made it's also clear that this PR causes a significant pessimization in a remap heavy configuration, something we're pushing for.

The one thing I would note here is that a few months ago — when both VRL and this work were in flux — it was decided to get this PR over the finish line without considering any performance regressions related to Vector <> VRL integration.

That is to say, a naïve implementation of `From<vrl::Path> for LookupBuf` and `impl vrl::Target for Event` was chosen to get this over the finish line as soon as possible.

The thought process behind that decision was that there is nothing technically stopping us from using the same performance optimizations w.r.t. path lookup in both VRL and Vector, by extracting out that logic into its own crate, having both VRL and Vector depend on it, and allowing us to pass paths between both without any conversion or allocation needed.

Obviously, this PR has lingered since, and VRL has shipped, so a lot has changed since then, but I never spent any large amount of time improving VRL's path lookup performance precisely because this work was on the horizon, so it _might_ still be worth pushing this PR over the finish line _if_ benchmarks show performance improvements leaving VRL out of the picture.

It would require us to schedule work to actually improve VRL <> Lookup integration, since we can't ship this work in a new release without tackling that first.

Having said that, the PR _is_ massive (a lot of work went into getting this done, as you mentioned), and it feels somewhat like a daunting task to get this in and then iterate on it further. One way forward is to take the idea proven in this PR, and build a new lookup crate that can be worked on in isolation, iterating on it in small PRs, until it's at a stage where both Vector and VRL can use it, and we can have one or two relatively larger PRs to do all the legwork of updating all path references to the new lookup API.

FungusHumungus:
I would be interested if a naive implementation of TryFrom:

```rust
impl TryFrom<&vrl::Path> for LookupBuf {
    type Error = LookupError;
    fn try_from(target: &vrl::Path) -> Result<Self, Self::Error> {
        Ok(LookupBuf {
            segments: target
                .segments()
                .iter()
                .map(|segment| match segment {
                    vrl::Segment::Field(vrl::Field::Regular(field)) => SegmentBuf::Field {
                        name: field.clone(),
                        requires_quoting: false,
                    },
                    _ => unimplemented!(),
                })
                .collect(),
        })
    }
}
```

had any impact. It's not perfect as there are still allocations, but it should be better than running it through pest. If it shifts things in the right direction, it would be indicative that the issue is here and would warrant further investigation.  

Note this only works with simple fields - no indexing or coalescing, but it looks like that is all the test conf uses anyway.

The 28% potential improvement is IMO well worth pursuing.

blt:
> Having said that, the PR _is_ massive (a lot of work went into getting this done, as you mentioned), and it feels somewhat like a daunting task to get this in and then iterate on it further. One way forward is to take the idea proven in this PR, and build a new lookup crate that can be worked on in isolation, iterating on it in small PRs, until it's at a stage where both Vector and VRL can use it, and we can have one or two relatively larger PRs to do all the legwork of updating all path references to the new lookup API.

This strikes me as the proper way forward. The ideas in this PR are _absolutely_ worth pursuing but we need to be careful to separate out the new work -- the lookup crate -- from the mechanical work of integrating it. These two things got integrated into one PR and it makes interrogating the new work quite hard. 

Personally, I'd be very happy to see an issue made to extract the path work here into a separate crate, PR and then a follow-up issue for integration into the wider project. Doing that we _should_ be able to capture the performance gain of this PR without imposing the loss. I would also vote to prioritize this project.

@FungusHumungus @JeanMertz @binarylogic I know we talked about this some out-of-ticket. Does this take capture your thoughts on the subject?

FungusHumungus:
> Does this take capture your thoughts on the subject?

Yes. I would suggest in the following order.

1. Pull it out into a separate crate. However, you (@blt) mentioned above that you would like to see `shared` pulled back into the root crate. That will not be possible as it needs sharing with Vrl. Is that still going to be an issue given we are now splitting the tasks over multiple stages?
2. Get Vrl working with the new crate. I think the main challenges here will involve enabling Vrl to use the same parsing code as the Lookup, bearing in mind Lookup likes to parse as a reference type (`Lookup<'a>`). The Vrl parser currently parses as an owned type. So it would need to handle both. Once Vrl is working with the new crate we should be able to do some hefty benchmarking. I think the main problematic config was pretty much entirely using lookups via Vrl. So I would expect to see a 28% performance gain right there.. If not more (fingers crossed).
3. Move the rest of the app to the new crate. I imagine we could temporarily support both apis, so `LogEvent` would have an `insert` and a `speedy_insert` function. This part could be done at a leisurely pace. When everything is moved over, we can delete `insert` and sed `speedy_insert` -> `insert`.

blt:
> > Does this take capture your thoughts on the subject?
> 
> Yes. I would suggest in the following order.
> 
>     1. Pull it out into a separate crate. However, you (@blt) mentioned above that you would like to see `shared` pulled back into the root crate. That will not be possible as it needs sharing with Vrl. Is that still going to be an issue given we are now splitting the tasks over multiple stages?

Could be I'm wrong there. I've only got vague notions for #7027 so far and this is part of that. As is, the only blocker I'm aware of with `shared` is the `Metrics::from_metrics_kv`. It does strike me that `Event` is a core concern of vector, but I do see that there's a dependency issue. 

>     2. Get Vrl working with the new crate. I think the main challenges here will involve enabling Vrl to use the same parsing code as the Lookup, bearing in mind Lookup likes to parse as a reference type (`Lookup<'a>`). The Vrl parser currently parses as an owned type.

FWIW if VRL could be made to parse by reference that would go a long way to addressing #6831. 

> So it would need to handle both. Once Vrl is working with the new crate we should be able to do some hefty benchmarking. I think the main problematic config was pretty much entirely using lookups via Vrl. So I would expect to see a 28% performance gain right there.. If not more (fingers crossed).
> 
>     3. Move the rest of the app to the new crate. I imagine we could temporarily support both apis, so `LogEvent` would have an `insert` and a `speedy_insert` function. This part could be done at a leisurely pace. When everything is moved over, we can delete `insert` and sed `speedy_insert` -> `insert`.

I buy the general approach, yeah. I like that it can be done gradually.

FungusHumungus:
> It does strike me that `Event` is a core concern of vector, but I do see that there's a dependency issue.

Vrl only needs the Lookup path data structures. It is absolutely doesn't want anything to do with the Event. That way it can remain agnostic to the data it is running on top of. So the new crate should really only contain the core Lookup api. The actual lookup code, Event, Metrics etc can remain within Vector core.

blt:
> > It does strike me that `Event` is a core concern of vector, but I do see that there's a dependency issue.
> 
> Vrl only needs the Lookup path data structures. It absolutely doesn't want anything to do with the Event. That way it can remain agnostic to the data it is running on top of. So the new crate should really only contain the core Lookup api. The actual lookup code, Event, Metrics etc can remain within Vector core.

Rad. Even better.

blt:
@binarylogic @JeanMertz @FungusHumungus I've made #7070 to replace this PR. I'd be grateful for your thoughts.

blt:
Closing in favor of #7070.

PR #5375, opened 2020-12-03
enhancement(remap): track inner container type definitions

JeanMertz:
This change allows our type check system to track the type definitions of values contained within collections.

For example:

```rust
let vec = vec![Value::Null, Value::Boolean(true)];
let expression = Array::from(vec);
let state = state::Compiler::default();

let def = expression.type_def(&state)

def.kind 			// Kind::Array
def.scalar_kind() 	// Kind::Null | Kind::Boolean
```

This also works for nested arrays.

### DIFF
--- a/lib/remap-lang/src/parser.rs
+++ b/lib/remap-lang/src/parser.rs
@@ -146,9 +146,9 @@ impl<'a> Parser<'a> {
         }
 
         if let Some(expression) = expressions.last() {
-            let type_def = expression.type_def(&self.compiler_state);
+            let kind = expression.type_def(&self.compiler_state).scalar_kind();
 
-            if !type_def.kind.is_all() && type_def.kind.contains_regex() {
+            if !kind.is_all() && kind.contains_regex() {
                 return Err(Error::RegexResult.into());
             }
         }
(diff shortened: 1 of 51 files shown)


lib/remap-lang/src/type_def.rs
```diff
@@ -16,13 +16,42 @@ pub struct TypeDef {
...
+    ///         kind: Kind::Array,
+    ///         inner_type_def: Some(TypeDef {
+    ///             fallible: false,
+    ///             kind: Kind::Null | Kind::Boolean,
+    ///             inner_type_def: None,
+    ///         }.boxed()),
+    ///     },
+    /// );
+    /// ```
+    pub inner_type_def: Option<Box<TypeDef>>,
```
JeanMertz, line 46: One thing I should note here is that we don't really have a reason to store `TypeDef`, we could just store `Kind`, but then we'd have to move (and rename) `inner_type_def` to add it to `Kind` (e.g. `kind.inner_kind`), but that in term is problematic because `Kind` is currently just an integer (using bitflags).

I started making this change, but the churn was too great to justify for a first iteration.

Basically, the result of this is that we get `fallible` on every level, even though there's no way to have an inner type be fallible, but an outer type not to be fallible.

Perhaps we'll tidy this up in the near future, but I found it to be too small of an issue to solve in this PR.

lib/remap-lang/src/parser.rs
```diff
@@ -146,9 +146,9 @@ impl<'a> Parser<'a> {
         }
 
         if let Some(expression) = expressions.last() {
-            let type_def = expression.type_def(&self.compiler_state);
+            let kind = expression.type_def(&self.compiler_state).scalar_kind();
 
-            if !type_def.kind.is_all() && type_def.kind.contains_regex() {
+            if !kind.is_all() && kind.contains_regex() {
```
JeanMertz, line 151: Technically, you can still _leak_ a regexp if its type is `any`.

We'll close this leak at some point as well, but for now, I considered it to be an edge-case (and one that has no implications, since Vector simply converts the regexp to its string representation), and the cost of plugging that hole was too great for now (denying returning `any` in a TRL script).

bruceg (commented):
LGTM so far, but it appears the new `inner_type_def` is missing from some more sources.

StephenWakely (approved):

PR #5379, opened 2020-12-04
enhancement: Add support for basic auth credendtials in URL for more sinks

### LINKED ISSUE #5138: Add support for basic auth credendtials in URL for more sinks
Follow up #5095, we should also add this feature to all sinks that support basic-auth.

- [ ] http
- [ ] clickhouse
- [ ] loki

juchiast:
close #5138

I expanded `UriSerde` to also extract credentials from the input.


src/sinks/clickhouse.rs
```diff
@@ -76,8 +76,13 @@ impl SinkConfig for ClickhouseConfig {
         let tls_settings = TlsSettings::from_options(&self.tls)?;
         let client = HttpClient::new(tls_settings)?;
 
+        let mut config = self.clone();
+        let uri = self.endpoint.parse::<UriSerde>()?;
```
jszwedko, line 80: Is there a reason we do the parsing down here rather than making `endpoint` `UriSerde` on the `ClickhouseConfig`?

src/sinks/logdna.rs
```diff
@@ -236,7 +236,11 @@ impl HttpSink for LogdnaConfig {
 
 impl LogdnaConfig {
     fn build_uri(&self, query: &str) -> Uri {
-        let host: Uri = self.endpoint.clone().unwrap_or_else(|| HOST.clone()).into();
+        let host: Uri = self
+            .endpoint
+            .clone()
+            .map(|uri| uri.uri)
```
jszwedko, line 242: Maybe `.map(Into::into)`?

src/sinks/util/uri.rs
```diff
@@ -1,18 +1,38 @@
...
+}
+
+impl UriSerde {
+    /// Used to combine existing authorization credentials with credentials in this URI.
+    /// If both is `Some`, return an error. Otherwise, choose one of them.
+    pub fn merge_auth_config(&self, auth: &mut Option<Auth>) -> crate::Result<()> {
+        if auth.is_some() && self.auth.is_some() {
+            Err("Two authorization credentials was provided.".into())
+        } else {
+            *auth = auth.take().or_else(|| self.auth.clone());
```
jszwedko, line 27: I think this function makes more sense on `Auth` since it mutates it.

src/sinks/util/uri.rs
```diff
@@ -44,27 +74,107 @@ impl<'a> Visitor<'a> for UriVisitor {
...
+    let user = url.username();
+    if !user.is_empty() {
+        let user = percent_decode_str(user).decode_utf8_lossy().into_owned();
+
+        let password = url.password().unwrap_or("");
+        let password = percent_decode_str(password)
+            .decode_utf8_lossy()
+            .into_owned();
+
+        url.set_username("").unwrap();
```
jszwedko, line 118: Could we make these `expect()`s to document why we expect them to never fail?

src/sinks/util/uri.rs
```diff
@@ -1,18 +1,38 @@
...
+    pub uri: Uri,
+    pub auth: Option<Auth>,
+}
+
+impl UriSerde {
+    /// Used to combine existing authorization credentials with credentials in this URI.
+    /// If both is `Some`, return an error. Otherwise, choose one of them.
+    pub fn merge_auth_config(&self, auth: &mut Option<Auth>) -> crate::Result<()> {
+        if auth.is_some() && self.auth.is_some() {
+            Err("Two authorization credentials was provided.".into())
```
jszwedko, line 25: I'd also be ok with preferring the `Auth` credentials over credentials in the URI.

src/sinks/clickhouse.rs
```diff
@@ -76,8 +76,13 @@ impl SinkConfig for ClickhouseConfig {
...
+        uri.merge_auth_config(&mut config.auth)?;
```
jszwedko, line 81: It's a bit unfortunate that we need to remember to call this merge method everywhere, but I'm having a hard time coming up with a better modeling given they are separate config fields.

jszwedko (commented):
Thanks for expanding support for this!

I left a few inline comments, but nothing major.

src/sinks/clickhouse.rs
juchiast: Yes it is unfortunate, I also don't have any ideas at the moment.

src/sinks/util/uri.rs
juchiast: I moved it to `fn merge_auth_config(&mut Option<Auth>, &Option<Auth>)` inside `impl Auth`. Unfortunately we cannot use an associated method because it need `self` to be `Option<Auth>`.

jszwedko: Instead of requiring `self` to be `Option<Auth>` could we just have callers use `maybe_auth.map(...)`?

src/sinks/util/uri.rs
jszwedko: These comments are good, but I was thinking you could just use `expect` to make the failure self documenting. Something like:

```
url.set_username("").expect("authority unexpectedly empty");
```

jszwedko: These comments are good, but I was thinking you could just use `expect` to make the failure self documenting. Something like:

```
url.set_username("").expect("authority unexpectedly empty");
```

src/sinks/util/uri.rs
jszwedko: Never mind, I see the function signature changed to merging two `Auth`s instead.

src/http.rs
```diff
@@ -172,46 +171,21 @@ impl<B> fmt::Debug for HttpClient<B> {
...
-            let password = percent_decode(password.as_bytes())
-                .decode_utf8_lossy()
-                .into_owned();
-
-            url.set_username("").ok()?;
-            url.set_password(None).ok()?;
-
-            Some((url.to_string(), Auth::Basic { user, password }))
+    /// If both is `Some`, return an error. Otherwise, choose one of them and store in `auth`.
+    pub fn merge_auth_config(auth: &mut Option<Auth>, auth1: &Option<Auth>) -> crate::Result<()> {
```
jszwedko, line 183: I realize there is some overhead associated with this, but I'm wondering if this shouldn't just return a `crate::Result<Auth>` instead of mutating the first passed parameter. It could be tied to the lifetime to avoid cloning `Auth`:

```
pub fn merge_auth_config<'a>(auth: &'a Option<Auth>, auth1: &'a Option<Auth>) -> &'a crate::Result<Auth> {
```

jszwedko: I personally would find that clearer as a user of this function.

src/sinks/util/uri.rs
juchiast: Thanks! I replaced them with `expect`.

src/http.rs
juchiast: Thanks! I changed the signature to
```rust
fn merge_auth_config(auth: &Option<Auth>, auth1: &Option<Auth>) -> crate::Result<Option<Auth>>
```

Needed to return Option<Auth> because both can be `None`. I don't think returning a reference could work because of the way the function is used, we need an owned value:
```rust
authorization = Auth::merge_auth_config(&authorization, &uri.auth)?;
```

bruceg: I think he meant something like this, that could be used the same way without requiring any cloning:
```rust
fn merge_auth_config<'a>(auth1: &'a Option<Auth>, auth2: &'a Option<Auth>) -> crate::Result<&'a Option<Auth>>
```
Having said that, AFAICT this is only called during sink build, so the extra clone isn't huge.

src/sinks/util/uri.rs
```diff
@@ -1,18 +1,26 @@
...
+use std::str::FromStr;
 
 /// A wrapper for `http::Uri` that implements the serde traits.
+/// Authorization credentials, if exist, will be removed from the URI and stored in `auth`.
+/// For example: "http://user:password@example.com".
 #[derive(Default, Debug, Clone)]
-pub struct UriSerde(Uri);
+pub struct UriSerde {
+    pub uri: Uri,
+    #[must_use]
```
jszwedko, line 17: This is nifty. I hadn't seen that macro before.

jszwedko (approved):
Thanks @juchiast !

src/http.rs
```diff
@@ -184,46 +183,23 @@ impl<B> fmt::Debug for HttpClient<B> {
     }
 }
 
-#[derive(Deserialize, Serialize, Clone, Debug)]
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq, Eq)]
 #[serde(deny_unknown_fields, rename_all = "snake_case", tag = "strategy")]
 pub enum Auth {
     Basic { user: String, password: String },
     Bearer { token: String },
 }
```
bruceg, line 191: It occurs to me that almost everywhere we use `enum Auth`, it is as `Option<Auth>`. Am I seeing that right? I appreciate the composability, but it prevents us from writing `fn merge(self: &Option<Self>). Would it make more sense for `None` to be one of the cases here, so that `fn merge_auth_config` could become something like the following.
```rust
impl Auth {
    fn or<'a>(&'a self, other: &'a Auth) -> &'a Self {
        match self {
            None => other,
            _ => self,
        }
    }
}
```
An alternative path would be to create a newtype `MaybeAuth(Option<Auth>)` and then you could write methods on that type.

bruceg (commented):
Looks better. I left a question and some suggestions for killing `mut`.

src/sinks/http.rs
```diff
@@ -119,12 +120,20 @@ impl SinkConfig for HttpSinkConfig {
...
+            Some(healthcheck_uri) => {
+                healthcheck(healthcheck_uri, self.auth.clone(), client.clone()).boxed()
+            }
+            None => future::ok(()).boxed(),
+        };
+
         let mut config = self.clone();
-        config.uri = build_uri(config.uri.clone()).into();
+        config.auth = Auth::merge_auth_config(&config.auth, &config.uri.auth)?;
+        config.uri.uri = build_uri(&config.uri.uri);
```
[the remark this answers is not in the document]
juchiast: I added a `with_default_parts` function and used `uri: self.uri.with_default_parts()`.

src/sinks/util/uri.rs
juchiast: Thanks for catching! I tried this to see if it would catch when we forgot to call `merge_auth_config`. I've removed it.

src/http.rs
juchiast: Yes, authentication is all optional at the moment. I don't think we should add `None` to `Auth` because some sink might want to use it as a hard requirement in the future, so `MaybeAuth` would be preferable.

There would be only one method for `MaybeAuth` at the moment, I'm not sure if it is worth introducing the type yet. I will take some time to try it out tomorrow.

juchiast: @bruceg I'm using a `MaybeAuth` at the moment, I prefer `choose_one` over `or`, let me know what you think!

bruceg (approved):
Good enough. :shipit:

PR #5391, opened 2020-12-04
 feat(new source): Initial `postgresql_metrics` source implementation

### LINKED ISSUE #3639: New `postgresql_metrics` source
Source implementation for the RFC in #3603.

fanatid:
Closes #3639 

Things for todo before PR is ready:
- [x] Use `included_databases` / `excluded_databases`
- [x] Support SSL
- [x] Handle errors on collect
- [x] Support reconnects
- [x] Investigate required minimal permissions for collecting metrics
- [x] Integration tests
  - [x] Password-based auth
  - [x] Unix socket
  - [x] Custom SSL certs


binarylogic:
@bruceg @jszwedko can we get a review on this please?

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,385 @@
...
+			}
+		}
+		namespace: {
+			description: "The namespace of metrics. Disabled if empty."
+			common:      false
+			required:    false
+			type: string: default: "postgresql"
+		}
+		included_databases: {
+			description: "A list of databases to match against `datname` column for which you want to collect logs from."
```
JeanMertz, line 82:
```suggestion
			description: "A list of databases to match against `datname` column for which you want to collect metrics from."
```

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,385 @@
...
+			common:      false
+			required:    false
+			type: array: {
+				default: null
+				items: type: string: examples: ["postgres", "vector"]
+			}
+		}
+		excluded_databases: {
+			description: "A list of databases to match against `datname` column for which you don't want to collect logs from."
```
JeanMertz, line 91:
```suggestion
			description: "A list of databases to match against `datname` column for which you don't want to collect metrics from."
```

JeanMertz: Also, should this describe what happens if I list the same database for both inclusion and exclusion?

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,385 @@
...
+		}
+	}
+
+	how_it_works: {
+		privileges: {
+			title: "Required Privileges"
+			body: """
+				PostgreSQL Metrics component collect metrics by making queries
+				to PostgreSQL Server. You need to be sure that provided user
+				allowed to make select queries for statistics views:
```
JeanMertz, line 125: > PostgreSQL Metrics component collects metrics by making queries to the configured PostgreSQL server. Ensure the configured user is allowed to make the following select queries:

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,385 @@
...
+		_postgresql_metrics_tags: {
+			endpoint: {
+				description: "PostgreSQL endpoint."
+				required:    true
+				examples: ["postgresql:///postgres?host=localhost&port=5432"]
+			}
+			host: {
+				description: "The hostname of the PostgreSQL server."
+				required:    true
+				examples: [_values.local_host]
```
JeanMertz, line 151: Is `_values.local_host` correct here? Shouldn't this just be `"localhost"`?

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,385 @@
...
+		_postgresql_metrics_tags_with_db: _postgresql_metrics_tags & {
+			type: {
+				description: "Database name."
+				required:    true
+				examples: ["postgres"]
+			}
+		}
+
+		up: {
+			description:       "If the PostgreSQL server is up or not."
```
JeanMertz, line 163:
```suggestion
			description:       "Whether the PostgreSQL server is up or not."
```

src/sources/mongodb_metrics/mod.rs
```diff
@@ -69,8 +69,6 @@ enum BuildError {
     CommandBuildInfoMongoError { source: MongoError },
     #[snafu(display("failed to parse `buildInfo` response: {:?}", source))]
     CommandBuildInfoParseError { source: bson::de::Error },
-    #[snafu(display("only `Mongod` supported right now, current: `{:?}`", node_type))]
-    UnsupportedNodeType { node_type: NodeType },
```
JeanMertz, line 73: Was this accidentally removed?

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,840 @@
...
+    };
+}
+
+macro_rules! counter {
+    ($value:expr) => {
+        MetricValue::Counter {
+            value: $value as f64,
+        }
+    };
+}
```
JeanMertz, line 52: Should we extract these into a shared utility module?

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,840 @@
...
+    ($value:expr) => {
+        MetricValue::Gauge {
+            value: $value as f64,
+        }
+    };
+}
+
+#[derive(Debug, Snafu)]
+enum BuildError {
+    #[snafu(display("invalid endpoint: {:?}", source))]
```
JeanMertz, line 64:
```suggestion
    #[snafu(display("invalid endpoint: {}", source))]
```

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,840 @@
...
+                endpoint.clone(),
+                datname_filter.clone(),
+                namespace.clone(),
+                self.tls.clone(),
+            )
+        }))
+        .await?;
+
+        let mut out =
+            out.sink_map_err(|error| error!(message = "Error sending mongodb metrics.", %error));
```
JeanMertz, line 147:
```suggestion
            out.sink_map_err(|error| error!(message = "Error sending postgresql metrics.", %error));
```

JeanMertz, line 147:
```suggestion
            out.sink_map_err(|error| error!(message = "Error sending postgresql metrics.", %error));
```

JeanMertz (commented):
Sweet!

I left a few comments, but didn't do an in-depth review (I'll leave that to @bruceg or @lukesteensen).

It does look like we might want to expand our test coverage though, as we're currently only testing that we got _an_ event back from Postgres, but we have no tests to validate that we return the expected events.

docs/reference/components/sources/postgresql_metrics.cue
fanatid: Yes, this is correct. `_values.local_host` refer to https://github.com/timberio/vector/blob/3b415d160de76e1ee4b600be5b475b922d38c221/docs/reference.cue#L14

src/sources/postgresql_metrics.rs
fanatid: We have an issue for this -- https://github.com/timberio/vector/issues/5229

src/sources/mongodb_metrics/mod.rs
fanatid: Yes, moved to https://github.com/timberio/vector/pull/5831

Makefile
```diff
@@ -763,6 +766,63 @@ ifeq ($(AUTODESPAWN), true)
...
+	/docker-entrypoint.sh postgres \
+	-c ssl=on \
+	-c ssl_key_file=/localhost.key \
+	-c ssl_cert_file=/docker-entrypoint-initdb.d/localhost.crt \
+	-c ssl_ca_file=/docker-entrypoint-initdb.d/localhost.crt"
+else
+	cp $(PWD)/tests/data/localhost.crt $(PWD)/tests/data/postgresql-local-socket-initdb/
+	cp $(PWD)/tests/data/localhost.key $(PWD)/tests/data/postgresql-local-socket-initdb/
+	$(CONTAINER_TOOL) $(CONTAINER_ENCLOSURE) create vector-test-integration-postgresql_metrics
+	$(CONTAINER_TOOL) run -d --$(CONTAINER_ENCLOSURE)=vector-test-integration-postgresql_metrics -p 5432:5432 --name vector_postgresql_metrics \
```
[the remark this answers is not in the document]
fanatid: I'm happy to not write the same commands twice but not sure is possible with podman/docker and we do this for each integration test. Maybe @bruceg has an answer?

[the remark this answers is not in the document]
jamtur01: I am going to fix this, this week.

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,836 @@
...
+
+impl PostgresqlMetrics {
+    async fn new(
+        endpoint: String,
+        datname_filter: DatnameFilter,
+        namespace: Option<String>,
+        tls_config: Option<PostgresqlMetricsTlsConfig>,
+    ) -> Result<Self, BuildError> {
+        let config: Config = endpoint.parse().context(InvalidEndpoint)?;
+        if config.get_hosts().len() > 1 {
```
jszwedko, line 293: Should we check for `config.get_hosts().len() == 0` too?

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,836 @@
...
+    }
+
+    fn get_client(&self) -> &Client {
+        self.client.as_ref().unwrap()
+    }
+
+    async fn collect(&mut self) -> stream::BoxStream<'static, Metric> {
+        let build_client = match self.client {
+            Some(_) => Ok(()),
+            None => self.build_client().await,
```
jszwedko, line 379: I think this would be a little simpler if we created the client on start-up rather than delaying to the first scrape. Then `self.client` doesn't need to be an `Option`. What do you think?

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,836 @@
...
+            .with_context(|| VersionFailed {
+                endpoint: config_to_endpoint(&self.config),
+            })?;
+        debug!(message = "Connected to server.", endpoint = %config_to_endpoint(&self.config), server_version = %version);
+
+        Ok(())
+    }
+
+    fn get_client(&self) -> &Client {
+        self.client.as_ref().unwrap()
```
jszwedko, line 373: I'm not crazy about this `unwrap` here. It would be solved by creating the client at start-up time, but if we choose not to do that, maybe would we maybe instead pass in the client to the functions that need it? Then they could be assured it exists. Right now, we are just depending on the functions being called in the right order.

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,836 @@
...
+        }
+    }
+}
+
+fn row_get_value<'a, T: FromSql<'a>>(row: &'a Row, key: &'static str) -> Result<T, CollectError> {
+    row.try_get::<&str, T>(key)
+        .map_err(|source| CollectError::PostgresGetValue { source, key })
+}
+
+fn config_to_endpoint(config: &Config) -> String {
```
jszwedko, line 666: Just to confirm, this will result in endpoints being stripped of the password, if it was set in the URL?

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,836 @@
...
+        params.push(("keepalives_idle", keepalives_idle.to_string()));
+    }
+
+    // target_session_attrs, ignore default value (TargetSessionAttrs::Any)
+    match config.get_target_session_attrs() {
+        TargetSessionAttrs::Any => {} // default, ignore
+        TargetSessionAttrs::ReadWrite => {
+            params.push(("target_session_attrs", "read-write".to_owned()))
+        }
+        _ => {} // non_exhaustive
```
jszwedko, line 723: Maybe we should add `unknown` for these rather than silently ignoring added enum vales?

jszwedko (commented):
Nice work @fanatid ! I appreciate that you added TLS and unix socket support at the start.

I left a few comments in-line, but nothing major.

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,391 @@
...
+			description: "The interval between scrapes."
+			common:      true
+			required:    false
+			type: uint: {
+				default: 15
+				unit:    "seconds"
+			}
+		}
+		namespace: {
+			description: "The namespace of metrics. Disabled if empty."
```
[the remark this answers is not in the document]
fanatid: This copy-paste from other sources, should I change it everywhere?

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,391 @@
...
+			}
+		}
+	}
+
+	how_it_works: {
+		privileges: {
+			title: "Required Privileges"
+			body: """
+				PostgreSQL Metrics component collects metrics by making queries to the configured PostgreSQL server.
+				Ensure the configured user is allowed to make the following select queries:
```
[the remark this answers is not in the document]
fanatid: :+1: I'll change but replace `tables` with `views`.

src/sources/postgresql_metrics.rs
fanatid: Yes, we need, otherwise, we will get `index out of bound` on the next steps =\

src/sources/postgresql_metrics.rs
fanatid: Yes, password will not be added to resulted endpoint

src/sources/postgresql_metrics.rs
fanatid: I'll add a warning message for `non_exhaustive` enums

src/sources/postgresql_metrics.rs
fanatid: I initially add connect on start-up so there was no `Option` but then moved to scrape because we need support reconnects.

src/sources/postgresql_metrics.rs
fanatid: Removed this function.

docs/reference/components/sources/postgresql_metrics.cue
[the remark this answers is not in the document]
jszwedko: Ah, if it is just copied, that's fine.

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,391 @@
...
+			description: "The namespace of metrics. Disabled if empty."
+			common:      false
+			required:    false
+			type: string: default: "postgresql"
+		}
+		include_databases: {
+			description: """
+				A list of databases to match against `datname` column for which you want to collect metrics from.
+				This can beused in conjunction with [`exclude_databases`](#exclude_databases).
+				In case if `exclude_databases` is not empty databases from there will be removed from `include_databases`.
```
jszwedko, line 85:
```suggestion
				A list of databases to match against the `datname` column for which you want to collect metrics from. If not set, metrics will be collected from all databases. This can be used in conjunction with [`exclude_databases`](#exclude_databases).
```

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,391 @@
...
+				items: type: string: examples: ["postgres", "vector"]
+			}
+		}
+		tls: {
+			common:      false
+			description: "TLS options to connect to the PostgreSQL Server."
+			required:    false
+			type: object: {
+				examples: []
+				options: {
```
jszwedko, line 112: A good future enhancement might be to allow for client side certificates too.

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,391 @@
...
+			common:   false
+			required: false
+			type: array: {
+				default: null
+				items: type: string: examples: ["postgres", "vector"]
+			}
+		}
+		exclude_databases: {
+			description: """
+				A list of databases to match against `datname` column for which you don't want to collect metrics from.
```
jszwedko, line 96:
```suggestion
				A list of databases to match against the `datname` column for which you don't want to collect metrics from.
```

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,851 @@
...
+    time::Instant,
+};
+use tokio::time;
+use tokio_postgres::{
+    config::{ChannelBinding, Host, SslMode, TargetSessionAttrs},
+    types::FromSql,
+    Client, Config, Error as PgError, NoTls, Row,
+};
+
+macro_rules! tags {
```
jszwedko, line 33: Another one! :smile: Linking to https://github.com/timberio/vector/issues/4651 to ensure we swap this one too.

docs/reference/components/sources/postgresql_metrics.cue
```diff
@@ -0,0 +1,391 @@
...
+
+	features: {
+		collect: {
+			checkpoint: enabled: false
+			from: {
+				service: {
+					name:     "PostgreSQL Server"
+					thing:    "an \(name)"
+					url:      urls.postgresql
+					versions: null
```
jszwedko, line 23: Do you know if there are any restrictions on Postgres versions with the queries we are making? It might be prudent to just call out all currently supported PostgreSQL versions here (9-13) assuming our queries work for all of them.

https://www.postgresql.org/support/versioning/

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,851 @@
...
+    }
+
+    async fn pg_stat_database(&self, client: &Client) -> Result<Vec<Row>, PgError> {
+        let mut conditions = "SELECT * FROM pg_stat_database".to_owned();
+        match self {
+            Self::Include {
+                need_null,
+                databases,
+            } => {
+                conditions += " WHERE datname = ANY($1)";
```
jszwedko, line 231: I still think it'd be useful to support wildcards for the include/exclude filters. This would let me write things like:

```
exclude = ["template*"]
```

To exclude all template dbs. I think it could be translated into a `LIKE` query here.

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,851 @@
...
+            return Err(BuildError::MultipleHostsNotSupported {
+                hosts: config.get_hosts().to_owned(),
+            });
+        }
+
+        let mut tags = BTreeMap::new();
+        tags.insert("endpoint".into(), config_to_endpoint(&config));
+        tags.insert(
+            "host".into(),
+            match &config.get_hosts().get(0).expect("host is not missed") {
```
jszwedko, line 308: This isn't bad since we check this invariant earlier, but I might just do something like:

```
&config.get_hosts().get(0).unwrap_or("unknown")
```

To avoid panic'ing.

Alternatively, we could store the config differently such that, when the invariant of 1 host is asserted, we pull it out and put it alongside the config so that other places can depend on it existing.

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,851 @@
...
+            Err(err) => Err(err.to_string()),
+        };
+
+        let (up_value, metrics) = match metrics {
+            Ok(metrics) => (1.0, stream::iter(metrics).boxed()),
+            Err(error) => {
+                self.client = None;
+                emit!(PostgresqlMetricsCollectFailed {
+                    error,
+                    endpoint: self.tags.get("endpoint").expect("should be defined"),
```
jszwedko, line 397: This too I'd like to see stored somewhere where it is guaranteed to exist rather than pulling from the key/value tags.

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,851 @@
...
+
+    async fn collect(&mut self) -> stream::BoxStream<'static, Metric> {
+        let build_client = match self.client {
+            Some(_) => Ok(()),
+            None => self.build_client().await,
+        };
+
+        let metrics = match build_client {
+            Ok(()) => self
+                .collect_metrics(self.client.as_ref().expect("should exists at this point"))
```
jszwedko, line 385: I still feel weird about structuring this this way where it is required to have called `self.build_client()` already. What would you think of something like:

```rust
 // get_client() would handle checking if the client already exists or making a new one if not
let metrics = match self.get_client().await { // Result<Client, _>
            Ok(client) => // do scrape and return those metrics plus an up metric of 1
            Err(err) => // emit error and return an up metric of 0
}
```

where `get_client` is something like:

```rust
async fn get_client(&self) -> Result<Client, _> {
    match self.client {
            Some(_) => Ok(client),
            None => self.build_client().await,
        };
} 
```

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,851 @@
...
+    if let Some(name) = config.get_application_name() {
+        params.push(("application_name", name.to_string()));
+    }
+
+    // ssl_mode, ignore default value (SslMode::Prefer)
+    match config.get_ssl_mode() {
+        SslMode::Disable => params.push(("sslmode", "disable".to_string())),
+        SslMode::Prefer => {} // default, ignore
+        SslMode::Require => params.push(("sslmode", "require".to_string())),
+        _ => {} // non_exhaustive
```
jszwedko, line 693: Maybe we could put the same warning here that we have for `get_target_session_attrs()`?

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,851 @@
...
+        }
+    };
+}
+
+#[derive(Debug, Snafu)]
+enum BuildError {
+    #[snafu(display("invalid endpoint: {}", source))]
+    InvalidEndpoint { source: PgError },
+    #[snafu(display("host missed"))]
+    HostMissed,
```
jszwedko, line 67:
```suggestion
    #[snafu(display("host missing"))]
    HostMissing,
```

jszwedko (commented):
This is looking good! Sorry for the delay in review. I left a few more comments, but I think this is close.

src/sources/postgresql_metrics.rs
fanatid: Yes! Missed this one =(

src/sources/postgresql_metrics.rs
fanatid: The problem with this approach a lifetime.
`build_client` require `&mut self` because of change `self.client` from `None` to `Some`
this means that `get_client` also will require `&mut self`
once we take `client` from `get_client` which required `&mut self` we can not call `collect_metrics` while result of `get_client` is not dropped
It's a little wired and I was surprised myself when tried the approach described above.

src/sources/postgresql_metrics.rs
fanatid: Hm, I think it would be cool, don't you mind if I add this in separate PR? I'd prefer to land this without adding optional enhancements.

jszwedko: :+1: I'm fine with that as a follow-up issue.

docs/reference/components/sources/postgresql_metrics.cue
fanatid: Good question! Checked and adjusted code for `9.6-13`.

src/sources/postgresql_metrics.rs
jszwedko: Ah, yeah, I see what you are saying.

I think one approach would be to split up some of the logic rather than putting it all on `self` (`PostgresqlMetrics`). My suggestion would be to wrap the `tokio_postgres::Client` with our own type that is able to reconnect on query failure. It seems like we only call `query()` so this shouldn't be too much additional implementation. Then we could avoid wrapping the `Client` in an `Option` just to drive that reconnect behavior.

An alternative would be to move `collect_metrics` out from `PostgresqlMetrics` so that we don't need to borrow it again. This would require passing through the dependencies of those functions.

It's just that something seems a bit off with the way it is currently structured in that we have implicit assumptions.

jszwedko (approved):
Talked with @fanatid offline and he is going to do the refactoring of the client reconnect handling I mentioned in https://github.com/timberio/vector/pull/5391/files#r558574724 as a follow-up PR.

src/sources/postgresql_metrics.rs
```diff
@@ -0,0 +1,867 @@
...
+        });
+
+        let event = time::timeout(time::Duration::from_secs(3), recv.next())
+            .await
+            .expect("fetch metrics timeout")
+            .expect("failed to get metrics from a stream");
+        let mut events = vec![event];
+        loop {
+            match time::timeout(time::Duration::from_millis(10), recv.next()).await {
+                Ok(Some(event)) => events.push(event),
```
jszwedko, line 829: Sorry, just seeing taking a closer look at the tests now, could we expand this test to verify some additional things about the metrics?

Maybe:

* Making sure `up` returns 1
* Making sure that some known metric is always returned
* Making sure they are all tagged correctly with any tags that are applied to all
* Testing the negative case to make sure `up` is reported with 0
* Making sure the namespace is applied correctly to all metrics

fanatid:
Summary of requested changes:
- add more asserts to tests: https://github.com/timberio/vector/pull/5391#discussion_r558581861
- refactor client reconnect (remove `expect`): https://github.com/timberio/vector/pull/5391#discussion_r558510249
- support wildcard in filters: https://github.com/timberio/vector/pull/5391#discussion_r558499855

@jszwedko thank you for the review!

binarylogic:
Nice work!

Issue #4882, opened 2020-11-05
Formalize the container start/stop stanzas in the Makefile


bruceg:
The vector `Makefile` contains significant overhead for handling our two container tools. For each integration test setup, there are two stanzas for starting and two for stopping the test containers and network. This is a lot of duplication, increasing the ways for the different parts to get out of sync. Additionally, any changes to the base invocations of the tools (ie adding a dynamic parameter to the `$(CONTAINER_TOOL) run` line) now has to be repeated in hundreds of places.

These could be integrated into a pair of scripts (start and stop) that read a common config file and handle the container tool differences internally. This would turn the invocations in the `Makefile` into simple one-liners.

For example:

```make
.PHONY: start-integration-elasticsearch
start-integration-elasticsearch:
ifeq ($(CONTAINER_TOOL),podman)
	$(CONTAINER_TOOL) $(CONTAINER_ENCLOSURE) create --replace --name vector-test-integration-elasticsearch -p 4571:4571 -p 9200:9200 -p 9300:9300 -p 9201:9200 -p 9301:9300
	$(CONTAINER_TOOL) run -d --$(CONTAINER_ENCLOSURE)=vector-test-integration-elasticsearch --name vector_localstack_es \
	 -e SERVICES=elasticsearch:4571 localstack/localstack@sha256:f21f1fc770ee4bfd5012afdc902154c56b7fb18c14cf672de151b65569c8251e
	$(CONTAINER_TOOL) run -d --$(CONTAINER_ENCLOSURE)=vector-test-integration-elasticsearch \
	 --name vector_elasticsearch -e discovery.type=single-node -e ES_JAVA_OPTS="-Xms400m -Xmx400m" elasticsearch:6.6.2
	$(CONTAINER_TOOL) run -d --$(CONTAINER_ENCLOSURE)=vector-test-integration-elasticsearch \
	 --name vector_elasticsearch-tls -e discovery.type=single-node -e xpack.security.enabled=true \
	 -e xpack.security.http.ssl.enabled=true -e xpack.security.transport.ssl.enabled=true \
	 -e xpack.ssl.certificate=certs/localhost.crt -e xpack.ssl.key=certs/localhost.key \
	 -e ES_JAVA_OPTS="-Xms400m -Xmx400m" \
	 -v $(PWD)/tests/data:/usr/share/elasticsearch/config/certs:ro elasticsearch:6.6.2
else
[...]

PR #5426, opened 2020-12-07
chore(ci): Prototype split Makefile into separate files

jamtur01:

@Hoverbear Any thoughts? I am trying to work out if this is easier or more complex?


bruceg:
By far the largest part of the refactor is `make/test.sh`, and that is no less complex than when it started. IMO refactoring those as described in #4882 would likely cut the size of `Makefile` in half and eliminate the worst of the complexity.

PR #5427, opened 2020-12-07
fix(tests): require features for generate_configfile test

fanatid:
Added in #4819 

Test should not be compiled without required features.


jszwedko (commented):
It looks like you might want to move `use tempfile::tempdir;` into `generate_configfile` to avoid the unused import for other component features..

fanatid:
Thanks @jszwedko! Stupid CI fail from me :facepalm:

jszwedko (commented):
Looks like `cleanup_configfile` might need a `cfg` block too :smile:

jszwedko (approved):
Looks good!

PR #5429, opened 2020-12-07
fix(config): fixing Compression parsing with tests

fanatid:
Moved from https://github.com/timberio/vector/pull/5281#discussion_r536383664

### DIFF
--- a/src/sinks/util/buffer/compression.rs
+++ b/src/sinks/util/buffer/compression.rs
@@ -108,7 +108,7 @@ impl<'de> de::Deserialize<'de> for Compression {
                                     Some(value) if value <= 9 => value as usize,
                                     Some(_) | None => {
                                         return Err(de::Error::invalid_value(
-                                            de::Unexpected::Other(&format!("{}", level)),
+                                            de::Unexpected::Other(&level.to_string()),
                                             &"0, 1, 2, 3, 4, 5, 6, 7, 8 or 9",
                                         ))
                                     }
@@ -127,7 +127,7 @@ impl<'de> de::Deserialize<'de> for Compression {
                                 },
                                 value => {
                                     return Err(de::Error::invalid_type(
-                                        de::Unexpected::Other(&format!("{}", value)),
+                                        de::Unexpected::Other(&value.to_string()),
                                         &"integer or string",
                                     ));
                                 }
--- a/src/sinks/util/buffer/compression.rs
+++ b/src/sinks/util/buffer/compression.rs
@@ -99,10 +99,7 @@ impl<'de> de::Deserialize<'de> for Compression {
                             if level.is_some() {
                                 return Err(de::Error::duplicate_field("level"));
                             }
-                            level = Some(match map.next_value::<usize>() {
-                                Ok(value) => value.to_string(),
-                                Err(_) => map.next_value::<&str>()?.to_string(),
-                            });
+                            level = Some(map.next_value::<&str>()?);
                         }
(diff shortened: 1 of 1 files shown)


src/sinks/util/buffer/compression.rs
```diff
@@ -99,9 +100,34 @@ impl<'de> de::Deserialize<'de> for Compression {
...
                             }
-                            level = Some(match map.next_value::<usize>() {
-                                Ok(value) => value.to_string(),
-                                Err(_) => map.next_value::<&str>()?.to_string(),
+                            level = Some(match map.next_value::<Value>()? {
+                                Value::Number(level) => match level.as_u64() {
+                                    Some(value) if value <= 9 => value as usize,
+                                    Some(_) | None => {
+                                        return Err(de::Error::invalid_value(
+                                            de::Unexpected::Other("0, 1, 2, 3, 4, 5, 6, 7, 8 or 9"),
```
juchiast, line 108: First parameter of `invalid_value` is the invalid value in the input, so we need to pass `level` here. I'm not sure what to do with the `None` case, maybe we convert it to a string if it is too cumbersome to conver `serde_json::Number` into `Unexpected`. 
 
Second parameter is what data was expected, using `self` as the second parameter will print "expected string or map". You can use a reference to a string here, so `&"0, 1, 2, 3, 4, 5, 6, 7, 8 or 9"` would be perfect.

src/sinks/util/buffer/compression.rs
```diff
@@ -99,9 +100,34 @@ impl<'de> de::Deserialize<'de> for Compression {
...
+                                        ))
+                                    }
+                                },
+                                Value::String(level) => match level.as_str() {
+                                    "none" => GZIP_NONE,
+                                    "fast" => GZIP_FAST,
+                                    "default" => GZIP_DEFAULT,
+                                    "best" => GZIP_BEST,
+                                    _ => {
+                                        return Err(de::Error::invalid_value(
```
juchiast, line 119: Same problem here.

src/sinks/util/buffer/compression.rs
```diff
@@ -99,9 +100,34 @@ impl<'de> de::Deserialize<'de> for Compression {
...
+                                            de::Unexpected::Other("none, fast, default or best"),
+                                            &self,
+                                        ))
+                                    }
+                                },
+                                _ => {
+                                    return Err(de::Error::invalid_type(
+                                        de::Unexpected::Other("Number or String"),
```
juchiast, line 127: Same problem here. I'm not sure how to get an appropriate `Unexpected` variant here.

juchiast:
Not introduced in this PR, but error message on `visit_str` is confusing:
```
invalid value: string "asdf", expected string or map at line 1 column 6
```

You can use
```rust
                    _ => Err(de::Error::invalid_value(
                        de::Unexpected::Str(s),
                        &r#""none" or "gzip""#,
                    )),
```

with will print
```
invalid value: string "asdf", expected none or gzip
````

Using `unknown_variant` is also fine, it will print 
```
unknown variant `asdf`, expected `none` or `gzip`
```

juchiast (changes requested):

fanatid:
@juchiast thanks! I changed errors and added tests for them.

juchiast (approved):
Look good! Some `format!("{}", ...)` could be replaced with `to_string` to avoid macro invocation.

fanatid:
@juchiast thank you for review!

PR #5432, opened 2020-12-07
chore(prometheus sink): add internal metrics for prometheus_writer

fanatid:
A possible solution for #4948.

We can not `emit!` event in `Service::call` because the request can fail and data will be resent. As result, we will encode data / emit event a few times what is not correct. My proposal is to add an additional Batch layer which will transform the result of the inner batch so encoding/emit will happen only once.

This is a draft of the code for approval before spending more time for full solution.

### DIFF
--- a/src/sinks/prometheus/remote_write.rs
+++ b/src/sinks/prometheus/remote_write.rs
@@ -88,9 +88,7 @@ impl SinkConfig for RemoteWriteConfig {
                 service,
                 EncodeBatch {
                     inner: MetricBuffer::new(batch.size),
-                    transform: Arc::new(Box::new(move |metrics: Vec<Metric>| {
-                        s2.encode_events(metrics)
-                    })),
+                    transform: Box::new(move |metrics: Vec<Metric>| s2.encode_events(metrics)),
                 },
                 batch.timeout,
                 cx.acker(),
@@ -111,19 +109,20 @@ impl SinkConfig for RemoteWriteConfig {
 
 ///
 use crate::sinks::util::batch::{Batch, BatchError, PushResult};
-use std::sync::Arc;
 
-struct EncodeBatch<B, I, O>
+struct EncodeBatch<B, I, O, F>
 where
     B: Batch<Output = I>,
+    F: Fn(I) -> O + Send + Sync + Clone,
 {
     inner: B,
-    transform: Arc<Box<dyn Fn(I) -> O + Send + Sync>>,
+    transform: F,
 }
 
-impl<B, I, O> Batch for EncodeBatch<B, I, O>
+impl<B, I, O, F> Batch for EncodeBatch<B, I, O, F>
 where
     B: Batch<Output = I>,
+    F: Fn(I) -> O + Send + Sync + Clone,
 {
     type Input = B::Input;
     type Output = O;
@@ -146,7 +145,7 @@ where
     fn fresh(&self) -> Self {
         Self {
             inner: self.inner.fresh(),
-            transform: Arc::clone(&self.transform),
+            transform: self.transform.clone(),
         }
     }
 
--- a/src/sinks/prometheus/remote_write.rs
+++ b/src/sinks/prometheus/remote_write.rs
@@ -81,11 +81,17 @@ impl SinkConfig for RemoteWriteConfig {
             buckets,
             quantiles,
         };
+        let s2 = service.clone();
         let sink = request
             .batch_sink(
                 HttpRetryLogic,
                 service,
-                MetricBuffer::new(batch.size),
+                EncodeBatch {
+                    inner: MetricBuffer::new(batch.size),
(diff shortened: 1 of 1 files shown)


src/sinks/prometheus/remote_write.rs
```diff
@@ -103,6 +109,57 @@ impl SinkConfig for RemoteWriteConfig {
...
+    fn finish(self) -> Self::Output {
+        (self.transform)(self.inner.finish())
+    }
+
+    fn num_items(&self) -> usize {
+        self.inner.num_items()
+    }
+}
+///
+
```
bruceg, line 162: This looks like it should live in `src/sinks/util/batch` instead of here.

src/sinks/prometheus/remote_write.rs
```diff
@@ -103,6 +109,57 @@ impl SinkConfig for RemoteWriteConfig {
...
+///
+use crate::sinks::util::batch::{Batch, BatchError, PushResult};
+use std::sync::Arc;
+
+struct EncodeBatch<B, I, O>
+where
+    B: Batch<Output = I>,
+{
+    inner: B,
+    transform: Arc<Box<dyn Fn(I) -> O + Send + Sync>>,
```
bruceg, line 121: How much does the `Arc` gain us here over just adding `Clone` to the list of traits? Could the thing in the `Box` be an `Arc` to have the same effect without requiring the `Arc`?

fanatid: Indeed, we can use `Clone` instead `Arc` :+1:

src/sinks/prometheus/remote_write.rs
fanatid: Yep, will move to util if an additional batch layer is ok for internal events.

fanatid:
@bruceg what do you think about this approach (outside of the wrong code location)? Is this look like a good solution for internal metrics?

bruceg:
There seem to be two issues here. The title issue is simply adding internal metrics. However, in order for those metrics to be accurate, they shouldn't double-count retries. This looks like an issue that would be common to all the batch sinks, or at least the metric collectors. As such, I think it belongs in the `PartitionBatchSink`, as a step between `batch.finish()` and `call(request)`. This step is separate from collecting the batch and sending the formatted request. I would tend to make it another trait so that `PartitionBatchSink` can match the input and output types, but we could make it always return `Vec<u8>` or `Bytes`.

ping @lukesteensen if you have any additional thoughts

lukesteensen:
I would prefer this not be tied to `Batch::finish`, since that happens before we even attempt to send the data. It would be better if we record the metrics once the request succeeds. This might be a little tricky, but we might be able to solve it in a layer within the batch sink that captures the relevant data (e.g. `Batch::num_items`), does `call(request)`, and then emits the data if it succeeds.

fanatid:
I checked sinks that emit events because not sure how better to implement them here.

Everything simple with `StreamSink`, because we emit events once data is sent. But with `BatchSink` things much complicated, we have nearly such events flow:
before batch -> batch -> BatchSink -> ServiceSink -> ARC / rate limit / retry logic -> sink service

Currently sinks emit events on `before batch` and `sink service` stages:
aws_kinesis_streams => sink service
aws_sqs => sink service
elasticsearch => before batch
http => before batch
splunk_hec => before batch

Both of these stages actually are incorrect. `before batch` is not correct because we are too far from actual data sending and can have some delay before data actually will be sent. `sink service` is not correct because if we will need to retry a request number of events will be duplicated in metrics.

If I understand @lukesteensen suggestion correctly we should emit events in BatchSink where we do ack. I do not see simple ways to do it :disappointed:, maybe add the processed number of bytes to service response (will need to do a lot of changes)? And here one more question for the future: how we will create internal metric when we will need to remove some events from the request if they were sent successfully (aws_sqs sink).

bruceg:
If we want to track the number of bytes and events that have been *successfully* transmitted, we cannot make the measurement at any point before the remote acknowledges (for those services that have acknowledgements, of course). This is, unfortunately, the hardest of the options. Counting the number of bytes and events that have been transmitted at the point of transmission, is also useful for resource utilization counting, but it is a different measurement.

jszwedko:
Indeed, this is a tricky problem.

If it helps, I think we can punt on `processed_bytes_total` for now and only do `processed_events_total`. It seems like we could do this in the manner @lukesteensen suggested.

I think we should open up issues for the discrepancies you identified with the other sinks (maybe one for each set using a given stage incorrectly; so two total). I agree with you that the way they are currently reporting is problematic.

Also, the issue this is referencing, #4948 , is actually about the `prometheus_exporter` sink rather than the `prometheus_writer` sink. I updated the title there; it was filed back when there was only one `prometheus` sink. We should instrument both, of course.

jszwedko:
Closing this as stale. We'll be revisiting this in the next few months.

PR #5434, opened 2020-12-07
enhancement(remap): allow multiple statements in boolean conditional

### LINKED ISSUE #5382: Update Remap grammar to allow expressions to be separated by `;` and newlines
Ref #3736 
Ref #2418 

Implements [rfc 2020-10-29-3736-remap-chain-parsing](https://github.com/timberio/vector/blob/master/rfcs/2020-10-29-3736-remap-chain-parsing.md).

This will enable us to have multiple statements within the condition field of an `if` statement, thus allowing mutation and conditional checking.

It will then be possible to chain multiple parsing strategies with minimal syntactic overhead:

```coffeescript
if ($match = matches(.foo, ...); !is_empty($match)) {
  -- Do something with $match
} else if ($match = matches(.foo, ...); !is_empty($match)) {
  -- Do something with this $match
}
```

FungusHumungus:
Closes #5382 

Implements [rfc 2020-10-29-3736-remap-chain-parsing](https://github.com/timberio/vector/blob/master/rfcs/2020-10-29-3736-remap-chain-parsing.md).

This allows if conditionals to contain multiple statements - the last statement of which has to be a boolean expression.  The expressions can be separated by a semicolon (`;`) or a newline.

This allows us to cleanly chain a number of parsing expressions like the following:

```
if ($match = matches(.message, /^Started (?P<method>[^\s]*) for (?P<remote_addr>[^\s]*)/)
    !is_empty($match)) {
  .method = $match.method
  .remote_addr = $match.remote_addr
  .source = "nginx"
} else if ($match = matches(.message, /^(?P<remote_addr>[^\s]*).*"(?P<method>[^\s]*).*"$/)
           !is_empty($match)) {
  .method = $match.method
  .remote_addr = $match.remote_addr
  .source = "haproxy"
} else {
  .source = "none"
}
```

(note: `matches` hasn't been implemented yet.)

### DIFF
--- a/lib/remap-lang/grammar.pest
+++ b/lib/remap-lang/grammar.pest
@@ -12,7 +12,8 @@ rule_string_inner = _{ SOI ~ string_inner ~ EOI }
 // Statements ------------------------------------------------------------------
 
 assignment   = { target ~ "=" ~ expression }
-if_statement = { "if" ~ boolean_expr ~ block ~ ("else if" ~ boolean_expr ~ block)* ~ ("else" ~ block)? }
+if_statement = { "if" ~ (boolean_expr | conditional_expr) ~ block ~ ("else if" ~ (boolean_expr | conditional_expr ) ~ block)* ~ ("else" ~ block)? }
+
 
 // Primary ---------------------------------------------------------------------
 
@@ -80,6 +81,7 @@ regex_flags =  { ("i" | "x" | "m")* }
 
 // Other ----------------------------------------------------------------------
 
+conditional_expr = { "(" ~ NEWLINE* ~ (expression ~ (NEWLINE+ | ";"))* ~ boolean_expr ~ NEWLINE* ~ ")" }
 ident  = @{ ASCII_ALPHANUMERIC ~ (ASCII_ALPHANUMERIC | "_")* }
 target = _{ variable | path }
 block  =  { "{" ~ NEWLINE* ~ (expression ~ (NEWLINE+ ~ expression)*) ~ NEWLINE* ~ "}" }
--- a/lib/remap-lang/src/error.rs
+++ b/lib/remap-lang/src/error.rs
@@ -79,6 +79,7 @@ impl fmt::Display for Rule {
             call,
             char,
             comparison,
+            conditional_expr,
             EOI,
             equality,
             expression,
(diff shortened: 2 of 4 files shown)


lib/remap-lang/grammar.pest
```diff
@@ -80,6 +81,7 @@ regex_flags =  { ("i" | "x" | "m")* }
 
 // Other ----------------------------------------------------------------------
 
+conditional_expr = { "(" ~ NEWLINE* ~ (expression ~ (NEWLINE+ | ";"))* ~ boolean_expr ~ NEWLINE* ~ ")" }
```
FungusHumungus, line 84: I'm not sure this is quite right. The final `NEWLINE*` in this doesn't seem to be picked up. So an expression like the following:

```
if (
   $$ook="nonk"; .plork == 2
   ) {
```

results in the following error:

```
Dec 07 16:46:50.288 ERROR vector::topology: Configuration error. error=Transform "remap": remap error: parser error:  --> 3:13
  |
3 |             ) {␊
  |             ^---
  |
  = expected assignment, if_statement, not, or block
```

Do you have any ideas why this might be?

JeanMertz (changes requested):
Thanks for this implementation @FungusHumungus.

I was a bit confused when I checked the implementation, because I remember I mentioned I'd prefer if we allow _all_ expressions to end with newlines and semicolons, not just the ones in an if-conditional.

This removes the special-casing we need to do, and allows people to write small single-line programs with multiple expressions as well.

I did see I mentioned it and you replied to it here: https://github.com/timberio/vector/pull/4808#issuecomment-719574769

But I can understand how my comment can be ambiguous.

Regardless, I took a stab at implementing this in https://github.com/timberio/vector/commit/eced0b5c9a15ec2bd0eadc43d795dd999999c60c. I would prefer we use that (or a derivative) implementation, without the need for special-casing.

The one thing missing in that implementation is the boolean check, but we shouldn't be doing that in the parser anyway, but should leverage our new type checking system to do so 
(basically call `type_def` on the resulting `Expr` we get from the conditional, and then see if it resolves to a boolean). But we can do that in a follow-up PR.

FungusHumungus:
Ah right, yeah I misunderstood. But that makes sense. 

Would you be able to add that commit to this branch? I'll then add in the type check.

JeanMertz:
@FungusHumungus I created a PR at https://github.com/timberio/vector/pull/5438. Feel free to push to that branch and update/review/merge the PR yourself 👍.

FungusHumungus:
Closing this in favour of #5438.

PR #5435, opened 2020-12-07
chore(splunk_hec sink): fix test splunk_index_is_interpolated

fanatid:
CI integration test fix (https://github.com/timberio/vector/pull/5418#issuecomment-739990577)

In our docker image used in CI we have only `custom_index`, so the rendered value should be `custom_index` instead `rendered_value`.


jamtur01 (approved):
Ah. Good catch. LGTM

PR #5436, opened 2020-12-08
enhancement(file source): Add encoding charset and line delimiter configuration

anupdhml:
This PR tackles the file source part of https://github.com/timberio/vector/issues/2166, and enables processing of non-utf8 encoded files from vector.

An example run illustrating the new features:

```
$ file /home/anup/sample_utf16 
/home/anup/sample_utf16: Little-endian UTF-16 Unicode text, with CRLF line terminators

$ cat /home/anup/sample_utf16 
1
2
  2.5
3

$ cat vector_config.toml 
[sources.file]
  type = "file"
  include = ["/home/anup/sample_utf16"]
  start_at_beginning = true
  # two new options!
  encoding.charset = "utf-16le"
  line_delimiter = "\r\n"

[sources.file.multiline]
  start_pattern = "^[^\\s]"
  mode = "continue_through"
  condition_pattern = "^[\\s]+"
  timeout_ms = 1000

[sinks.console]
  type = "console"
  inputs = ["file"]
  encoding.codec = "json"

$ cargo run -- -c vector_config.toml
...
{"file":"/home/anup/sample_utf16","host":"virtualbox","message":"1","source_type":"file","timestamp":"2020-12-04T04:21:43.746004544Z"}
{"file":"/home/anup/sample_utf16","host":"virtualbox","message":"2\n  2.5","source_type":"file","timestamp":"2020-12-04T04:21:43.747066995Z"}
{"file":"/home/anup/sample_utf16","host":"virtualbox","message":"3","source_type":"file","timestamp":"2020-12-04T04:21:44.738389017Z"}
```

If the new options are disabled (same as the state before this PR), multiline aggregation logic we have wouldn't work as intended (along with the not-so-ideal json output):
```
{"file":"/home/anup/sample_utf16","host":"virtualbox","message":"��1\u0000\r\u0000","source_type":"file","timestamp":"2020-12-08T05:35:45.317950125
Z"}
{"file":"/home/anup/sample_utf16","host":"virtualbox","message":"\u00002\u0000\r\u0000","source_type":"file","timestamp":"2020-12-08T05:35:45.31905
8162Z"}
{"file":"/home/anup/sample_utf16","host":"virtualbox","message":"\u0000 \u0000 \u00002\u0000.\u00005\u0000\r\u0000","source_type":"file","timestamp
":"2020-12-08T05:35:45.319327416Z"}
{"file":"/home/anup/sample_utf16","host":"virtualbox","message":"\u00003\u0000\r\u0000","source_type":"file","timestamp":"2020-12-08T05:35:46.32438
2976Z"}
```

## Encoding Transcode

The read file bytes are transcoded *from* the specified encoding *to* utf-8 iff the new config option `encoding.charset` is set. Utf-8 is the encoding vector assumes internally for string-like data and thus converting the encoding ensures that processing like in the above example continues to work even for file sources with non utf-8 encodings.

The encoding transcode functionality is not coupled with the file source, since we ultimately want to use it in other sources too. All the core logic for this task is packaged in new module `encoding_transcode`, which also has facility for encoding *to* a specified encoding *from* utf-8. This latter facility is currently used for proper line delimiter handling in file source (see section below), and it should be usable in sinks too later when we implement the new encoding option there.

## Line Delimiter Changes

The PR also introduces a new `line_delimiter` option to file source, that allows customization of the file's line delimiter as string (eg: as `\r\n` for reading CRLF terminated files found in Windows systems).

For line splitting to work when the new `encoding.charset` option is set, the line delimiter (even for default value of `\n`) has to be converted to the specified encoding and depending on the encoding in use (eg: `utf-16le`), it can span multiple bytes. Accordingly, the internal representation for the line delimiter has now changed from `u8` to `Bytes`, which also has the nice side-benefit of enabling line delimiter sequences that span multiple characters. The new file source option `line_delimiter` just exposes this internal flexibility one level up to the user.

## New Dependencies

For the encoding transcode operation, [encoding_rs](https://docs.rs/encoding_rs/0.8.26/encoding_rs/) is used, as suggested in the accompanying issue. For searching the line delimiter bytes on the data read from the file, we now rely on [bstr](https://docs.rs/bstr/0.2.14/bstr/). Both of these are [already](https://github.com/timberio/vector/blob/00431fecde6001c416b2ba878a95ff6243adb0b0/Cargo.lock#L1671-L1677) [present](https://github.com/timberio/vector/blob/00431fecde6001c416b2ba878a95ff6243adb0b0/Cargo.lock#L705-L714) indirectly in vector's dependency chain. As part of the PR, I ended up bumping them to latest versions as well.

One potential issue with the bstr introduction: the library is marked as experimental, but compared to other alternatives I researched for the task (twoway, memmem, jetscii), it seems to be more popular and regularly maintained.

## Other Notes

Some unresolved questions as I was implementing this:

* There's sources like [http](https://vector.dev/docs/reference/sources/http/#encoding) that specify `encoding` in a similar fashion to `encoding.codec` in sinks -- should we consolidate this as `codec` too under the new `encoding` config namespace introduced for `encoding.charset`? This would make the encoding configuration consistent with that in sinks, but maybe there's other ways of organizing these configurations on the source side.

* Should I add benches for file source while transcoding? I think I saw general benchmarks for the source as part of CI and I am hoping that these changes don't cause a major regression there :crossed_fingers: 

Appreciate any/all feedback, thanks!


anupdhml:
Looked into the CI failures and the two left are for the cargo-deny check (same issue is there in master too) and  docs publishing (looks like the failure is because I don't have access to the website repo). Ignoring them for now, but please let me know if there's anything I need to do there!

lukesteensen (commented):
Thank you so much for this @anupdhml! I gave it an initial skim and it looks like excellent work. We'll try to get a good thorough review and testing of this done soon.

docs/reference/components/sources.cue
```diff
@@ -60,6 +60,25 @@ components: sources: [Name=string]: {
...
 
+		if sources[Name].features.encoding.enabled {
+			encoding: {
+				common:      false
+				description: "Configures the encoding specific source behavior."
+				required:    false
+				type: object: options: {
+					charset: {
+						common:      false
+						description: "Encoding of the source messages. Takes one of the encoding [label strings](\(urls.encoding_charset_labels)) defined as part of the [Encoding Standard](\(urls.encoding_standard)). When set, the messages are transcoded from the specified encoding to utf-8, which is the encoding vector assumes internally for string-like data. So it makes sense to rely on this operation only if you need your data to be in utf-8 for further processing. At the time of transcoding, any malformed sequences (that can't be mapped to utf-8) will be replaced with [replacement character](\(urls.unicode_replacement_character)) and warnings will be logged."
```
bruceg, line 71:
```suggestion
						description: "Encoding of the source messages. Takes one of the encoding [label strings](\(urls.encoding_charset_labels)) defined as part of the [Encoding Standard](\(urls.encoding_standard)). When set, the messages are transcoded from the specified encoding to UTF-8, which is the encoding vector assumes internally for string-like data. Enable this transcoding operation if you need your data to be in UTF-8 for further processing. At the time of transcoding, any malformed sequences (that can't be mapped to UTF-8) will be replaced with [replacement character](\(urls.unicode_replacement_character)) and warnings will be logged."
```

docs/reference/components/sources/file.cue
```diff
@@ -160,6 +161,15 @@ components: sources: file: {
 			required:    true
 			type: array: items: type: string: examples: ["\(_directory)/**/*.log"]
 		}
+		line_delimiter: {
+			common:      false
+			description: "String sequence used to separate one file line from another"
```
bruceg, line 166: Would it make more sense to be able to handle any of the standard line delimiters (`\n`, `\r\n`, and `\r`), or are there other patterns that are in use that we should also handle that can't be detected so easily?

src/sources/file.rs
```diff
@@ -200,12 +205,23 @@ pub fn file_source(
...
 
+    let encoding_charset = config.encoding.clone().and_then(|e| e.charset);
+
+    // if file encoding is specified, need to convert the line delimiter (present as utf8)
+    // to the specified encoding, so that delimiter-based line splitting can work properly
+    let line_delimiter_as_bytes = if let Some(e) = encoding_charset {
+        Encoder::new(e).encode_from_utf8(&config.line_delimiter)
+    } else {
+        Bytes::from(config.line_delimiter.clone())
+    };
```
bruceg, line 216:
```suggestion
    let line_delimiter_as_bytes = match encoding_charset {
        Some(e) => Encoder::new(e).encode_from_utf8(&config.line_delimiter),
        None => Bytes::from(config.line_delimiter.clone()),
    };
```

src/sources/file.rs
```diff
@@ -235,9 +251,21 @@ pub fn file_source(
...
+        let rx = rx
+            .map(futures::stream::iter)
+            .flatten()
+            .map(move |(line, src)| {
+                // transcode each line from the file's encoding charset to utf8
+                if let Some(d) = encoding_decoder.as_mut() {
+                    (d.decode_to_utf8(line), src)
+                } else {
+                    (line, src)
+                }
```
bruceg, line 267:
```suggestion
                if match encoding_decoder.as_mut() {
                    Some(d) => (d.decode_to_utf8(line), src),
                    None => (line, src),
                }
```

src/sources/file.rs
```diff
@@ -1493,6 +1536,105 @@ mod tests {
...
+
+        let path = dir.path().join("file");
+        let mut file = File::create(&path).unwrap();
+
+        sleep_500_millis().await; // The files must be observed at their original lengths before writing to them
+
+        writeln!(&mut file, "hello i am a line\r").unwrap();
+        writeln!(&mut file, "and i am too\r").unwrap();
+        writeln!(&mut file, "CRLF is how we end\r").unwrap();
+        writeln!(&mut file, "please treat us well\r").unwrap();
```
bruceg, line 1608: I suggest making the line endings explicit (ie `write!(&mut file, "…\r\n")`). This may behave differently if `writeln!` uses a different line ending on a non-Unix platform.

src/encoding_transcode.rs
```diff
@@ -0,0 +1,432 @@
...
+                    continue;
+                }
+            }
+        }
+
+        if total_had_errors {
+            warn!(
+                message = "Replaced malformed sequences with replacement character while decoding to utf8.",
+                from_encoding = %self.inner.encoding().name()
+            );
```
bruceg, line 73: This should probably be moved into an internal event, so we can also track it as a metric. I also suggest adding a rate limit to avoid spamming the logs.

src/encoding_transcode.rs
```diff
@@ -0,0 +1,432 @@
...
+        // any more use for us, since the source encoding is already pre-identified
+        // as part of decoder initialization.
+        if output
+            .get(..BOM_UTF8_LEN)
+            .map_or(false, |start| start == BOM_UTF8)
+        {
+            trace!(
+                message = "Removing initial BOM bytes from the final output while decoding to utf8.",
+                from_encoding = %self.inner.encoding().name()
+            );
```
bruceg, line 95: Ditto above on internal events.

src/encoding_transcode.rs
```diff
@@ -0,0 +1,432 @@
...
+            self.output.extend_from_slice(&self.buffer[..written]);
+
+            match result {
+                CoderResult::InputEmpty => {
+                    // we have consumed all of the given input so we are done!
+                    break;
+                }
+                CoderResult::OutputFull => {
+                    continue;
+                }
```
bruceg, line 65: The exhaustive match is good. However, this is the end of the loop, so the `continue` appears superfluous:
```suggestion
                CoderResult::OutputFull => ()
```

src/encoding_transcode.rs
```diff
@@ -0,0 +1,432 @@
...
+    pub fn encode_from_utf8(&mut self, input: &str) -> Bytes {
+        // alternate logic if the encoder is for a utf-16 encoding variant
+        if let Some(variant) = self.utf16_encoding {
+            return self.encode_from_utf8_to_utf16(input, &variant);
+        }
+
+        let mut total_read_from_input = 0;
+        let mut total_had_errors = false;
+
+        loop {
```
bruceg, line 161: This looks like almost exactly the same loop as in `decode_to_utf8`. Could these be unified into one implementation used twice?

src/encoding_transcode.rs
```diff
@@ -0,0 +1,432 @@
...
+            d.decode_to_utf8(Bytes::from(test_data_utf16le_123())),
+            Bytes::from("123")
+        );
+        assert_eq!(
+            d.decode_to_utf8(Bytes::from(test_data_utf16le_crlf())),
+            Bytes::from("\r\n")
+        );
+        assert_eq!(
+            d.decode_to_utf8(Bytes::from(test_data_utf16le_vector_devanagari())),
+            Bytes::from("भेक्टर")
```
bruceg, line 251: Vector in Devanagari, cool!

src/encoding_transcode.rs
```diff
@@ -0,0 +1,432 @@
...
+        );
+    }
+
+    #[test]
+    fn test_transcode_symmetry() {
+        let encoding = UTF_16LE;
+        let mut encoder = Encoder::new(encoding);
+        let mut decoder = Decoder::new(encoding);
+
+        let input = "οὐροβόρος";
```
bruceg, line 422: Hah! Nice in joke.

bruceg (changes requested):
Looks pretty good. I left a few recommendations.

docs/reference/components/sources/file.cue
anupdhml: I think the standard line-delimiters will suffice most of the time, but Unicode does specify other characters too there:
https://en.wikipedia.org/wiki/Newline#Unicode
           
I left it as an arbitrary string so that we don't have to pre-specify all the possible characters. Also (but maybe not as important), leaving the configuration open can allow creative uses of the option -- eg: breaking down file lines by some string pattern recurring in the file.

If you prefer to make this a string enum though, I can do the change. fwiw, there are both kind of implementations out there in the wild:

https://www.elastic.co/guide/en/logstash/current/plugins-inputs-file.html#plugins-inputs-file-delimiter
https://github.com/elastic/beats/blob/v7.10.1/filebeat/_meta/config/filebeat.inputs.reference.yml.tmpl#L96-L98

---

A third option is to try and auto-detect line-endings for the file, but that might get complicated if we want to support all the possible characters, and also if the file somehow uses line endings inconsistently, it can throw off the heuristics we use. This strategy would also not work if the file starts out as empty, for example.

src/sources/file.rs
anupdhml: good catch! will change it

src/encoding_transcode.rs
anupdhml: makes sense -- doing the move here and in other places.

src/encoding_transcode.rs
anupdhml: I can put the loop logic in a macro and share it in both places, if you like that. I can also try and build something generic to capture the logic, but the effort/indirection may not be worth it since it will be used only in two places (plus the loop logic here is not that long).

src/encoding_transcode.rs
anupdhml: I don't get to write in it much these days so making the most of it here :smile:

src/encoding_transcode.rs
bruceg: I would discourage the macro approach. I was thinking a function that would be generic on the encode/extend parameters, either as closures or something else.

You are correct the logic isn't long, though, so it's probably not worth a large effort if it turns into something complex.

docs/reference/components/sources/file.cue
bruceg: I would think auto-detection, either when opening the file or just allowing any of a number of line ending patterns, would be the preferred default, making this *mostly* unnecessary. I do agree however that allowing for creative uses here would be a benefit to some users, so this is more a question of whether the default should be a fixed single pattern or something more sophisticated.

@lukesteensen, you've dealt with area this more, would appreciate your thoughts.

anupdhml: My instinct was to let the line-delimiter be user-determined only (either restricted as a fixed enum or open for arbitrary entries), but I can see the appeal of an automatic handling option. Like you suggested, there's two ways we can achieve it:

1. We can auto-default to the first line-delimiter string used by the file. This is the simpler/cheaper option, and will probably cover most of the usecases.
2. We can allow all possible delimiters (either the [unicode list](https://en.wikipedia.org/wiki/Newline#Unicode) or just `\n,\r\n,\r`) and emit lines any time one of the allowed delimiters is seen. This will handle files with inconsistent line-endings too, at the added cost of needing to search all possible patterns for every line-splitting.

If users don't like this sort of "smart" handling (or they know their delimiters beforehand), they will still have the ability to choose a single line-delimiter. One thing worth noting though: if we change the default away from the current `\n` to the new auto option, it would be a breaking change and we would need to communicate the changes as such.

Also, if we are adding the auto option (either as a default or one of the choices), we need to decide the strategy to pick there from the above two. We can also expose both the variants as separate options from configuration side, though that might be one too many option.

ktff (commented):
Looks good, and regarding 
> should we consolidate this as codec too under the new encoding config namespace introduced for encoding.charset?

this PR is large as it is so let's leave that for later.

src/sources/util/encoding_config.rs
```diff
@@ -0,0 +1,8 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct EncodingConfig {
+    #[serde(default)]
+    pub charset: Option<&'static encoding_rs::Encoding>,
```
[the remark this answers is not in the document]
anupdhml: In the current form of `EncodingConfig`, it's not strictly needed (since it's the only item here and using the struct implies specifying a value to `charset`) but if we place the source-level encoding configuration as `codec` here (eg: from the http source), `charset` would need to be optional, since it does not need to take a value always. I was also trying to keep this in sync with the [sink encoding configuration](https://github.com/timberio/vector/blob/3a48a0a47dd1e1b59113041a96cdced29937cc19/src/sinks/util/encoding/config.rs#L23-L34) -- when `charset` support is added there, it will be as an `Option`.

If you would prefer to turn this into an `Option` later when the struct gets other items, I can remove it from here for now.

docs/reference/components/sources.cue
```diff
@@ -60,6 +60,25 @@ components: sources: [Name=string]: {
 			}
 		}
 
+		if sources[Name].features.encoding.enabled {
```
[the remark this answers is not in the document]
anupdhml: thanks for the tip! Added the check and I like how changes to all the other source cue files are gone now :smile_cat:

docs/reference/components/sources/file.cue
anupdhml: I am fine with leaving the auto-detection for later. If we are all in agreement about this, I can file the issue for it and we can continue the conversation there.

When we do implement it, perhaps a keyword like `auto` can stand in for that option, but we may need to rethink it in context of arbitrary strings allowed here now (or we can just treat `auto` string as special). Is there anything we should do as part of this PR, to avoid potential pain/config breakage if we are to implement the auto-detection later? There might be other configuration modes for supporting it and still enabling arbitrary strings, that I am not be seeing. Or we can drop the support for arbitrary strings and restrict the configuration to fixed string enums standing in for various standard separators, which will simplify things.

We can also choose to remove the `line_delimiter` configuration altogether from this PR and just rely on the earlier hardcoded default of `\n` for now. However, the immediate end-goal of this PR was to read non-utf8 encoded log files on Windows systems and they are mostly coupled with `\r\n` as line separator so would like to ideally handle the splitting for it too here (especially since support for wider delimiter sequence would still be there internally, for line-splitting to work consistently across encodings).

src/sources/util/encoding_config.rs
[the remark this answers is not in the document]
ktff: > If you would prefer to turn this into an Option later when the struct gets other items, I can remove it from here for now.

Yep, we should prefer this, a bit of complexity less for now.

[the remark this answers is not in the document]
anupdhml: Fair enough, removed the `Option` here as discussed.

ktff (approved):
Looks good. 

cc. @bruceg

Hoverbear:
(Removed some reviewers so this can merge sooner)

jamtur01:
@anupdhml Can I get you to please do a quick update to fix the merge conflicts and @bruceg will take a look next week for a final review! Thanks!

anupdhml:
> @anupdhml Can I get you to please do a quick update to fix the merge conflicts and @bruceg will take a look next week for a final review! Thanks!

Sorry for the late reply! I was on a break and couldn't get to it earlier, but this should be all set now. Will keep an eye out for CI failures (if any).

CONTRIBUTING.md
```diff
@@ -95,11 +95,11 @@ expanding into more specifics.
 2. One approved, [fork the Vector repository][urls.fork_repo] in your own
    Github account.
 3. [Create a new Git branch][urls.create_branch].
-4. Review the Vector [workflow](#workflow) and [development](#development).
+4. Review the Vector [change control](#change-control) and [development](#development).
```
bruceg, line 98:
```suggestion
4. Review the Vector [change control](#change-control) and [development](#development) work flows.
```

MOZGIII:
I've activated k8s E2E tests for this PR, just for the extra certainty as it affects Kubernetes integration.

The change to the `kubernetes_logs` looks fine to me.

src/encoding_transcode.rs
```diff
@@ -0,0 +1,422 @@
...
+pub struct Decoder {
+    buffer: [u8; BUFFER_SIZE],
+    output: BytesMut,
+    inner: encoding_rs::Decoder,
+}
+
+impl Decoder {
+    pub fn new(encoding: &'static Encoding) -> Self {
+        Self {
+            buffer: [0; BUFFER_SIZE],
```
bruceg, line 24: This buffer does not appear to actually retain any data between calls to `decode_to_utf8`. Could it be a local in that function instead?

src/encoding_transcode.rs
```diff
@@ -0,0 +1,422 @@
...
+            output.slice(BOM_UTF8_LEN..)
+        } else {
+            output
+        }
+    }
+}
+
+/// Helps transcoding to the specified encoding from utf8
+pub struct Encoder {
+    buffer: [u8; BUFFER_SIZE],
```
bruceg, line 101: Ditto this buffer.

bruceg (commented):
Looks pretty good, I had one question about the `buffer` in the codec state.

src/encoding_transcode.rs
anupdhml: Moving it to be local in the function will work but this was done to save on the buffer initialization/allocation when `decode_to_utf8` is called repeatedly (like in the file source, which initializes the `Decoder` once and calls `decode_to_utf8` for each line).

bruceg: :+1: That makes sense. You might be able to get an uninitialized buffer with `MaybeUninit`, but probably only by adding `unsafe` blocks.

bruceg (approved):

anupdhml:
The only CI failure repeating here is due to credentials issue, when uploading the linux benchmark results to S3: https://github.com/timberio/vector/pull/5436/checks?check_run_id=1736646882

```
Run scripts/upload-benchmarks-s3.sh
upload failed: - to s3://test-artifacts.vector.dev/benches/benches_version=1/environment_version=1/branch= /machine=x86_64/operating_system=linux/libc=gnu/threads=3/year=2021/month=01/day=20/rev_count=1/sha=92c0aad87a816d14bc32bc2201492b9563aa7307/timestamp=1611173849/raw.csv Unable to locate credentials
Error: Process completed with exit code 1.
```

Since this is an external PR, makes sense for it to not have access to the credentials. Should we ignore the check, or is there a way to run it successfully here?

binarylogic:
@anupdhml we can ignore it. Thank you for your contribution!

anupdhml:
> @anupdhml we can ignore it. Thank you for your contribution!

np! happy to contribute to a project like vector :smiley: 

Thanks all for the reviews and merge!

PR #5438, opened 2020-12-08
enhancement(remap): allow multiple statements in boolean conditional

### LINKED ISSUE #5382: Update Remap grammar to allow expressions to be separated by `;` and newlines

JeanMertz:
Alternative implementation to #5434.

Closes #5382 

Implements [rfc 2020-10-29-3736-remap-chain-parsing](https://github.com/timberio/vector/blob/master/rfcs/2020-10-29-3736-remap-chain-parsing.md).

This allows if conditionals to contain multiple statements - the last statement of which has to be a boolean expression.  The expressions can be separated by a semicolon (`;`) or a newline.

This allows us to cleanly chain a number of parsing expressions like the following:

```
if ($match = matches(.message, /^Started (?P<method>[^\s]*) for (?P<remote_addr>[^\s]*)/)
    !is_empty($match)) {
  .method = $match.method
  .remote_addr = $match.remote_addr
  .source = "nginx"
} else if ($match = matches(.message, /^(?P<remote_addr>[^\s]*).*"(?P<method>[^\s]*).*"$/)
           !is_empty($match)) {
  .method = $match.method
  .remote_addr = $match.remote_addr
  .source = "haproxy"
} else {
  .source = "none"
}
```

(note: `matches` hasn't been implemented yet.)


FungusHumungus:
It should be noted this does not perform a compile time Boolean type on the If conditional. It will ensure the conditional is a boolean at runtime however.

It isn't clear to me how we can easily do this, as the If statement is instantiated directly within the Parser, so doesn't have a compile step.

JeanMertz:
> the If statement is instantiated directly within the Parser, so doesn't have a compile step.

The parser has (mutable) access to the compilation state.

We have other built-in expressions (such as the assignment expression) that take a mutable reference to this state on initialization. You should be able to do something similar for the if-statement expression.

PR #5439, opened 2020-12-08
fix(elasticsearch sink): Add Accept-Encoding header when compression is enabled

smlgbl:
So that responses from ES are compressed as well (traffic and costs are saved) we need to set the "Accept-Encoding" header as well.


fanatid:
It's not correct, see https://github.com/timberio/vector/pull/5355#issuecomment-740645178

PR #5444, opened 2020-12-08
chore(remap): improve test helpers

JeanMertz:
High-level changes:

* Allow function tests to fail individually (this was introduced in an earlier PR, but now applied to the `flatten` function tests as wel)
* Make it easier to work with commonly used remap types in tests (`Value`, `Array`, and soon `Map`)

Specific changes:

* add `value` macro to quickly create `remap::Value` types in tests
* change `array` macro to create an `expression::Array` (again often used in tests)
* remove `map` macro (will be reintroduced in #5279 to create `expression::Map`)
* add a few more `{Try}From` impls to go to/from different wrapper types
* update `flatten` function tests to use new set-up

### DIFF
--- a/lib/remap-lang/src/expression.rs
+++ b/lib/remap-lang/src/expression.rs
@@ -158,7 +158,12 @@ expression_dispatch![
 
 impl<T: Into<Value>> From<T> for Expr {
     fn from(value: T) -> Self {
-        Literal::from(value.into()).into()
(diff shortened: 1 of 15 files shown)


FungusHumungus (approved):
There's a few ci errors, but this is a work of art! The `test_function` macro is beautiful!

PR #5450, opened 2020-12-09
feat(remap): add initial "trl" CLI

JeanMertz:
This adds a simple `trl` (Timber Remap Language) CLI to help debug programs:

```
echo '{ "foo": "bar" }` | trl .foo
bar
```

_(note: it should print `"bar"`, but that's an improvement that will be made to TRL itself, in a separate PR)_

To run this from the main Vector project, use:

```
echo '{ "foo": "bar" }' | cargo run --no-default-features --package remap-cli .foo
bar
```

There is still much to do here (I added a `TODO.md` since I didn't want to necessarily create issues in such an early state), but it's a start.

Here's the current `-h` output:

```
TRL 0.1.0
Timber Remap Language CLI

USAGE:
    trl [FLAGS] [OPTIONS] [PROGRAM]

FLAGS:
    -h, --help            Prints help information
    -o, --print-object    Print the (modified) object, instead of the result of the final expression
    -V, --version         Prints version information

OPTIONS:
    -i, --input <input-file>        File containing the object to manipulate, leave empty to use stdin
    -p, --program <program-file>    File containing the program to execute, can be used instead of "PROGRAM"

ARGS:
    <PROGRAM>    Program to execute
```

### DIFF
--- a/lib/remap-cli/src/main.rs
+++ b/lib/remap-cli/src/main.rs
@@ -1,0 +1,120 @@
+use remap::{state, Object, Program, Runtime, Value};
+use std::collections::BTreeMap;
+use std::fs::File;
+use std::io::{self, Read};
+use std::iter::IntoIterator;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "TRL", about = "Timber Remap Language CLI")]
+struct TRL {
+    /// Program to execute
+    ///
+    /// For example, ".foo = true" will set the object's `foo` field to `true`.
+    #[structopt(name = "PROGRAM")]
+    program: Option<String>,
+
+    /// File containing the object to manipulate, leave empty to use stdin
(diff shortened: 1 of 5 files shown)


lib/remap-cli/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+    print_object: bool,
+}
+
+#[derive(thiserror::Error, Debug)]
+enum Error {
+    #[error("io error")]
+    Io(#[from] io::Error),
+
+    #[error("remap error: {0}")]
+    Remap(#[from] remap::RemapError),
```
JeanMertz, line 41: TRL error improvements are on my list, at which time this will simply be `remap::Error` and doesn't require the `{0}` on the line above — don't ask why this is currently needed, it's complicated (read: I should've RTFM) 😞.

StephenWakely (approved):
Yes. Great idea!

PR #5452, opened 2020-12-09
fix(syslog source): priority now defaults to 13, so is never missing

### LINKED ISSUE #5441: Support parsing syslog RFC 3164 messages lacking a priority
Discovered the lack of this via @spencergilbert in discord: https://discord.com/channels/742820443487993987/746070591097798688/785914040458412073

### Current Vector Version

```
vector 0.12.0 (gb1d0189 x86_64-unknown-linux-musl 2020-12-08)
```

### Use-cases

Attempting to parse syslog RFC 3164 messages from a file where they lack a PRI value.

### Attempted Solutions

```
[sources.stdin]
  type = "stdin"

[transforms.remap]
  type = "remap"
  inputs = ["stdin"]
  source = """
    .parsed = parse_syslog( .message)
  """

[sinks.console]
  type = "console"
  inputs = ["remap"]
  encoding.codec = "json"
```

This fails to parse syslog messages being read from a file where they lack the priority specifier:

```
$ echo 'Dec 8 16:48:11 node-01 auditd[38707]: Audit daemon rotating log files' | /tmp/vector-x86_64-unknown-linux-musl/bin/vector --config /tmp/tmp.toml                                
Dec 08 17:50:12.690  INFO vector::app: Log level is enabled. level="vector=info,codec=info,file_source=info,tower_limit=trace,rdkafka=info"                                                    
Dec 08 17:50:12.690  INFO vector::app: Loading configs. path=[("/tmp/tmp.toml", None)]                                                                                                         
Dec 08 17:50:12.692  INFO vector::sources::stdin: Capturing STDIN.                                                                                                                             
Dec 08 17:50:12.692  INFO vector::topology: Running healthchecks.                                                                                                                              
Dec 08 17:50:12.692  INFO vector::topology: Starting source. name="stdin"             
Dec 08 17:50:12.692  INFO vector::topology: Starting transform. name="remap"
Dec 08 17:50:12.692  INFO vector::topology: Starting sink. name="console"
Dec 08 17:50:12.692  INFO vector: Vector has started. version="0.12.0" git_version="v0.11.0-40-gb1d0189" released="Tue, 08 Dec 2020 04:45:05 +0000" arch="x86_64"
Dec 08 17:50:12.692  INFO vector::app: API is disabled, enable by setting `api.enabled` to `true` and use commands like `vector top`.
Dec 08 17:50:12.692  INFO vector::topology::builder: Healthcheck: Passed.
Dec 08 17:50:12.694  INFO vector::shutdown: All sources have finished.   
Dec 08 17:50:12.694  INFO source{component_kind="source" component_name=stdin component_type=stdin}: vector::sources::stdin: Finished sending.                   
Dec 08 17:50:12.694  INFO vector: Vector has stopped.                                                                                                                                          
{"host":"a-8cktst39ij0a","message":"Dec 8 16:48:11 node-01 auditd[38707]: Audit daemon rotating log files","parsed":{"message":"Dec 8 16:48:11 node-01 auditd[38707]: Audit daemon rotating log
 files"},"source_type":"stdin","timestamp":"2020-12-08T17:50:12.694308282Z"}            
```

### Proposal

RFC 3164 does [state that the messages can lack a priority](https://tools.ietf.org/html/rfc3164#section-4.3.3) (where it should be defaulted to `<13>`).

We may want to consider this for RFC 5424 messages as well, even though the spec requires them, as I believe it is common for messages to be lacking this field when written to a file by rsyslog or another daemon.

This may require a change to the upstream `syslog_loose` crate.

### References

https://tools.ietf.org/html/rfc3164#section-4.3.3

FungusHumungus:
Closes #5441 

Updates `syslog_loose` to v0.6.0 which handles missing priority fields. Now if that field is missing or invalid it defaults to 13 - `LOG_USER` and `SEV_NOTICE`. As a result, these fields are now always present in every message.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6301,9 +6301,9 @@ dependencies = [
 
 [[package]]
 name = "syslog_loose"
-version = "0.5.0"
+version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bc84750f2634660415240f7ab6c67022b79d3eccdd9c8f6b969d95e5e409a2c5"
+checksum = "59105a46f64e6e0c60a07061ffd07ef42d137735b5d9796572fe969b21342772"
 dependencies = [
  "chrono",
  "nom 5.1.2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -142,7 +142,7 @@ http = "0.2"
 typetag = "0.1"
 toml = "0.5.7"
 syslog = "5"
-syslog_loose = { version = "0.5.0" }
+syslog_loose = { version = "0.7.0" }
 derive_is_enum_variant = "0.1.1"
 leveldb = { version = "0.8", optional = true, default-features = false }
 db-key = "0.0.5"
--- a/lib/remap-functions/src/parse_syslog.rs
+++ b/lib/remap-functions/src/parse_syslog.rs
@@ -183,6 +183,18 @@ mod tests {
                     r#"<133>Jun 13 16:33:35 haproxy[73411]: Proxy sticky-servers started."#,
                 ))),
             ),
+            (
+                map![],
+                Ok(map![
+                        "message": "I am missing a pri.",
+                        "timestamp": DateTime::<Utc>::from(chrono::Local.ymd(Utc::now().year(), 6, 13).and_hms_milli(16, 33, 35, 0)),
+                        "appname": "haproxy",
+                        "procid": 73411
+                ]),
+                ParseSyslogFn::new(Box::new(Literal::from(
+                    r#"Jun 13 16:33:35 haproxy[73411]: I am missing a pri."#,
+                ))),
+            ),
         ];
 
         let mut state = state::Program::default();


jszwedko (commented):
Thanks @FungusHumungus !

Having thought about this a bit more, I am wondering if we might want to adjust the implementation slightly to be:

* For the `syslog` source where it accepts syslog packets: default PRI to `<13>` for RFC 3164 messages if missing as per the RFC; reject RFC 5424 messages lacking a priority as this doesn't seem to be allowed for by the RFC
* When using `parse_syslog` in `remap`, allow priority to be missing (perhaps conditionally?), but don't default it, instead having `None` values for `severity` and `facility`

Partially for my clarification, the issue is, that when rsyslog (or another daemon) writes messages to a file, it often will write them as RFC-style syslog messages just without the priority (sometimes the severity/facility is instead encoded by the filename; like `auth.log`). We'd like to support parsing these messages. I think we can more strictly parse syslog messages coming into the `syslog` source though.

It could be the case that the user is reading raw syslog messages from another source though (like kafka) that will result in `.message` encoding the priority and so `parse_syslog` in `remap` should handle that correctly.

I'm curious what you think.

FungusHumungus:
>     * When using `parse_syslog` in `remap`, allow priority to be missing (perhaps conditionally?), but don't default it, instead having `None` values for `severity` and `facility`

Yes, I agree. I think this is the way forward. The goal of `syslog_loose` is to extract as much useful information out of as many messages as it can rather than strictly stick to the standards. Choosing the defaults shouldn't be the role of the parser.

I don't think it should default to 13 for 3164.. If someone needs this to happen, it should be possible to do in Remap.

I'll make these changes.

FungusHumungus:
Ok, have updated so that if the PRI field is missing it will still parse the rest of the field, but will not default the values.

Most of the changes were in [`syslog_loose`](https://github.com/FungusHumungus/syslog-loose). The changes here are just to update the `syslog_loose` version and add a new test.

I'm pretty sure the CI failures are not related to these changes.

jszwedko (approved):
Looks good, thank you!

PR #5469, opened 2020-12-09
feat(remap): add `ok` error handling function

JeanMertz:
This one  requires a bit of context.

From the start of Remap/TRL, we've been discussing error handling in the language. This is because right now, any error in TRL results in the program terminating. This is fine for small scripts, but there are real-world use-cases (as reported on Discord) that require some way to handle errors.

One such example is this simple program:

```rust
```

This is a common situation: take a string value, and try to parse that string as a JSON value, then overwrite the string with the new JSON type.

It's very common to have some events that can be parsed as JSON, while others can't, and so this program will not always work as-is.

How should we handle this case?

---

One or two months ago by now, I proposed adding `try/catch` blocks to the language to allow catching errors. Most people weren't fans of this approach, and so we discussed many alternatives.

At some point, an `ok_or` functional approach was proposed. This didn't get as much pushback, but the consensus was we could do better.

This is when (after many lengthy discussions) RFC #4840 proposed adding "infallible functions". Unfortunately, there were still flaws in this proposal, and it somewhat died down.

After this, @lukesteensen and I had a lengthy discussion on Discord to further experiment with potential solutions.

In the end, we came close to something similar as rust's `let Ok(_) = ...` pattern that leverages our existing if-statement control-flow to allow acting on successful function calls.

That's where this PR comes in.

---

This PR adds _three_ new functions:

* `ok_or(Expr, Expr) :: Result<Value>`
* `ok(Expr) :: bool`
* `err(Expr) :: bool`

To demonstrate use-cases for these functions, we'll circle back to the original example:

```rust
```

There are three patterns you might want to do here:

1. **assign parsed JSON value, or keep original message**

   This is where `ok_or` comes in:

   ```rust
   .message = ok_or(parse_json(.message), .message)
   ```

   Or, alternatively:

   ```rust
   .message = ok_or(parse_json(.message), { "error": "unable to parse json", "source": .message })
   ```

2. **try to parse JSON, and use the resulting value**

   This uses the `ok` function:

   ```rust
   if ok($json = parse_json(.message)) {
   	.my_field = $json.my_field
   }
   ```

3. **try to assign parsed JSON, or log + assign error**

   ```rust
   if err(.message = parse_json(.message)) {
   	log("failed!")
   	.error.code = 42
   	.error.message = "failed to parse message field"
   }
   ```

---

I've put these three functions up because I think the carry their weight, were easy to add, and unblock error handling. With #5459 we could mark all or some of these functions as experimental.

Of the three, `ok_or` is probably(?) the least favourite for some. I like how it makes that specific use-case more terse to write, but I can also see us starting with `ok` and `err` (or maybe even only `ok`), and go from there. Still though, I'd love to get all three merged as experimental functions that allow us to gather real-word data on their usage.

I'm not confident this is the final chapter on error handling, but I don't feel bad about merging these as is.

If others are, I'm happy to continue discussing this further, don't feel pressured to accept this because I created a PR. Creating an RFC would have taken the same amount of time, so I'm happy to close this PR and consider it a valid but unsuccessful experiment.


binarylogic (approved):
I like it, I think it's a simple first step. Pairing it with https://github.com/timberio/vector/issues/5474 would make it much more obvious when these functions should be used.

FungusHumungus:
Would it hurt to call it `error`? I don't think we need to save bytes that much, so would prefer to avoid abbreviations.

StephenWakely (approved):
Yes, I like the simplicity of this approach.

bruceg:
Without looking at the PR, I have a surface comment on the approaches:

> Of the three, `ok_or` is probably(?) the least favourite for some. I like how it makes that specific use-case more terse to write, but I can also see us starting with `ok` and `err` (or maybe even only `ok`), and go from there.

As much as I like the functional solutions within Rust, I think the `ok_or` function makes less sense for TRL. I do like the simplicity of the `ok` approach.

An `err` function (or `error` as @FungusHumungus suggested) really only adds value if it can provide the error value somehow. That is, to be able to show what kind of error happened rather than just that it happened. The latter is just `!ok(…)`.

I'm also wondering if this isn't hiding a control-flow operation in a function form, similar to the discussion on the `walk` operation. Consider `err(.message = parse_json(.message))`: Does the assignment happen if `parse_json` errors? If so, we have overwritten `.message`. If not, we have turned what is normally an unconditional assignment into a conditional one, which is a flow control change.

The `ok` function doesn't have the same problem, as the temporary `$json` can have junk on failure without causing data loss.

> Still though, I'd love to get all three merged as experimental functions that allow us to gather real-word data on their usage.

The problem I foresee with this approach is that of how we might *drop* one of these experiments that we think isn't working but that users have started to rely on.

lukesteensen (approved):

JeanMertz:
> As much as I like the functional solutions within Rust, I think the `ok_or` function makes less sense for TRL. I do like the simplicity of the `ok` approach.

Fair point. I do think we need _some way_ to quickly go from "I have this fallible expression" to "if it fails, just default to this other expression".

> An `err` function (or `error` as @FungusHumungus suggested) really only adds value if it can provide the error value somehow. That is, to be able to show what kind of error happened rather than just that it happened. The latter is just `!ok(…)`.

Another fair point. I'll go into more depth on my alternative proposal down below.

> I'm also wondering if this isn't hiding a control-flow operation in a function form, similar to the discussion on the `walk` operation. 

Another good point, and one I tend to agree with, but unfortunately, we've got somewhat stuck on that discussion.

---

After some more experimenting, I've come to an alternative approach that borrows heavily from Rust, and considers error handling as control-flow:

Given this field:

```rust
.bar = "{ \"foo\": true }"
```

1. If I want to call a fallible function safely:

	```rust
	// ok result handling, with else block.
	if ok($int) = to_int(.bar) {
	    .foo = $int + 5;
	} else {
	    .foo = 0;
	}
	```

2. If I want to deal with the error message of a fallible function:

	```rust
	// error result handling
	if err($err) = del(.bar) {
	    log("del error: " + $err)
	}
	```

3. If I want to handle the ok-case, but also log an error:

	```rust
	// try statement
	//
	// Try to parse `.bar`, if it succeeds, assign the value to
	// the `.bar` field, if it fails, assign the error to the
	// `$err` value and log the error.
	//
	// If we want to, we can name this `match` or `switch`, in
	// case we want to broaden this to full pattern matching.
	try parse_json(.bar) {
	    ok(.bar)
	    // could also be written as:
	    // ok($bar) -> .bar = $bar
    	err($error) -> log("parse json error: " + $error)
	}
	```

4. Finally, I want to be able to have a (built-in?) _ok-or_ pattern:

	```rust
	// try coalescence
	//
	// Sometimes you just want to use a fallible expression, and
	// fall back to a default value if the expression fails.
	.foo = parse_json(.bar) ? {}
	```

If we go this route, I'd say we definitely need (1), we probably also want (2), there is a strong but not urgent case for (3), and while (4) improves ergonomics, it doesn't unlock any features that 1, 2 and 3 don't cover. (4) does become more relevant when we start working on https://github.com/timberio/vector/issues/5479.

jszwedko:
How would we use both `ok` and `err` in an `if/else` block?

Something like:

```coffeescript
if ok($int) = to_int(.bar) {
    .foo = $int + 5
} else if err($err) = to_int(.bar) {
    log($err)
    .foo = 0
}
```

?

I can imagine it'll be common to want to handle both the success value and the error value.

I'm personally not a huge fan of using pattern matching for error handling unless we plan to expand it and make it more central to the language (for example by adding `match` statements) rather than just a special case for error handling.

My personal preference is still multiple return values :) This is should be fairly straight-forward for users (especially those unfamiliar with pattern matching constructs) to understand.

Example:

```coffeescript
if $int, $err = to_int(.bar); !$err {
    .foo = $int + 5;
} else {
    .foo = 0
}
```

I realize this error handling pattern is not universally liked (i.e. in the context of Go), but I think for a language like TRL which should remain fairly simple and the scripts fairly short, it is a good fit. I think the pattern starts to fall flat whenever you are building full-fledged applications using it (where things like union types offer some advantages). Again, this is similar to the error handling pattern in Lua (https://www.lua.org/pil/8.4.html) which may offer some advantage in letting people more easily translate their `lua` transforms to `remap`.

If we pursue this approach, we could also introduce `_` to let people ignore errors where desired, and/or `!` function variants to cause the script to fail on errors instead.

JeanMertz:
Alright. Given the feedback, here's what I'll be doing:

1. I'm going to **merge** the `ok` function into master. This will unblock a basic form of error-handling for people. I will **not** merge the `err` and `ok_or` functions, as those are merely convenience functions, and given that we're not entirely sure we'll stick with the `ok` function, it doesn't make sense to merge those in yet (and even if we did, I agree with @bruceg that `err` isn't useful in its current form).

2. I've created a [**pre-RFC** discussion board](https://github.com/timberio/vector/discussions?discussions_q=category%3APre-RFCs) on GitHub. I'll be using that to create a new discussion, outlining the current story of error-handling in TRL, and allowing us (and anyone interested in the topic) to weigh in and discuss the best way forward.

3. Once that discussion settles down, and we've reached a point where we either have consensus, or have to make a choice between two fleshed-out proposals, I'll go ahead and write an RFC for one of those proposals to get in. (there's obviously a chance here it won't be the proposal you like, but that's not always avoidable).

4. If the final error-handling solution doesn't require the `ok` function to exist any more, we'll deprecate it, or remove it if this happens before the 1.0 release.

JeanMertz:
For those reading along in this PR, I thought of a clean alternative built-in solution for which I implemented a PoC at https://github.com/timberio/vector/pull/5593. Will clean that PR up tomorrow, but happy to get some initial thoughts in that PR.

PR #5492, opened 2020-12-11
fix(observability): Emit `processed_events_total` in topology for transforms

### LINKED ISSUE #4454: Unify span context when components are nested

ktff:
Closes #4454 
Replaces #5424, relevant https://github.com/timberio/vector/pull/5424#issuecomment-742089007

To solve double emitting metrics this PR:
  1. Raises emitting of `processed_events_total` for transforms to topology level.
       * This also means all `processed_events_total` emits, for transforms, and their internal metrics are removed.
  2. Fixes other cases:
       * Removes double emit in `vector` sink.
       * Removes double emit in `prometheus_remote_write` source.
       * Although `kubernetes_logs` reuses file server, it isn't emitting any problematic metric.
  
### Todo

- [ ] Open follow up for before/after events.
- [ ] Open follow up for preciser way to measure.


src/topology/builder.rs
```diff
@@ -120,9 +121,11 @@ pub async fn build_pieces(
 
         let (output, control) = Fanout::new();
 
+        let filtered = input_rx
+            .filter(move |event| filter_event_type(event, input_type))
+            .inspect(|_| emit!(EventProcessed));
```
MOZGIII, line 126: Won't this print before the event is actually processed, rather than after?

src/topology/builder.rs
```diff
@@ -275,16 +273,10 @@ pub async fn build_pieces(
...
+fn filter_event_type(event: &Event, data_type: DataType) -> bool {
     match data_type {
-        DataType::Any => Box::new(stream), // it's possible to not call any comparing function if any type is supported
-        DataType::Log => Box::new(stream.filter(|event| matches!(event, Event::Log(_)))),
-        DataType::Metric => Box::new(stream.filter(|event| matches!(event, Event::Metric(_)))),
+        DataType::Any => true,
+        DataType::Log => matches!(event, Event::Log(_)),
+        DataType::Metric => matches!(event, Event::Metric(_)),
     }
 }
```
bruceg, line 282: I would tend to think this a step backwards, for performance at least. Can this still be done with a stream adapter like before for all the places it's used above?

bruceg (commented):
I share @MOZGIII's concern about the sequence of when the event is emitted and have one of my own, but this is a great simplification of this event handling! :fire:

src/topology/builder.rs
MOZGIII: It looked very suspicious to me too - that the code before the change was duplicated like that, but I don't get the idea behind it. Could you please elaborate more on where will the performance difference come from?

bruceg: Thinking about it a bit more, it probably doesn't make a big difference on modern CPUs, and might actually be a win. Before, each pipeline would be instantiated with a fixed stream filter function (single conditional branch on the event type) but will likely require an indirect jump to get to that function. With this change, each pipeline is instantiated with a generic stream filter function (two conditional branches). Mispredicted indirect jumps are likely more expensive than mispredicted branches. This is pretty down in the weeds of micro-benchmarking, though.

ktff: Well, regarding performance, with the old there would have to be two dyn boxing which this replaces for an integer switch. Two data memory jumps + two code jumps vs one local integer switch.

src/topology/builder.rs
ktff: Yes it will. 

First, about half of transforms were emitting this metric before doing anything with the event, and half after it. 

Now, this more of an issue of what do we want to count as processed. I see two versions:
* Only events that are processed through a proper, non error, code path is to be counted.
* Or, all events being processed are counted, even if we encounter some error with them. 

The second definition is what is being counted in this PR. The most practical reason as to why is because we cant count the first definition on the topology level. If we would count on the output of the transform we would miscount all sampling/filtering/dedup based transforms.

ktff: As a side note, that's why no `emit_logs` is added to `EventProcessed`.

src/topology/builder.rs
bruceg: Nit: the replacement code has two switches: first on `data_type` and then on the `matches!` inside the match.

src/topology/builder.rs
ktff: But if the issue is more of just the timing of emit, we can do something about it.

ktff: I don't see a way of emitting the metric after it was processed. The main reason is that there is currently no way of injecting code to `Trannsform::Task` that would run once per event and after event was processed. 

Any idea?

MOZGIII: How about changing the logic here to emit two events: before and after? It makes way more sense from the observability standpoint imo, if we apply this logic at the topology layer. I.e. if we ever have a `blackhole` transform, it will show a lot of events went in and none went out. For the `sampler` transform, it would be obvious how many events were filtered out.

ktff:
It seams the two api tests aren't failing because of this PR.

src/topology/builder.rs
MOZGIII: > I don't see a way of emitting the metric after it was processed. The main reason is that there is currently no way of injecting code to `Trannsform::Task` that would run once per event and after event was processed.
> 
> Any idea?

I'll think about it... Also, maybe @Hoverbear can help?

Hoverbear: Maybe we can stick it onto `output` somehow?

src/topology/builder.rs
ktff: That's true I just compared the diff.

src/topology/builder.rs
ktff: > Maybe we can stick it onto output somehow?

Nope, the number of events going into output doesn't correspond to the number of events that went into the transform.

> How about changing the logic here to emit two events: before and after?

That would be pretty useful, but we can expand with that in a separate PR.

---

I think the difference between the orderings is not relevant. Currently that event doesn't print anything, and if we change that, we can adjust the log message accordingly. And regarding the metric, that should only be observable from the `internal_metrics` source and the api, right? So that metric would only be out of order relative to other metrics emitted by that transform, for which I don't think we need to care.

bruceg (approved):

src/topology/builder.rs
ktff: @MOZGIII I would like to get your take on this before merging.

lukesteensen: > > How about changing the logic here to emit two events: before and after?
> 
> That would be pretty useful, but we can expand with that in a separate PR.

Yes, I think the goal here should be to match the current behavior. We can discuss transitioning to multiple events separately.

lukesteensen (approved):
Looks great, thank you @ktff!

I agree with @MOZGIII that we can consider a more precise way to measure, but this is already a huge improvement in that we're measuring consistently across all transforms.

src/topology/builder.rs
MOZGIII: > And regarding the metric, that should only be observable from the `internal_metrics` source and the api, right? So that metric would only be out of order relative to other metrics emitted by that transform, for which I don't think we need to care.

We also expose those metrics via prometheus sink in k8s, so users will be able to see raw metrics, and we should try and make them consistent if possible.

I agree that we can transition to the before/after in a separate PR - for what it's worth, it'd be a cleaner change and easier to review.

src/sinks/vector.rs
```diff
@@ -75,10 +74,6 @@ fn encode_event(event: Event) -> Option<Bytes> {
     let event_len = event.encoded_len();
     let full_len = event_len + 4;
 
-    emit!(VectorEventSent {
-        byte_size: full_len
-    });
-
```
MOZGIII, line 81: We might want to convert this one to a sink-specific emit/metric to preserve the bytes sent data point.

src/sources/prometheus/remote_write.rs
```diff
@@ -115,10 +115,9 @@ impl HttpSource for RemoteWriteSource {
         _header_map: HeaderMap,
         _query_parameters: HashMap<String, String>,
     ) -> Result<Vec<Event>, ErrorMessage> {
-        let byte_size = body.len();
```
MOZGIII, line 118: In this case, `body` is `Bytes`, which has the length known upfront. No reason not to send pass this data to the metric.

src/internal_events/prometheus.rs
```diff
@@ -129,19 +129,13 @@ impl InternalEvent for PrometheusRemoteWriteSnapError {
...
     pub count: usize,
 }
 
 impl InternalEvent for PrometheusRemoteWriteReceived {
     fn emit_logs(&self) {
         debug!(message = "Received remote_write events.", count = ?self.count);
     }
-
-    fn emit_metrics(&self) {
-        counter!("processed_events_total", self.count as u64);
```
MOZGIII, line 142: We could convert this to sink-specific metrics to preserve the data point

MOZGIII (approved):
I think we're losing some data points with this PR in a couple of places, which is not good and we probably should spend some effort to try to preserve them.
Other than that looks good!
Should we create a follow-up issue to transition to before/after metrics for transfroms?

src/sinks/vector.rs
ktff: We have that data point emitted in `BytesSink` which is being used by `TcpSink` which is in turn being used by this sink.
https://github.com/timberio/vector/blob/ae18f8f7790c56bd916cc634a0cb818d1e6db9c4/src/sinks/util/socket_bytes_sink.rs#L70-L74

Or do you mean some other data point?

src/internal_events/prometheus.rs
ktff: We have the same metrics being emitted in `HttpSource` being used by this source.
https://github.com/timberio/vector/blob/ae18f8f7790c56bd916cc634a0cb818d1e6db9c4/src/sources/util/http.rs#L180-L183

Am I missing something here?

ktff:
@MOZGIII I'll merge this to not prolong it any further, we can address your comments with a follow up.

PR #5502, opened 2020-12-11
chore(remap): add owners to `.github/CODEOWNERS`

fanatid:
I added Jean (@JeanMertz) for all remap functions and Stephen (@FungusHumungus) only for components that he added.

@FungusHumungus is this looks good for you or change for all remap components?

PR #5522, opened 2020-12-12
chore(remap): pull functions out of Vector into their own crate

JeanMertz:
This PR moves out all Remap functions from the `vector` crate into the new `remap-functions` crate. This is needed so that we can start embedding the functions into `remap-cli`, without having to depend on the `vector` crate.

Doing so, I had to move some shared logic out of the `vector` crate and into its own `shared` crate.

I guess this last part is of most interest to others, so here's the gist:

* introduced a new (unpublished) `shared` crate
* pulled out the bare minimum needed to make things compile
* put everything behind feature flags so other crates can pick what they need
* made sure the dependencies match up between crates
* re-exported the dependencies at their original locations in Vector

Here's the list of features that got pulled out of vector:

* ~~`vector::event::Value` -> `shared::value`~~
* `vector::types` -> `shared::conversion`
* ~~`vector::trace` -> `shared::trace`~~
* `vector::tokenize` -> `shared::tokenize`
* ~~some of `vector::test_util` moved to `shared::test_util`~~
* ~~some of `vector::transform::lua` moved to `shared::lua`~~

(**update:** I was able to [reduce the amount of code](https://github.com/timberio/vector/pull/5522#pullrequestreview-550868977) pulled into the `shared` crate)

(**update 2:** [more features moved back into Vector](https://github.com/timberio/vector/pull/5522/commits/229deca164122e25e8467c329efdd778271b0c30))

(again, these are re-exported, so there is minimal/no churn to dependent code)

I know @bruceg is interested in this work, as he too needs to pull out some shared logic to use in his Prometheus work. Also cc'ing @lukesteensen and @Hoverbear since Bruce told me there were some discussions before on doing/not doing this.

I'm happy to make any changes, or I can just copy/paste the required code into `remap-functions` so that the new `shared` crate is no longer needed (but I would obviously like to avoid that situation).

_(note: Vector builds and unit tests work locally, but I'm sure the CI will catch a few integration tests that I still need to fix, as I didn't run those locally)_


lib/shared/src/value.rs
```diff
@@ -20,6 +21,34 @@ pub enum Value {
     Null,
 }
 
+macro_rules! value_impl {
```
JeanMertz, line 24: Ah, this got snuck in from `remap::Value`. I wanted to introduce this in a follow-up PR (the rename from `as_bytes` to `as_bytes_lossy` is related to this change).

I'm happy to pull it out, but if others are fine with it, I'll sneak it in through this PR 😄

lib/shared/src/conversion.rs
```diff
@@ -0,0 +1,395 @@
...
+    #[snafu(display("No matching timestamp format found for {:?}", s))]
+    AutoTimestampParseError { s: String },
+}
+
+impl Conversion {
+    /// Use this `Conversion` variant to turn the given `bytes` into a new `T`.
+    pub fn convert<T>(&self, bytes: Bytes) -> Result<T, Error>
+    where
+        T: From<Bytes> + From<i64> + From<f64> + From<bool> + From<DateTime<Utc>>,
+    {
```
JeanMertz, line 117: This method was updated to be generic over `T`, so that it no longer relies on `vector::Value`.

JeanMertz (commented):
I made some changes to the function impls and to the `Conversion` type, making it so that we no longer require `Value` to be shared between crates. Also, because of this the Lua logic has all been moved back into Vector as well.

lib/shared/src/value.rs
JeanMertz: Never mind, reverted this change for now.

JeanMertz:
Another round of trimming, the `trace` feature is no longer needed in the `shared` crate.

This makes the logic pulled out of Vector into the `shared` crate very minimal, and probably a lot less contentious to agree on.

lib/remap-functions/src/parse_url.rs
```diff
@@ -58,34 +56,32 @@ impl Expression for ParseUrlFn {
...
+    map.insert(
+        "query",
+        url.query_pairs()
+            .into_owned()
+            .map(|(k, v)| (k, v.into()))
+            .collect::<BTreeMap<String, Value>>()
+            .into(),
+    );
+
+    Value::from_iter(map.into_iter().map(|(k, v)| (k.to_owned(), v)))
```
bruceg, line 84: This `map` has a fixed number of elements, and the tree form is not actually used other than to iterate over. I think this could be done as an array (`[(&str, Value)]`) and avoid the extra tree allocation work for nothing.

Yes, this is just a code move, didn't realize this before.

lib/remap-functions/src/round.rs
```diff
@@ -82,6 +82,7 @@ impl Expression for RoundFn {
 }
 
 #[cfg(test)]
+#[allow(clippy::approx_constant)]
```
bruceg, line 85: Can this be avoided by using a different constant?

src/remap/mod.rs
```diff
@@ -1,7 +1,5 @@
-pub(crate) mod function;
-
-pub use function::*;
 use lazy_static::lazy_static;
+use remap_functions::*;
```
bruceg, line 2: Does this depend on all features in `remap-functions` being enabled?

src/sinks/influxdb/logs.rs
```diff
@@ -225,14 +225,12 @@ impl InfluxDBLogsConfig {
...
-            _ => Field::String(self.to_string_lossy()),
-        }
+fn value_to_field(value: &Value) -> Field {
+    match value {
+        Value::Integer(num) => Field::Int(*num),
+        Value::Float(num) => Field::Float(*num),
+        Value::Boolean(b) => Field::Bool(*b),
+        _ => Field::String(value.to_string_lossy()),
     }
 }
```
bruceg, line 235: Could this be `impl From<&Value> for Field`? `enum Field` is defined within Vector.

bruceg (commented):
I have a few questions, but otherwise I approve. Also, does this require any changes to `.github/CODEOWNERS` due to the moves?

FungusHumungus:
With these files moving the `.github/CODEOWNERS` file will need updating too (once merged with #5502).

lib/remap-functions/src/ceil.rs
```diff
@@ -122,6 +122,7 @@ mod tests {
     ];
 
     #[test]
+    #[allow(clippy::approx_constant)]
```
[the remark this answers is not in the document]
JeanMertz: Will do 👍 (I should note, we also have this disabled globally in Vector itself, which is why I followed the same pattern here)

lib/remap-functions/src/parse_url.rs
JeanMertz: Yeah, this can definitely be improved. We still have a "performance pass" to do over all of our functions, but I'll leave that for another PR.

I created an issue to track these at https://github.com/timberio/vector/issues/5573.

src/sinks/influxdb/logs.rs
JeanMertz: I've reverted these changes, since `Value` was moved back into Vector again 👍

src/remap/mod.rs
JeanMertz: Good point. It does, but this has been updated in https://github.com/timberio/vector/pull/5531, which adds an `all()` method to `remap_functions`, which returns whatever list of functions you've enabled through the feature flags.

PR #5525, opened 2020-12-12
fix(kubernetes platform): Implement improvements to the Pod recreation on Helm release upgrades

### LINKED ISSUE #5524: Improve the support for Pod recreation upon Helm chart update
While working on the k8s dev env, it became evident that Helm charts need some improvements with the Pod recreation upon the helm release update.
I'm going to address this.

Ref: https://github.com/helm/helm/issues/9120

MOZGIII:
This PR adds two new parameters to the Helm charts: `podRollmeAnnotation` and `podValuesChecksumAnnotation`. See the diff to learn more about them.

It also restructures the internals of the Helm charts to share the annotation logic.

It also corrects the config checksum calculation logic by switching it to checksumming the evaluated config, rather than some odd behviour. The code we had there came from the official Helm docs, so was a bit surprising to find issues with it.

Closes #5524.

### DIFF
--- a/distribution/helm/vector-agent/templates/daemonset.yaml
+++ b/distribution/helm/vector-agent/templates/daemonset.yaml
@@ -14,9 +14,7 @@ spec:
   template:
     metadata:
       annotations:
-        {{- if not .Values.externalConfigMap }}
-        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
-        {{- end }}
+        {{- include "libvector.configTemplateChecksumAnnotation" . | nindent 8 }}
         {{- include "libvector.rollmeAnnotation" . | nindent 8 }}
         {{- include "libvector.metricsPrometheusPodAnnotations" . | nindent 8 }}
         {{- with .Values.podAnnotations }}
--- a/distribution/helm/vector-agent/templates/daemonset.yaml
+++ b/distribution/helm/vector-agent/templates/daemonset.yaml
@@ -14,9 +14,7 @@ spec:
   template:
     metadata:
       annotations:
-        {{- include "libvector.configTemplateChecksumAnnotation" . | nindent 8 }}
-        {{- include "libvector.valuesChecksumAnnotation" . | nindent 8 }}
-        {{- include "libvector.rollmeAnnotation" . | nindent 8 }}
+        {{- include "libvector.rerollAnnotations" . | nindent 8 }}
         {{- include "libvector.metricsPrometheusPodAnnotations" . | nindent 8 }}
         {{- with .Values.podAnnotations }}
           {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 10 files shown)


spencergilbert:
I think in general _most_ changes to values would trigger updates itself - what values changes do you think would not be covered by default behavior @MOZGIII? The Rollme annotation would definitely be useful for my implementation because I'm providing a configmap myself so the checksum annotation doesn't apply there.

MOZGIII:
> I think in general _most_ changes to values would trigger updates itself - what values changes do you think would not be covered by default behavior @MOZGIII? The Rollme annotation would definitely be useful for my implementation because I'm providing a configmap myself so the checksum annotation doesn't apply there.

Well, the `ConfigMap` changes are the most frequent, yet they won't trigger a reroll on their own.

distribution/helm/vector-agent/templates/daemonset.yaml
```diff
@@ -14,9 +14,7 @@ spec:
   template:
     metadata:
       annotations:
-        {{- if not .Values.externalConfigMap }}
-        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
```
spencergilbert, line 18: @MOZGIII shouldn't this annotation cause the pods to require a redeploy on any change in Values that causes the configmap to change?

MOZGIII: No, this only sum changes when the *template* changes, unfortunately. It's reading the raw template file and just sums it.

MOZGIII: Hmm, maybe a better fix would be to evaluate that template instead?

spencergilbert: Oh, word - didn't realize that!

spencergilbert (approved):
I think both are good changes, and the values based hash being on by default is good default behavior. 

EDIT: I can think of something that _wouldn't_ need the pods to restart, such as some changes to the statefulset's service. It may be less disruptive if only the .Values.{values that affect config} are hashed

distribution/helm/vector-agent/templates/daemonset.yaml
spencergilbert: I think either hashing _just_ the values that affect the configuration or hashing the final configuration would probably be optimal

MOZGIII: Yep, fixed it, and disabled the `podValuesChecksumAnnotation` by default.

The code we had there is from the Helm chart tips and tricks docs. Well, I guess they were just confused with their own tool, happens.

MOZGIII:
> EDIT: I can think of something that _wouldn't_ need the pods to restart, such as some changes to the statefulset's service. It may be less disruptive if only the .Values.{values that affect config} are hashed

I figured it can be further improved, see https://github.com/timberio/vector/pull/5525#discussion_r541613649

distribution/helm/vector-agent/templates/daemonset.yaml
spencergilbert: Is `podValuesChecksum` still needed if the configmap based checksum properly works?

eeyun (approved):

PR #5526, opened 2020-12-12
fix(elasticsearch sink): Fix trailing '/' at the end of URL

juchiast:
Introduced in #5379


juchiast:
Some context: ES sink assumed that `endpoint` does not end with `/`, but `Uri::to_string` produces a string ending with `/`.

jszwedko (approved):
I was just digging into this issue :smile: 

I'll merge to get master back to green. The CI failures seem unrelated.

PR #5532, opened 2020-12-14
enhancement(docker_logs source): Add host and TLS configuations

### LINKED ISSUE #5413: Enhance docker_logs source with additional configuration for Docker socket
Currently, the `docker_logs` source is configured via the `DOCKER_HOST` environment variable or the socket default of `/var/run/docker.sock`. We should allow this to be configured in the `docker_logs` configuration because:

1. We shouldn't rely on setting an environment variable.
2. The socket default varies on various platforms, for example `/run/containerd/containerd.socket`.

### LINKED ISSUE #5488: Add and document TLS configuration for docker_logs source
Documentation should mention how env var `DOCKER_CERT_PATH` and `DOCKER_CONFIG` is being used to get certificates when connecting via TLS.

In the same spirit as #5413, we should add  TLS configuration options for `docker_logs` source.

Three required options:
- `tls.ca_file`
- `tls.crt_file`
- `tls.key_file`

Official docker client allows some of the options to be missing, but `bollard` does not, therefore all three options are required.

juchiast:
close #5413 
close #5488 


docs/reference/components/sources/docker_logs.cue
```diff
@@ -14,21 +14,32 @@ components: sources: docker_logs: {
...
 		DOCKER_HOST: {
-			description: "The Docker host to connect to."
+			description: "The Docker host to connect to when `docker_host` configuration is absent."
 			type: string: {
 				default: null
 				examples: ["unix:///var/run/docker.sock"]
 			}
 		}
 
-		DOCKER_VERIFY_TLS: {
```
juchiast, line 24: Removed because `bollard` does not implement this features.

docs/reference/components/sources/docker_logs.cue
```diff
@@ -84,6 +95,51 @@ components: sources: docker_logs: {
 	}
 
 	configuration: {
+		docker_host: {
+			common: true
+			description: """
+				The Docker host to connect to. Use an HTTPS URL to enable TLS encryption.
```
jszwedko, line 101: Would it make sense to call out that the default ('null') will result in it automatically trying `/var/run/docker.sock` on unix or `\\.\pipe\docker_engine` on Windows?

I'd also call out that this will default to `$DOCKER_HOST`.

docs/reference/components/sources/docker_logs.cue
```diff
@@ -84,6 +95,51 @@ components: sources: docker_logs: {
...
+				"""
+			required: false
+			type: string: {
+				default: null
+				examples: ["http://localhost:2375", "https://localhost:2376", "/var/run/docker.sock"]
```
jszwedko, line 106: Maybe we could add `\\.\pipe\docker_engine` as another example?

docs/reference/components/sources/docker_logs.cue
```diff
@@ -84,6 +95,51 @@ components: sources: docker_logs: {
...
+			}
+		}
+		tls: {
+			common:      false
+			description: "TLS options to connect to the Docker deamon. This has no effect unless `docker_host` is an HTTPS URL."
```
jszwedko, line 111: Can we call out the env vars and path (`~/.docker`) that it will default to here?

jszwedko (approved):
Looks good! Just some notes on the documentation.

juchiast:
@jszwedko I've updated the docs.

jszwedko (approved):
Looks good!

PR #5533, opened 2020-12-14
chore(ci): Fix conditions::remap::test::check_remap null check

jszwedko:
Broken in master: https://github.com/timberio/vector/runs/1547837589#step:9:1063

It is a bit odd that the error message differs from the tests below it in that includes "to an error", but this fixes it at least.

### DIFF
--- a/src/conditions/remap.rs
+++ b/src/conditions/remap.rs
@@ -120,8 +120,8 @@ mod test {
             ),
             (
                 log_event![],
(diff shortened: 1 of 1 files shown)


jszwedko:
Opened issue for OSX flake: https://github.com/timberio/vector/issues/5536

JeanMertz (approved):

PR #5537, opened 2020-12-14
chore(ci): Run benchmarks on GA runners

jszwedko:
Until we sort out a way to isolate the builds as part of #5394


jszwedko:
I had intended not to switch it to the self-hosted runners, but must have forgot to revert that change.

jamtur01 (approved):
LGTM

PR #5540, opened 2020-12-14
fix(kubernetes platform): Correct the prometheus format parsing at K8s E2E tests for the metrics pipeline

MOZGIII:
This PR fixes the failing Kubernetes E2E tests.

The issue was in the test code, not in our app code, and the patch improves the test code to allow for more flexibility, effectively allowing the current app state (which is valid really and shouldn't fail the tests).

### DIFF
--- a/lib/k8s-e2e-tests/src/metrics.rs
+++ b/lib/k8s-e2e-tests/src/metrics.rs
@@ -57,7 +57,7 @@ pub async fn get_processed_events(url: &str) -> Result<u64, Box<dyn std::error::
 pub async fn assert_vector_started(url: &str) -> Result<(), Box<dyn std::error::Error>> {
     let metrics = load(url).await?;
     if !extract_vector_started(&metrics) {
-        return Err(format!("vector_started metric was not found:\n{}", metrics).into());
+        return Err(format!("`vector_started`-ish metric was not found:\n{}", metrics).into());
     }
     Ok(())
 }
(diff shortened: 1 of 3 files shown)


lib/k8s-e2e-tests/src/metrics.rs
```diff
@@ -9,7 +9,7 @@ pub async fn load(url: &str) -> Result<String, Box<dyn std::error::Error>> {
 
 fn metrics_regex() -> regex::Regex {
     regex::RegexBuilder::new(
-        r"^(?P<name>[a-zA-Z_:][a-zA-Z0-9_:]*)(?P<labels>\{[^}]*\})? (?P<value>.+)$",
+        r"^(?P<name>[a-zA-Z_:][a-zA-Z0-9_:]*)(?P<labels>\{[^}]*\})? (?P<value>[^ ]+?)( (?P<timestamp>[^ ]+?))?$",
```
bruceg, line 12: I don't think the `value` is optional as above, and `\S` is a nice short-hand for `[^ ]`:
```suggestion
        r"^(?P<name>[a-zA-Z_:][a-zA-Z0-9_:]*)(?P<labels>\{[^}]*\})? (?P<value>\S+)( (?P<timestamp>[^ ]+?))?$",
```

bruceg (approved):

lib/k8s-e2e-tests/src/metrics.rs
MOZGIII: The `?` here is a lazy qualifier for `+`, not an optional qualifier for a group.
I didn't know about `\S`, thx!

MOZGIII: Added `\S`.

jszwedko (approved):
Thanks @MOZGIII !

PR #5556, opened 2020-12-16
chore(tests): Test for errors_total

### LINKED ISSUE #5329: Add `_errors_total` API test
There's currently no assertion to check that error metrics are being picked up by the API.

This should be caught by https://github.com/timberio/vector/blob/master/src/api/schema/metrics/mod.rs#L172-L188 but @FungusHumungus reported that events aren't updating in `vector top`.

leebenson:
Test to assert that `errors_total` is picked up correctly in the API. This is the aggregate error count, and not per component.

This doesn't answer why specific components aren't reporting errors, which requires a little more diving after some recent metrics changes. That's next.

Closes #5329


FungusHumungus (approved):
Nice!

PR #5559, opened 2020-12-16
chore(ci): use rustflags instead of rustc wrappers

japaric:
now that
- the aarch64-musl `cross` image respects the rustflags setting specified in .cargo/config, and
- rustc 1.48 supports the -C link-self-contained flag

we can remove the rustc wrappers in favor of rustflags settings in `.cargo/config`


jszwedko:
Yeah, the feature check should be fixed by https://github.com/timberio/vector/pull/5553

Hoverbear (approved):

PR #5561, opened 2020-12-16
fix(docker_logs source): Update hyper to work around the docker EOF errors

MOZGIII:
This PR fixes (or rather is supposed to be fixing) the issue with docker that was reported earlier today.

Once merged, we'd want to backport it to 0.11 and ship with 0.11.1.

This is a hotfix, a proper resolution will be to simply bump the hyper once they cut a release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -570,3 +570,6 @@ required-features = ["remap-benches"]
 
 [patch.'https://github.com/tower-rs/tower']
 tower-layer = "0.3"
+
+[patch.crates-io]
+hyper = { version = "0.13", git = "https://github.com/timberio/hyper", rev = "4ec7442fb74634ea9d128ad58bb741fa1dabefc7" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2672,8 +2672,7 @@ dependencies = [
 [[package]]
 name = "hyper"
 version = "0.13.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6ad767baac13b44d4529fcf58ba2cd0995e36e7b435bc5b039de6f47e880dbf"
+source = "git+https://github.com/timberio/hyper?rev=4ec7442fb74634ea9d128ad58bb741fa1dabefc7#4ec7442fb74634ea9d128ad58bb741fa1dabefc7"
 dependencies = [
  "bytes 0.5.6",
  "futures-channel",


eeyun (approved):

jamtur01 (approved):
LGTM.

jamtur01:
@binarylogic As discussed.

Cargo.toml
```diff
@@ -570,3 +570,6 @@ required-features = ["remap-benches"]
 
 [patch.'https://github.com/tower-rs/tower']
 tower-layer = "0.3"
+
+[patch.crates-io]
```
jamtur01, line 574: We should add a comment here explaining what we need to bump hyper at some point.

jamtur01 (changes requested):

Cargo.toml
eeyun: Indeed - I'd also like to wait to merge this until we verify that it's the solution. I think based on our morning detective work it _is_ but if its at all possible to avoid merging an unnecessary crate patch that would be my preference.

MOZGIII: We can merge the patch and then switch to the released version. If we merge it the solution will be in the nightly and it'll be easier to test. We'll have to bump to publushed version either way, but we're imo in a good position to work with such patches.

I'll add a comment, and will merge if you agree.
We can already move from a fork to the upstream git since my PR was accepted.

eeyun:
That sounds good to me!

On December 17, 2020, MOZGIII <notifications@github.com> wrote:
> @MOZGIII commented on this pull request.
>

> In Cargo.toml
> <https://github.com/timberio/vector/pull/5561#discussion_r545048034>:

>

> @@ -570,3 +570,6 @@ required-features = ["remap-benches"]
[patch.'https://github.com/tower-rs/tower'] tower-layer = "0.3" +
+[patch.crates-io]
> We can merge the patch and then switch to the released version. If we

> merge it the solution will be in the nightly and it'll be easier to
> test. We'll have to bump to publushed version either way, but we're
> imo in a good position to work with such patches.
>
> I'll add a comment, and will merge if you agree.
>  We can already move from a fork to the upstream git since my PR was
> accepted.
>

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/timberio/vector/pull/5561#discussion_r545048034>, or

unsubscribe <https://github.com/notifications/unsubscribe-
auth/ADAH33B2JQOB5TPNH6MLEK3SVHZUBANCNFSM4U6H4FSQ>.

Cargo.toml
eeyun: Sounds good to me!

PR #5572, opened 2020-12-17
chore(observability): Humanized bytes

leebenson:
Fixes the display of abbreviated bytes metrics when using the `-h` flag in `vector top`.

Previously, byte stats were suffixed with `K`, `G`, `T`, etc, used with non-bytes metrics via `human_format`.

Bytes metrics now use the `human_format_bytes` extension in all places.

Result:

**With no `-h` flag:**

image

**With `-h`:**

image


sghall (approved):
Looks good 👍

Issue #5129, opened 2020-11-19
Audit dependency licenses


jszwedko:
We currently have dependencies with the following licenses:

```
BSD-2-Clause
BSD-3-Clause
Apache-2.0
MPL-2.0
OpenSSL
MIT
ISC
CC0-1.0
GPL-2.0
Unlicense
```

We should audit thees and ensure we are satisfying their restrictions (for example republishing licenses of dependencies where prescribed).

As part of this, we should add checking licenses to our usage of `cargo-deny-action` in CI to ensure we deal with any new licenses that surface

My initial survey surfaced requirements like:

- BSD-2-Clause and BSD-3-Clause: Requires redistribution of licenses with binary form
- Apache-2.0: Requires redistribution of licenses with binary form
- MPL-2.0: Requires conveying to users how to obtain sources for binary form
- OpenSSL: Requires redistribution of license as well as a notice on advertising materials that we use openssl
- GPL-2.0: Compatible with our MPL-2.0
- MIT: Seems like it does not require redistribution of license with binary form
- ISC: Seems like it does not redistribution of license with binary form
- Unlicense: No requirements
- CC0-1.0: No requirements

fanatid:
What we should do with code and licenses which was moved from crates to vector? Examples:
- https://github.com/timberio/vector/blob/947319bb59f0841d245c47683d0279056f19b7e8/src/sinks/util/retries.rs#L182
- https://github.com/timberio/vector/blob/947319bb59f0841d245c47683d0279056f19b7e8/src/sink.rs#L1

jszwedko:
I think we should just make sure to satisfy the requirements of the license. In this case, given we _are_ redistributing the source including the licenses with the code seems like the right approach to me for MIT licenses.

PR #5592, opened 2020-12-17
chore(ci): Change license check default to warn

jamtur01:


jszwedko (approved):
This seems reasonable for now until we hit #5129

PR #5603, opened 2020-12-18
chore(ci): Always pass benches workflow

jszwedko:
Until we can reduce measurement noise as part of
https://github.com/timberio/vector/issues/5394 as, even if the workflow is marked to allow failures, it still marks the suite of checks as failing.


jamtur01 (approved):
LGTM

jszwedko:
Mac failure is fixed by #5609

Issue #5589, opened 2020-12-17
Metrics batch buffer does not pass through a single counter


bruceg:
While writing test cases for #5227, I discovered that the metrics batch buffer does not pass through single counter values. All of the internal tests contain multiple values for the same counter, and they pass, but trying to send a single counter disappears.

I wrote up a test case (along with some other refactoring) on this branch: https://github.com/timberio/vector/tree/fix-metric-batch-buffer

bruceg:
I'm having difficulty reasoning about the correct behavior for this feature, other than that the current behavior is not correct. The metric batch buffer saves the last known state of each counter in addition to the current buffer, used to aggregate repeated metrics. Given that saved state, a buffer, and a new counters to buffer, the following lists all the possible states (have to list them all to make sure all angles are covered). Is the new desired behavior the correct behavior?

| Saved state | In buffer              | New counter | Current behavior | Changed behavior |
| ----------- | ---------------------- | ----------- | --- | --- |
| None        | None                   | Absolute    | Nothing(1) | Emit new absolute |
| None        | None                   | Incremental | Emit new incremental |
| None        | Absolute               | N/A         | *unreachable*
| None        | Absolute + Incremental | N/A         | *unreachable*
| None        | Incremental            | Absolute    | Nothing(3) | Replace with new absolute |
| None        | Incremental            | Incremental | Update incremental |
| Absolute    | None                   | Absolute    | Emit delta | Emit absolute |
| Absolute    | None                   | Incremental | Emit incremental |
| Absolute    | Absolute               | Absolute    | Emit delta(2) | Update existing |
| Absolute    | Absolute               | Incremental | Emit new incremental(2) | Update existing |
| Absolute    | Absolute + Incremental | Absolute    | Update incremental | *unreachable* |
| Absolute    | Absolute + Incremental | Incremental | Update incremental | *unreachable* |
| Absolute    | Incremental            | Absolute    | Update incremental | Replace with new absolute |
| Absolute    | Incremental            | Incremental | Update incremental |
| Incremental | N/A                    | N/A         | *unreachable* |

1. This is the behavior that first triggered this work, as I was trying to test a single counter being passed through a metric sink.
2. This results in both an absolute and an incremental in the same buffer.
3. This is obviously wrong too.

Hopefully I got all the lines above right. Who knew simple counters could be so complex!

jszwedko:
[...]

PR #5615, opened 2020-12-19
chore(metrics): Refactor metrics batch buffer tests

bruceg:
While investigating #5589, I did some reductive work to simplify the tests for the metrics batch buffer. This doesn't actually change what is tested, it just reduces the amount of code needed to do so.


ktff (approved):
Now this is much nicer. I think it's for the better that the tests are now more focused on the things that they are testing.

